PR #4, opened 2021-06-05
check status before proceeding to add plan to map

eqy:


include/cudnn_frontend_find_plan.h
```diff
@@ -55,7 +55,10 @@ time_sorted_plan(cudnnHandle_t handle, executionPlans_t plans, VariantPack &vari
         float min_time_ms   = std::numeric_limits<float>::max();
 
         // Warm-up run
-        ::cudnnBackendExecute(handle, plan.get_raw_desc(), variantPack.get_raw_desc());
+        cudnnStatus_t stat = ::cudnnBackendExecute(handle, plan.get_raw_desc(), variantPack.get_raw_desc());
+        if (stat != CUDNN_STATUS_SUCCESS) {
```
Anerudhan, line 59: Hi @eqy ,

Thanks for the feedback. We have the above change in works and will be part of next release.

Thanks

PR #6, opened 2021-07-16
Fix move constructor on Tensor_v8.

jlebar:


jlebar:
cc @awpr

Anerudhan:
Thank you for your feedback. We will incorporate the said changes and add it as part of our next release

Anerudhan:
Above changes are part of the patch release.
Thanks for your continued feedback.

PR #10, opened 2021-10-04
Fix compilation on gcc 11.1

zasdfgbnm:


Anerudhan:
Thank you for the message. We will add the required changes in the next release. 
For now cuDNN frontend does not accept direct pull requests.

PR #27, opened 2022-04-09
V0.6.1

Anerudhan:
Fix missing return statement for cuDNN versions less than cuDNN v8.4


Anerudhan:
Fix missing return statement for cuDNN versions less than cuDNN v8.4

PR #57, opened 2024-01-22
Update Ifdef for no_half_converison__

drisspg:
# Summary

There are number of cuda 11.4.2 builds that set this env variable and while trying to update them I ran across this compilation issue. I am not entirely sure if this macro makes sense for the value.


Anerudhan:
Hi @drisspg ,

Thank you for the message. cuDNN team has added the changes in 1.0.3

At this point we do not accept external PRs directly.

Please re-open this if you need.

Thanks

PR #65, opened 2024-02-27
1.1.2 release.

Anerudhan:
[Bug fix] Fixed an issue where the heuristic, when returning 0 results was throwing an error. This is considered a success as this can be appended with the fallback heuristics to build a valid plan.


tfogal (approved):
LGTM! Thank you for the quick fix.

Sadly the issue only reproduces for me in CI, but I did try this version of cudnn locally and it works great with our internal project. Thanks!

PR #90, opened 2024-07-28
Fix incorrect usage of CMAKE_{SOURCE,BINARY}_DIR

junaire:
When importing cudnn-frontend as a 3rd party library using FetchContent, cmake incorrectly parse the file path. This is because we incorrectly using CMAKE_SOURCE_DIR variable, which should be PROJECT_SOURCE_DIR (The former one refers to the top-level source directory that contains a CMakeLists.txt, while the latter refers to the source directory of the most recent project() command


junaire:
CC @Anerudhan

PR #125, opened 2025-01-18
cmake: install config with headers

ConnorBaker:
This PR contains a number of changes I made while packaging this repo for use with Nixpkgs.


Anerudhan:
Hi @ConnorBaker ,

Firstly, thanks for the pull request.

- We have a release 1.10 scheduled later this month, after which I will try to resolve the changes and merge it in.  
- I see you have added `CUDA::cublasLt` and `CUDA::nvrtc` as `target_link_dependencies` for  the samples. However, these are dynamically loaded and not dynamically linked. May I understand the motivation for this? 


Thanks,
Anerudhan

ConnorBaker:
Dynamic loading is problematic in Nixpkgs because it hides the dependency from Nix. As a result, users need to provide the libraries themselves and patch the samples so they can find them (NixOS and nix-built binaries don't use FHS).

By dynamically linking instead of dynamically loading, Nix is able to provide the required libraries with the samples without the need to manually patch anything.

I'm also in favor of dynamically linking over dynamically loading because it moves would-be runtime errors to build errors.

BwL1289:
@Anerudhan is there an update on merging this?

BwL1289:
@ConnorBaker I'm about to submit another PR for `python/CMakeLists.txt` code so that system `dlpack` can be used instead of fetched. I believe Nix would also benefit from it. 

Want me to push to this branch? If not I'll create a separate PR.

Thanks.

ConnorBaker:
Push away! Just added you to the repo.

BwL1289:
Awesome. Just opened a PR on your branch. Tried to change as few lines as necessary. LMK what you think

BwL1289:
@Anerudhan good to merge?

Anerudhan:
Hi @BwL1289 ,
We are in the process of releasing the 1.12.0 this week (by 5/20) I will merge this on top of that after some final testing by 5/21. 
Thanks and sorry for the delay.

BwL1289:
Thank you. Will keep an eye out. We're maintaining a number of nvidia library forks right now and trying to reduce that as much as possible.

PR #130, opened 2025-02-20
Update helpers.h

atupone:
cudnnGetVersion is not found when doing test.

Added #include <cudnn_ops_infer.h>


Anerudhan:
Hi, 
Thank you for your PR. 

I believe, the right header to include is `cudnn.h`, which should be considered the entry point for all headers.

PR #140, opened 2025-04-28
Fix hardcoded `libcudart.so` to support libraries from PyPI

ksivaman:
`libcudart.so` is hardcoded in `get_cuda_symbol`, which breaks workflows that use wheels provided cudaRT's [PyPI](https://pypi.org/project/nvidia-cuda-runtime-cu12/) package.


ksivaman:
Moved

PR #183, opened 2025-11-30
Fix issues in warmup function.

Anerudhan:
- Make sure doing a cuda graph capture is not allowed when one is already in progress
- Teardown the cuda graph to prevent future errors when execute fails


include/cudnn_frontend/graph_interface.h
```diff
@@ -963,7 +968,8 @@ class Graph : public ICudnn, public INode {
         CHECK_CUDNN_FRONTEND_ERROR(
             extend_tensor_map_with_pass_by_value_tensors_(tensor_uid_to_pointer_map, deserialized_pass_by_value));
 
-        CHECK_CUDNN_FRONTEND_ERROR(execute(handle, tensor_uid_to_pointer_map, tmp_pointer));
+        auto cudnn_status = execute(handle, tensor_uid_to_pointer_map, tmp_pointer);
+        (void)cudnn_status;  // No need to check bad executes
```
Skylion007: Isn't the minimum C++17? use [maybe_unused]?

PR #223, opened 2026-04-12
[bugfix] Fix block_descale_b batch stride in NVFP4/MXFP8 block scale matmul samples

### LINKED ISSUE #222: Is the stride for block_descale_b set incorrectly?
https://github.com/NVIDIA/cudnn-frontend/blob/a91f0e04dcea10515f0f776fc5a89535e316a9c8/samples/cpp/matmul/blackwell_nvfp4_mxfp8_block_scale_matmul.cpp#L360-L365

In the code above, block_descale_b's stride[0] is set to block_scale_dim_m * block_scale_dim_k, which is the same as block_descale_a. However, looking at the dimensions, shouldn't it be block_scale_dim_n * block_scale_dim_k instead?

haowen-han:
fix https://github.com/NVIDIA/cudnn-frontend/issues/222
The batch stride (stride[0]) of block_descale_b was incorrectly using block_scale_dim_m instead of block_scale_dim_n. Since block_descale_b has dim {b, block_scale_dim_k, block_scale_dim_n}, the correct batch stride should be block_scale_dim_n * block_scale_dim_k.

This bug was masked when m == n (e.g., 128x128 cases) but produces incorrect strides when m != n.


yanqinz2:
Thanks for catching this issue. @Anerudhan this fix looks good to me, can we merge it?

PR #248, opened 2026-05-21
Fix varpack-template lifecycle bugs and add defensive checks

Anerudhan:
Two pre-existing bugs in the VariantPackTemplate, plus one defensive guard:

1. Graph copy -> dangling host pointers. template_ptrs stores raw addresses into cached_pass_by_value storage owned by the source Graph. Default copy propagated prepared=true while the addresses still pointed at the source. Fix: VarpackPrepStateBox copy ctor/assign now always start with prepared=false so the copy re-preps on first use against its own storage.

2. Re-deserialize on the same Graph -> stale template. deserialize(handle,...) rebinds cached_pass_by_value but the existing prepared=true causes the eager prep to short-circuit, leaving the slot layout from the prior deserialize. Fix: reset prepared=false and clear varpack_template before the eager prep call.

3. Null device_ptrs in raw-ptr create_variant_pack overloads. Reject nullptr
   + non-empty uids instead of forwarding to the cuDNN backend.

Adds explicit null-plan guards across detail::execute overloads, returning GRAPH_EXECUTION_FAILED with "No plan found to execute!" instead of dereferencing plan via plan->getTag().

Authored by : gstefanakis@nvidia.com

### DIFF
--- a/include/cudnn_frontend/graph_interface.h
+++ b/include/cudnn_frontend/graph_interface.h
@@ -1676,6 +1676,11 @@ class Graph : public ICudnn, public INode {
         cached_pass_by_value           = deserialized_pass_by_value;
         cached_workspace_modifications = deserialized_workspace_modifications;
 
+        // Reset prep state in case this Graph is being re-deserialized; otherwise the
+        // eager prep below would early-return with the old slot layout.
+        varpack_prep_state->prepared.store(false, std::memory_order_release);
+        varpack_template = {};
+
         // Eager prep, matching what build_plans() does for fresh-build graphs.
         CHECK_CUDNN_FRONTEND_ERROR(prepare_variant_pack_template());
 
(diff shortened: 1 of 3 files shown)


Anerudhan:
@cudnn-ci-bot run

include/cudnn_frontend/graph_interface.h
```diff
@@ -1676,6 +1676,11 @@ class Graph : public ICudnn, public INode {
         cached_pass_by_value           = deserialized_pass_by_value;
         cached_workspace_modifications = deserialized_workspace_modifications;
 
+        // Reset prep state in case this Graph is being re-deserialized; otherwise the
+        // eager prep below would early-return with the old slot layout.
+        varpack_prep_state->prepared.store(false, std::memory_order_release);
+        varpack_template = {};
+
```
hwanseoc, line 1691: I think the re-deserialize fix is still incomplete.

For this Graph::deserialize(), the lines 1679-1682 here only clears varpack_prep_state->prepared and varpack_template.

But this does not clear these deserialize owned containers:

- deserialized_tensor_properties: lines 1631-1637 insert new tensors
- deserialized_pass_by_value: lines 1648-1655 assign new entries but keep old unmatched keys
- deserialized_workspace_modifications: lines 1658-1668 assign new entries but keep old unmatched keys
- tensors_to_dump: lines 1687-1697 emplace_back new entries

Then lines 1675-1677 copy deserialized_pass_by_value / deserialized_workspace_modifications into cached_pass_by_value / cached_workspace_modifications, so stale entries from a previous deserialize(handle, old_data) can still feed prepare_variant_pack_template() at line 1685.

Can we clear those deserialize-owned containers near line 1625, right after json j = json::from_ubjson(data), before reading the new JSON?

hwanseoc (changes requested):
See https://github.com/NVIDIA/cudnn-frontend/pull/248/changes#r3278873614

hwanseoc:
@cudnn-ci-bot status

hwanseoc:
@cudnn-ci-bot run

include/cudnn_frontend/graph_interface.h
Anerudhan: Pushed 4f3546c0 which clears `deserialized_tensor_properties`, `deserialized_pass_by_value`, `deserialized_workspace_modifications`, and `tensors_to_dump` right after `json::from_ubjson(data)` (before any of the deserialize logic that repopulates them), so a re-deserialize(handle, …) on the same Graph can't feed prepare_variant_pack_template() with stale entries from the prior call.

Anerudhan:
@cudnn-ci-bot run

hwanseoc (approved):
LGTM, let's merge when both branches are in sync

PR #251, opened 2026-05-26
Add row-scale support to grouped GEMM quant

zianglih:
## Summary
@humansand

Add an optional `row_scale_tensor` input to the SM100 grouped GEMM quant path. The new tensor is a contiguous FP32 vector with one scale per valid output row. When provided, the epilogue multiplies FP32 accumulators by `alpha[expert] * row_scale[m]` before output conversion.

This is intended for integrations that need an additional per-row accumulator scale, such as row-scaled NVFP4. The immediate motivation is TransformerEngine's row-scaled NVFP4 work in NVIDIA/TransformerEngine#2931. The kernel treats `row_scale_tensor` as an epilogue multiplier and does not interpret how the caller produced those values.

The existing bias and probability semantics are preserved: the row scale applies only to the accumulator path, while fused bias keeps the existing `bias * prob` behavior.

## Changes

- Add `sample_row_scale` / `row_scale_tensor` plumbing to `GroupedGemmQuantSm100`.
- Add scheduler tensor mapping for the row-scale input.
- Apply row-scale multiplication in the grouped GEMM quant epilogue.
- Include the row-scale tensor in wrapper validation and compile-cache signatures.
- Extend grouped GEMM quant tests and Python reference coverage for compile/execute, dense wrapper, and discrete wrapper paths.

## Validation

Current upstream-facing branch:

- `python3 -m py_compile python/cudnn/grouped_gemm/grouped_gemm_quant/api.py python/cudnn/grouped_gemm/grouped_gemm_quant/grouped_gemm_quant.py python/cudnn/grouped_gemm/moe_sched_extension.py test/python/fe_api/test_grouped_gemm_quant.py test/python/fe_api/test_grouped_gemm_quant_utils.py`
- `git diff --check upstream/develop...HEAD`
- B200 devbox: `python3 -m pytest -q test/python/fe_api/test_grouped_gemm_quant.py -k row_scale --tb=short` -> `5 passed, 309 deselected`

Notes:

- `pre-commit run --all-files` is unavailable in this checkout because `.pre-commit-config.yaml` is not present.

### DIFF
--- a/python/cudnn/grouped_gemm/grouped_gemm_quant/api.py
+++ b/python/cudnn/grouped_gemm/grouped_gemm_quant/api.py
@@ -90,6 +90,7 @@ class GroupedGemmQuantSm100(APIBase):
         sample_amax: Optional[torch.Tensor] = None,
         sample_norm_const: Optional[torch.Tensor] = None,
         sample_prob: Optional[torch.Tensor] = None,
+        sample_row_scale: Optional[torch.Tensor] = None,
         # Configuration
         acc_dtype: torch.dtype = torch.float32,
         mma_tiler_mn: Tuple[int, int] = (256, 256),
@@ -121,6 +122,9 @@ class GroupedGemmQuantSm100(APIBase):
         :param sample_amax: Optional amax tensor for quantization
         :param sample_norm_const: Optional normalization constant
         :param sample_prob: Optional probability tensor for gating
+        :param sample_row_scale: Optional 1-D FP32 row-scale tensor. When
+            provided, the epilogue scales GEMM accumulators by
+            ``alpha[expert] * row_scale[m]`` before output conversion.
         :param acc_dtype: Accumulator data type
         :param mma_tiler_mn: MMA tiler shape (M, N)
         :param cluster_shape_mn: Cluster shape (M, N)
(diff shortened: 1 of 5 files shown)


Anerudhan:
@cudnn-ci-bot run

Anerudhan:
Thanks for the PR @zianglih, I have launched an internal review. I will let you know on its progress.

Anerudhan:
@cudnn-ci-bot run

python/cudnn/grouped_gemm/grouped_gemm_quant/api.py
```diff
@@ -886,6 +924,7 @@ def _compile_discrete(self, gemm_quant, max_active_clusters, fake_stream) -> Non
             stride=self.prob_desc.stride,
             assumed_align=16,
         )
+        row_scale_tensor = self._make_fake_cute_tensor_from_desc(self.row_scale_desc, assumed_align=16)
```
saltyminty, line 927: I believe the fake tensor needs to also be constructed with symbolic `valid_m` as the 0th dimension, similar to prob_tensor above.

saltyminty (approved):
Approved but please see above comment about dynamic shapes.

zianglih:
Follow-up from validating the row-scale integration in NVIDIA/TransformerEngine#3042 on B200:

The `row_scale_tensor` API is present in v1.26.0 and the row-scale wrapper contract is correct, but the MoE scheduler does not compile under the release's pinned `nvidia-cutlass-dsl[cu13]==4.5.0` without explicit `Int32` conversions. Both dynamic and static scheduler attempts hit a DSL type-structure error. The working cuDNN-side changes are explicit `Int32(...)` conversions for:

- scheduler state updates in `_advance_to_expert_containing`
- tile-index initialization and division/modulo results in `_convert_local_idx_to_tile_coords`
- offset loads/subtraction results in `get_expert_offset_and_tokens`

With those conversions applied temporarily, the cuDNN discrete FP4 row-scale control and all 8 focused TE integration cases pass. The TE PR intentionally includes no scheduler workaround or fallback and remains draft pending a cuDNN-side fix.

PR #253, opened 2026-05-26
Expose Cost Model Predicted Times Through Public API

Anerudhan:
* cleanup tests

* add test for uncached predicted time

* cost model api


Anerudhan:
Close pull request as a one-off. Branch still exists.

PR #257, opened 2026-05-28
Bypass cuteDSL d=256 path on cuDNN 9.23+

vedaanta:
cuDNN 9.23.0 added native d=256 SDPA fprop and bprop support in the graph backend, so the OSS (cuteDSL) kernels at
`cudnn.experimental.ops.sdpa` are no longer required when the linked backend is recent enough.

Add `_cudnn_supports_native_d256()` gated on
`cudnn.backend_version() >= 92300` and require it to be `False` before routing fprop/bprop through the SM100 OSS wrappers. The pre-existing SM100+ device check is kept so older cuDNN versions still light up the OSS path on Blackwell.

The `test_d256_uses_oss_forward_path` test now skips on cuDNN 9.23+ since the OSS bypass is intentional, and a new
`test_d256_uses_graph_path_on_cudnn_9_23_plus` asserts that fprop/bprop populate the cuDNN graph cache (proving the OSS path is bypassed).

Also: `_skip_if_unsupported_d256` and `test_d256_uses_oss_forward_path` used `import cudnn.sdpa` inside the function body, which made `cudnn` a local variable and shadowed the module-level import as soon as any earlier line referenced `cudnn` (e.g. the new `cudnn.backend_version()` check). Switch to `importlib.import_module("cudnn.sdpa")` to avoid the binding.


python/cudnn/experimental/ops/sdpa.py
```diff
@@ -89,6 +89,16 @@ def _device_supports_d256_oss(device: torch.device) -> bool:
     return major * 10 + minor >= 100
 
 
+# cuDNN 9.23.0 added native d=256 SDPA fprop and bprop support in the graph
+# backend. From this version on the OSS (cuteDSL) kernels are no longer needed.
+_CUDNN_NATIVE_D256_VERSION = 92300
```
Anerudhan: Can we call it cudnn_backend instead of cudnn native?

test/python/test_cudnn_sdpa_op.py
```diff
@@ -188,9 +197,13 @@ def test_basic(self, D):
 
     @pytest.mark.L0
     def test_d256_uses_oss_forward_path(self):
+        if cudnn.backend_version() >= _CUDNN_NATIVE_D256_VERSION:
```
Anerudhan: Can we make OSS backend opt-in ? Currently there is not way to select this

test/python/test_cudnn_sdpa_op.py
```diff
@@ -213,6 +226,34 @@ def test_d256_uses_oss_forward_path(self):
...
+
+        B, H, S, D = 2, 8, 128, 256
+        _fprop_cache.clear()
+        _bprop_cache.clear()
+
+        q = torch.randn(B, H, S, D, dtype=torch.float16, device="cuda", requires_grad=True)
+        k = torch.randn(B, H, S, D, dtype=torch.float16, device="cuda", requires_grad=True)
+        v = torch.randn(B, H, S, D, dtype=torch.float16, device="cuda", requires_grad=True)
+
+        o = scaled_dot_product_attention(q, k, v)
```
Anerudhan, line 258: How are you verifying it chose the cudnn backend and not OSS backend?

Anerudhan (changes requested):

Anerudhan:
Should we auto direct user to cudnn backend (or) allow user to choose?

Anerudhan:
@cudnn-ci-bot run

vedaanta:
closing for https://github.com/NVIDIA/cudnn-frontend/pull/272 instead

PR #259, opened 2026-05-29
feat(python): add get_engine_and_knobs_at_index for structured plan pinning

YangXu1990uiuc:
## What

Adds `pygraph.get_engine_and_knobs_at_index(index) -> (engine_id, {knob_type: value})` — the **structured counterpart of `get_plan_name_at_index`**.

`get_plan_name_at_index` returns a formatted `engN_kT=V` tag built (in `detail::get_engine_tag`) from the engine global index and knob choices. Tools that want to **persist a tuned plan and replay it later** are currently forced to either store the bare plan *index* (which drifts when the `policy=ALL` plan list is re-enumerated across frontend/backend versions or after `deselect_engines`) or parse the tag string.

This exposes the same data structurally, reading the identical backend attributes (`CUDNN_ATTR_ENGINE_GLOBAL_INDEX`, `CUDNN_ATTR_ENGINECFG_KNOB_CHOICES`, `CUDNN_ATTR_KNOB_CHOICE_KNOB_TYPE/VALUE`) that `get_engine_tag` stringifies. The result feeds straight into `create_execution_plan(engine_id, knobs)` to rebuild the exact same kernel on a fresh graph **without a heuristics query**.

## Changes

- `detail::get_engine_id_and_knobs` (`cudnn_frontend_utils.h`): structured reader (mirrors `get_engine_tag`)
- `Execution_plan_list::get_engine_and_knobs_at_index` (`plans.h`): converts backend knob types → `KnobType_t` via the existing `convert_from_backend_knob_type`
- `Graph::get_engine_and_knobs_at_index` (`graph_interface.h`)
- `PyGraph` binding (`pygraph.h` / `pygraph.cpp`)

Purely additive read-only API; no existing behavior changes.

## Validation

Builds clean. Verified end-to-end on an A100 (cuDNN backend 9.30, frontend built from this branch):
enumerate a `policy=ALL` matmul graph → for plan *i* read `(engine_id, knobs)` → `create_execution_plan(engine_id, knobs)` + `build_plans()` on a fresh graph yields exactly that plan (`get_plan_name_at_index(0)` of the pinned graph == `get_plan_name_at_index(i)` of the original). Round-trip matches.

## Motivation

Lets downstream autotuners (e.g. FlashInfer) store a **version-robust, self-describing** cuDNN tactic and pin it on replay without re-running heuristics, instead of a positional plan index that can silently point at a different engine after a library update.

### DIFF
--- a/include/cudnn_frontend/plans.h
+++ b/include/cudnn_frontend/plans.h
@@ -2,8 +2,11 @@
 
 #include <optional>
 #include <string>
+#include <unordered_map>
+#include <utility>
 #include <vector>
 
+#include "knobs.h"
 #include "../cudnn_frontend_EngineConfig.h"
 #include "../cudnn_frontend_Logging.h"
 #include "graph_helpers.h"
@@ -474,6 +477,29 @@ class Execution_plan_list {
     error_t
     get_name_at_index(int64_t index, std::string& name) const {
         name = detail::get_engine_tag(engine_configs[index]);
+        return {error_code_t::OK, ""};
+    }
+
+    // Structured counterpart of get_name_at_index(): returns the engine global
+    // index and the (KnobType_t, value) choices for the plan at `index`.
+    // The returned pair can be fed straight back into
+    // Graph::create_execution_plan(engine_id, knobs) to rebuild the exact same
+    // kernel without a heuristics query -- and, unlike a positional plan index,
+    // it survives plan-list re-enumeration across cudnn-frontend versions.
+    error_t
+    get_engine_and_knobs_at_index(int64_t index,
+                                  int64_t& engine_id,
+                                  std::unordered_map<KnobType_t, int64_t>& knobs) const {
+        std::vector<std::pair<cudnnBackendKnobType_t, int64_t>> backend_knobs;
+        auto status = detail::get_engine_id_and_knobs(engine_configs[index], engine_id, backend_knobs);
+        RETURN_CUDNN_FRONTEND_ERROR_IF(
+            status != CUDNN_STATUS_SUCCESS,
+            error_code_t::CUDNN_BACKEND_API_FAILED,
+            "Failed to query engine id / knob choices for plan at index " + std::to_string(index));
+        knobs.clear();
+        for (auto const& backend_knob : backend_knobs) {
+            knobs[convert_from_backend_knob_type(backend_knob.first)] = backend_knob.second;
+        }
         return {error_code_t::OK, ""};
     }
 
--- a/include/cudnn_frontend/plans.h
+++ b/include/cudnn_frontend/plans.h
@@ -480,16 +480,15 @@ class Execution_plan_list {
         return {error_code_t::OK, ""};
     }
 
-    // Structured counterpart of get_name_at_index(): returns the engine global
-    // index and the (KnobType_t, value) choices for the plan at `index`.
-    // The returned pair can be fed straight back into
-    // Graph::create_execution_plan(engine_id, knobs) to rebuild the exact same
-    // kernel without a heuristics query -- and, unlike a positional plan index,
-    // it survives plan-list re-enumeration across cudnn-frontend versions.
+    // Structured counterpart of get_name_at_index(): engine index + knob choices,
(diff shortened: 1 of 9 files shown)


Anerudhan:
@cudnn-ci-bot run

include/cudnn_frontend/plans.h
```diff
@@ -477,6 +480,29 @@ class Execution_plan_list {
...
+    // The returned pair can be fed straight back into
+    // Graph::create_execution_plan(engine_id, knobs) to rebuild the exact same
+    // kernel without a heuristics query -- and, unlike a positional plan index,
+    // it survives plan-list re-enumeration across cudnn-frontend versions.
+    error_t
+    get_engine_and_knobs_at_index(int64_t index,
+                                  int64_t& engine_id,
+                                  std::unordered_map<KnobType_t, int64_t>& knobs) const {
+        std::vector<std::pair<cudnnBackendKnobType_t, int64_t>> backend_knobs;
+        auto status = detail::get_engine_id_and_knobs(engine_configs[index], engine_id, backend_knobs);
```
Anerudhan, line 493: There is a missing check for index < engine_configs.size()

Anerudhan (commented):
Missing tests. (Preferably add it to the tests/cpp section (unit tests).

include/cudnn_frontend/plans.h
YangXu1990uiuc: Done in e86eb985e — added an out-of-range check (mirrors `check_support_at_index`) returning `GRAPH_EXECUTION_FAILED` before indexing `engine_configs`.

YangXu1990uiuc:
Thanks for the review! Addressed both in e86eb985e:

- **Bounds check**: `get_engine_and_knobs_at_index` now rejects out-of-range `index` (mirrors `check_support_at_index`) instead of indexing `engine_configs` out of bounds.
- **Unit test**: added `test/cpp/get_engine_and_knobs.cpp` (registered in `test/cpp/CMakeLists.txt`). It builds a matmul graph, reads `(engine_id, knobs)` for every plan, and confirms re-pinning via `create_execution_plan` reproduces the same plan (matching `get_plan_name_at_index`); also checks out-of-range indices error. Verified on an SM100 GPU (120 assertions pass).

Note: the re-pin round-trip is best-effort with a `>= 1` floor — not every heuristic-enumerated engine is standalone-constructable via `create_execution_plan`, but when one is, it reproduces the exact same kernel.

Also trimmed the new doc comments to match the surrounding style.

YangXu1990uiuc:
Follow-up (c5d04e646): I dug into *why* some plans didn't round-trip earlier and it turned out to be a real gap worth fixing here, not an engine limitation.

`KnobType_t` (and `convert_to/from_backend_knob_type`) stopped at `WARP_SPEC_CFG` (42). Engines using `SWAP_AB` (43, @9.18), `INPUT_TMA_ENABLE` (44) or `OUTPUT_TMA_ENABLE` (45, @9.22) had those knobs silently mapped to `NOT_SET` by `convert_from_backend_knob_type`; feeding `NOT_SET` back into `create_execution_plan` then failed `convert_to_backend_knob_type` with `INVALID_VALUE`. So a plan carrying one of these knobs couldn't be pinned.

Added the three knob types to the enum + both converters (version-gated to the backend `@since`) + the pybind `knob_type` enum. On an SM100 box (backend 9.30), `create_execution_plan` now succeeds for **all** enumerated matmul plans (15/15), up from 12/15.

I also switched the test to compare the **structured identity** (engine id + knob map) rather than the plan-name tag: the tag serializes knobs in engine-config order, which differs between the heuristic config and the pinned one even though the kernel is identical (e.g. `eng11_k2=29_k27=0...` vs `eng11_k43=0_k38=0...k2=29`). `create_execution_plan` is now asserted for every plan; *building* the pinned plan stays best-effort since it can fail for unrelated environment reasons (e.g. a ptxas older than the engine's target).

Anerudhan (approved):

Anerudhan:
@cudnn-ci-bot run

PR #268, opened 2026-06-02
Add linear attention API

jhjpark:
## Summary by CodeRabbit

* **New Features**
  * Added the `gated_delta_net` linear-attention operator to the experimental ops package public namespace.
  * Supports optional `initial_state` and can return a `final_state` when requested.
  * Added CUDA-only test coverage for forward correctness, gradient behavior, default-vs-explicit scaling, empty final-state handling, and `torch.compile` parity.
* **Bug Fixes**
  * Improved shape/dtype handling for linear-attention inputs and recurrent state outputs.


Anerudhan:
Can you add documentation? PR description ..

Also, seems like we are building the op out of pure Pytorch ops.. Is that the expectation?

PR #270, opened 2026-06-02
fix: allow override libcudart via CUDNN_FRONTEND_CUDART_LIB_NAME

### LINKED ISSUE #267: cuDNN frontend tries to load the libcudart shared library and detects multiple versions in the system path
- Is there any way to work around this limitation perhaps by specifying the library it should load with an environment variable?

- Would it be possible to optionally specify the cudart version cuDNN frontend should use and override this check via an environment variable setting?

Use case:
This occurs on a Kubernetes cluster (GKE) when running a MaxText/TransformerEngine multi-node workload, where a user-container ships with libcudart.so.13, while GKE’s NCCL net plugin (TCPXO) has a dependency on its own shared library libcudart.so.12 which is installed on the host (and mounted into the container).

Anerudhan:
CUDNN_FRONTEND_CUDART_LIB_NAME

When dynamic loading is enabled, load_cudart_so() searches for the supported libcudart major versions and aborts with "Multiple libcudart libraries found" when more than one is visible on the library search path. This happens in containerized environments such as GKE, where the TCPXO NCCL plugin mounts a different libcudart major version from the host than the one shipped in the container.

Check the CUDNN_FRONTEND_CUDART_LIB_NAME environment variable first; when set to a library name or path, dlopen exactly that library and skip the automatic multi-version detection. Behavior is unchanged when the variable is unset.

Fixes #267

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,24 @@ export CUDNN_FRONTEND_LOG_FILE=execution_log.txt
 
 Alternatively, you can control logging programmatically via `cudnn_frontend::isLoggingEnabled()`.
 
+### Overriding the CUDA runtime library
+
+When the frontend is built with dynamic loading enabled, it locates the CUDA runtime
+(`libcudart.so.*`) at runtime by searching for the supported major versions. In some
+environments (for example, containers such as GKE where the TCPXO NCCL plugin mounts a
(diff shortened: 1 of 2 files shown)


Anerudhan:
@cudnn-ci-bot run

PR #274, opened 2026-06-03
fix(windows): consolidate getenv access and fix C4996/C4005 on MSVC

### LINKED ISSUE #139: MSVC warnings on macro redefinition
In `cudnn_frontend_Logging.h`, there's this block of code:

```c++
#ifdef WIN32
#pragma warning(disable : 4996)
#define _CRT_SECURE_NO_WARNINGS
#endif
```

When the library in included into a project that already has `_CRT_SECURE_NO_WARNINGS` defined, that produces warnings:
```
[build] ...\cudnn-frontend\include\cudnn_frontend_Logging.h(35): warning C4005: '_CRT_SECURE_NO_WARNINGS': macro redefinition
[build] ...\cudnn-frontend\include\cudnn_frontend_Logging.h(35): note: '_CRT_SECURE_NO_WARNINGS' previously declared on the command line
```

Please wrap it into `#ifndef _CRT_SECURE_NO_WARNINGS` to avoid that.

Also, it would be nice to use `#pragma warning(push)` and `#pragma warning(pop)` later to avoid changing the warning settings for the code including cudnn_frontend.

Anerudhan:
The Windows wheel build fails because the `std::getenv` call added to load_cudart_so() in cudnn_frontend_shim.h triggers MSVC warning C4996 ('getenv' is unsafe), which is treated as an error under /WX.

Root cause and fixes:
- Move get_environment() to cudnn_frontend_shim.h (the lowest-level header, included by utils.h before Logging.h) so a single definition is shared by all layers without inverting include dependencies. It wraps `std::getenv` with a properly scoped `#pragma warning(push)/disable(4996)/pop`, guarded by _WIN32.
- Route all getenv call sites through get_environment(): shim.h, graph_properties.h, scaled_dot_product_flash_attention.h, and sm100_rms_norm_silu_engine.h. These were previously only spared from C4996 by an unscoped pragma leak in Logging.h, and would have started failing once that leak was fixed.
- Remove the duplicate get_environment() from cudnn_frontend_Logging.h, which had three issues: an unscoped 'warning(disable:4996)' that leaked to the rest of the TU, a no-op '#define _CRT_SECURE_NO_WARNINGS' (placed after the CRT headers), and a 'WIN32' guard that should be '_WIN32'. Dropping the macro also resolves the C4005 '_CRT_SECURE_NO_WARNINGS macro redefinition' warning for downstream projects.

Fixes NVIDIA/cudnn-frontend#139

## Summary by CodeRabbit

* **Refactor**
  * Standardized environment-variable access across the library for improved consistency and cross-platform behavior.
  * Reworked logging and dynamic library load handling to be more robust: reduces risky failures by warning and continuing when multiple candidates are found, improving runtime resilience.

### DIFF
--- a/include/cudnn_frontend/experimental/sm100_rms_norm_silu_engine.h
+++ b/include/cudnn_frontend/experimental/sm100_rms_norm_silu_engine.h
@@ -310,9 +310,9 @@ class Sm100RmsNormSiluEngine : public IOssNormEngine {
 
         // Add CUDA Toolkit include paths for NVRTC to resolve #include <cuda_bf16.h> etc.
         std::string cuda_include = "/usr/local/cuda/include";
-        if (auto env0 = std::getenv("CUDA_HOME")) {
+        if (auto env0 = get_environment("CUDA_HOME")) {
             cuda_include = std::string(env0) + "/include";
-        } else if (auto env1 = std::getenv("CUDA_PATH")) {
+        } else if (auto env1 = get_environment("CUDA_PATH")) {
             cuda_include = std::string(env1) + "/include";
         }
         flags.push_back("--include-path=" + cuda_include);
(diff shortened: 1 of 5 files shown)


Anerudhan:
@cudnn-ci-bot run

PR #277, opened 2026-06-04
bench: add autoregressive video DiT SDPA config + GB200/GB300 results

vedaanta:
Adds a new benchmark config for the autoregressive (world-model / next-frame) video DiT shape: short query (one new frame, s_q ∈ {985, 1024, 2048, 4096, 8192}) attending a long cached KV history (s_kv=62208) with h=9, d=128 and no operator-level mask. This is a class of workload that prior DiT configs (LTX-2, Wan 2.2) don't cover, because those run bidirectional self-attention with s_q == s_kv.

Captured on GB200 and GB300 (cuDNN 9.24.0, FAv4 from the CuTe-DSL build). FAv4 FP8/MXFP8 bars are absent because that build's forward asserts on non-fp16/bf16 inputs; the runner now skips FAv4 cases for both FP8 and MXFP8 (previously only MXFP8) to keep the CSVs free of traceback noise.


Anerudhan (commented):
- Please correct the cudnn version in the images.

Anerudhan (approved):

PR #285, opened 2026-06-05
Skip TensorIR MemBound / compile-time-const samples on consumer Blackwell (SM12x)

YangXu1990uiuc:
## What

On consumer Blackwell (SM120 / RTX 50-series, and SM121 / GB10), 8 C++
sample test cases **FAIL** (not skip) with `No valid engine configs
returned from heuristics`:

| File | Test case(s) |
|---|---|
| `membound/transpose.cpp` | Membound transpose permutes dims |
| `membound/reshape.cpp` | Membound reshape … LOGICAL mode |
| `membound/slice.cpp` | Membound slice window with step |
| `membound/concat.cpp` | Membound concatenate on channel axis |
| `membound/membound_fusion.cpp` | Fusion reshape then ReLU / Fusion transpose then add bias |
| `membound/boolean_fusion.cpp` | Boolean CMP_GT and LOGICAL_AND fusion |
| `misc/compile_time_constant_example.cpp` | Compile-time constant scalar multiply and add |

## Root cause

These samples gate on `check_device_arch_newer_than("blackwell")` /
`is_blackwell_arch()`, both of which are **true for cc ≥ 100** — so SM120
consumer Blackwell falls through. But the TensorIR MemBound engine
(`cudnnTensorIrMemBoundEngine`) only supports **SM100–SM109** (data center
Blackwell): its arch gate is `[SM_100, SM_110)` and the DKG cubins it emits
are the `sm_100f` family-portable target, which the CUDA driver will not
load on `sm_120`. No engine serves the graph (the kernelgen runtime-fusion
fallback only targets SM70/80/90), so `create_execution_plans()` returns no
configs and the `REQUIRE` fails.

## Change

Narrow the guard to the existing `is_blackwell_computing_arch()` helper
(`100 ≤ cc < 110`), matching the backend engine's actual support range:

```cpp
-    if (!check_device_arch_newer_than("blackwell")) {
-        SKIP("TEST requires device blackwell or newer");
+    if (!is_blackwell_computing_arch()) {
+        SKIP("TensorIR MemBound engine is only supported on SM100-SM109 (data center Blackwell)");
```

Behavior change: SM100–109 unchanged (still run), Hopper/Ampere unchanged
(still skip); only SM120+ goes from **fail → clean skip**.

## Verification

Built and re-ran on RTX 5080 (`sm_120`, cuDNN 9.30): all 8 cases now SKIP
cleanly. 7 files / 16 lines, clang-formatted.

This is the same class of fix as #283 (flexible-graph SDPA backward skipped
on SM120+).

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized GPU gating across samples and a notebook to require supported data‑center Blackwell hardware (SM100–SM109) and updated skip/notification text. This ensures MemBound, cuDNN, and related examples only run on compatible GPUs, preventing unsupported executions and reducing misleading test results.

### DIFF
--- a/samples/cpp/membound/boolean_fusion.cpp
+++ b/samples/cpp/membound/boolean_fusion.cpp
@@ -39,8 +39,8 @@ TEST_CASE("Boolean CMP_GT and LOGICAL_AND fusion", "[membound][boolean][pointwis
     if (cudnn_frontend::detail::get_backend_version() < 92200) {
         SKIP("Boolean fusion sample requires cuDNN backend 9.22.0 or newer at runtime.");
     }
-    if (!is_blackwell_arch()) {
-        SKIP("Boolean fusion requires Blackwell (SM100+) architecture.");
+    if (!is_blackwell_computing_arch()) {
+        SKIP("Boolean fusion (TensorIR MemBound engine) is only supported on SM100-SM109 (data center Blackwell)");
     }
 
     constexpr int64_t d0 = 4, d1 = 8, d2 = 16;
--- a/samples/cpp/membound/concat.cpp
+++ b/samples/cpp/membound/concat.cpp
@@ -37,8 +37,8 @@ TEST_CASE("Membound concatenate on channel axis (no in-place index)", "[membound
     return;
 #endif
 
-    if (!check_device_arch_newer_than("blackwell")) {
-        SKIP("TEST requires device blackwell or newer");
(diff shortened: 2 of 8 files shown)


vedaanta (approved):

Anerudhan:
@cudnn-ci-bot run

Anerudhan (approved):

PR #297, opened 2026-06-10
Align DSA indexer kernels and fix dense score-grad clipping

jiayus-nvidia:
## Summary

  - Align SM90/SM100 indexer backward launch grids to use `(seqlen, batch)` ordering.
  - Add `topk_indices_global` plumbing so backward can handle local per-batch top-k indices by default.
  - Rework SM90 indexer forward to write reduced logits directly to global memory, removing the separate score-store path.
  - Fix dense score-grad clipping to use log-domain mask checks, and apply the same behavior to SM90 dense score-grad.
  - Keep SM90 dense GEMM compile caching independent of runtime `seqlen_k`.
  - Update the dense DSA reference to backprop through logits with an explicit grad signal matching clipped-log KL behavior.

## Summary by CodeRabbit

* **New Features**
  * Added `topk_indices_global` parameter for flexible top-k indexing control
  * Introduced `clean_logits` option for improved masking configuration

* **Performance Optimizations**
  * Optimized kernel grid scheduling for better memory efficiency
  * Replaced torch-based computation with specialized kernel implementation
  * Streamlined forward indexer pipeline

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_backward/api.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_backward/api.py
@@ -147,6 +147,7 @@ class IndexerBackward(APIBase):
         sample_topk_indices: torch.Tensor,  # (B, S_q, topk) INT32
         sm_scale: float = 1.0,
         block_I: int = 128,
+        topk_indices_global: bool = False,
     ):
         super().__init__()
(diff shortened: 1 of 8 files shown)


Anerudhan:
@cudnn-ci-bot run

saltyminty (approved):

PR #298, opened 2026-06-10
DSA bwd SM100 fix dropped dKV gradients when topk width > total_S_kv

Jie-Fang:
(global_row_idx) against max_seqlen_kv. A column position >= total_S_kv is not invalid -- with a non-compact topk_idxs layout (-1 sentinels, width > total_S_kv) valid indices can sit at any column. Entries past column total_S_kv were silently treated as -1 and their dKV contributions dropped, while dQ (whose load path correctly judges validity by the index value) stayed correct. With a [window | compressed] layout this zeroes the entire original-KV region of dkv bit-exactly. Drop the position-vs-seqlen comparison; the < topk bound plus the topk_idx >= 0 sentinel check in the store helpers already match the load-side and FlashMLA-forward semantics. Remove the now-unused max_seqlen_kv parameter from reduce_dKV.
Also fix the test reference _make_topk_mask: without topk_length it clamped -1 sentinels to index 0, spuriously marking KV row 0 as attended, which corrupted out/lse/gradient references for non-compact inputs.
Verified on B200: topk width 1024 > S_kv 256 now gives cos_sim(dkv) 0.9996 (was 0.498); wide non-compact layouts pass FP32 autograd checks; fe_api/dsa pytest suite passes (16 tests). Co-Authored-By: Claude Fable 5 noreply@anthropic.com

## Summary by CodeRabbit

* **Refactor**
  * Optimized sparse attention backward computation to improve kernel efficiency by streamlining index validation logic during gradient reduction operations.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100.py
@@ -1078,7 +1078,6 @@ class FlashAttentionDSABackwardSm100:
                 mdKV_acc,
                 mTopkIdxs,
                 sTopkIdxs,
-                max_seqlen_kv,
                 tile_count,
                 topk,
                 mma_reduce_dKV_pipeline,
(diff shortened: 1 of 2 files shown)


Anerudhan:
@cudnn-ci-bot run

saltyminty (approved):

PR #312, opened 2026-06-16
fix: IMA on indexer_topk_wrapper

Hyaloid:
cute_dsl lowers the per-block slice buffer = extra_buffer[bidx, ...] into a 32-bit-by-32-bit -> 32-bit multiply followed by a sign-extending widening accumulate. This pr chunk the launch in cute_dsl_topk_wrapper along the row dimension so that within each chunk bidx < chunk_rows and `(chunk_rows − 1) * buffer_numbers * num_cols < 2**31`.

When run python repro_indexer_topk_ima.py, an IMA error does not exist.

```python
# repro_indexer_topk_ima.py
import torch
from cudnn import DSA

S, S_kv, top_k, next_n = 4096, 262209, 2048, 1
device = torch.device("cuda")

g = torch.Generator(device=device).manual_seed(0)
logits   = torch.randn((S, S_kv), dtype=torch.float32, device=device, generator=g)
seq_lens = torch.full((S // next_n,), S_kv, dtype=torch.int32, device=device)

result = DSA.indexer_top_k_wrapper(
    logits, seq_lens,
    top_k=top_k, next_n=next_n, return_val=False,
)
torch.cuda.synchronize()  # raises cudaErrorIllegalAddress
print(result["indices"].sum().item())
```
No IMA occurs. And the output looks good.
When use compute-sanitizer to double check, the output show 0 errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved variable-length sequence handling for sparse attention decoding by adjusting how work is partitioned and buffered, especially for larger workloads.
  * Enhanced extra-buffer access and pointer alignment behavior to improve correctness and efficiency when specific memory-store modes are enabled.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_decode_varlen.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_decode_varlen.py
@@ -14,6 +14,8 @@
 
 """SM90+ CuTe DSL indexer top-K decode kernel."""
 
+import math
+
 import cuda.bindings.driver as cuda
 import cutlass
 import cutlass.cute as cute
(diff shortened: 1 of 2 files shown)


Anerudhan:
Thanks for the PR @Hyaloid . Will review and get back to you soon.

Anerudhan:
@cudnn-ci-bot run

limin2021:
Great thanks for reporting this bug. The root cause for this bug is the addr offset for extra buffer out of int32. Insdead of splitting the rows into multi row-chunks and running each row-chunk in one kernel, I prefer to change the addr offset to int64, which solve the root cause. 

You could refer to or just apply the fix here: [buffer-addr-ima-fix](https://github.com/NVIDIA/cudnn-frontend/compare/develop...limin2021:fix/dsa-indexer-topk-offset64?expand=1)

Anerudhan:
Thanks @limin2021 . @Hyaloid can you update your PR to reflect this  ?

Hyaloid:
Hi @limin2021, @Anerudhan,

I agree with @limin2021's point. The root cause here is indeed that the offset can exceed what int32 can represent. And I  have already updated the pr. 

The change in _slice_row_64bit (together with the explicit cutlass.Int64(...) casts on bidx, grid_dim_y, and bidy) promotes the offset computation from int32 to int64, so expressions like bidx * stride[0] no longer overflow inside the kernel.

That said, I don't think the chunking path is redundant. It solves a different problem and is still necessary.

The key issue is memory usage. extra_buffer has shape (num_rows, buffer_numbers, num_cols) and uses fp32 storage. For one of our representative cases:

- S = 16384
- S_kv = 1M
- dtype = fp32
- buffer_numbers = 2

the scratch buffer size is: 16384 × 2 × 1048576 × 4B ≈ 128 GiB

A B200 has 180 GiB of HBM. After accounting for the fp32 logits (16384 × 1M × 4B ≈ 64 GiB) and normal framework overhead, we typically see only ~113 GiB free at the call site. In practice, a single contiguous 128 GiB torch.empty(...) for extra_buffer fails with OOM before the kernel is even launched.

So even with the int64 fix, this workload would still not run — we'd simply replace the indexing issue with an allocation failure.

Chunking along num_rows addresses the memory side of the problem. By enforcing:

chunk_rows * buffer_numbers * num_cols <= INT32_MAX

each chunk allocation is bounded to roughly 8 GiB, which fits comfortably alongside the 64 GiB logits on a B200. As a side effect, each launch also operates on a smaller logical range, so the common single-CTA indexing path naturally stays within int32 limits.

So my recommendation would be to keep both changes. The int64 conversion fixes the overflow root cause, while chunking is what makes these large shapes actually runnable on hardware such as B200. Removing either one would reintroduce a failure mode for the 16k × 1M workload that we're targeting.

limin2021:
@Hyaloid Hi Hyaloid, 

For the OOM issue, the chunk method is only a workaround.

I could update the kernel impl to remove the extra buffer latter.

We could merge this version first.

Anerudhan:
@cudnn-ci-bot run

Anerudhan (approved):
Approved by the DSA team

PR #319, opened 2026-06-22
Fix sort order in block_scale_quantize.h

dimitar-asenov:
If I compile and run the `samples/cpp/norm/norm_block_scale.cpp` sample with clang in debug mode I get this error:

```
strict_weak_ordering_check.h:50: libc++ Hardening assertion !__comp(*(__first + __a), *(__first + __b)) failed: Your comparator is not a valid strict-weak ordering
```

The comparator indeed violates strict weak ordering. I.e. it in this case it will report that index 0 is smaller than index 1 and also that index 1 is smaller than index 0:

```
X_stride = {10, 10}
X_dim = {1, 1}
```

The fix makes the comparator a strict weak order.

## Release Notes

* **Bug Fixes**
  * Refined stride index sorting logic for block-scale quantization operations with transposed and packed memory layouts.


Anerudhan:
Thank you for the PR @dimitar-asenov . We will run through the CI and get this merged and released.

Anerudhan:
@cudnn-ci-bot run

Anerudhan (approved):

Anerudhan:
LG submitting

PR #329, opened 2026-06-26
Add run_warmup opt-out and reuse-parsed-json overload to Graph::deser…

vincejhan:
Changes to graph_interface.h graph deserialization logic exist to cut avoidable per-entry cost of deserialize path, where every cached cuDNN graph is deserialized up front. The original deserialize(handle, data, enforce_precompiled) forced two unavoidable costs on each entry. This PR makes both opt‑outable, while keeping the default behavior identical.

1. New json-overload — avoid a second from_ubjson parse.
The old function took the serialized blob and called json::from_ubjson(data) internally. But the caller was required to already parse that same blob into a json (it reads cudnn_backend_version / cudnn_frontend_version from it for the per‑entry version‑compatibility check). Calling the old API then re‑parsed the identical buffer a second time.

The blob overload is now a thin wrapper that parses once and delegates:
```
deserialize(handle, std::vector<uint8_t> const& data, bool enforce_precompiled = false, bool run_warmup = true)
    -> deserialize(handle, json::from_ubjson(data), enforce_precompiled, run_warmup);
```
and a new overload accepts the already‑parsed json directly:
`deserialize(handle, json const& j, bool enforce_precompiled = false, bool run_warmup = true);`

Callers that already hold the parsed json (the runtime cache, via TRT_RTC_REUSE_PARSED_JSON) hand it straight in and skip the redundant from_ubjson. The original body moved verbatim into the json overload — no logic changed.

The #ifndef CUDNN_FRONTEND_SKIP_JSON_LIB guard was restructured to match: the blob wrapper now carries the "unavailable when compiled with CUDNN_FRONTEND_SKIP_JSON_LIB" fallback, and the entire json overload lives inside the guard (it can't exist without the json library).

2. run_warmup parameter — skip the throwaway warmup capture.
`deserialize `unconditionally called `warmup(handle)`, which does a throwaway CUDA graph capture (a fake execute over dummy device pointers) purely to prime the plan, plus the deserialized_tensor_properties build that exists only to feed it (and tensors_to_dump). This is pure overhead for eager users who do not need it.

`run_warmup = false `(used by the runtime cache via TRT_RTC_SKIP_WARMUP) skips only the warmup capture and the tensor‑properties build that feeds it. Everything the executable plan depends on is still produced unconditionally — plans.build_plans(...), variant_pack_uids, the pass‑by‑value / workspace caches, and the eager prepare_variant_pack_template() — so the graph is still fully usable.

Safety / compatibility:
- Both new parameters default to their original behavior (run_warmup = true, blob path still parses internally), so every existing caller is unchanged and behaves bit‑for‑bit identically. The opt‑outs are exercised only by the runtime cache's experimental, env‑gated fast path.
- The two side‑effect containers (deserialized_tensor_properties, tensors_to_dump) are left empty when warmup is skipped — documented in the function comment.
- This is an additive parameter change

New test in `test/cpp/serialize.cpp`

Adds `TEST_CASE("Plan deserialize with run_warmup=false still prepares template", "[graph][serialize][deserialize]")`, a sibling to the existing "Plan deserialize prepares variant pack template" case. It builds a small matmul graph, runs it through build → serialize, then deserializes with warmup disabled and asserts the resulting plan state is intact. Three sections:

1. blob overload, run_warmup=false — deserialize succeeds and the variant‑pack template is still populated, proving template prep is independent of warmup.
2. json overload, run_warmup=false — the same assertion via the new pre‑parsed‑json overload, covering the from_ubjson‑avoiding path.
3. equivalence — deserializes the same blob with run_warmup=true and run_warmup=false and asserts both produce the identical variant‑pack template, confirming warmup is purely a priming step with no effect on the executable plan.

## Summary by CodeRabbit

* **New Features**
  * Added an option to skip warmup work during graph deserialization.
  * Deserialization now supports this option for both binary and JSON inputs.

* **Bug Fixes**
  * Improved flexibility when restoring graphs by allowing faster load paths without warmup preparation.

* **Tests**
  * Added coverage for deserializing graphs with warmup disabled, including consistency checks against the default behavior.

### DIFF
--- a/include/cudnn_frontend/graph_interface.h
+++ b/include/cudnn_frontend/graph_interface.h
@@ -1626,11 +1626,24 @@ class Graph : public ICudnn, public INode {
 #endif
     }
 
-    // Parse the blob then delegate. Callers that already parsed it should call the
-    // json overload to skip this second parse.
+    /**
+     * @brief Deserialize an execution plan from a serialized byte blob.
+     *
+     * Parses @p data with from_ubjson and delegates to the json overload. Callers that
+     * have already parsed the blob should call the json overload directly to avoid a
+     * second parse.
+     *
+     * @param handle              cuDNN handle used to rebuild the execution plan.
+     * @param data                UBJSON blob previously produced by serialize().
+     * @param enforce_precompiled When true, fail unless the blob carries a precompiled plan.
+     * @param run_warmup          When false, skip the throwaway warmup capture.
+     * @return error_t OK on success, otherwise an error code describing the failure.
+     */
     error_t
-    deserialize(cudnnHandle_t handle, std::vector<uint8_t> const &data, bool const enforce_precompiled = false,
-                bool run_warmup = true) {
+    deserialize(cudnnHandle_t handle,
+                std::vector<uint8_t> const &data,
+                bool const enforce_precompiled = false,
+                bool run_warmup                = true) {
 #ifndef CUDNN_FRONTEND_SKIP_JSON_LIB
         return deserialize(handle, json::from_ubjson(data), enforce_precompiled, run_warmup);
 #else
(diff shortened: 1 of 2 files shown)


vincejhan:
@cudnn-ci-bot

Anerudhan:
@cudnn-ci-bot run

Anerudhan (changes requested):
Can you please fix the pipeline - clang failure? 

Thanks

vincejhan:
@cudnn-ci-bot run

Anerudhan (approved):

PR #345, opened 2026-07-05
cmake: include GNUInstallDirs before install interface paths

fallintoplace:
## Why
`CMAKE_INSTALL_INCLUDEDIR` could be empty when used in `target_include_directories()` before `GNUInstallDirs` is included, causing invalid installed include paths for package consumers.

## What changed
- Move `include(GNUInstallDirs)` to immediately after `project()` so `CMAKE_INSTALL_LIBDIR`, `CMAKE_INSTALL_BINDIR`, and `CMAKE_INSTALL_INCLUDEDIR` are defined before any usage.

## Validation
No tests were run per request.

## Summary by CodeRabbit

* **Chores**
  * Streamlined the build setup for installation paths, making standard install directories available earlier in the configuration process.
  * Removed a redundant configuration step to keep packaging and export handling cleaner and more consistent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.23)
 
 project(cudnn_frontend VERSION 1.26.0)
 
+# Introduce variables:
+# * CMAKE_INSTALL_LIBDIR
+# * CMAKE_INSTALL_BINDIR
+# * CMAKE_INSTALL_INCLUDEDIR
+include(GNUInstallDirs)
+
 option(CUDNN_FRONTEND_SKIP_JSON_LIB "Defines whether FE should not include nlohmann/json.hpp." OFF)
 option(CUDNN_FRONTEND_BUILD_SAMPLES "Defines if samples are built or not." ON)
 option(CUDNN_FRONTEND_BUILD_TESTS "Defines if unittests are built or not." ON)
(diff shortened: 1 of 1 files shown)


YangXu1990uiuc (approved):

Reviewed `d0937540037c` against base `2a0343f079b0`.

The ordering change makes sense: initialize GNUInstallDirs after project() and before expanding the install-interface variables. No actionable regression found. One nuance for the description: with CMake 3.31 the installed HEADERS file set already supplies the include directory, so the default consumer also works on the baseline; this removes the initialization-order dependency rather than fixing a reproduced default-consumer failure.

Validation: CMake 3.31.4 / GCC 11: configured and installed the exact PR head, current develop and their clean prospective merge using both default and custom include/library directories. All six installed packages could be relocated and found by a downstream CMake consumer, with the expected exported include directories. A downstream translation unit including cudnn_frontend.h compiled against all three default-directory installations after relocation. Fresh-configure export comparison confirms the explicit install-interface include is absent on the baseline and present with this change.

Limitations: Custom INCLUDEDIR compilation fails in all three versions because cudnn_frontend_Errata.h includes ../include/cudnn_frontend_Logging.h. This is pre-existing and independent of this ordering change. No GPU/performance test applies to this configuration-only move. No Actions run exists for the exact head; validation here is the targeted Linux packaging/consumer check, not a Windows build.

PR #346, opened 2026-07-05
cmake: switch python linker flags to target_link_options

fallintoplace:
## Why
`_compiled_module` set `LINK_FLAGS` three times in `set_target_properties`, which means earlier values can be overwritten and final exported flags may miss required linker options.

## What changed
- Replace repeated `LINK_FLAGS` assignments with a single `target_link_options(_compiled_module PRIVATE ...)` call.
- Preserve `LINK_WHAT_YOU_USE` in its existing `set_target_properties(...)` call.

## Verification
`cmake` configure/build in this environment is blocked by missing CUDA (`nvcc` not found).

Please validate with:

```bash
cmake -S . -B build-link-check -G Ninja \\\n  -DCUDNN_FRONTEND_BUILD_PYTHON_BINDINGS=ON \\\n  -DCUDNN_FRONTEND_BUILD_TESTS=OFF \\\n  -DCUDNN_FRONTEND_BUILD_SAMPLES=OFF
cmake --build build-link-check --target _compiled_module --verbose
```

Confirm the verbose link line contains all three flags:
- `-Wl,--no-as-needed`
- `-Wl,--enable-new-dtags`
- `-Wl,-rpath,\x27$ORIGIN\x27,-rpath,\x27$ORIGIN/../lib\x27,-rpath,\x27$ORIGIN/../nvidia/cudnn/lib\x27`

## Summary by CodeRabbit

* **Chores**
  * Updated the Python build configuration to use a more standard linker options setup.
  * Preserved existing build behavior while improving how linker settings are applied during module compilation.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -104,13 +104,17 @@ else()
     add_compile_options(-fvisibility=hidden)
 endif()
 
+target_link_options(
+    _compiled_module
+    PRIVATE
+    -Wl,--no-as-needed
+    -Wl,--enable-new-dtags
+    -Wl,-rpath,'$ORIGIN',-rpath,'$ORIGIN/../lib',-rpath,'$ORIGIN/../nvidia/cudnn/lib'
+)
+
 set_target_properties(
     _compiled_module
-
     PROPERTIES
-    LINK_FLAGS "-Wl,--no-as-needed"
-    LINK_FLAGS "-Wl,--enable-new-dtags"
-    LINK_FLAGS "-Wl,-rpath,'$ORIGIN',-rpath,'$ORIGIN/../lib',-rpath,'$ORIGIN/../nvidia/cudnn/lib'"
     LINK_WHAT_YOU_USE TRUE
 )
 


YangXu1990uiuc:

I found no test CI run for `448c810ab98a` and will request `python_tests` with the next standalone command to exercise the Python build. The focused Make/Ninja check has already exposed a RUNPATH quoting regression, which I am documenting in the review; a build-only pass would not exclude it. For future updates, request this target with `@cudnn-ci-bot run python_tests` in its own comment, or ask a maintainer to send it if your account cannot trigger CI.

YangXu1990uiuc:

YangXu1990uiuc (approved):

Reviewed `448c810ab98a` against base `2a0343f079b0`.

Moving repeated LINK_FLAGS assignments to target_link_options is sensible, but the shell quoting must be adapted to the new API. Please address the reproduced RUNPATH regression below.

**[P2] Remove the shell single quotes when moving RUNPATH into target_link_options** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/448c810ab98aa31fcf3ea5a56757858edf10d13e/python/CMakeLists.txt#L112))

target_link_options escapes each list item itself, so the single quotes that LINK_FLAGS previously let the shell consume now become literal characters in the ELF RUNPATH. With both Make and Ninja, readelf reports `[$ORIGIN:$ORIGIN/../lib:$ORIGIN/../nvidia/cudnn/lib]` before this change but `['$ORIGIN':'$ORIGIN/../lib':'$ORIGIN/../nvidia/cudnn/lib']` afterwards. A module calling dlopen on a sibling library succeeds on the baseline and fails with these new options when LD_LIBRARY_PATH is unset. This breaks deployments that rely on the intended relative-library lookup (including FE's CUDA runtime loader). Remove the single quotes around the three $ORIGIN entries; CMake preserves the dollar sign correctly. That one-line adjustment restores both generators in the reproduction. Check the resulting ELF and a loader probe, since the verbose link command alone still looks plausible.

Evidence: CMake 3.31.4 / GCC11, exact baseline/head CMake linker blocks on isolated modules, Make and Ninja; baseline/head/fixed ELF inspection plus actual dlopen (six builds, expected success/failure/success for each generator). Attribution: Introduced by this PR; the raw LINK_FLAGS baseline does not preserve the single quotes in RUNPATH. The clean prospective merge retains the same changed block.

Validation: Reviewed the complete change and its current-develop integration; the merge is clean. Reproduced the RUNPATH/loader regression with both Unix Makefiles and Ninja using the exact CMake blocks. Removing only the single quotes restores relative lookup on both. Started python_tests CI and verified that its mirrored commit identifies the full reviewed GitHub SHA. The pipeline is still running.

Limitations: The focused reproduction uses small shared modules to isolate the linker configuration; it is not a full FE extension or Windows build. No execute/kernel code changes, so a steady-state CPU/GPU performance comparison is not applicable.

YangXu1990uiuc:

CI follow-up for `448c810ab98a`: all six development/release Python test lanes on SM80, SM90 and SM100 passed. The remaining pipeline failure is a source-comment scan outside this PR. These passes do not exercise the relative-library lookup from the review: the quoted RUNPATH regression still needs the linker-option fix and a loader/relocation check.

PR #347, opened 2026-07-05
benchmark: compute causal SDPA FLOP counts without allocating masks

fallintoplace:
## Why
The benchmark FLOP helper allocated `torch.ones((q_seqlen, kv_seqlen))` for causal masks, which can allocate huge CPU masks (e.g. 32768 x 32768) before counting non-masked elements.

## What changed
- Added `count_causal_nonmasked_elems(...)` to compute causal non-masked element counts using row-bound arithmetic.
- Replaced `torch.tril(...).sum()` usage in `flops(...)` with the new helper.
- Added unit tests covering:
  - square masks
  - rectangular masks
  - top-left causal and bottom-right causal
  - sliding-window variants
  - reference cross-check against a brute-force boolean-mask computation

## Test coverage
Added `test/python/test_sdpa_benchmark_single_sdpa.py` with explicit assertions and a reference cross-check.

## Summary by CodeRabbit

* **New Features**
  * Added FLOP counting support for causal attention with top-left and bottom-right mask alignment.
  * Added support for optional sliding-window attention and rectangular query/key-value dimensions.
  * Unsupported mask alignments are now reported clearly.

* **Tests**
  * Added coverage for square, rectangular, sliding-window, empty-row, and reference-comparison scenarios.
  * Improved confidence in benchmark FLOP calculations across supported attention configurations.

### DIFF
--- a/benchmark/attention_training/benchmark_single_sdpa.py
+++ b/benchmark/attention_training/benchmark_single_sdpa.py
@@ -31,6 +31,11 @@ from typing import Optional, Dict, Any
 
 from torch.profiler import profile, record_function, ProfilerActivity
 
+if __package__:
+    from .flops import count_causal_nonmasked_elems
+else:
+    from flops import count_causal_nonmasked_elems
+
 # torch.profiler measures through CUPTI. If CUPTI cannot attach -- for any
 # reason: a device it does not recognise, a driver mismatch, missing profiling
 # permissions, or simply not being present -- it records no events at all, and
@@ -1477,21 +1482,7 @@ else:
         if attn_mask == "no_mask":
             num_nonmasked_elems = q_seqlen * kv_seqlen
         elif attn_mask in ("top_left", "bottom_right"):
-            if sliding_window_size is not None:
-                # With sliding window: each query attends to at most W keys
-                # (left_bound is exclusive, right_bound is inclusive)
-                # For positions 0 to W-1: 1, 2, ..., W keys (partial window)
-                # For positions W to S-1: W keys each (full window)
-                W = sliding_window_size
-                S = min(q_seqlen, kv_seqlen)
-                if S <= W:
-                    # Sequence shorter than window, use full causal
-                    num_nonmasked_elems = S * (S + 1) // 2
-                else:
-                    # Partial window (first W positions) + full window (remaining positions)
-                    num_nonmasked_elems = W * (W + 1) // 2 + (S - W) * W
-            else:
-                num_nonmasked_elems = torch.tril(torch.ones((q_seqlen, kv_seqlen), dtype=torch.bool)).sum()
+            num_nonmasked_elems = count_causal_nonmasked_elems(q_seqlen, kv_seqlen, attn_mask, sliding_window_size)
         else:
             raise ValueError(f"Unknown attn_mask: {attn_mask}")
         # BMM FLOPs: 2 * M * N * K.
--- a/benchmark/attention_training/flops.py
+++ b/benchmark/attention_training/flops.py
@@ -1,0 +1,45 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for calculating attention operation counts."""
+
+from typing import Optional
+
+
+def count_causal_nonmasked_elems(
+    q_seqlen: int,
+    kv_seqlen: int,
+    attn_mask: str,
+    sliding_window_size: Optional[int] = None,
+) -> int:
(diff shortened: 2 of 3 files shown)


vedaanta (commented):
The math here is correct — and it actually fixes two latent bugs beyond the stated allocation issue — but the PR will red-line CI as written. Two blockers, both in the new test file.

---

## :red_circle: Blocker 1 — the new test aborts the entire Python test run

`test/python/test_sdpa_benchmark_single_sdpa.py` does:

```python
from benchmark.sdpa_benchmark_training.benchmark_single_sdpa import count_causal_nonmasked_elems
```

That import cannot resolve under CI's invocation. Reproduced on a clean checkout with bare `pytest`:

```
ImportError while importing test module '.../test/python/test_sdpa_benchmark_single_sdpa.py'.
E   ModuleNotFoundError: No module named 'benchmark'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

Why:

* `test/python/pytest.ini` exists, so pytest's rootdir is `test/python`. Under the default `prepend` import mode only `test/python` is placed on `sys.path` — the repo root never is.
* `benchmark/` has no `__init__.py` (only `benchmark/norms/` and `benchmark/sdpa_benchmark_training/` do), so there's no installed-package fallback either.
* `ci/run_python_tests.sh:14` and `ci/build_test_wheels.yml:108` both invoke bare `pytest test/python …`.

Because this is a collection **error** rather than a failure, it aborts the whole suite, not just this file. It does happen to work under `python -m pytest` from the repo root (which implicitly puts cwd on `sys.path`) — likely how this was tested locally.

Suggested fix: move the test next to the code it tests (`benchmark/sdpa_benchmark_training/test_*.py`, which already has an `__init__.py`), or prepend the repo root to `sys.path` from `test/python/conftest.py`.

## :red_circle: Blocker 2 — even with the import fixed, the test never executes

`test/python/pytest.ini` sets `addopts = -m L0`. The new tests carry no marker, so they are deselected in every CI lane. 55 files in the suite use `pytest.mark.L0` — please add `@pytest.mark.L0` (or the appropriate level).

## :yellow_circle: Pulling the benchmark harness into collection is fragile

`benchmark_single_sdpa.py` imports `torch`, `torch.profiler`, and optionally `cutlass.cute` at module scope — all to exercise a pure-integer helper. Even after the path fix, this couples a trivial arithmetic test to the availability of the entire benchmark dependency stack.

---

## :yellow_circle: Undisclosed behavior change (worth adding to the PR description)

The PR is framed as "don't allocate a 32768×32768 mask", but it also changes the reported numbers:

* The old code applied `torch.tril` (top-left aligned) even when `attn_mask == "bottom_right"` — wrong for rectangular shapes.
* The old sliding-window path used `S = min(q_seqlen, kv_seqlen)` triangular arithmetic for *both* alignments, ignoring `bottom_right` entirely.

The new code is right on both counts, and matches what the benchmark actually asks cuDNN for — see `benchmark_single_sdpa.py:665-667`, where `diagonal_band_left_bound = sliding_window_size` with left bound exclusive and right bound `0` inclusive, i.e. exactly `W` keys anchored to the alignment diagonal.

But that means reported TFLOP/s for `bottom_right` and for rectangular sliding-window configs will shift, so the checked-in numbers under `benchmark/sdpa_benchmark_training/results*/` are no longer directly comparable to fresh runs. Worth calling out explicitly so nobody chases a phantom perf regression.

## :green_circle: Math verified

I checked the implementation against the test's reference across both alignments, including the clamp edges:

* For `bottom_right`, `q_idx + (kv_seqlen - q_seqlen)` never exceeds `kv_seqlen - 1`, so `row_end` never actually clamps.
* For `top_left` with `q_seqlen > kv_seqlen`, the clamp correctly yields zero for rows whose window has slid past the end of KV (`row_start > row_end`).

The return type also improves from a 0-d tensor to a plain `int`.

Formatting is fine — the 130-char signature at line 268 is comfortably under the repo's black `--line-length 160` (`.pre-commit-config.yaml:18`).

---

## Nits

* The commit headline says "closed-form", but this is an `O(q_seqlen)` Python loop. It's fast enough in practice (~32k iterations, and `flops()` is called twice per run), so no objection to the approach — the name just oversells it. A genuine closed form is a handful of arithmetic lines if you'd prefer one.
* `_reference_causal_nonmasked_elems` in the test is the same formula re-expressed with broadcasting, so the cross-check is close to circular. A stronger reference would build the mask from the `diagonal_alignment` / `diagonal_band_*` semantics used at lines 665-667 rather than restating the helper's own row-bound logic.

---

**Recommendation:** the `benchmark_single_sdpa.py` change looks good to me. Please fix the test file's import path and add the `L0` marker before merge — as-is it breaks `ci/run_python_tests.sh` outright.

vedaanta (approved):

PR #355, opened 2026-07-07
Preserve the CUDA default stream in DSA wrappers

kunlunl:
## What

Use `torch.cuda.get_stream_from_external()` when entering a caller-supplied CUDA stream.

## Why

`torch.cuda.ExternalStream(0)` does not preserve the legacy default-stream handle. It produces a different pooled stream, so PyTorch setup or copy operations performed inside the wrapper can race a CuTe kernel launched on stream 0.

`get_stream_from_external()` preserves both the zero default-stream handle and nonzero external stream handles.

## Validation

Run on an H100 (SM90) with PyTorch 2.12, CUDA 13.2, and cuDNN Frontend built from current `develop`:

- New stream-identity regression on unpatched `develop`: **failed as expected** because the active handle was nonzero.
- Same regression with this change: **1 passed**.
- Complete `test/python/fe_api/dsa` suite: **22 passed**.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/utils/runtime.py
+++ b/python/cudnn/deepseek_sparse_attention/utils/runtime.py
@@ -56,5 +56,5 @@ def torch_stream_context(current_stream: Optional[cuda.CUstream] = None) -> Iter
     if current_stream is None:
         yield
         return
-    with torch.cuda.stream(torch.cuda.ExternalStream(int(current_stream))):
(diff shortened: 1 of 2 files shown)


kunlunl:
Superseded by #354, which now contains this fix as a separate signed-off commit and passed the combined H100 DSA validation (3 targeted regressions and 24/24 DSA tests).

PR #356, opened 2026-07-07
Make DSA dense indexer backward CUDA Graph safe

kunlunl:
## What changed

- Keep `grad_loss` on the GPU in dense indexer backward instead of reading it with `.item()`.
- Pass the device scalar to the SM90 and SM100 dense score-gradient kernels and combine it with the host normalization scale in the kernel.
- Add a CUDA Graph test that replays the same graph with two different `grad_loss` values and checks all gradients against the PyTorch reference.

## Why

`DenseIndexerBackward.execute()` synchronously copied the CUDA `grad_loss` tensor to the host. That operation is forbidden during CUDA Graph capture and made the dense-loss DSA backward path impossible to capture. The scalar is runtime data, so consuming it directly in the score-gradient kernel also preserves its value across graph replays without changing any compile-cache key or GEMM metadata.

## Validation

Validated on one H100 80GB (SM90):

- Unpatched `develop`: the new targeted test fails at `grad_loss.item()` with `operation not permitted when stream is capturing`.
- Patched targeted CUDA Graph test: `1 passed`.
- Complete dense indexer backward test file: `2 passed`.
- Complete DSA test directory: `22 passed`, `0 skipped`.
- Repository `black` and `black-jupyter` pre-commit hooks passed for all changed files.

The runtime validation covers SM90. The SM100 score-gradient path receives the same shared API fix and was checked for matching compile/runtime argument order, but was not run on SM100 hardware in this validation.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_backward/api.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_backward/api.py
@@ -380,8 +380,8 @@ class DenseIndexerBackward(APIBase):
     ) -> None:
         backend_stream = None if self._uses_current_stream_pipeline else current_stream
         with _torch_stream_context(backend_stream):
-            grad_loss_value = float(_as_grad_loss_tensor(grad_loss, index_q.device).item())
-            grad_scale = float(loss_coeff) * grad_loss_value / max(int(self.normalization_tokens), 1)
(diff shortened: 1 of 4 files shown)


kunlunl:
Superseded by #354, which now contains this fix as a separate signed-off commit and passed the combined H100 DSA validation (3 targeted regressions and 24/24 DSA tests).

PR #364, opened 2026-07-08
Serialize selected plan behavior notes

yeliu-oss:


hwanseoc (approved):
Fix looks good, optionally add C++ catch regression test

PR #365, opened 2026-07-08
Fix cutlass DSL deprecation: use .ptr for cute.struct scalar fields

Anerudhan:
## What

Fixes the cutlass DSL deprecation warning emitted from `cutlass/cute/core.py` when running the DSA kernels with nvidia-cutlass-dsl >= 4.5.0:

```
DeprecationWarning: Use explicit `struct.scalar.ptr` for pointer instead.
```

## Why

cutlass-dsl 4.5+ deprecates using a scalar field of a `@cute.struct` directly as a pointer (via the `_ScalarData.value` property). The warning fires whenever `tmem_holding_buf` / `tmem_dealloc_mbar_ptr` struct scalars are passed to `cute.arch.alloc_tmem` / `cute.arch.retrieve_tmem_ptr` (directly or through `cutlass.utils.TmemAllocator`), which read `.value` on their pointer argument.

## How

Append the explicit `.ptr` accessor at the remaining call sites, matching the pattern already used by the other DSA kernels (e.g. `dense_gemm_persistent_swiglu.py`). No functional change; `.ptr` returns the same underlying pointer without going through the deprecated property.

- grouped_gemm: dsrelu / srelu / wgrad `TmemAllocator(...)` arguments
- deepseek_sparse_attention: indexer_forward, indexer_backward (sparse + dense), score_recompute (sparse + dense)

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_backward/dense_indexer_backward_sm100.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_backward/dense_indexer_backward_sm100.py
@@ -788,9 +788,9 @@ class DenseIndexerBackward2QGemmSm100:
         Q_q1_mbar_ptr = storage.Q_q1_mbar.data_ptr()
         K_mbar_ptr = storage.K_mbar.data_ptr()
         mbar = storage.mbar.data_ptr()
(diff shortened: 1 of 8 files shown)


Anerudhan:
@cudnn-ci-bot run

PR #371, opened 2026-07-09
Make plan structure serialization optional within serialize() to construct symmetry with deserialize logic

vincejhan:
### Summary
This PR makes plan serialization symmetric with the handle-based deserialize path and adds regression coverage for both payload modes.

### Changes
1. Updated Graph::serialize(...) in include/cudnn_frontend/graph_interface.h to accept serialize_structure (default true):
- true: include structural graph metadata (context/nodes/tensors) plus plan payload.
- false: emit plan-only payload.
2. In plan-only re-serialize scenarios (sub_nodes.empty() after deserialize(handle, ...)), serialize(...) now sources:
- pass_by_values from cached_pass_by_value
- workspace_modifications from cached_workspace_modifications instead of walking subtree nodes.
3. Added tests in test/cpp/serialize.cpp:
- serialize_structure flag controls structural payload
- Plan re-serialize preserves pass-by-value and workspace modifications

### Reason
Previously, serialization always emitted structure and always collected pass-by-value/workspace data from the node subtree. That could degrade plan-only round trips after deserialize(handle, ...), where structural graph nodes may not be present; particularly in cases where `deserialize(const json&, ...)` isn't called. This change preserves correctness for plan-reload workflows while keeping default behavior backward-compatible.

## Summary by CodeRabbit

* **New Features**
  * Added an option to serialize a compact graph package without embedding full structural metadata.
  * Serialization can now be used to produce either a full graph payload or a plan-only payload for later reload.
* **Bug Fixes**
  * Ensures pass-by-value tensor and workspace modification metadata is retained when re-serializing a previously loaded plan.
  * Keeps variant-pack identifiers consistent across save/load cycles.
* **Tests**
  * Added automated coverage for full vs. compact serialization and reload/metadata consistency.

### DIFF
--- a/include/cudnn_frontend/graph_interface.h
+++ b/include/cudnn_frontend/graph_interface.h
@@ -1563,12 +1563,30 @@ class Graph : public ICudnn, public INode {
         return {error_code_t::OK, ""};
     }
 
+    /**
+     * @brief Serialize graph execution-plan payload, optionally with graph structure.
+     *
+     * When @p serialize_structure is true, the payload includes structural graph metadata
+     * (context/nodes/tensors) in addition to plan fields. When false, the payload is
+     * plan-only and is suitable for handle-based plan reload paths that do not rebuild
+     * graph structure. Plan execution remains valid in plan-only mode.
+     *
+     * Recommendation: if the consumer does not call explicit structural deserialize
+     * (deserialize(const json&, ...)), prefer serialize_structure=false.
+     *
+     * @param data Output UBJSON payload.
+     * @param serialize_structure Include structural graph metadata when true.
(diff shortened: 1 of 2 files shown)


vincejhan:
@cudnn-ci-bot run

vincejhan:
@cudnn-ci-bot run

hwanseoc (approved):
Approved. This works standalone and the serialization/repro tests pass. A later PR currently assumes graph JSON v2 structural fields (`json_version`, `gid`, and `tensors`) are present during handle-based deserialization; that PR will be adjusted after this merges to explicitly support the plan-only payload produced by `serialize_structure=false`.

PR #383, opened 2026-07-12
Separate graph TIDs from backend UIDs in JSON repros

hwanseoc:
@coderabbitai ignore

## Why

Graph JSON v1 uses tensor names as references. Duplicate and anonymous names can collapse distinct tensors and make extracted repros incorrect.

This draft is the hybrid-identity alternative to #280:

- `tid`: graph-local structural identity used by graph JSON and repro extraction
- `uid`: user/backend identity used by variant packs and execution plans

## Architecture

- emit graph JSON `2.0`
- assign graph-local TIDs at tensor creation
- reference node ports and ragged offsets by TID
- preserve UID and whether it was explicitly assigned
- emit both TID and UID in tensor dumps; join repro dumps by TID
- use process-local `gid` to correlate serialized graphs with executions
- leave backend UID assignment behavior unchanged

## Behavior shared with #280

- tensor-list graph schema with optional names
- typed rejection of malformed versions, tensor tables, duplicate identities, and dangling references
- versioned plan-only serialization without `gid` or graph structure
- per-execution repro correlation
- simple, FP8, and MXFP8 SDPA extraction, including sequence lengths and ragged offsets
- migrated XLA serialization sample and native lowered-JSON consumers
- canonicalization of shared ragged-offset objects during deserialization

## Intentional difference from #280

TID and UID are explicit, non-interchangeable namespaces. Structural serialization does not assign missing backend UIDs. #280 instead uses UID for both structure and execution identity.

## Compatibility

- Structural graph JSON v1 is rejected; no ambiguous upgrade is attempted.
- Plan/backend consumers remain UID-based.
- `gid` is process-local execution correlation, not persistent identity.

## Dependencies

#386 contains the architecture-independent runtime tensor-dump and FP8/MXFP8 tester fixes.

## Verification

Based on `develop` `526020e`; B200, CUDA 13.1, cuDNN backend 9.30:

- C++ graph/serialization suite: 75 assertions in 10 cases passed
- chained two-operation TID allocation torture test: 11 assertions passed
- shared ragged-offset canonicalization: 5 assertions passed
- native lowered-JSON regression: 1 passed
- deviceless deserialize API-contract regression: 1 passed
- complete repro-tool suite: 71 passed, 2 MXFP8 target skips
- live parity against #280 after normalizing only UID/TID identity: forward, ragged forward, backward, and FP8 payloads match

## Draft question

Should graph-local TID allocation remain at tensor creation sites, or move to a centralized graph-owned allocator before merge?

PR #280, opened 2026-06-04
Use UID-based graph JSON v2 for repro extraction

hwanseoc:
@coderabbitai ignore

## Why

Graph JSON v1 used tensor names when UIDs were not assigned. Anonymous tensors could be dropped and duplicate names could collapse into one tensor, so repro logs could not reliably join graph tensors, executions, and tensor dumps.

This draft is the unified-UID alternative to #383.

## Architecture

- emit graph JSON `2.0`
- use required integer `uid` for tensor-table identity, node references, ragged-offset references, and repro dump joins
- assign missing UIDs before validation, build, or structural serialization while preserving user UIDs
- use process-local `gid` only to correlate serialized graphs with executions
- serialize tensors as a list; tensor names remain optional metadata
- canonicalize shared ragged offsets during structural deserialization

## Behavior shared with #383

- versioned graph JSON v2 and typed rejection of malformed versions, tensor tables, duplicate identities, and dangling references
- versioned plan-only serialization without `gid` or graph structure
- per-execution repro correlation
- simple, FP8, and MXFP8 SDPA extraction, including sequence lengths and ragged offsets
- migrated XLA serialization sample and native lowered-JSON consumers
- shared-ragged-offset deserialization coverage

## Intentional difference from #383

This PR unifies structural and backend identity. Serializing or validating a graph can permanently assign missing tensor UIDs. #383 instead uses graph-local TIDs for structure and leaves backend UID assignment behavior unchanged.

## Schema

```js
{
  "json_version": "2.0",
  "gid": 1,
  "nodes": [{
    "tag": "SDPA_FWD",
    "inputs": { "Q": 1 },
    "outputs": { "O": 4 }
  }],
  "tensors": [{
    "uid": 1,
    "name": "...",
    "dim": [ ... ],
    "stride": [ ... ],
    "ragged_offset_uid": 7
  }]
}
```

## Compatibility

- Structural v1 JSON and unversioned historical plan blobs are rejected; no migration adapter is included.
- `gid` is process-local execution correlation, not persistent identity.
- Plan-only payloads retain `json_version` but omit `gid`, `context`, `nodes`, and `tensors`.

## Dependencies

- #386 contains the architecture-independent runtime tensor-dump and FP8/MXFP8 tester fixes.

## Verification

Rebased on `develop` `526020e`; B200, CUDA 13.1, cuDNN backend 9.30:

- C++ test target rebuilt successfully
- plan/schema validation: 18 assertions passed
[...]

PR #387, opened 2026-07-13
Fix large FP8 and MXFP8 performance repros

hwanseoc:
@coderabbitai ignore

## Why

`--perf` measures graph execution and should not construct full attention references. FP8 and MXFP8 performance repros still allocated or compared references, causing avoidable OOMs after successful cuDNN execution.

Large MXFP8 inputs also exceed Transformer Engine's CUDA grid-y limit when quantized as one flattened 2-D tensor. TE 2.15 passes at 4,194,240 rows and fails at 4,194,272 rows.

## Changes

- Skip FP8 and MXFP8 forward/backward reference construction and comparison under `--perf`
- Keep deterministic backward exact-equality checks enabled
- Use GPU forward output and stats as backward inputs in MXFP8 perf mode
- Avoid expanded FP32 MXFP8 scale references in perf mode
- Quantize oversized MXFP8 inputs in aligned chunks, concatenate compact results, then apply the global TE GEMM scale swizzle
- Add an MXFP8 perf integration test that fails if either reference function is called

This is tester-only. It does not change graph JSON, tensor identity, extraction, or cuDNN graph construction.

## Verification

NVIDIA B200, cuDNN backend 9.30:

- FP8 forward `--perf`: passed
- MXFP8 perf integration: passed; forward and backward references were not called
- MXFP8 deterministic backward: dQ/dK/dV and all three amax outputs matched bitwise
- MXFP8 non-perf reference comparison: passed
- Exact DSV3 sequence-16384 repro that previously exceeded TE's launch limit: passed, zero CUDA OOMs

MXFP8 tests used the supported PyTorch 26.05 container with PyTorch 2.12.0a0, CUDA 13.2, and Transformer Engine 2.15. During the original investigation, the same patch moved the supplied set from 2/60 to 60/60 passes and the deduplicated raw-log set from 385/445 to 445/445 passes.

### DIFF
--- a/test/python/sdpa/fp8.py
+++ b/test/python/sdpa/fp8.py
@@ -311,6 +311,7 @@ def create_paged_container_and_block_table(tensor, block_size):
 def exec_sdpa_fp8(cfg, request, cudnn_handle):
     if request.config.option.dryrun:
         pytest.skip("dryrun")
+    perf = request.config.getoption("--perf")
 
     cudnn_version = LooseVersion(cudnn.backend_version_string())
     if cudnn_version < "9.14.0":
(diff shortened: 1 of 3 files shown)


hwanseoc:
Superseded by #386. Its two focused commits were preserved there, so #386 now contains all architecture-independent repro fixes shared by #280 and #383.

PR #389, opened 2026-07-14
Add MoE + expert-parallel Python API surface, PyTorch reference, and tests

Anerudhan:
## Summary

Proposes the public contract for a fused SwiGLU MoE operator with expert parallelism (`cudnn.MoeEp`), ahead of the device implementation:

- **`python/cudnn/moe_ep/`** — `MoeEp` (plain class: constructor + `__call__` + `backward`) with bf16, mxfp8, and nvfp4 public outputs. Quantized operands travel as `BlockScaledTensor` (data + scale bundled, logical/unswizzled layouts). Both `__call__` and `backward` currently allocate correctly-shaped results without launching a kernel.
- **`test/python/fe_api/moe_ep/moe_ep_reference.py`** — executable pure-PyTorch semantic reference: routing (contiguous expert sharding, `-1` = dropped route), variable-size `all_to_all_single` EP dispatch, local SwiGLU experts (`silu(gate) * up`, optional DeepSeek-style clamp, router weight in FC1 or FC2 epilogue), Form-A per-top-k combine with quantized-combine round-trips, and block-scaled output encoding.
- **Training integration (forward)** — `generate_c=True` returns `(output, fc1_c, route_metadata)`: the expert-rank-local pre-SwiGLU BF16 accumulator and its row-aligned `(local_expert, src_rank, src_token, src_slot)` routing words.
- **Backward API** — `backward(grad_output, forward inputs..., fc1_c, route_metadata)` returns FP32 `(grad_activation, grad_fc1_weight, grad_fc2_weight, grad_topk_weights)`. It is a collective (every EP rank participates): `MoeEpReference.backward` re-dispatches gradient rows along the identical forward routes, recomputes SwiGLU from the `fc1_c` stash (no post-SwiGLU intermediate is saved), reconstructs the row permutation purely from `route_metadata`, and returns/scatters route gradients to their source `(token, slot)`. Quantization round-trips are straight-through.
- **`docs/fe-oss-apis/moe_ep.md`** — design proposal: mapping to the MegaMoE kernel interface (why workspaces, peer pointer mappers, streams, and launch config are not public arguments), the `fc1_c`/`route_metadata` contract, the backward call contract (argument/return tables), the saved-tensors-for-backward table, and the numerical conventions a fused kernel must match (straight-through quantization round-trips, bf16 stash recompute, inclusive clamp bounds).

## Testing

`pytest test/python/fe_api/moe_ep` — 25 tests:

- Forward validated against naive per-token loops (bf16/fp32 and mxfp8/nvfp4 block-scaled inputs; both `apply_topk_in_fc1` modes; clamp).
- Backward validated against torch autograd replicas for fp32 and block-scaled inputs, including zero gradient at `-1` routes.
- 4-rank/4-GPU NCCL expert-parallel test (forward, `fc1_c`, `route_metadata`, and backward gradients cross-checked from deterministic seeds); skips below 4 GPUs.
- 7 strict-xfail API-vs-reference gates (forward per output format, quantized inputs, `generate_c`, and backward) that flip to hard failures once the device kernels are wired.

Verified on a single-GPU SM100 machine and on 8x B200 nodes.

### DIFF
--- a/test/python/fe_api/moe_ep/test_moe_ep.py
+++ b/test/python/fe_api/moe_ep/test_moe_ep.py
@@ -1,0 +1,771 @@
+# Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+from datetime import timedelta
+
+import pytest
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+import torch.nn.functional as F
+
+from fe_api.moe_ep.moe_ep_reference import (
+    BlockScaledTensor,
+    MoeEpReference,
+    MoeFormat,
+    quantize_blockwise,
+)
+
+pytestmark = pytest.mark.L0
+
+
+def _output_as_float(output):
+    if isinstance(output, torch.Tensor):
+        return output.float()
+    return output.dequantize()
+
+
+def _naive_reference(
+    activation,
+    fc1_weight,
+    fc2_weight,
+    topk_idx,
+    topk_weights,
+    *,
+    apply_topk_in_fc1,
+    clamp=None,
+    combine_format=MoeFormat.BF16,
+):
+    token_count, top_k = topk_idx.shape
+    hidden_size = activation.shape[1]
+    intermediate_size = fc2_weight.shape[1]
+    combine = torch.zeros(token_count, top_k, hidden_size, dtype=torch.float32)
+    for token in range(token_count):
+        for slot in range(top_k):
+            expert = int(topk_idx[token, slot])
+            if expert == -1:
+                continue
+            gate_up = activation[token].float() @ fc1_weight[expert].float()
+            gate, up = gate_up.split(intermediate_size)
+            if clamp is not None:
+                gate = gate.clamp(max=clamp)
+                up = up.clamp(-clamp, clamp)
+            intermediate = F.silu(gate) * up
+            route_weight = topk_weights[token, slot].float()
+            if apply_topk_in_fc1:
+                intermediate = intermediate * route_weight
+            result = intermediate @ fc2_weight[expert].float()
+            if not apply_topk_in_fc1:
+                result = result * route_weight
+            if combine_format is MoeFormat.BF16:
+                result = result.to(torch.bfloat16).float()
+            else:
+                result = quantize_blockwise(result, combine_format).dequantize()
+            combine[token, slot] = result
+    return combine.sum(dim=1).to(torch.bfloat16)
+
+
+@pytest.mark.parametrize(
+    "format,expected_data_shape,expected_scale_shape,scale_dtype",
+    [
+        (MoeFormat.MXFP8, (3, 64), (3, 2), torch.float8_e8m0fnu),
+        (MoeFormat.NVFP4, (3, 32), (3, 4), torch.float8_e4m3fn),
+    ],
+)
+def test_block_scaled_round_trip(format, expected_data_shape, expected_scale_shape, scale_dtype):
+    values = torch.linspace(-4.0, 4.0, 3 * 64).reshape(3, 64)
+    quantized = quantize_blockwise(values, format)
+
+    assert isinstance(quantized, BlockScaledTensor)
+    assert quantized.format is format
+    assert quantized.logical_shape == (3, 64)
+    assert tuple(quantized.data.shape) == expected_data_shape
+    assert tuple(quantized.scale.shape) == expected_scale_shape
+    assert quantized.scale.dtype == scale_dtype
+    assert quantized.dequantize().shape == values.shape
+    assert torch.isfinite(quantized.dequantize()).all()
+
+
+def test_quantization_along_weight_reduction_axis():
+    values = torch.randn(2, 64, 48)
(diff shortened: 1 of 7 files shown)


test/python/fe_api/moe_ep/test_moe_ep.py
```diff
@@ -0,0 +1,855 @@
...
+    grad_output = torch.randn(tokens, hidden, device=device)
+
+    kwargs = dict(
+        num_experts=experts,
+        hidden_size=hidden,
+        intermediate_size=intermediate,
+        top_k=top_k,
+        generate_c=True,
+    )
+    api = MoeEp(**kwargs)
```
Anerudhan, line 663: Usage and reference

YangXu1990uiuc (changes requested):

Reviewed `50df079e8e0f` against base `08b9a909ceac`.

The explicit EP ownership, combine-versus-output encoding, and expert-local backward stash form a coherent proposal. I am requesting changes on the public compute stub: it must report that execution is unavailable instead of returning unwritten results. There is also a reproduced public/reference tensor-type mismatch. Two existing validation findings remain confirmed below.

**[P1 · blocks approval] Make the public compute entrypoints fail explicitly until execution exists** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/50df079e8e0f954e1d1eaca5568c7eb47e31d905/python/cudnn/moe_ep/api.py#L327))

Keeping the API proposal ahead of its kernel is reasonable, but exporting cudnn.MoeEp with __call__/backward that successfully return torch.empty buffers lets a caller consume uninitialized activations or gradients as a computed result. Documentation and xfail markers do not guard runtime use. With valid all-dropped routes, zero inputs and zero incoming gradients, the reference returns exact zeros; enabling PyTorch deterministic uninitialized-memory filling makes the public output and all four gradient tensors NaN without raising. This is a high-risk P1 because the normal public call can silently corrupt a model computation. Keep any shape/allocation helper private or explicitly named, and raise NotImplementedError from the compute entrypoints until they execute a backend (or provide an explicit reference mode). The proposal can land without requiring the fused kernel to be finished.

Evidence: SM100 probe on unchanged api.py: torch.use_deterministic_algorithms(True) plus torch.utils.deterministic.fill_uninitialized_memory=True; T3/H32/E2/I16/K2, all topk_idx=-1. Output and four gradients all NaN; independent reference outputs/gradients exact zero. No allocator monkeypatch. Attribution: Forward behavior introduced in c159fb64; backward public allocation stub added in 50df079e. Intentional placeholder behavior in this PR, but unsafe as a successful public compute call.

**[P2] Exercise the public BlockScaledTensor in the semantic reference** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/50df079e8e0f954e1d1eaca5568c7eb47e31d905/test/python/fe_api/moe_ep/moe_ep_reference.py#L299))

The reference defines a different BlockScaledTensor class from cudnn.moe_ep.api. A valid public MXFP8 bundle therefore misses this isinstance branch and reaches tensor.is_floating_point(), raising AttributeError. I reproduced it by rewrapping the reference quantizer's exact same data/scale/shape/axis in cudnn.moe_ep.BlockScaledTensor. Existing quantized-input tests only construct the reference class, so they miss the public integration boundary. Share the representation/validation or add an explicit adapter, and test the public class through both the API and oracle.

Evidence: SM100 valid public MXFP8 activation: AttributeError("'BlockScaledTensor' object has no attribute 'is_floating_point'"); identical payload works in the reference class. Attribution: Introduced with the two separate representations and their tests in c159fb64.

- **[P2] Still open:** [Validate data and scale storage together](https://github.com/NVIDIA/cudnn-frontend/pull/389#discussion_r3581510380).

- **[P2] Still open:** [Keep route validation independent of generate_c](https://github.com/NVIDIA/cudnn-frontend/pull/389#discussion_r3581510402).

Validation: Focused suite on SM100:17 passed,7 expected failures,1 four-GPU NCCL test skipped. The expected failures cover the unimplemented numerical API; they are not passing kernel evidence. Four-process CPU/Gloo probe:all four scenarios passed on every rank, checking forward, stash/metadata and all gradients against per-route math/autograd. Covered unequal token counts including a zero-token rank, all-dropped and duplicate routes, all-to-one expert-rank skew, and both router-weight placements. SM100 contract probes reproduced uninitialized public output/gradients, rejection of the public quantized representation by the reference, mixed-device bundle acceptance, and generate_c-dependent route validation. NVFP4 finite inputs at 2688/3000/10000 encoded finite in the tested PyTorch build; no overflow finding. Tests used a prospective merge with current develop, resolving only the additive lazy-import-map conflict locally by retaining both sides. All three new implementation/test files are identical to the reviewed head, and the reused native extension has matching native sources. Rechecked after develop advanced to 08b9a909 (#1068): the same focused suite again reports17 passed,7 expected failures,1 NCCL skip. The MoE proposal files and discussion are unchanged.

Limitations: No four-GPU NCCL or fused-device execution was independently validated in this review. Gloo coverage establishes reference communication semantics only. No performance measurement applies to a public stub and semantic reference. The PR needs an import-map rebase before test CI can run on the integration revision. After rebasing, request the OSS test target with a standalone `@cudnn-ci-bot run oss` comment, or ask a maintainer to send it if needed; include the four-GPU test explicitly when validating NCCL coverage.

PR #392, opened 2026-07-15
BSA: remove dead SM100 block-128 fragment allocations

jiayus-nvidia:
## Summary

  - Remove an unused TMEM scale-load setup chain from the SM100 block-128 correction warp.
  - Remove an RMEM scale tensor allocation whose result was never consumed.
  - Remove an outer `tOrO_frg` allocation that was unconditionally overwritten before use.
  - Keep the per-iteration `make_rmem_tensor` allocation to preserve the original fragment layout, dtype, and lifetime.

  ## Motivation

  `nvidia-cutlass-dsl` 4.6 removed the deprecated top-level `cute.make_fragment` API and private `cute.core.ThrMma` type. The BSA implementation had
  already migrated its live register tensors and type annotations to `cute.make_rmem_tensor` and `cute.ThrMma`.

  That migration exposed several pre-existing dead allocations in the SM100 block-128 correction path. This change removes only those unused
  constructs, following the cleanup in [FlashInfer PR #3922](https://github.com/flashinfer-ai/flashinfer/pull/3922).

  No copies, barriers, reductions, indexing operations, TMEM loads/stores, or architecture instructions are changed.

  ## Compatibility

  Verified with:

  - `nvidia-cutlass-dsl==4.5.2`
  - `nvidia-cutlass-dsl==4.6.1`

  The `block_sparse_attention` tree contains no remaining uses of:

  - `cute.make_fragment(...)`
  - `cute.core.ThrMma`
  - `cute.core.ThrCopy`

  MMA object methods such as `make_fragment_A/B/C` remain unchanged because they are still valid public APIs.

  ## Validation

  Tested on NVIDIA B200 / SM100:

  - BSA forward and backward correctness tests: `17 passed` with CUTLASS DSL 4.5.2.
  - BSA forward and backward correctness tests: `17 passed` with CUTLASS DSL 4.6.1.
  - Cold-cache 4.6.1 JIT trace and compilation of the SM100 block-128 forward kernel succeeded.
  - All 39 files under `python/cudnn/block_sparse_attention` parsed and imported successfully with both versions.
  - Black 26.3.1 formatting check passed.
  - `git diff --check` passed.

  The tests cover SM100 block-128 and block-64 forward/backward kernels. SM90 and SM120 modules were statically imported under both DSL versions but
  were not executed because the validation machine was SM100.

### DIFF
--- a/python/cudnn/block_sparse_attention/csrc/fwd/sm100_blk128/bsa_fwd_sm100.py
+++ b/python/cudnn/block_sparse_attention/csrc/fwd/sm100_blk128/bsa_fwd_sm100.py
@@ -1506,16 +1506,6 @@ class BlockSparseAttnForwardSm100Blk128:
         tidx = cute.arch.thread_idx()[0] % (cute.arch.WARP_SIZE * len(self.correction_warp_ids))
         warp_idx = cute.arch.make_warp_uniform(cute.arch.warp_idx()) % 4
 
-        tScS = thr_mma_qk.partition_C(cute.make_identity_tensor(self.mma_tiler_qk[:2]))
-        tStScale_layout = cute.composition(tStS.layout, cute.make_layout((self.m_block_size, 1)))
-        tStScales = tuple(cute.make_tensor(tStS.iterator + self.tmem_vec_offset[stage], tStScale_layout) for stage in range(self.s_stage))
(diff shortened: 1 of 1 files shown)


Anerudhan:
@cudnn-ci-bot run

Anerudhan (approved):

PR #401, opened 2026-07-16
Add large-tensor convolution fuzzer

msalasooNV:
## Before submitting

## Affected area

- CI or test infrastructure

## Summary

Adds `test/python/test_conv_large_tensor_fuzzer.py`, a deterministic Python
large-tensor convolution fuzzer for FPROP, DGRAD, and WGRAD. The test generates
large tensor/offset/reduction cases, compares cuDNN execution against a PyTorch
float32 reference, and skips generated graphs when cuDNN reports no supported
engine configuration for the current architecture.

FPROP and WGRAD target `C*R*S > 2^27` boundary coverage when the per-worker
memory budget permits. DGRAD still exercises large tensors and filters, but
uses a conservative runtime bound rather than claiming the same reduction
boundary.

The fuzzer includes configurable L0/L1 testcase counts, a per-case work cap to
avoid pathological long-running generated problems, exact configuration JSON
emission for failures, and data policies for very large reductions that keep
reference comparisons stable while still exercising large tensor shapes.
Integer-policy comparisons use fixed dtype-specific bounds, while dense-random
comparisons scale with each operation's effective reduction depth. Developer
diagnostics can restrict generation to one operation and optionally select one
or more public graph engine indices for that operation.

Operation selection happens before shape-generation retries so budget
rejections do not skew FPROP, DGRAD, and WGRAD coverage. Output tensors are
prefilled with NaNs so incomplete writes fail comparison.

## Why

Large tensor convolution shapes can expose backend and frontend issues that are
not covered by smaller deterministic tests. This adds reproducible fuzz coverage
for those cases.

The generator bounds runtime with a work cap, uses deterministic seeds, emits
copy-pasteable repro context, and treats unsupported generated graphs as skips
instead of hard failures. GPU allocation failures are reported separately as
insufficient-memory skips with configuration and lifecycle-phase context.

## Related issues

None.

None. This adds a Python test file only and does not change public frontend APIs
or compatibility behavior.

## Testing

The latest default-count validation used four pytest workers:

```bash
CUDNN_FUZZ_NUM_TESTS_L0=64 \
CUDNN_FUZZ_NUM_TESTS_L1=448 \
CUDNN_FUZZ_RUNTIME_WORK_BUDGET=1e14 \
python -m pytest test/python/test_conv_large_tensor_fuzzer.py \
  -o addopts= -m "L0 or L1" -n 4 --tb=short --durations=20 -ra
```

The runs used a local development build of the cuDNN backend.

| Architecture | Result | Runtime |
| --- | ---: | ---: |
| SM90 | `447 passed`, `65 skipped` | `1218.80s (0:20:18)` |
| SM100 | `447 passed`, `65 skipped` | `2010.97s (0:33:30)` |

Both runs completed without numeric or execution failures. The skipped cases
were generated graphs unsupported on the corresponding architecture.

Developer configuration is exposed through these environment variables:

| Variable | Accepted value and behavior |
| --- | --- |
| `CUDNN_FUZZ_NUM_TESTS_L0` | Positive integer L0 testcase count; default `64`. |
| `CUDNN_FUZZ_NUM_TESTS_L1` | Positive integer L1 testcase count; default `448`. |
| `CUDNN_FUZZ_RUNTIME_WORK_BUDGET` | Positive numeric per-case runtime-work cap; default `1e14`. |
| `CUDNN_FUZZ_ENGINE_OP` | Restrict generation to `fprop`, `dgrad`, or `wgrad`; aliases `fp`, `dg`, and `wg` are accepted. It can be used without an engine index. |
| `CUDNN_FUZZ_GRAPH_ENGINE_INDICES` | Unique comma-separated public graph engine indices for the selected operation. Requires `CUDNN_FUZZ_ENGINE_OP`; each index is validated against the generated graph and one is selected deterministically per testcase. |
| `CUDNN_FUZZ_REGEN_ON_UNSUPPORTED` | With an operation and engine indices, retry generated configs that do not support the selected engine. Accepted true values are `1`, `true`, `yes`, and `on`; false values are `0`, `false`, `no`, `off`, or empty. Other values fail during collection. |
| `CUDNN_FUZZ_REGEN_ATTEMPTS` | Integer regeneration-attempt cap; default `50` and effective minimum `1`. |
| `CUDNN_FUZZ_REPRO_CONFIG` | Inline emitted configuration JSON for `test_conv_large_tensor_repro`. Takes precedence over `CUDNN_FUZZ_REPRO_FILE`. |
| `CUDNN_FUZZ_REPRO_FILE` | Path to emitted configuration JSON for `test_conv_large_tensor_repro`. |

Standard pytest controls remain available, including `-m "L0 or L1"`,
`-n <workers>`, and `-k test_conv_large_tensor_repro`.

Focused repro validation:

- Inline and file-based repro configurations passed on both architectures; the
  inline configuration took precedence when both were set.
- Malformed repro JSON failed with a clear parser error.
- Marker-filtered L0/L1 runs excluded the unmarked manual repro test.
- Representative integer-policy FPROP, DGRAD, and WGRAD repros passed on both
  architectures.

Focused engine-filter and argument validation:

- Operation-only filtering with a non-default work cap passed on both
  architectures.
- Public graph engine index `0` executed successfully for FPROP, DGRAD, and
  WGRAD on both architectures.
- A comma-separated `0,1` FPROP engine list executed successfully on both
  architectures.
- Missing operation selectors, duplicate or out-of-range indices, invalid
  operations/counts/work budgets/regeneration limits, regeneration without
  indices, and all produced the expected diagnostics.

Notes:

- Runtime depends on GPU architecture, cuDNN plan selection, pytest worker
  count, and selected testcase counts.
- Unsupported generated graphs are skipped when cuDNN has no valid engine
  configuration for the current architecture.

## Summary

* **Tests**
  * Added large-tensor cuDNN convolution regression/fuzz coverage for 2D and 3D across forward, data-gradient, and weight-gradient.
  * Introduced two configurable randomized test tiers with engine/op selection, optional regeneration, and work/memory budgets.
  * Added deterministic failure repro support with detailed shape/config context and computed tolerance information.
  * Improved handling of unsupported or insufficient-resource scenarios by skipping appropriately; unexpected runtime errors still fail.

### DIFF
--- a/test/python/test_conv_large_tensor_fuzzer.py
+++ b/test/python/test_conv_large_tensor_fuzzer.py
@@ -254,8 +254,8 @@ def _create_filtered_execution_plans(graph, graph_engine_index: int) -> None:
 MEMORY_BUDGET_BYTES = _device_memory_budget()
 WORK_BUDGET_FLOPS = _work_budget_flops()
 WORKSPACE_OVERHEAD = 0.15  # fraction added for cuDNN workspace + allocator slack
-DEFAULT_NUM_TESTS_L0 = _num_tests(_NUM_TESTS_L0_ENV, 48)  # default smoke slice
-DEFAULT_NUM_TESTS_L1 = _num_tests(_NUM_TESTS_L1_ENV, 64)  # longer expansion slice
+DEFAULT_NUM_TESTS_L0 = _num_tests(_NUM_TESTS_L0_ENV, 64)  # default smoke slice
+DEFAULT_NUM_TESTS_L1 = _num_tests(_NUM_TESTS_L1_ENV, 448)  # longer expansion slice
 DEFAULT_SEED_L0 = 42
 DEFAULT_SEED_L1 = 12345
 MAX_REGEN_ATTEMPTS = 50
--- a/test/python/test_conv_large_tensor_fuzzer.py
+++ b/test/python/test_conv_large_tensor_fuzzer.py
@@ -361,11 +361,14 @@ class LargeTensorConfig:
 
 
 def _cudnn_dtype(t: torch.dtype):
-    return {
+    dtype_map = {
         torch.float16: cudnn.data_type.HALF,
         torch.bfloat16: cudnn.data_type.BFLOAT16,
         torch.float32: cudnn.data_type.FLOAT,
-    }[t]
+    }
+    if t not in dtype_map:
+        raise ValueError(f"Unsupported tensor dtype: {t}")
+    return dtype_map[t]
 
 
 def _estimate_bytes(cfg: LargeTensorConfig) -> int:
@@ -702,8 +705,8 @@ def _repro_payload(
     test_id = None
     if test_num is not None and total_tests is not None and config_seed is not None:
         test_id = _test_id((test_num, total_tests, config_seed, cfg))
-    graph_engine_op = _graph_engine_filter_op()
-    graph_engine_indices = _graph_engine_indices()
+    graph_engine_op = _FORCED_CONV_TYPE
+    graph_engine_indices = _COLLECTED_GRAPH_ENGINE_INDICES
     selected_graph_engine_index = _select_graph_engine_index(cfg, graph_engine_indices) if graph_engine_indices else None
 
     metadata = {
@@ -1095,8 +1098,8 @@ def _run_cudnn(cfg: LargeTensorConfig, X: torch.Tensor, W: torch.Tensor, Y: torc
 
         graph.validate()
         graph.build_operation_graph()
-        filtered_op = _graph_engine_filter_op()
-        filtered_graph_engine_indices = _graph_engine_indices()
+        filtered_op = _FORCED_CONV_TYPE
+        filtered_graph_engine_indices = _COLLECTED_GRAPH_ENGINE_INDICES
         if filtered_graph_engine_indices:
             if cfg.conv_type != filtered_op:
                 raise ValueError(f"{_GRAPH_ENGINE_OP_ENV}={filtered_op.name.lower()} does not match " f"test op {cfg.conv_type.name.lower()}")
@@ -1308,6 +1311,8 @@ def _run_single_config(
 
 def _run_test(cfg: LargeTensorConfig, cudnn_handle, test_num: int, total_tests: int, config_seed: int, allow_unaligned: bool, include_extras: bool) -> None:
     if _REGEN_ON_UNSUPPORTED_ENABLED:
+        # Replay from config_seed so regeneration can advance past unsupported
+        # candidates; cfg is the first candidate from the same generator.
         _run_test_with_regen(
             config_seed=config_seed,
             test_num=test_num,
@@ -1331,7 +1336,7 @@ def _run_test(cfg: LargeTensorConfig, cudnn_handle, test_num: int, total_tests: 
 
 
 def _run_test_with_regen(*, config_seed: int, test_num: int, total_tests: int, allow_unaligned: bool, include_extras: bool, cudnn_handle) -> None:
-    filtered_op = _graph_engine_filter_op()
+    filtered_op = _FORCED_CONV_TYPE
     if filtered_op is None:
         raise ValueError(f"{_REGEN_ON_UNSUPPORTED_ENV} requires {_GRAPH_ENGINE_OP_ENV}")
 
@@ -1340,7 +1345,7 @@ def _run_test_with_regen(*, config_seed: int, test_num: int, total_tests: int, a
         allow_unaligned=allow_unaligned,
         include_extras=include_extras,
         forced_conv_type=filtered_op,
-        force_plain_fprop=bool(_graph_engine_indices()),
+        force_plain_fprop=bool(_COLLECTED_GRAPH_ENGINE_INDICES),
     )
     max_attempts = _REGEN_ATTEMPT_LIMIT
     last_unsupported = "no configs generated"
--- a/test/python/test_conv_large_tensor_fuzzer.py
+++ b/test/python/test_conv_large_tensor_fuzzer.py
@@ -1221,16 +1221,12 @@ def _to_cuda_conv_layout(cpu_tensor: torch.Tensor, memory_format) -> torch.Tenso
 
 
 def _make_x_cpu_tensor(cfg: LargeTensorConfig, gen: torch.Generator) -> torch.Tensor:
-    # DGRAD writes X as the output buffer, so keep its prefill random to help
-    # expose partial writes.
     if _uses_integer_data(cfg) and cfg.conv_type in (ConvType.FPROP, ConvType.WGRAD):
         return _dense_small_integer_cpu_tensor(cfg.x_shape, cfg.dtype, gen)
     return _dense_random_cpu_tensor(cfg.x_shape, cfg.dtype, gen)
 
 
 def _make_w_cpu_tensor(cfg: LargeTensorConfig, gen: torch.Generator, sparse_rng: random.Random) -> torch.Tensor:
-    # WGRAD writes W as the output buffer, so keep its prefill random to help
-    # expose partial writes.
     if _uses_sparse_filter(cfg):
         return _sparse_filter_cpu_tensor(cfg, sparse_rng)
     return _dense_random_cpu_tensor(cfg.w_shape, cfg.dtype, gen)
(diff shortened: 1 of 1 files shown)


test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1456 @@
...
+    """
+    input_spatial = math.prod(cfg.input_spatial)
+    filter_spatial = math.prod(cfg.filter_spatial)
+    output_spatial = math.prod(cfg.y_shape[2:])
+
+    if cfg.conv_type == ConvType.FPROP:
+        return 2 * cfg.n * cfg.k * output_spatial * cfg.c * filter_spatial
+    if cfg.conv_type == ConvType.DGRAD:
+        return 2 * cfg.n * cfg.c * input_spatial * cfg.k * filter_spatial
+    return 2 * cfg.k * cfg.c * filter_spatial * cfg.n * output_spatial
```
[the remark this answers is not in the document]
msalasooNV: This is intentionally a conservative runtime estimate, not an exact FLOP count. In local testing, huge-filter DGRAD cases took hundreds of seconds despite having small outputs.

I renamed the helper and added a comment to clarify this.

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+        pytest.skip(f"Set {_REPRO_CONFIG_ENV} or {_REPRO_FILE_ENV} to run an exact configuration")
+
+    ok, msg = _run_single_config(cfg, cudnn_handle)
+    if ok:
+        return
+    if msg.startswith("insufficient_memory"):
+        pytest.skip(f"Insufficient GPU memory: {msg}" f"{_format_repro_context(cfg, message=msg)}")
+    if msg.startswith("not_supported"):
+        pytest.skip(f"Graph not supported on this arch: {msg}")
+    pytest.fail(f"cuDNN execution error: {msg}" f"{_format_repro_context(cfg, message=msg)}")
```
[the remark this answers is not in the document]
msalasooNV: This is intentionally unmarked because it is intended for manual repros

Anerudhan (approved):

Anerudhan:
@cudnn-ci-bot run

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+            c, k = self._ch(1, 16), self._ch(1, 4)
+
+        # Input spatial chosen so output spatial matches out_spatial (or 1 if None).
+        out_h = 1 if out_spatial is None else out_spatial
+        out_w = 1 if out_spatial is None else out_spatial
+        h = stride[0] * (out_h - 1) + dil[0] * (r - 1) + 1
+        w = stride[1] * (out_w - 1) + dil[1] * (s - 1) + 1
+        return dict(n=n, k=k, c=c, input_spatial=[h, w], filter_spatial=[r, s], padding=[0, 0], stride=stride, dilation=dil, epilogue=epi)
+
+    def _raw3d(self, shape_family: ShapeFamily) -> dict:
```
chmayer-gpu, line 914: Code duplication with _raw2D. Consider something like:

```
def _raw(self, shape_family: ShapeFamily, sdims: int) -> dict:
    ...
    if sdims == 2:
        r, s = ...; filter_spatial = [r, s]; padding = [0, 0]
    else:
        t, r, s = ...; filter_spatial = [t, r, s]; padding = [0, 0, 0]
```

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+            del bias
+        if actual is not None:
+            del actual
+        if ref is not None:
+            del ref
+        torch.cuda.empty_cache()
+
+
+def _run_test(cfg: LargeTensorConfig, cudnn_handle, test_num: int, total_tests: int, config_seed: int, allow_unaligned: bool, include_extras: bool) -> None:
+    if _REGEN_ON_UNSUPPORTED_ENABLED:
```
chmayer-gpu, line 1361: It is not straight forward to me why we can ignore the LargeTensorConfig when _REGEN_ON_UNSUPPORTED_ENABLED is True. A comment here would be helpful to understand why the cfg no longer matters.

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+    rtol: Optional[float] = None,
+    atol: Optional[float] = None,
+    message: Optional[str] = None,
+    attempt: Optional[int] = None,
+) -> dict:
+    test_id = None
+    if test_num is not None and total_tests is not None and config_seed is not None:
+        test_id = _test_id((test_num, total_tests, config_seed, cfg))
+    graph_engine_op = _graph_engine_filter_op()
+    graph_engine_indices = _graph_engine_indices()
```
chmayer-gpu: Is it intentional to re read the env variables here or could you just use the constants, e.g. _FORCED_CONV_TYPE

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+    except ValueError as e:
+        raise ValueError(f"{_GRAPH_ENGINE_INDICES_ENV} must be a comma-separated list of graph engine indices") from e
+    if not indices:
+        raise ValueError(f"{_GRAPH_ENGINE_INDICES_ENV} must contain at least one graph engine index")
+    seen = set()
+    for index in indices:
+        if index in seen:
+            raise ValueError(f"{_GRAPH_ENGINE_INDICES_ENV} contains duplicate graph engine index {index}")
+        seen.add(index)
+    return indices
```
chmayer-gpu, line 116: A more pythonic alternative to this loop would be something like: 
if len(indices) != len(set(indices)):
    dupes = [x for x in indices if indices.count(x) > 1]
    raise ValueError(f"... duplicate indices: {dupes}")

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _cudnn_dtype(t: torch.dtype):
+    return {
+        torch.float16: cudnn.data_type.HALF,
+        torch.bfloat16: cudnn.data_type.BFLOAT16,
+        torch.float32: cudnn.data_type.FLOAT,
+    }[t]
```
chmayer-gpu: consider raising a more descriptive exception here if encountering an unsupported data type.

test/python/test_conv_large_tensor_fuzzer.py
```diff
@@ -0,0 +1,1443 @@
...
+    reduction_size = _fprop_reduction_size(cfg)
+    nonzero_count = min(_SPARSE_INTEGER_TARGET_NONZERO, reduction_size)
+    W = torch.zeros(cfg.w_shape, dtype=cfg.dtype)
+    if nonzero_count == 0:
+        return W
+
+    flat = W.view(cfg.k, reduction_size)
+    for k_idx in range(cfg.k):
+        positions = rng.sample(range(reduction_size), nonzero_count)
+        signs = [1 if rng.getrandbits(1) else -1 for _ in range(nonzero_count)]
```
chmayer-gpu: more pythonic version would look something like: 
`rng.choices((-1, 1), k=nonzero_count).`

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: I’m keeping these separate because the params have different ranges and constraints for nearly every shape family. Combining them would reduce some duplication but introduce dim-branches throughout, reducing readability.

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: Added a comment

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: Updated to read env vars once into constants

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: Added

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: Updated, I am getting more familiar with pythonic expressions

test/python/test_conv_large_tensor_fuzzer.py
msalasooNV: Updated with a similar expression

PR #405, opened 2026-07-17
DSA: fix 576-wide d_sink, test/document top-k index semantics, B300 benchmarks

vedaanta:
Three small, independent improvements to the DeepSeek Sparse Attention backward — **no API changes**; the existing `cudnn.DSA.sparse_attention_backward_wrapper` surface is untouched.

## 1. Bug fix: zero `d_sink` for the 576-wide latent

The kernel's `sum_dSink` postprocess only runs when `D_qk == D_v`, so the 576-wide MLA latent (512 value dims + 64 RoPE dims) silently returned a zero sink gradient. Fixed with a closed-form epilogue in `_interface_sm100.py` (`d_sink_h = -sum_t sigmoid(sink_h - lse[t,h]) * (out . dout)[t,h]`); the failing `test_DSA_sparse_attention_backward_wrapper[False-512-576-512-32-...]` now passes.

Known pre-existing issue (reproducible on clean develop, unchanged by this PR): `[True-512-576-512-32-...]` fails on dq accuracy when `topk_length` is combined with `D_qk=576`; kernel-level fix tracked separately.

## 2. Test + docs: top-k index semantics

The kernel supports two index conventions that were undocumented and untested: `-1` sentinels without `topk_length` (the FlashMLA convention) vs the compact-prefix fast path with it. Added a parity test that the same valid index set produces the same gradients through both compiled variants, and documented the semantics in `docs/fe-oss-apis/dsa.md` — including that `topk_length` values must be `>= 1` and the inert-row (padding) recipe: `topk_length=1`, any valid index, and a zeroed `dout` row (verified to produce exactly-zero `dq` rows and no `dkv`/`d_sink` contribution).

## 3. Benchmarks: B300 results with MMA-utilization analysis

Wall-clock sweeps for `d_qk=512` (topk 512/2048) and `d_qk=576` (topk 2048) across seqlens 1k–32k, plus Nsight Compute tensor-pipe speed-of-light numbers for the main backward kernel. Highlights: MMA utilization is set by `topk` (44% at topk=2048 / d512, ~37% at d576), is sequence-length invariant, the kernel is memory-pipe-bound (top-k gather + fp32 dKV accumulation), and the seqlen-32k dip is the KV working set exceeding L2.

## Testing (SM100-class GPU)

`test_DSA_sparse_attention_backward.py`: 4 passed, 1 skipped, 1 pre-existing failure (the dq issue above, identical on clean develop).

## Summary by CodeRabbit

* **Documentation**
  * Clarified sparse attention backward top-k index rules, including sentinel values, valid index ranges, and padding behavior.
  * Added B300 benchmark results, tensor-pipe utilization metrics, and performance observations.

* **Bug Fixes**
  * Corrected sink-gradient calculation for mismatched query and value head dimensions.

* **Tests**
  * Added coverage verifying equivalent backward results for sentinel-padded and explicit top-k length inputs.

### DIFF
--- a/benchmark/dsa/README.md
+++ b/benchmark/dsa/README.md
@@ -88,6 +88,72 @@ repository.
 |     8192 |      8192 | 1024 |  4.538 |     605.76 |
 |     8192 |      8192 | 2048 |  8.562 |     642.06 |
 
+### B300
+
+Generated on an NVIDIA B300 (`nheads=64`, bf16, attention sink and
+`topk_length` enabled, `warmup=10`, `repeat=50`), using `torch 2.13.0+cu130`,
+`nvidia-cutlass-dsl 4.5.2`, CUDA 13.0, and `nvidia-cudnn-frontend` built from
+this repository.
+
+`d_qk = d_v = 512`:
+
+| seqlen_q | seqlen_kv | topk | BWD ms | BWD TFLOPS |
+|---------:|----------:|-----:|-------:|-----------:|
+|     1024 |      1024 |  512 |  0.344 |     500.08 |
+|     2048 |      2048 |  512 |  0.648 |     530.63 |
+|     4096 |      4096 |  512 |  1.238 |     555.26 |
+|     8192 |      8192 |  512 |  2.406 |     571.31 |
+|    16384 |     16384 |  512 |  4.789 |     574.04 |
+|    32768 |     32768 |  512 | 10.415 |     527.85 |
+|     2048 |      2048 | 2048 |  1.993 |     689.57 |
+|     4096 |      4096 | 2048 |  3.949 |     696.08 |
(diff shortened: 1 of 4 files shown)


jiayus-nvidia:
Thanks for the fix. I don't think adding a PyTorch epilogue to patch `d_sink` is the best approach, since this logic belongs in the existing SM100 kernel and the PyTorch fallback introduces additional launches and temporary tensors.

I opened [#421](https://github.com/NVIDIA/cudnn-frontend/pull/421), which fixes both the sink normalization and `d_sink` computation directly in the kernel. With #421, the host-side fallback in this PR should no longer be necessary. The documentation, top-k semantics tests, and benchmark changes here remain useful and independent.

PR #407, opened 2026-07-17
DSA indexer_top_k: fall back to scalar stores for odd top_k

### LINKED ISSUE #406: [Bug]: DSA indexer_top_k: odd top_k > threads-per-CTA fails JIT compilation despite check_support allowing any top_k <= 2048
### Before submitting

- [x] I searched existing issues and did not find a duplicate.

### Affected area

FE OSS kernels or CuTeDSL

### Description

`DSA.IndexerTopK` / `DSA.indexer_top_k_wrapper` fail at JIT-compile time for odd `top_k` values larger than the kernel's `num_threads_per_cta`. The documented contract ([docs/fe-oss-apis/dsa.md](https://github.com/NVIDIA/cudnn-frontend/blob/develop/docs/fe-oss-apis/dsa.md): `top_k <= 2048`) and `IndexerTopK.check_support()` (`top_k must be in (0, 2048]`) both permit these values, so `check_support()` returns True and compilation then dies on a bare assert.

### Steps to reproduce

```python
import torch
from cudnn import DSA

scores = torch.randn(235, 705, dtype=torch.float32, device="cuda")
seq_lens = torch.full((235,), 705, dtype=torch.int32, device="cuda")
DSA.indexer_top_k_wrapper(scores, seq_lens, top_k=705, next_n=1, return_val=True)
```

Any odd `top_k > num_threads_per_cta` (256–1024 depending on `max_num_cols`/dtype) reproduces. Even values, and odd values up to `num_threads_per_cta`, compile fine.

### Expected behavior

Compiles and returns the top 705 entries per row.

### Actual behavior

`AssertionError` from the output phase of `python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_varlen_util.py` (`IndexerTopKKernelVarlen.indexer_topk_kernel_per_row`, "Phase 3: Output phase"):

```python
vecsize_out = cutlass.const_expr(
    min(
        self.top_k,
        cute.ceil_div(self.top_k, self.num_threads_per_cta),
        self.num_copy_bits // self.dtype.width,
        # TODO: only tested for float32. need to check for other dtypes.
        2,
    )
)
assert self.top_k % vecsize_out == 0
```

`vecsize_out` resolves to 2 whenever `top_k > num_threads_per_cta`, and the width-2 vectorized index/value store below it has no scalar tail (the store loop bound `col < self.top_k // vecsize_out` would silently drop the last element), so odd `top_k` cannot compile. The assert is a correct guard for the codegen, but the parity restriction is undocumented and not validated by `check_support()`.

### Why this is rarely hit

Every major DSA integration passes the model config's fixed `index_topk = 2048` (even), so parity never comes up. It surfaces as soon as a caller clamps `top_k = min(index_topk, visible_keys)` and the visible-key count is odd and below 2048 — e.g. Megatron-core's cuDNN fused DSA path under context parallelism, where a CP rank's zigzag shard sees only a prefix of the keys. That is how we hit the exact repro above (`scores [235, 705]`, `top_k = 705`: a 15k-token sequence at CP32 gives 64 chunks of 235 tokens, and one rank's front segment sees 3 × 235 = 705 keys).

### Environment

Observed with the `nvidia-cudnn-frontend==1.26.0` wheel on B200 (SM100). The assert is unchanged on current `develop`.

I'll open a PR that falls back to scalar output stores for odd `top_k`.

JackRao123:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Cap the DSA indexer_top_k output-store vector width at 1 when `top_k` is odd, so odd `top_k` values JIT-compile instead of failing `assert self.top_k % vecsize_out == 0`. Adds `top_k=705` (odd, greater than threads-per-CTA) to the indexer_top_k test parametrization as a regression case.

## Why

Fixes #406. `check_support()` and the documented constraint permit any `top_k` in `(0, 2048]`, but the output phase selected `vecsize_out = 2` whenever `top_k > num_threads_per_cta`, and the width-2 vectorized index/value store has no scalar tail — so odd `top_k` above that threshold asserted at compile time. Real workloads hit this when clamping `top_k = min(index_topk, visible_keys)` on context-parallel shards (see the issue for the exact `[235, 705]` / `top_k=705` production case).

Scalar stores (`vecsize_out == 1`) are the path every `top_k <= num_threads_per_cta` config already takes; this change extends that path to odd `top_k` rather than adding a tail case to the vectorized writer. Only the output-phase store width changes — the selection phases and all even-`top_k` codegen are untouched.

## Related issues

Fixes #406

None for previously-working configs: `vecsize_out` is unchanged for even `top_k` and for odd `top_k <= num_threads_per_cta`. Previously-asserting odd `top_k` values now compile, using scalar output stores.

## Testing

- `black --line-length 160` on both changed files.
- I don't have an SM90/SM100 machine available in this environment, so I could not run the GPU tests myself. The added `top_k=705` parametrization exercises the previously-asserting configuration in both `test_DSA_indexer_top_k_compile_execute` and `test_DSA_indexer_top_k_wrapper` (default shapes give 256–512 threads per CTA, so 705 previously resolved `vecsize_out = 2` and failed the assert). Happy to iterate on CI results or run additional cases on a B200 and report back.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed top-k attention processing for odd top-k values, improving correctness when results exceed standard vectorized widths.

* **Tests**
  * Expanded coverage to include an additional odd top-k scenario, helping prevent regressions in variable-length attention processing.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_varlen_util.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_varlen_util.py
@@ -965,7 +965,9 @@ class IndexerTopKKernelVarlen:
                     cute.ceil_div(self.top_k, self.num_threads_per_cta),
                     self.num_copy_bits // self.dtype.width,
                     # TODO: only tested for float32. need to check for other dtypes.
-                    2,
+                    # Odd top_k cannot be tiled by width-2 vectorized stores (there is
+                    # no scalar tail), so fall back to scalar stores in that case.
+                    2 if self.top_k % 2 == 0 else 1,
                 )
             )
             assert self.top_k % vecsize_out == 0
--- a/test/python/fe_api/dsa/dsa_utils.py
+++ b/test/python/fe_api/dsa/dsa_utils.py
@@ -43,7 +43,9 @@ DSA_INDEXER_FORWARD_PARAM_MARKS = [
 ]
 
 DSA_INDEXER_TOP_K_PARAM_MARKS = [
-    pytest.mark.parametrize("top_k", [512]),
+    # 705: regression for odd top_k > threads-per-CTA, which used to fail the
+    # width-2 vectorized-store assert at JIT compile time.
+    pytest.mark.parametrize("top_k", [512, 705]),
     pytest.mark.parametrize("next_n", [1]),
     pytest.mark.parametrize("return_val", [True, False]),
 ]


Anerudhan (approved):

Anerudhan:
Thanks @JackRao123  merging

PR #416, opened 2026-07-21
DSA indexer forward: add a lean SM100 fast path for the head_dim=128 / qhead_per_kv_head=64 regime (up to 1.46x)

zkyue:
## What
Adds an opt-in "lean" SM100 kernel for the DSA indexer forward (lightning-indexer scoring), plus a transparent dispatch hook in `indexer_forward_wrapper`: eligible configs route to the lean kernel; everything else falls through to the existing kernel **unchanged**. Escape hatch: `CUDNNFE_DSA_INDEXER_FWD_DISABLE_LEAN`.

## Why
On B200 (SM100), at DeepSeek-V4-like shapes (`head_dim=128`, `qhead_per_kv_head=64`, bf16-in/fp32-out, ratio-causal), the existing indexer-forward spends ~2x the machine instructions of a leaner schedule for the same math (ncu: 41.6M vs 20.3M `sm__inst_executed`, ~80% FMA+ALU -- CLC per-tile address replay and the dual-stage epilogue account for it). A static single-wave persistent grid (`min(sm_count x CTAs/SM, num_tiles)` CTAs, runtime-queried, grid-stride, reversed-LPT order) with an N=256 query tile and register-resident head-sum removes that overhead in the grid-saturated regime:

| Shape (S_q, S_kv) | existing | lean | speedup |
|---|---|---|---|
| 4096, 2048 | 88.6 us | 59.2 us | 1.50x |
| **8192, 4096** | **298.4 us** | **205.7 us** | **1.45x** |
| 12288, 6144 | 666.0 us | 473.6 us | 1.41x |
| 16384, 8192 | 1172.2 us | 833.3 us | 1.41x |

(nsys pure-kernel per-launch median, 50 warmup/100 reps, fp64-oracle-gated both kernels on identical inputs. Sweep across S_q 256-8192, ratio in {2,4}: 1.27-1.48x, no losing point measured. Calibration data is B200/148-SM; the wave gate is runtime-`sm_count`-based.)

## Precision / determinism
Same contract as the existing kernel (bf16 tensor-core, fp32 accumulate, fp32 scores, `-inf` outside the causal window). fp64-oracle rel-L2: lean 1.41e-7 vs existing 1.45e-7. Deterministic run-to-run (no atomics, fixed reduction order). Dispatched output is bitwise-identical to direct lean invocation; bf16-W ingest is bitwise-identical to fp32-W.

## Support surface (checked in `check_support`; everything else -> existing kernel)
SM100 only; uniform-length BSHD (any B); `head_dim=128`, `qhead_per_kv_head=64`, `h_kv=1`; bf16 Q/K, bf16-or-fp32 W, fp32 out; contiguous, 16B-aligned base pointers; `S_q % 4 == 0`, any `S_kv >= 1`; `sm_scale` and `q_causal_offsets` supported (offsets restricted to the int32-safe range); grid-saturation gate `num_m_tiles >= sm_count`. THD/varlen (`cu_seqlens`) intentionally not routed -- the static grid cannot honor per-batch KV offsets; those stay on the existing path.

## Testing
`test/python/fe_api/dsa/test_DSA_indexer_forward_lean.py`: check_support accept/reject matrix, numerics vs reference (B in {1,2,3}, causal offsets incl. int32 boundary, sm_scale, fp32/bf16 W bitwise-equal, ragged tails S_kv in {1,127,129,999} with full-visibility offsets, all-masked windows, run-to-run determinism, dispatch-routing assertions incl. env hatch and non-default-knob fallback). Existing indexer-forward tests pass unchanged. `black -l 160` clean.

## Summary by CodeRabbit

* **New Features**
  * Added a new SM100 “IndexForwardLean” fast path for DeepSeek Sparse Attention `indexer_forward`.
  * Exposed new lean API entries (`IndexerForwardLean` and `indexer_forward_lean_wrapper`) and updated the `indexer_forward` module exports.
  * Workloads meeting the documented constraints are routed to the lean implementation; the legacy path remains the fallback, with an environment-variable escape hatch.

* **Documentation**
  * Updated DSA API docs with specialization rules, dispatch/caching behavior, output guarantees, and error handling.

* **Tests**
  * Added extensive pytest coverage for eligibility, numerics, determinism, offsets/masking, and correct dispatch/fallback routing (including THD/varlen).

### DIFF
--- a/docs/fe-oss-apis/dsa.md
+++ b/docs/fe-oss-apis/dsa.md
@@ -174,16 +174,26 @@ substantially faster than the generic SM100 kernel for the large-`S_q`
 H=64 regime. The lean path is **additive**: any configuration outside its
 support gate keeps using the existing kernel unchanged, and setting the
 environment variable `CUDNNFE_DSA_INDEXER_FWD_DISABLE_LEAN` (to any
-non-empty value) forces the legacy kernel for every configuration.
+non-empty value) forces the legacy kernel for every transparently
+dispatched configuration. Calls made explicitly through
+`indexer_forward_lean_wrapper` do not consult the variable; in particular
+its THD mode has no legacy fallback (THD lean scores use a different
+output layout — see below), so opting a THD call site out of the lean
+path means calling `indexer_forward_wrapper` instead.
 
 The lean gate (`IndexerForwardLean.check_support()`) requires all of:
 
 - `head_dim == 128`, `qhead_per_kv_head == 64`, `H_kv == 1`
-- uniform-length batched BSHD (any `B`); THD/varlen (`cu_seqlens_*`)
-  always uses the legacy path — the lean schedule is a static
-  reversed-LPT single-wave persistent grid built around one uniform
-  triangular work distribution, and per-batch KV storage offsets would
-  break its 128-row-aligned dense KV-tile TMA sweep
+- uniform-length batched BSHD (any `B`) on the transparent wrapper.
+  THD/varlen (`cu_seqlens_*`) is served by the same lean kernel, but only
+  through the explicit `indexer_forward_lean_wrapper`: per-row absolute
+  compressed-KV windows carry the segment isolation and the ratio-causal
+  mask, and the scores come back in a global-compressed-KV-column
+  `(T_q, m_total)` layout (each segment's finite scores sit in its own
+  absolute column block), which intentionally differs from the legacy
+  `(total_q, max_seqlen_k)` local-column layout —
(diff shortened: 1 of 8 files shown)


zkyue:
Closing this as stale. It has been open without review for a while, the branch now conflicts with `develop`, and a ~2.7k-line diff against a moving base is not worth carrying.

The kernel still works. If a lean SM100 indexer-forward fast path for the head_dim=128 / topk envelope is of interest later, I am happy to refile a rebased and narrower version.

PR #421, opened 2026-07-22
DSA: fix SM100 sink normalization for 576/512 backward

jiayus-nvidia:
## Before submitting

  - [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
  - [x] I ran `pre-commit run` and committed any formatting changes.

  ## Affected area

  FE OSS kernels or CuTeDSL

  ## Summary

  - Fold the attention sink into the FlashMLA KV-only LSE for every SM100 head-dimension specialization.
  - Launch the existing `sum_dSink` reduction for the `D_qk=576, D_v=512` path.
  - Add deterministic SM100 regression coverage for both full top-k and per-query `topk_length`, including lengths around the 64-entry tile boundary.

  ## Why

  `same_hdim_kv` is intended to control dimension-dependent layouts and the 64-dimensional RoPE tail. It was also incorrectly used to guard two
  dimension-independent sink operations.

  For `D_qk=576, D_v=512`, this caused:

  1. The backward kernel to reconstruct attention probabilities using the KV-only denominator:

  Z_KV

  instead of the sink-inclusive denominator:

  Z_KV + exp(sink)

  This produced incorrect `dq` and `dkv`, especially for queries with short `topk_length`, where the sink probability is significant.

  2. The existing `sum_dSink` kernel to be skipped entirely, leaving `d_sink` zero.

  This change reuses the existing stable sink-folding implementation and `sum_dSink` kernel. It does not add a PyTorch-side epilogue or duplicate the
  `O * dO` reduction.

  ## Related issues

  Fixes #418.

  Related to #405.

  ## API and compatibility impact

  No public API changes.

  The existing `D_qk=D_v` path is unchanged semantically. The `576/512` specialization now performs the same sink normalization and sink-gradient
  reduction as the equal-dimension path.

  ## Testing

  Tested on NVIDIA B300, SM103:

  - `test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py`
  - 7 passed
  - 1 skipped (SM90-only test)
  - All four existing sparse-attention backward parameter combinations passed, including the two previously failing `576/512` cases.
  - Added regression coverage for:
  - full top-k without `topk_length`
  - `topk_length = [1, 63, 64, 65, 512]`
  - `dq`, `dkv`, and `d_sink`
  - `black --check --fast --line-length 160 ...`
  - `git diff --check`

## Summary by CodeRabbit

[...]

PR #429, opened 2026-07-23
Fix stream-ordering and input-validation holes in the SM100 DSA backward interface

zkyue:
### Before submitting

### Affected area

FE OSS kernels or CuTeDSL

### Summary

Three correctness/hardening fixes for the Python interface of the SM100 DSA
backward (`sparse_attention_backward/_interface_sm100.py`, `api.py`), one
commit each. All three failure modes were confirmed empirically on
develop @3a9ed3f (B200); each commit carries regression tests that fail on
the parent commit and pass with the fix.

1. **Stream-order the allocations with the launch stream** (185728d).
   `flash_attn_bwd_sm100` allocates dq/dkv/d_sink and the two workspaces
   (and makes contiguity copies) with plain torch calls, which enqueue on
   the *ambient* torch stream, while the kernel launches on the
   caller-provided `current_stream`. With a non-default `current_stream`
   the semantically required zero-fills of dkv/d_sink and the workspaces
   are unordered with the kernel. Observed on develop: with the ambient
   stream busy, the zero-fills land after the kernel and the returned dkv
   comes back **all-zero** (3/3 trials; the other interleaving lets the
   kernel accumulate into uninitialized memory). Fix: resolve the stream
   first and scope the normalization/allocation section with
   `torch_stream_context(current_stream)` — the same pattern the
   score_recompute / indexer_forward / indexer_backward interfaces already
   use.

2. **Reject fp16 on SM100** (05c0965).
   The dtype checks (`api.py` `check_support`, interface assert) accept
   fp16, but `FlashAttentionDSABackwardSm100` hardcodes BF16 as its element
   type and never receives the input dtype. Observed on develop: fp16
   inputs compile, run **without any error**, and return silently wrong
   gradients — ~96% of dq elements outside 5e-2 tolerances against the
   fp16 autograd reference, on the same harness where bf16 passes. Fix:
   gate the SM100 dispatch to bf16 (the SM90 kernels are
   dtype-parameterized and keep fp16), and update the two in-repo surfaces
   that advertised fp16 unconditionally (`docs/fe-oss-apis/dsa.md`, the
   `benchmark/dsa` backward benchmark — which now skips the
   fp16-on-non-SM90 combination). Plumbing the dtype through the SM100
   kernel would restore fp16 there and is left as a follow-up.

3. **Validate the input contract** (2536e6d).
   Every kernel dimension is derived from `q` and the compiled kernel
   treats dimensions as dynamic values, so mis-shaped companion tensors do
   not fail. Observed on develop: a transposed `dout` or `lse` runs without
   any error and returns silently corrupted gradients (rel-L2 vs the
   correct result: ~1.1 and ~45 respectively). Fix:
   - assert the shape contract of kv/out/dout/lse/attn_sink/topk_idxs/
     topk_length against `q` in the interface (style of the existing
     dq/dkv out-parameter asserts), and require all inputs on `q`'s device;
   - enforce the same contract in `SparseAttentionBackward.check_support`,
     the advertised metadata-only support gate (it previously accepted any
     companion shapes and omitted out/dout/topk_length dtype checks);
   - extend the contiguity normalization (previously q/kv/out/dout/lse
     only) to attn_sink/topk_idxs/topk_length, which today escape to the
     CuTe DSL layer and fail there with low-level stride errors (signature
     mismatch against the shared compile-cache entry on the warm path, a
     leading-stride assert on the cold path);
   - require caller-provided dq/dkv to be contiguous: the compile cache is
     keyed without output strides, so a strided out-parameter would be
     written through the wrong layout (silently copying would break
     out-parameter identity).

### Related issues

None filed; found during an interface audit and confirmed empirically on
develop.

### API and compatibility impact

- **Default-stream path is numerically unchanged**: on fixed seeded inputs
  (S=512, S_kv=2048, H=64, D=576, topk=128; both with and without
  topk_length), dq and d_sink are **bitwise-identical** before/after this
  PR in both configurations. dkv is an FP32 atomic accumulation followed
  by BF16 storage and is not bitwise-stable even against itself
  (base-vs-base rerun: rel-L2 9.2e-6 / 80 of 1.18M elements differ without
  topk_length, 8.6e-6 / 27 with); base-vs-fixed sits inside that jitter
  (7.5e-6 / 87 and 1.5e-6 / 21 respectively).
- fp16 on SM100 now raises `ValueError` from `check_support` /
  `AssertionError` from the interface instead of silently returning wrong
  gradients. SM90 fp16 acceptance is unchanged (code-inspected; we had no
  SM90 machine to run it, maintainers' CI covers that arch).
- Mis-shaped companion tensors now fail loudly (named assert at the
  interface, `ValueError` from `check_support`) instead of running with
  corrupted results; non-contiguous attn_sink/topk_idxs/topk_length are
  normalized instead of failing with DSL-level stride errors;
  non-contiguous caller-provided dq/dkv are rejected.
- The SM90 interface (`_interface_sm90.py`) shares the
  allocation-before-stream-resolve pattern and the narrower contiguity
  normalization; it is left untouched here (no SM90 hardware available to
  validate a change) and could receive the same treatment as a follow-up.

### Testing

All on B200 (SM100), CUDA 13.3 toolchain, nvidia-cutlass-dsl 4.6.1, torch 2.13.

1. **New regression tests**:

   ```
   test_DSA_sparse_attention_backward_nondefault_stream_zero_init_ordering
   test_DSA_sparse_attention_backward_rejects_fp16_on_sm100
   test_DSA_sparse_attention_backward_noncontiguous_aux_inputs   (cold + warm cache paths)
   test_DSA_sparse_attention_backward_cross_shape_validation
   test_DSA_sparse_attention_backward_check_support_validates_contract
   ```

   All five fail on the parent commit (the stream test with dkv returned
   all-zero; the fp16 and contract tests because nothing raises; the
   noncontiguous test with a DSL stride error) and pass with the fixes.
   The stream test parks the ambient stream on `torch.cuda._sleep` and is
   marked `gpu_exclusive` (a concurrently busy GPU could mask the
   ordering violation).

2. **Full DSA suite** `pytest test/python/fe_api/dsa -q`:
   before: `4 failed, 26 passed, 4 skipped` — after: the **same 4** failed,
   `31 passed, 4 skipped`. The pre-existing failures
   (2× `test_DSA_indexer_top_k[705]`,
   2× `test_DSA_sparse_attention_backward_wrapper[...576...]`) fail
   identically on the unpatched base; the wrapper[576] pair looks related
   to the attn-sink issue addressed by #421.

3. **Default-path bitwise check** as described under compatibility impact.

## Summary by CodeRabbit

* **New Features**
  * Expanded SM100 sparse-attention backward to support both FP16 and BF16 element dtypes.
  * Added stricter head-dimension support and stronger validation of dtype, device placement, contiguity, and cross-tensor shape contracts.
* **Bug Fixes**
  * Improved SM100 stream semantics to ensure correct ordering of gradient/auxiliary zero-initialization relative to kernel launches.
* **Tests**
  * Added SM100-gated tests covering FP16 numerics, non-contiguous auxiliary inputs, stream ordering, and contract enforcement (including rejecting unsupported head dimensions).

### DIFF
--- a/benchmark/dsa/README.md
+++ b/benchmark/dsa/README.md
@@ -57,8 +57,7 @@ Options:
 - `--topks` — comma-separated top-k values.
 - `--nheads` — number of query heads (default 64).
 - `--head-dim` — QK head dim, `512` or `576`; `head_dim_v` is derived (512).
-- `--dtype` — `bfloat16` (default) or `float16` (`float16` is SM90-only;
-  the SM100 backward kernel is BF16-only and the combination is skipped).
+- `--dtype` — `bfloat16` (default) or `float16`.
 - `--no-attn-sink` — disable the attention sink (passes `-inf` sink logits).
 - `--no-topk-length` — omit the `topk_length` tensor. Kernels with and
   without `topk_length` are different compiled variants; the default
--- a/benchmark/dsa/README.md
+++ b/benchmark/dsa/README.md
@@ -57,7 +57,8 @@ Options:
 - `--topks` — comma-separated top-k values.
 - `--nheads` — number of query heads (default 64).
 - `--head-dim` — QK head dim, `512` or `576`; `head_dim_v` is derived (512).
-- `--dtype` — `bfloat16` (default) or `float16`.
+- `--dtype` — `bfloat16` (default) or `float16` (`float16` is SM90-only;
+  the SM100 backward kernel is BF16-only and the combination is skipped).
 - `--no-attn-sink` — disable the attention sink (passes `-inf` sink logits).
 - `--no-topk-length` — omit the `topk_length` tensor. Kernels with and
   without `topk_length` are different compiled variants; the default
--- a/benchmark/dsa/benchmark_dsa_sparse_attention_backward.py
+++ b/benchmark/dsa/benchmark_dsa_sparse_attention_backward.py
@@ -263,9 +263,6 @@ def main():
     if major not in (9, 10):
         print(f"SKIP: DSA backward requires a Hopper (SM90) or Blackwell (SM100) GPU, found SM{major}{minor}")
         return
-    if args.dtype == "float16" and major != 9:
-        print("SKIP: float16 is only supported by the SM90 DSA backward kernel (the SM100 kernel is BF16-only)")
-        return
     if DSA is None:
         print(f"SKIP: cudnn is not importable ({_CUDNN_IMPORT_ERROR})")
         print("Install with: pip install nvidia-cudnn-frontend[cutedsl]")
--- a/benchmark/dsa/benchmark_dsa_sparse_attention_backward.py
+++ b/benchmark/dsa/benchmark_dsa_sparse_attention_backward.py
@@ -263,6 +263,9 @@ def main():
     if major not in (9, 10):
         print(f"SKIP: DSA backward requires a Hopper (SM90) or Blackwell (SM100) GPU, found SM{major}{minor}")
         return
+    if args.dtype == "float16" and major != 9:
+        print("SKIP: float16 is only supported by the SM90 DSA backward kernel (the SM100 kernel is BF16-only)")
+        return
     if DSA is None:
         print(f"SKIP: cudnn is not importable ({_CUDNN_IMPORT_ERROR})")
         print("Install with: pip install nvidia-cudnn-frontend[cutedsl]")
--- a/docs/fe-oss-apis/dsa.md
+++ b/docs/fe-oss-apis/dsa.md
@@ -110,7 +110,7 @@ Backward pass for DeepSeek Sparse Attention. Expects the forward outputs
 (`out`, `lse`) from FlashMLA (or the PyTorch reference).
 
 - **Inputs**
-  - `q`: `(total_S_q, H, D)` BF16 (FP16 additionally accepted on SM90; the SM100 kernel is BF16-only)
+  - `q`: `(total_S_q, H, D)` BF16/FP16
   - `kv`: `(total_S_kv, D)` (K = V; MQA)
   - `out`, `dout`: `(total_S_q, H, D_v)`
   - `lse`: `(total_S_q, H)` FP32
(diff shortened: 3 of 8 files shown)


Anerudhan:
Thanks @zkyue for the contribution. 
Have asked the relevant folks to review your contribution.

jiayus-nvidia:
Thanks @zkyue for tracking these issues down and adding the regression coverage. I reviewed the three commits and ran focused checks on B200.

The stream-ordering fix looks correct to me.

For FP16 on SM100, I would prefer enabling it rather than rejecting it. I prepared a small commit based directly on the current head of this PR:

https://github.com/jiayus-nvidia/cudnn-frontend/commit/bdbb731cfca6ece90c534d12b6f674480563aee6

Both cache layers already include the dtype, so the change only threads the mapped input dtype into FlashAttentionDSABackwardSm100 and replaces the hardcoded BF16 element type. It also restores the FP16 documentation/benchmark behavior and replaces the rejection test with an FP16 numerical regression.

I tested D=512 on B200 both with and without topk_length; the dq/dkv/d_sink relative L2 errors against the FP32 autograd reference were approximately 2.5e-4–3.0e-4. The existing BF16 and stream/contract regression tests also pass.

Please feel free to cherry-pick bdbb731c if this direction looks good.

zkyue:
Thanks @jiayus-nvidia — I verified bdbb731 independently on a B200 before taking it, and everything checks out:

- Both cache layers are indeed already keyed on dtype (`api.py:196` and the `compile_key` in `_interface_sm100.py:165`), and the diff only threads the mapped dtype through to `FlashAttentionDSABackwardSm100` — accumulation stays FP32 and the stream-ordering/contract paths from the earlier commits are untouched.
- Re-measured the FP16 rel-L2 against an FP32 autograd reference (D=512, with and without `topk_length`): dq 2.8e-4 / 2.9e-4, dkv 2.5e-4 — consistent with your numbers.
- BF16 output is byte-identical before/after the commit, and the stream/contract regression tests stay green.

Cherry-picked onto the branch with your authorship preserved (now `4d90c19`). Since the head SHA changed, CI will need a re-authorization when convenient.

zkyue:
Also addressed the two CodeRabbit findings from the latest scan (both verified real, both in the validation layers this PR owns): `check_support` now validates device placement (all descriptors CUDA + same device, including the optional `topk_length`), and both validation layers now reject `head_dim` outside {512, 576}. Negative tests added; full regression green. Head is now `651e3c1`.

jiayus-nvidia:
Thanks, LGTM.

Anerudhan (approved):

PR #366, opened 2026-07-08
Expose cu_seq_len_q/kv on the sdpa_fp8 python binding

egilliam-nv:
Expose cu_seq_len_q/kv on the sdpa_fp8 python binding, and enable testing of cu_seq_len_q/kv and ragged offset multiplier for fp8.

The unified-engine FP8/MXFP8 forward (cuDNN 9.25+) accepts cumulative sequence lengths, and the C++ API has supported them on the fp8 node since 1.25 (`SDPA_fp8_attributes` aliases `SDPA_attributes`), but the python `sdpa_fp8` binding hardcoded `cu_seq_len_q/kv` to nullptr. This PR exposes them as kwargs (appended last to preserve positional backward compatibility) so python callers can use fp8 + cu_seq_len; the python-native pygraph capture/replay layer forwards them without changes.

Changes:
- `python/pygraph/{pygraph.h,sdpa.cpp}`: add `cu_seq_len_q/kv` parameters to `PyGraph::sdpa_fp8` and its `m.def`, with docstring entries (requires cuDNN 9.25+ and the UNIFIED implementation). Removes a stale "Deprecated, use sdpa_unified instead" comment — implementation selection is automatic (or explicit via the `implementation` attribute), and python fp8 users are expected to call `sdpa_fp8`.
- `docs/operations/Attention.md`: document `cu_seq_len_q/kv` on the fp16/bf16 C++ and python APIs (missed in #266), the ragged offset multiplier (missed in #290), and the fp8 varlen surface including the new kwargs.
- `test/python/sdpa/fp8.py`: support `is_cu_seq_len` and `with_ragged_offset_multiplier` configs (mirroring `fp16.py`): cu_seq_len graph tensors, token-coarse offsets with per-tensor multipliers on Q/K/V/O, version gating at 9.25.
- `test/python/test_mhas_v2.py`: `test_sdpa_fp8_fwd_ragged_L0` now draws `ragged` / `cu_ragged` / `cu_ragged_mult` with equal weight.

`sdpa_mxfp8` is intentionally untouched: it has no varlen surface at all (no padding mask or seq_len kwargs), so cu_seq_len support there is a separate feature.

Validated against cuDNN 9.25 (`test_sdpa_fp8_fwd_ragged_L0`): H100 10 passed / 22 skipped (pre-existing Hopper config limits), Blackwell 24 passed / 8 skipped (head-dim limits); the passing draws include 23 `is_cu_seq_len=True` and 9 multiplier configs, zero failures.

[...]

PR #425, opened 2026-07-22
test_mhas_v2: extend bwd random head-dim coverage to d=256

vedaanta:
## What

Extends the randomized SDPA backward test coverage in `test/python/test_mhas_v2.py` to head dims up to 256:

- The three bwd suites (`test_sdpa_random_bwd_L0`, `test_sdpa_random_bwd_ragged_L0`, `test_sdpa_random_bwd_bias_L0`) already listed `(256,256)` in `with_high_probability`, but their random-draw branch was still capped at `d_qk_max=192, d_v_max=128` — so backward configs with 128 < d <= 256 outside the fixed high-probability shapes were never generated. This raises the caps to 256/256, matching the fwd suites.
- Bumps `test_sdpa_random_bwd_L0` from 256 to 384 seeds to keep sampling density over the larger shape space.

## Why

A head_dim=256 dense backward plan-build failure recently escaped to PyTorch integration testing (`test_cudnn_attention_dense_d256_sm90_sm10x`) because the FE test suites never exercised d=256 backward: v2's random ranges were capped as above, and v1's `test_sdpa_backward` samples d only from [32, 56, 64, 128]. A cuDNN backend fix is in flight; this change closes the test-coverage side.

## Validation

- With this change `test_sdpa_random_bwd_L0` draws `d_qk=d_v=256` in 66/384 configs and d>128 in 206/384 (previously 0 outside the three fixed shapes).
- Sampled (256,256) configs verified end-to-end (build + execute + reference and determinism checks) against a cuDNN build with d=256 bprop support; on backends without support the configs are waived by the existing `cudnnGraphNotSupportedError` handling, so no new failures are introduced on older cuDNN versions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Expanded cuDNN SDPA attention backward test coverage.
  * Increased generated test scenarios from 256 to 384.
  * Added coverage for larger query/key and value hidden dimensions, up to 256.

PR #430, opened 2026-07-23
Support mixed-form sequence lengths in SDPA forward (cuDNN 9.25+)

egilliam-nv:
## Before submitting

## Affected area

- C++ frontend API or graph construction (SDPA support surface, diagonal band mask)
- Python API or bindings (SDPA test harness)

## Summary

Allow the Q and KV sides of SDPA forward to independently choose their sequence-length representation: per-batch (`seq_len_*`) or cumulative (`cu_seq_len_*`). Previously the two forms were mutually exclusive across the whole operation.

- `sdpa_support_surface`: validation is now per-side (exactly one form per side; both sides provided together); the paged requirement accepts either form per side; mixed forms are gated on cuDNN 9.25+ in the UNIFIED surface. A `has_input` helper is hoisted for the presence checks.
- `diagonal_band_mask`: the node's metadata seqlen check is relaxed to per-side exclusivity, matching the SDPA rules.
- Tests: per-side form control in the SDPA test harness (`cu_seq_len_sides`) plus a deterministic mixed-form test (`test_sdpa_mixed_seq_len_forms_L0`) with non-uniform per-batch lengths, including bottom-right-causal cases that guard the backend's band-mask alignment derivation.

## Why

Paged-attention integrations naturally hold a cumulative Q prefix sum (`qo_indptr`) alongside per-batch KV lengths. The old same-form-both-sides rule forced them to materialize a KV-side prefix sum purely to satisfy the pairing constraint. Permitting mixed forms removes that unnecessary conversion.

Mixed forms require the per-side sequence-length support added to the cuDNN backend in 9.25, so they are gated on `min(compiled, runtime) >= 92500` in the UNIFIED support surface and rejected with a clear error below that.

## Related issues

Backend support merged separately (cuDNN backend MRs, internal). Related to #366 (expose `cu_seq_len_q/kv` on the `sdpa_fp8` Python binding).

No public API additions or removals. Purely relaxes an existing validation constraint on the `cu_seq_len_*` / `seq_len_*` SDPA inputs; all previously-valid graphs remain valid. New capability (mixed forms) is version-gated to cuDNN 9.25+ and rejected with a clear error below that.

## Testing

On H100 (sm90) against a cuDNN 9.27 dev build with the backend support:

- `pytest test/python/test_mhas_v2.py -k mixed_seq_len_forms` — 4/4 pass (both mixed directions x {top-left, bottom-right-causal}).
- `pytest test/python/test_sdpa_edge_cases.py` — 10/10 pass (no harness regression).
- `pytest test/python/test_mhas_v2.py::test_sdpa_random_fwd_L0` (slice) — 203 passed, 9 skipped.
- Standalone end-to-end check: non-paged mixed (both directions), and paged ragged-Q + page tables with `cu_seq_len_q` + `seq_len_kv` and bottom-right causal, all match the both-per-batch reference.

Below cuDNN 9.25 the mixed-form tests skip via the version gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined SDPA sequence-length validation to detect conflicts per side (per-batch vs cumulative) with more precise errors.
  * Strengthened padding-mask and paged-attention validation to require consistent sequence-length availability across Q and KV.
  * Improved implementation-compatibility handling for mixed Q/KV sequence-length representations.
* **Tests**
  * Added deterministic test coverage for mixed-form sequence lengths (cumulative on Q or KV) across alignment and right-bound scenarios.
* **Documentation**
  * Updated Python bindings, operation docs, and sample guidance to clarify `cu_seq_len_*` rules, mixed-form support, and version/implementation requirements.


egilliam-nv:
@cudnn-ci-bot run

egilliam-nv:
Thanks @coderabbitai.

- **UNIFIED path**: done — the test now sets `implementation=cudnn.attention_implementation.UNIFIED` explicitly (both the `SDPATestConfig` and the nested `ExecConfig`), so it no longer relies on `AUTO` resolving to it.
- **Black on `RandomHiddenDimSize` (L192/L481/L654)**: skipping — those lines are not part of this PR's diff (our changes are confined to the mixed-form test near L1024+ and the per-side plumbing in `sdpa/`). They're pre-existing long lines introduced by #425's `d=256` coverage; `black --line-length 160 --check` reports our changed files clean. Better addressed in a separate formatting change so this PR stays scoped.

egilliam-nv:
@cudnn-ci-bot run

egilliam-nv:
Good catch on the stale sample doc — fixed in 62424a9b6.

`fp16_fwd_with_cu_seq_len.cpp` no longer claims cu_seq_len_q/kv must both be set or that cu_seq_len_* are mutually exclusive with seq_len_*; it now documents the per-side rules (each side independently uses one form; both sides present but forms may differ; mixed forms require 9.25+, plain cu_seq_len on both sides remains 9.24+).

I also swept the same now-stale wording out of the parallel parameter docs while here: `docs/operations/Attention.md` (C++ and Python sections) and the sdpa pybind docstrings in `python/pygraph/sdpa.cpp`.

egilliam-nv:
@cudnn-ci-bot run

egilliam-nv:
Thanks — addressed in c28249219.

- **Regular `use_padding_mask` summary** (`docs/operations/Attention.md`): no longer says "Must also provide seq_len_q and seq_len_kv"; now accepts either per-batch or cumulative lengths on each side.
- **Regular `sdpa` pybind docstring** (`python/pygraph/sdpa.cpp`): enriched the `cu_seq_len_q/kv` entries with opposite-side pairing, `use_padding_mask=True`, and the mixed-form 9.25 requirement, to match the detailed docs.
  - Note on the site labels: `sdpa.cpp:1307-1308` is actually the **`sdpa_fp8`** docstring (the regular `sdpa` one is at 1141) — I updated both.

One deliberate divergence from the suggestion on the FP8 summary: I did **not** state 9.24+ for cumulative forms there. The FP8 unified engine itself requires cuDNN 9.25+, so cu_seq_len on the FP8 path needs 9.25+ in *any* form; the 9.24-base / 9.25-mixed split applies only to fp16/bf16. The previous doc commit had incorrectly carried 9.24 into the FP8 docs — that's now corrected to 9.25+.

egilliam-nv:
Addressed in a1772cb4e — but shortened rather than wrapped, for a structural reason.

These `cu_seq_len_q/kv` lines live inside a single `R"pbdoc(...)pbdoc"` raw string, and the block documents **one parameter per line**. A source-level line break inside a raw string becomes a literal newline in the rendered `help()` text, so wrapping would split a parameter description mid-sentence in the output (and would be inconsistent with every other entry in the block — 80+ of these docstring lines already exceed 120 columns, and clang-format leaves raw-string contents untouched).

Instead I condensed the wording so the lines are no longer outliers (~550 → ~370 chars, in line with existing long entries like `dropout`/`generate_stats`), while keeping all the constraints: same-side exclusivity, opposite-side pairing, `use_padding_mask=True`, and the version floors (regular 9.24+ / mixed 9.25+; fp8 9.25+).

vedaanta (approved):

PR #433, opened 2026-07-24
DSA backward SM100: reject topk_length entries < 1 (a zero entry hangs the kernel)

zkyue:
## Before submitting

- [x] I ran the repo's Python formatting hook (black 26.3.1 `-l 160`) and committed any changes — files unchanged. (clang-format n/a: Python only.)

## Affected area

FE OSS kernels or CuTeDSL

## Summary

`sparse_attention_backward` accepts a per-query `topk_length`, but the SM100
kernel has no defined behavior for a zero entry. In `dsa_bwd_sm100.py` the
per-token tile count is

```python
topk = mTopkLength[token_idx]
...
tile_count = cute.ceil_div(topk, self.block_tile)
```

with no zero guard. In practice a `topk_length` containing zero entries
**hangs the launch**: on the configuration we measured, the call never
returns — 100% SM utilization until the process is SIGKILLed — while the
identical call with every length >= 1 completes normally. Code inspection
suggests a zero-tile pipeline deadlock (the warp-specialized stages have no
tiles to hand through for that token), and that a zero-tile token reaching
the dq epilogue would store accumulator TMEM that no MMA has written; the
mathematically correct dq for a query attending to nothing is exactly 0.

This PR adds minimal interface validation, with no kernel changes:

- `_interface_sm100.flash_attn_bwd_sm100` raises `ValueError` when any
  `topk_length` entry is `< 1`. Zero entries are the measured hazard;
  negative values are rejected by the same check. The check needs a
  device-to-host sync, so it runs **outside CUDA graph capture only** —
  under capture the caller must uphold the precondition, which the
  docstrings state explicitly;
- the precondition is documented on the interface and on
  `sparse_attention_backward_wrapper`;
- a test locks the guard in (`0` and a negative value both raise).

We deliberately did **not** implement kernel-side support (skipping the
pipeline handshake and zero-filling dq for zero-tile tokens): that is a
semantic decision about whether empty rows are supported input, and it
touches the same kernel region #421 is currently changing. If you would
rather support `topk_length == 0` in the kernel, we are happy to follow up
with that variant instead — this guard simply becomes removable in that
world.

## Why

Repro on unmodified develop @3a9ed3f (B200, CUDA 13.3, nvidia-cutlass-dsl
4.6.1, torch 2.13): s_q=128, s_kv=4096, h=64, head_dim=head_dim_v=512,
topk=512; `topk_length[{3, 64, 127}] = 0` (3 of 128 rows), all other rows
512:

```
STEP: priming...                                   # all topk_length = 512
priming run finished; max|dq_ok| = 5.71875         # ~40 s incl. first-call JIT
STEP: bad run 1...                                 # topk_length[{3,64,127}] = 0
<never returns — killed after 66 minutes>
```

While hung: GPU at 100% utilization, host thread parked in
`torch.cuda.synchronize()` (py-spy), no forward progress. Reproduces from a
fresh process. With this PR the same call raises `ValueError` up front.

The current test generator floors lengths at 1
(`torch.randint(1, topk_k + 1, ...)`), so nothing in CI exercises 0 — the
case is reachable only from user input, which is exactly the case that
should fail loudly rather than hang the device.

## Related issues

None filed; found while auditing the DSA OSS kernel contracts. Adjacent but
independent of #421 (which changes the same kernel's sink normalization and
extends the same test file — this PR touches only the Python interface layer
plus one appended test; trivial to rebase in either order).

- Calls with any `topk_length` entry `< 1` now raise `ValueError` outside
  CUDA graph capture (previously: zero entries hung the device on the
  configuration we measured). Under capture the check cannot run and the
  documented precondition applies. Calls satisfying the precondition are
  unaffected in behavior.
- Cost on valid calls: one `(topk_length < 1).any()` reduction + D2H sync
  per call on the SM100 path (skipped under capture). If that is unwanted on
  hot paths, we can gate it behind an env var or debug flag instead —
  maintainer preference.
- SM90 path: not investigated (we have no SM90 evidence either way), so the
  guard is scoped to the SM100 interface where the failure is demonstrated.

## Testing

Environment: B200 (SM100), CUDA 13.3, nvidia-cutlass-dsl 4.6.1, torch 2.13.

- Full `test/python/fe_api/dsa` suite on this branch: **4 failed, 27 passed,
  4 skipped** — the failure set is identical to unpatched develop @3a9ed3f
  (2× `test_DSA_indexer_top_k[705]`, 2×
  `test_DSA_sparse_attention_backward_wrapper[576]`, pre-existing and in
  #421's scope); the +1 passed is the new guard test.
- New test: `test_DSA_sparse_attention_backward_topk_length_zero_raises`
  (the guard raises before any kernel compile/launch, so it runs in ~1 s).
  The hang itself was demonstrated with the separate repro configuration
  above, not by this test.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Added SM100-specific pre-launch validation for sparse-attention backward to reject `topk_length` entries less than 1 (including `0` and negatives), raising a clear `ValueError`.
  * Clarified that this validation is skipped during CUDA graph capture due to undefined behavior for `topk_length == 0`.
  * Improved behavior guidance when invalid values are provided outside capture (callers should prevent empty rows upstream).

* **Tests**
  * Added regression coverage for invalid `topk_length` rejection.
  * Added capture-aware tests, including scenarios with an explicitly provided non-current launch stream.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm100.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm100.py
@@ -8,7 +8,7 @@ import cutlass
 import cutlass.cute as cute
 
 from cudnn.deepseek_sparse_attention.utils.compiler import compile_options
-from cudnn.deepseek_sparse_attention.utils.runtime import resolve_stream
+from cudnn.deepseek_sparse_attention.utils.runtime import resolve_stream, torch_stream_context
 from cudnn.deepseek_sparse_attention.utils.tensor_conversion import to_cute_tensor
 from .dsa_bwd_sm100 import FlashAttentionDSABackwardSm100
 
@@ -57,9 +57,11 @@ def flash_attn_bwd_sm100(
 
     Raises:
         ValueError: if any entry of ``topk_length`` is < 1 (validated outside
-            CUDA graph capture only -- during capture the required
-            device-to-host sync is not possible, and the caller must uphold
-            the precondition). A row with ``topk_length == 0`` gives the
+            CUDA graph capture only -- capture is detected on the resolved
+            launch stream, i.e. the explicit ``current_stream`` when one is
+            given, and the required device-to-host sync runs on that stream;
+            during capture the sync is not possible, and the caller must
+            uphold the precondition). A row with ``topk_length == 0`` gives the
             kernel zero KV tiles to process, for which it has no defined
             behavior: on the configuration we tested (B200, head_dim 512,
             topk 512) the kernel never terminates -- the launch hangs at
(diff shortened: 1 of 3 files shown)


Anerudhan:
Thanks @zkyue for the contribution. 
Have asked the relevant folks to review your contribution.

jiayus-nvidia:
Thanks @zkyue for identifying and carefully reproducing this issue. After exploring the implementation options, I opened #439 as a kernel-side alternative that supports empty rows directly, avoids the per-call D2H synchronization, and behaves consistently during CUDA graph capture.

zkyue:
Closing in favor of #439, which supports empty rows inside the kernel — no host-side synchronization and consistent behavior under CUDA graph capture. Thanks for the quick turnaround!

PR #435, opened 2026-07-24
Remove alibi, score_max/sum_exp, and dropout randomization from test_…

vedaanta:
…mhas_v2

Drop is_alibi, with_score_max, with_score_sum_exp, and is_dropout from the randomized SDPA test configurations. The plumbing in the sdpa test package is kept; only the test file stops exercising these options.

## Before submitting

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attention operation reliability by refining randomized test coverage across standard, ragged, paged, and specialized scenarios.

* **Tests**
  * Added coverage for ALiBi masking with causal constraints.
  * Added validation for score maximum and score-sum-exponential outputs.
  * Added forward and backward coverage for dropout with float16 inputs.


vedaanta:
@cudnn-ci-bot run

Anerudhan (changes requested):
I am hesitant to remove this. New engines can return not supported. As these are tested into upstream integrations.

vedaanta:
@Anerudhan Agreed that we have users for these, but we do already run TE and PyT tests directly.
In case something breaks for dropout/max/sum_exp/alibi, TE/PyT tests will flag it.

Given that we want to focus on important fusions, current testing surface only clobber stats for all the new kernels we are bringing up.

---

I will add a separate individual test for all these fusions. Separate from test_mhas.

vedaanta:
Added `test/python/test_sdpa_custom_features.py` with dedicated deterministic coverage for the features removed from the randomized configs: ALiBi, `score_max`/`score_sum_exp`, and dropout — each with a fwd and bwd case, so basic lowering for these fusions stays exercised. Verified locally on SM100 with cuDNN 9.30: 6/6 pass.

vedaanta:
@cudnn-ci-bot run

Anerudhan (approved):

PR #439, opened 2026-07-27
DSA backward SM100: support zero-length top-k rows in the kernel

jiayus-nvidia:
## Summary

  Add native SM100 kernel support for sparse-attention backward rows whose
  `topk_length` is zero.

  The kernel now detects an empty row before initializing any asynchronous
  pipeline or allocating TMEM. All threads in the CTA cooperatively write zero
  to the corresponding `dQ` row and then exit. The row contributes nothing to
  `dKV` or `d_sink`, matching the mathematical gradient of a query attending
  to no KV entries.

  Malformed negative lengths take the same defensive early-exit path so they
  cannot enter the zero-tile pipeline path.

  ## Why

  The SM100 backward kernel previously computed:

  ```python
  tile_count = cute.ceil_div(topk, self.block_tile)
  ```

  For topk == 0, this produces zero tiles, but the warp-specialized pipeline
  assumes that at least one tile is handed between its producer and consumer
  stages. On B200 this can deadlock the launch at 100% SM utilization. If
  execution reached the epilogue, dQ could also be written from TMEM
  accumulators that no MMA instruction had initialized.

  Handling the row inside the kernel:

  - avoids a per-call device reduction and device-to-host synchronization;
  - behaves consistently during CUDA graph capture;
  - gives empty rows a defined and mathematically natural result;
  - leaves the existing positive-topk_length path unchanged.

  ## Implementation

  For each query-token/head-block CTA, the kernel reads the CTA-uniform
  topk_length before setting up pipelines or TMEM.

  When the value is nonpositive, it:

  1. writes zero to every valid dQ element owned by the CTA;
  2. skips all KV, MMA, reduction, and pipeline work;
  3. exits the CTA before any asynchronous synchronization state is created.

  The normal path reuses the same topk value, so positive rows do not incur
  an additional load.

  - topk_length == 0 is now supported on SM100.
  - The corresponding dQ row is exactly zero.
  - The row contributes zero to dKV and d_sink.
  - Calls with positive lengths are unchanged.
  - Negative lengths are handled defensively as empty rows rather than being
    allowed to enter the deadlocking pipeline path.

  ## Related work

  This is an alternative kernel-side resolution for #433.

  Thanks to @zkyue for identifying the zero-tile failure mode, providing the
  B200 hang reproducer, and analyzing the likely pipeline deadlock. That
  investigation directly motivated this implementation.

  Unlike the interface guard proposed in #433, this change supports empty rows
  without adding a host synchronization or capture-specific behavior.

  ## Testing

  Tested on an NVIDIA B200 (SM100):

  PYTHONPATH=/code/github/cudnn-frontend/python \
  pytest -q -rs \
    fe_api/dsa/test_DSA_sparse_attention_backward.py \
    -k sm100_zero_topk_length

  Result:

  2 passed, 6 deselected

  Coverage includes:

  - mixed lengths around the 64-row tile boundary;
  - zero and defensive negative lengths;
  - all-empty inputs;
  - head dimensions 512 and 576;
  - full and partial head CTAs;
  - caller-provided dQ and dKV buffers;
  - numerical comparison with the PyTorch reference;
  - exact-zero checks for all-empty gradients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sparse attention backward processing for empty or invalid attention rows.
  * Empty rows now produce zero gradients without affecting valid rows.
  * Improved handling of invalid indices in empty rows and prevented potential GPU synchronization issues.

* **Tests**
  * Added regression coverage for zero and negative attention lengths across multiple configurations.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100.py
@@ -788,6 +788,25 @@ class FlashAttentionDSABackwardSm100:
 
         max_seqlen_q, max_seqlen_kv, head_dim, (num_heads, batch_size) = problem_shape
 
+        if cutlass.const_expr(mTopkLength is not None):
+            topk = mTopkLength[token_idx]
+        else:
+            topk = mTopkIdxs.shape[0]
+
+        # topk is CTA-uniform (one value per query token). Handle an empty
+        # sparse row before initializing any async pipeline or allocating
+        # TMEM: the row contributes nothing to dKV and its dQ tile is zero.
+        # Treat malformed negative lengths as empty as well so they cannot
+        # enter the same zero-tile pipeline path.
+        if topk <= 0:
+            for linear_idx in cutlass.range(tidx, self.head_dim * self.block_tile, self.threads_per_cta):
+                head_offset = linear_idx // self.head_dim
+                dim_idx = linear_idx % self.head_dim
+                head_idx = head_block_idx * self.block_tile + head_offset
+                if head_idx < num_heads:
+                    mdQ[dim_idx, head_idx, (token_idx, batch_idx)] = mdQ.element_type(0.0)
+            cute.arch.nvvm.exit()
+
         if warp_idx == self.load_warp_id:
             cpasync.prefetch_descriptor(tma_atom_Q)
             cpasync.prefetch_descriptor(tma_atom_dO)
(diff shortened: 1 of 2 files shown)


zkyue:
Thanks @jiayus-nvidia — this is a better resolution than the interface guard in #433. Handling the empty row inside the kernel avoids the per-call D2H sync, keeps CUDA-graph capture behavior consistent, and the early exit before any pipeline/TMEM setup addresses the deadlock at its root.

I also looked at the sm100 `score_recompute` path while reviewing this: its block-skip loop participates in the mbarrier handshake for every skipped slot, so it is already safe for zero-length rows — with this PR the whole family is closed for empty rows.

Closing #433 in favor of this.

Anerudhan:
@cudnn-ci-bot run

Anerudhan (approved):

PR #445, opened 2026-07-28
DSA indexer_top_k: exclude OOB tile-fill lanes from radix selection

### LINKED ISSUE #446: [Bug]: DSA indexer_top_k: OOB tile-fill lanes counted as candidates — data-dependent illegal memory access when scores exceed fp16 range
### Before submitting

- [x] I searched existing issues and did not find a duplicate.

### Affected area

FE OSS kernels or CuTeDSL

### Description

`DSA.indexer_top_k_wrapper` / the CuTe-DSL radix top-k kernel (`indexer_top_k_varlen_util.py`, `IndexerTopKKernelVarlen.indexer_topk_kernel_per_row`) reads each row in fixed-width vector tiles and fills the final partial tile's out-of-bounds lanes with `-inf` via `_fill_oob` so the predicated copy is memory-safe — but the histogram and candidate-collection loops then iterate **every** fragment lane without a bounds check, counting those phantom `-inf` lanes as real elements.

This is harmless while a row's top-k threshold sits above the phantom bin (phantom `-inf` lanes sort last and are dropped). However, `to_coarse_key` derives the coarse radix bin from the **fp16 conversion** of the fp32 score, so every score below fp16's −65504 minimum collapses into the fp16 `-inf` bin — the same bin the phantom lanes occupy. When a row's top-k threshold lands inside that bin (fewer than `top_k` values above ~−65504), the threshold bin's candidate list becomes `real_count + up-to-tile-width phantom lanes`:

- in the `large_occupancy` compile (`num_rows > 148`), the per-row candidate buffers (512-entry smem + `num_cols` gmem) overflow → out-of-bounds shared/global writes → `cudaErrorIllegalAddress`;
- in any configuration, phantom lanes can be selected as winners, yielding silently out-of-range indices.

### Steps to reproduce

```python
import torch
from cudnn import DSA

# 149 rows (> 148 selects the large_occupancy compile) x 4310 cols.
scores = torch.empty(149, 4310, dtype=torch.float32)
scores[1:] = torch.arange(4310, dtype=torch.float32).expand(148, -1)
row0 = torch.empty(4310, dtype=torch.float32)
row0[:1122] = torch.arange(1122, dtype=torch.float32)
# 3000 values below fp16's -65504 floor -> collapse into the fp16 -inf bin,
# so the top-2048 threshold lands inside that bin.
row0[1122:4122] = torch.linspace(-66000.0, -200000.0, 3000)
row0[4122:] = float("-inf")  # beyond seq_len; the kernel must not read these
scores[0] = row0
seq_lens = torch.full((149,), 4310, dtype=torch.int32)
seq_lens[0] = 4122

DSA.indexer_top_k_wrapper(scores.cuda(), seq_lens.cuda(), top_k=2048, next_n=1, return_val=True)
```

### Expected behavior

Returns the top-2048 in-range indices per row.

### Actual behavior

`RuntimeError: CUDA Error: cudaErrorIllegalAddress` from `cute_dsl_topk_wrapper`. `compute-sanitizer` reports `Invalid __shared__ write of size 2 bytes` (the per-row candidate buffers are `Uint16`).

If the 3000 collapsed values are made identical (`row0[1122:4122] = -100000.0`), the same call does not crash but silently returns out-of-range indices (e.g. `4872` with `seq_len 4122`) — phantom OOB lanes selected as top-k winners.

### Why this is rarely hit

The crash needs all of: (a) the >148-row `large_occupancy` compile, (b) a partial final vector tile, and (c) a row whose top-`k` threshold sits in the fp16 `-inf` bin with enough collapsed values that real + phantom candidates exceed the 512-entry smem + `num_cols` gmem per-row candidate capacity. We hit it in production on a GLM-5.2 context-parallel (CP32) training run at 55k-token context — roughly once per several hours of training: watchdog-first async signature, Xid 43, no ECC errors, self-healing on restart. The trigger is the score distribution, not the shape, which is why longer sequences ran fine afterward.

The captured production input reproduces the crash deterministically in a fresh process — same call, same iteration, on both B200 and B300 — so this is data-dependent, not architecture- or scheduling-dependent.

### Environment

`nvidia-cudnn-frontend==1.26.0` wheel, torch 2.11.0+cu128, observed on B200 (SM100) and B300 (SM103). The affected loops are unchanged on current `develop`.

Fix PR: #445 (excludes OOB lanes from the three vectorized histogram/collection loops using the same `aligned_size` bound the predicated copy already computes; L0 regression test included).

JackRao123:
## Before submitting

- [x] I agree to license this contribution under the terms of LICENSE.txt.
- [x] I ran `black --line-length 160` on the changed files.

Fixes #446

## Affected area

FE OSS kernels or CuTeDSL (DSA `indexer_top_k`).

## Summary

The CuTe-DSL radix top-k kernel reads each row in fixed-width vector tiles. The final partial tile's out-of-bounds lanes are filled with `-inf` by `_fill_oob` so the predicated copy is memory-safe — but the histogram and candidate-collection loops then iterate **every** fragment lane without a bounds check, counting those phantom `-inf` lanes as real elements.

This is harmless while a row's top-k threshold sits above the phantom bin, because phantom `-inf` lanes sort last and are dropped. However, `to_coarse_key` computes the coarse radix bin from the **fp16 conversion** of the fp32 score, so every score below fp16's −65504 minimum collapses into the fp16 `-inf` bin — the same bin the phantom lanes occupy. When a row's top-k threshold lands inside that bin (fewer than `top_k` values above ~−65504), the threshold bin's candidate list becomes `real_count + up-to-tile-width phantom lanes`:

- in the `large_occupancy` compile (`num_rows > 148`), the per-row candidate buffers (512-entry smem + `num_cols` gmem) overflow → out-of-bounds shared/global writes → `cudaErrorIllegalAddress`;
- in any configuration, phantom lanes can be selected as winners, yielding silently out-of-range indices.

## Observed in production

A GLM-5.2 context-parallel (CP32) training run on B200s crashed with Xid 43 (`illegal memory access`) mid-`forward_backward` — once in 5 hours, sequence-length-independent, no ECC errors. The crash is **deterministically data-dependent**: the captured score tensor reproduces it in a fresh process, on both B200 and B300, at the same call for the same data. `CUDA_LAUNCH_BLOCKING=1` attributes the fault to this kernel; `compute-sanitizer` reports invalid 2-byte shared writes; bisection isolated the trigger to a single row (seq_len 4122 of width 4310, 2242 values below −65504) plus the >148-row compile.

## Fix

Skip lanes whose tile column coordinate is out of bounds (`col < aligned_size`) in the three vectorized histogram/collection loops — the same bound the predicated copy already computes via `predicate_tile`. The scalar prologue/leftover loops are exact-bounded and unchanged. Selection semantics for in-bounds elements are untouched.

## Testing

- New L0 regression test `test_DSA_indexer_top_k_oob_tile_lanes` deterministically drives a row into the flood regime (149 rows → `large_occupancy`; 4310-wide; a 4122-long row with < `top_k` values above the fp16 negative-overflow point), in two variants: identical collapsed values (fails pre-patch with out-of-range indices) and distinct collapsed values (pre-patch: `cudaErrorIllegalAddress`). Both pass post-patch.
- GPU validation was performed against the v1.26.0 release (the affected version; the patched region is byte-identical on `develop`) on B200 and B300: previously-crashing inputs complete with in-range indices, `compute-sanitizer` reports no invalid accesses, and 12/12 randomized parity trials against `torch.topk` pass (rows 37–862, widths 4097–51720, fp16-overflow and exact-tie distributions, both kernel compiles).
- Kernel microbench, (862, 51720) k=2048: 0.115 → 0.129 ms/call (+12% on a 0.1 ms kernel).

## Related

- Complements #407 (odd `top_k` store width) — same kernel, disjoint code (that fix is the phase-3 output store; this one is phases 1–2 selection).
- #312 fixed a different IMA (int32 scratch overflow); this one is the data-dependent candidate flood.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_varlen_util.py
+++ b/python/cudnn/deepseek_sparse_attention/indexer_top_k/indexer_top_k_varlen_util.py
@@ -525,12 +525,18 @@ class IndexerTopKKernelVarlen:
                     -tXrX.element_type.inf,
                 )
 
+                # Skip out-of-bounds lanes: the last tile can extend past
+                # aligned_size, and its -inf-filled lanes must not be counted
+                # as real elements.
+                cur_tXcX = tXcX[None, None, None, tile_idx]
                 for i in cutlass.range(cute.size(tXrX), unroll_full=True):
-                    bin_val = self.to_coarse_key(tXrX[i])
-                    atomicAdd(
-                        s_histogram.iterator + cutlass.Int32(bin_val),
-                        val_one,
-                    )
+                    col = cur_tXcX[i // vec_size][1] + i % vec_size
+                    if col < aligned_size:
+                        bin_val = self.to_coarse_key(tXrX[i])
+                        atomicAdd(
+                            s_histogram.iterator + cutlass.Int32(bin_val),
+                            val_one,
+                        )
 
             # for initial scalar load part.
             for j in range(tidx, prologue_elems, self.num_threads_per_cta):
@@ -594,11 +600,13 @@ class IndexerTopKKernelVarlen:
                     )
                     for i in cutlass.range(cute.size(tXrX), unroll_full=True):
                         cur_tXcX = tXcX[None, None, None, tile_idx]
-                        bin_val = self.to_coarse_key(tXrX[i])
-                        if bin_val < threshold_bin:
-                            pos = atomicAdd(s_counter.iterator, val_one)
-                            idx = self.index_type(cur_tXcX[i // vec_size][1] + i % vec_size + vec_start)
-                            s_indices[pos] = idx
+                        col = cur_tXcX[i // vec_size][1] + i % vec_size
+                        if col < aligned_size:
+                            bin_val = self.to_coarse_key(tXrX[i])
+                            if bin_val < threshold_bin:
+                                pos = atomicAdd(s_counter.iterator, val_one)
+                                idx = self.index_type(col + vec_start)
+                                s_indices[pos] = idx
 
                 # for initial scalar load part.
                 for j in range(tidx, prologue_elems, self.num_threads_per_cta):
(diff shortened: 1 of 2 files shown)


JackRao123:
Closing as a duplicate of #410, which predates this PR and carries an essentially identical fix (same OOB-lane guards in the same three loops). Production evidence and an additional trigger route (fp16-collapse of scores below −65504 into the `-inf` coarse bin) are noted in the #410 comments. Whichever lands first works for us.

PR #462, opened 2026-07-31
Fix ragged SDPA backward workspace under-allocation for non-token-major stats layouts

YangXu1990uiuc:
## What

A ragged (THD) `sdpa_backward` graph that sets `max_total_seq_len_q` silently returns **all-zero dQ/dK/dV** when the `Stats` tensor is head-major. The forward `O` and `stats` are correct and no error is raised, so this corrupts training without ever surfacing. Present unchanged in 9.22, 9.23, 9.24, 9.25 and 9.26, on both sm90 and sm100.

Head-major `[h, total_q]` is FlashAttention's and PyTorch varlen's `softmax_lse` layout, so this is reachable from exactly the integrations that would reach for `max_total_seq_len` in the first place.

## Root cause

`max_total_seq_len_*` is documented as a workspace-sizing hint. It is not — setting it makes `CompositeSDPABackwardNode` re-lay-out its non-virtual intermediates onto the corresponding user tensor's stride and ragged offset, and then size them from a single stride element:

```cpp
softmax_sum->set_stride(Stats->get_stride());
softmax_sum->set_ragged_offset(Stats->get_ragged_offset());
softmax_sum_size = max_total_seq_len_q.value() * (Stats->get_stride())[2] * sizeof(float);
```

`stride[2]` is the *sequence* stride, so the formula assumes the per-token footprint is `stride[2]`. That is true only for token-major `[t, h, 1]` stats, where `stride[2] == h_q`. For head-major `[h, t]` stats `stride[2] == 1`, and the buffer is under-allocated by exactly `h_q`.

`softmax_sum` then writes `(h_q - 1) * t_q * 4` bytes past its slot. The frontend places the backend engine's own workspace immediately after the FE region (`graph_interface.h`: `engine_workspace = workspace + fe_workspace_size`), and the first word of that region is the persistent scheduler's tile counter. Every CTA claims work with `tile_id = atomicAdd(tile_id_counter, 1)` and retires unless `tile_id < num_tiles`; `tile_id` is `uint32_t`, so any nonzero float bit pattern is astronomically out of range. Every CTA therefore retires without storing anything — no dQ accumulation, no dK/dV TMA store — and the gradients are simply never written. No fault, no error, no hang.

Measured directly, comparing the exact `int32` bits of the clobbered word against the true `softmax_sum` value at the same element from a clean run at the same geometry:

```
observed at engine workspace[0] = -1067984384   (f32 -3.3726807)
true softmax_sum(b=0, h=1, s=0) = -1067984512   (f32 -3.3726501)
difference                      = +128
```

A float bit pattern plus exactly 128 integer increments — one `atomicAdd` per CTA, every one of which then failed the tile bound and retired.

The mask plays no part in the trigger: `TOP_LEFT` and the completely unmasked case fail identically. It only decides whether the clobbering bytes happen to be zero — a fully-masked leading query row has row-sum 0, which is what the memset already put there. That is why a handful of geometries appear to work, and why one intermediate case produces `|dQ| ≈ 6e36` garbage rather than zeros.

## The fix (commit 1)

Size the workspace from every axis instead of from `stride[2]` alone:

```cpp
static int64_t
ragged_workspace_size(int64_t max_total_seq_len,
                      std::vector<int64_t> const& dim,
                      std::vector<int64_t> const& stride,
                      int64_t elem_size) {
    return ((max_total_seq_len - 1) * stride[2] + (dim[1] - 1) * stride[1] + (dim[3] - 1) * stride[3] + 1) *
           elem_size;
}
```

applied to `softmax_sum`, `dQ_accum`, `dK_fullhead` and `dV_fullhead`.

For token-major inputs the new expression is **bit-identical** to the old one — `softmax_sum` gives `(maxT-1)·h_q + (h_q-1)·1 + 0 + 1 = maxT·h_q`, `dQ_accum` gives `(maxT-1)·h_q·d + (h_q-1)·d + (d-1) + 1 = maxT·h_q·d`. Allocations are unchanged for every layout that works today; only non-token-major layouts move.

## Why this was never caught (commit 2)

`test/python/sdpa/fp16.py` only ever allocated token-major ragged stats, the one layout for which the old formula happens to be exactly right. There was no coverage of any other packed `Stats` layout, so the suite was structurally blind to this class of defect.

Commit 2 adds a `ragged_stats_layout` axis to `ExecConfig`, randomized in the ragged fwd/bwd tests, choosing between token-major and head-major. The stats allocation and the stats ragged offset now both derive from `cfg.stride_stats` rather than hardcoding the token-major assumption.

Against the **unpatched** frontend the new axis reproduces the defect immediately — 11 tests fail with `Backward workspace overwritten outside its boundaries` (the suite's existing guard-band check) before the run aborts. The first failing config has `stride_stats = (58880, 7360, 1, 1)`, i.e. sequence stride 1.

## Verification

| | unpatched | with fix |
|---|---|---|
| `-k ragged` (fwd + bwd), sm100 | 11 failures, then abort | 370 passed, 78 skipped |
| head-major direct probe, sm100 | 3/4 all-zero | 4/4 correct |
| head-major direct probe, sm90 | 3/4 all-zero | 4/4 correct |
| token-major workspace sizes | — | byte-identical to unpatched |

Sanity check on the numbers rather than just pass/fail: with the fix, a `TOP_LEFT` 2048/512 case with `max_total_seq_len` set gives `|dQ| = 5.53e5`, exactly matching the same case with `max_total_seq_len` unset; a `BOTTOM_RIGHT` 512/512 case gives `1.976e5` in both.

## Not addressed here

- `docs/operations/Attention.md` still describes `max_total_seq_len_*` as a workspace allocation hint. It changes intermediate tensor layout and can change results, and should say so.
- The formula still assumes `ragged_offset == cum_token × stride[2]`. Offsets carrying inter-sequence padding would still under-allocate, and the frontend cannot detect this since the offsets are device data. This wants either a documented contract or for `max_total_seq_len` to move into the backend attention descriptor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace sizing for ragged attention operations, accounting for full tensor strides and dimensions.
  * Corrected ragged statistics offsets and memory capacities for packed token layouts.
  * Added support for both token-major and head-major ragged statistics layouts.
  * Improved handling of ragged training and statistics tensors across varied sequence configurations.

* **Tests**
  * Expanded randomized attention testing to cover multiple ragged statistics layouts and packed token capacities.

### DIFF
--- a/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
+++ b/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
@@ -1091,6 +1091,19 @@ class CompositeSDPABackwardNode : public NodeCRTP<CompositeSDPABackwardNode> {
     std::shared_ptr<Tensor_attributes> alibi_slopes;
     int64_t alibi_slopes_size = 0;
 
+    // Byte span of a workspace tensor packed (ragged) over the token axis: the sequence axis is
+    // bounded by max_total_seq_len instead of dim[2], every other axis contributes its own extent.
+    // Sizing from stride[2] alone under-allocates any layout whose per-token footprint is not
+    // stride[2] -- e.g. head-major [h, t] stats, where stride[2] == 1.
+    static int64_t
+    ragged_workspace_size(int64_t max_total_seq_len,
+                          std::vector<int64_t> const& dim,
+                          std::vector<int64_t> const& stride,
+                          int64_t elem_size) {
+        return ((max_total_seq_len - 1) * stride[2] + (dim[1] - 1) * stride[1] + (dim[3] - 1) * stride[3] + 1) *
+               elem_size;
+    }
+
     mutable bool has_workaround_padding_mask         = false;  // Will be edited in pre_validate_node()
     mutable int32_t s_q_for_workaround_padding_mask  = 0;      // Will be edited in pre_validate_node()
     mutable int32_t s_kv_for_workaround_padding_mask = 0;      // Will be edited in pre_validate_node()
@@ -1711,8 +1724,10 @@ class CompositeSDPABackwardNode : public NodeCRTP<CompositeSDPABackwardNode> {
             // sized TH1 softmax_sum
             softmax_sum->set_stride(attributes.inputs[input_names::Stats]->get_stride());
             softmax_sum->set_ragged_offset(attributes.inputs[input_names::Stats]->get_ragged_offset());
-            softmax_sum_size = attributes.max_total_seq_len_q.value() *
-                               (attributes.inputs[input_names::Stats]->get_stride())[2] * sizeof(float);
+            softmax_sum_size = ragged_workspace_size(attributes.max_total_seq_len_q.value(),
+                                                     softmax_sum->get_dim(),
+                                                     softmax_sum->get_stride(),
+                                                     sizeof(float));
         } else {
             // sized BHS1 softmax_sum
             softmax_sum->set_stride({h_q * s_q, s_q, 1, 1});
@@ -1921,7 +1936,8 @@ class CompositeSDPABackwardNode : public NodeCRTP<CompositeSDPABackwardNode> {
                 dV_fullhead->set_ragged_offset(attributes.outputs[output_names::dV]->get_ragged_offset());
                 // non virtual dV full head
                 dV_fullhead->set_is_virtual(false);
-                dV_fullhead_size = attributes.max_total_seq_len_kv.value() * dV_fullhead_stride[2] * sizeof(float);
+                dV_fullhead_size = ragged_workspace_size(
+                    attributes.max_total_seq_len_kv.value(), dV_fullhead->get_dim(), dV_fullhead_stride, sizeof(float));
             } else {
                 // sized BHSD dQ_accum
                 dV_fullhead->set_stride({h_q * s_kv * d_v, s_kv * d_v, d_v, 1});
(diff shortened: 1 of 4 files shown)


vedaanta:
@cudnn-ci-bot run

vedaanta (approved):

vedaanta:
@cudnn-ci-bot run backend,frost

YangXu1990uiuc:
Thanks @vedaanta. Triaging the two CodeRabbit comments — I checked both against the code, and I'd like to land this as-is and carry them separately.

**On the test coverage comment (`test_mhas_v2.py:368`) — it's correct.** `is_train` is `not is_infer`, the forward ragged test never sets `is_infer = False`, and `with_score_max`/`with_score_sum_exp` default to `False` and aren't randomized there. So none of the three `set_ragged_offset(stats_ragged_offset)` call sites in `sdpa/fp16.py` execute and `stride_stats` reaches no tensor under test — the randomization only perturbs an unconsumed graph input. Harmless but useless. The backward test, which does set `is_infer = False`, is where the real coverage is, and that half is exercising both layouts as intended. I'll drop the forward-side randomization in a follow-up rather than respin an approved PR for a no-op test line.

**On the partially-packed workspace comment (`scaled_dot_product_flash_attention.h:1105`) — the concern is real, but it's orthogonal to this PR and not introduced by it.** The token-axis term is `max_total_seq_len` in both the old and the new expression, so the exposure is identical before and after. Concretely, at b=4 h=8 t=512:

| stats layout | old | new | |
|---|---|---|---|
| token-major | 4096 | 4096 | identical — no regression |
| head-major | 512 | 4096 | old under-allocates by `h_q`×, which is what this PR fixes |

The underlying issue is that `max_total_seq_len` is documented as the *sum of sequence lengths*, and that quantity cannot bound a partially-packed buffer — for the layout in our own docs (`[0, 4·H·D, 7·H·D]`, `Q = aa00bbb0`) there are five valid tokens but data reaches token index 6.

The suggested remedy of computing a physical-capacity bound isn't available to the frontend at build time: the gaps live in the ragged-offset *device* tensor and aren't known until execute. The only lever here is the host-side `max_total_seq_len`. So the fix is a contract change — document it as the packed *capacity* (the last ragged offset expressed in tokens) rather than the sum of sequence lengths, or add a separate capacity attribute — which deserves its own PR and its own regression test with inter-batch padding, as CodeRabbit suggests. Filing that separately.

YangXu1990uiuc:
<sub>note to self: claude::88367f80-c82c-4c8f-93a2-e66ed5a89e07 — "审计PyTorch中CuDNN Attention相关问题"<br>
cwd /home/scratch.yanxu_gpu/cudnn · workspace /home/scratch.yanxu_gpu/cudnn_pt191576 (ROOTCAUSE_maxtot_zero_grad.md, probe_maxtot2/)</sub>

PR #483, opened 2026-08-05
Fix default-stream race: resolve _get_default_stream(None) to torch's current stream

vedaanta:
## Problem

`APIBase._get_default_stream(None)` returns `cutlass.cuda.default_stream()` (legacy CUDA stream 0). Every `APIBase`-derived `execute()` called without an explicit stream therefore launches its TVM-FFI kernel on stream 0, while the surrounding torch-side ops (input copies, FP8 `amax_*.zero_()` resets, output allocations) run on torch's **current** stream. Under `with torch.cuda.stream(s):` — the exact usage the `execute()` docstrings advertise — the kernel races those ops.

Verified empirically on SM100 (B200): an SDPA forward called on a side stream reads stale inputs 7-8 times out of 8 once JIT warmup no longer masks the window. This also covers the FP8 per-tensor amax reset ordering flagged by CodeRabbit (`amax_o_buf.zero_()` vs kernel launch).

## Fix

Resolve `None` to `torch.cuda.current_stream()` instead, so the kernel and its surrounding torch ops land on the same stream. One change covers all 29 `_get_default_stream` resolution sites (SDPA DSL incl. the FP8 amax path, gemm/cutedsl, NSA). The SM120 execute path already carried this exact fallback locally; this hoists it into the shared helper.

## Tests

`test/python/sdpa/frost/test_sdpa_stream_ordering.py`:
- Unit test: `_get_default_stream(None)` matches `torch.cuda.current_stream()` inside and outside a side-stream context.
- Side-stream ordering tests (d128/d256/d512): poison Q with NaN, enqueue [~1s spin, restore Q, SDPA] on a side stream, require the output to reflect the restored Q. An unordered (stream-0) launch overtakes the restore and reads the poison.

All 4 tests **fail** on develop without this change and **pass** with it (SM100/B200). Full `test/python/sdpa/frost` suite run in progress; result will be posted as a comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA stream handling to respect PyTorch’s currently active stream, helping operations execute in the expected order.
  * Improved FROST SDPA behavior when work is scheduled on non-default CUDA streams, reducing the risk of incorrect results caused by stream-ordering issues.

* **Tests**
  * Added coverage for current-stream selection and side-stream SDPA execution ordering on supported SM100 devices.

### DIFF
--- a/python/cudnn/api_base.py
+++ b/python/cudnn/api_base.py
@@ -526,8 +526,9 @@ class APIBase(ABC):
         """Get default CUDA stream if none provided.
 
         This is a convenience helper to handle optional stream parameters.
-        If a stream is provided, it is returned as-is. If None, the default
-        CUDA stream is returned.
+        If a stream is provided, it is returned as-is. If None, the caller's
+        current PyTorch stream is returned, so kernels stay ordered with
+        surrounding torch ops under ``with torch.cuda.stream(s):``.
 
         :param stream: CUDA stream or None
         :type stream: cuda.CUstream or None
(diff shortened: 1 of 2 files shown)


vedaanta:
Full `test/python/sdpa/frost` regression on this branch, SM100 (B200), torch 2.x cu130: **357 passed, 36 skipped, 5 deselected, 0 failures** in 9:57. The 4 new stream-ordering tests were additionally verified to fail 4/4 with the fix reverted.

vedaanta:
@cudnn-ci-bot run

Anerudhan (approved):

PR #485, opened 2026-08-05
frost(sdpa): causal right-band widening + per-sequence THD bottom-right diagonal on SM100

vedaanta:
Two mask-machinery features for the SM100 FROST SDPA fwd engines, closing the two biggest single-reason routing gaps after bwd (per a test_mhas_v2 mismatch census on SM100: right-band widening solely blocked 220 graphs, THD bottom-right 31+). Supersedes GitLab MR !2308 (closed).

**Causal right-band widening** (`diagonal_band_right_bound > 0`)
- Lowers as `MASK_CAUSAL` with a compile-time `BAND_RIGHT` offset on the diagonal, applied in both `compute_kv_loop_bounds` (hi/lo tile limits) and `apply_mask_chunk` (per-element), so all five SM100 rows — f16 d128/d256/d512, fp8, mxfp8 — serve it (caps flipped).
- The analyzer records the bound as `facts.window_right`; a new `_band_covers_kv_tail` helper accounts for the widened band in the S_kv%128 tail rule (top-left: covered iff `s_q + R <= s_kv`; bottom-right: iff `R == 0`).
- The DSL adapter grows `window_size_right` (validated `>= 0`, requires the causal band); SM120 rejects it explicitly (not plumbed there).

**THD bottom-right per-sequence diagonal**
- New `resolve_seqlen_q` helper reads each sequence's actual Q length from the packed cu_seqlen metadata (the same buffer `_thd_decode` reads) and feeds both the KV-loop bounds and the mask diagonal; it folds out entirely unless `THD_VARLEN && CAUSAL_BOTTOM_RIGHT`.
- The `thd_bottom_right` capability gap flag and its gate are deleted (SM120 already anchored per sequence).

**Validation (B200, cuDNN 9.23, cutlass-dsl internal)**
- `test_mhas_v2` full run: **1951 pass / 0 fail**, routing **526/2874 graphs on FROST (18.3%)**, up from 244/2753 (8.9%) — d128 103→221, d256 86→224, fp8 55→81.
- frost sm100 suites: 447 pass pre-rebase, 390/390 post-rebase. Combo sweep extended with `band`, `band_br`, `band_swa` (dense) and `causal_br`, `band`, `band_br` (THD); analyzer probe tests flipped to accept, plus a negative-bound rejection test.

## Summary by CodeRabbit

* **New Features**
  * Added support for right-side widening in causal attention windows on supported SM100 configurations.
  * Improved bottom-right and variable-length attention handling with per-sequence query lengths.

* **Bug Fixes**
  * Improved masking and KV-bound calculations for causal, sliding-window, and variable-length attention.
  * Added validation for invalid window values and unsupported configurations.

* **Tests**
  * Expanded coverage for right-banded masks, bottom-right alignment, variable-length sequences, and validation scenarios.

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -23,10 +23,6 @@ from cudnn.frost.tile_dsl.constants import (
     DTYPE_E4M3,
     DTYPE_E5M2,
     DTYPE_FP16,
-    MASK_CAUSAL,
-    MASK_NONE,
-    MASK_PADDED,
-    MASK_SWA,
     SCHED_LPT,
     SCHED_NATURAL,
 )
@@ -260,6 +256,13 @@ class SdpaFwdDsl(APIBase):
         # widened by R columns. None = no right band; requires is_causal (the
         # band is the causal diagonal, possibly widened).
         self.window_size_right = window_size_right
+        # The ONE canonical band every adapter lowers (same model as the
+        # analyzer facts / config TemplateParams): per-side offsets from the
+        # diagonal, None = unbounded. is_causal means "right bound 0";
+        # window_size_right widens it (check_support validates it requires
+        # is_causal). The padding mask stays orthogonal (seq_kv_lens_present).
+        self.window_left: Optional[int] = window_size_left
+        self.window_right: Optional[int] = (window_size_right or 0) if self.is_causal else None
         self.scale_softmax = scale_softmax
         self.seq_kv_lens_present = bool(seq_kv_lens_present)
         # Dense padded-Q trim (cuDNN >= 9.14): q rows >= seq_len_q[b] write
@@ -463,8 +466,6 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
 
     def _initialize_implementation(self) -> None:
         self.flavor: Optional[tuple[int, int]] = None
-        self.mask_flags = 0
-        self.swa_window_runtime = 0
         self.thd_stats_head_major = False
         self.thd_stats_head_stride = 0
         self._k_mod = None
@@ -626,9 +627,9 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
             band_right is not None and not self.is_causal,
             "SM100 DSL SDPA: window_size_right widens the causal diagonal and requires is_causal=True",
         )
-        # The kernels' bottom-right diagonal path supports plain causal only:
-        # CAUSAL_BOTTOM_RIGHT requires MASK_CAUSAL and excludes MASK_SWA
-        # (see config._validate_knobs).
+        # The kernels' bottom-right diagonal path excludes a left bound:
+        # bottom_right requires a right bound and rejects window_left
+        # (see config_sm100._validate_params).
         self._value_error_if(
             self.causal_bottom_right and not self.is_causal,
             "SM100 DSL SDPA: causal_bottom_right requires is_causal=True",
@@ -646,19 +647,8 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
             "supported (kernel anchors the BR diagonal at the global S_q, not "
             "seq_len_q[b])",
         )
-        if self.is_causal:
-            self.mask_flags = MASK_CAUSAL | (MASK_SWA if swa_left is not None else 0)
-            self.swa_window_runtime = swa_left if swa_left is not None else 0
-        elif swa_left is not None:
-            self.mask_flags = MASK_SWA
-            self.swa_window_runtime = swa_left
-        else:
-            self.mask_flags = MASK_NONE
-            self.swa_window_runtime = 0
         if self.thd:
             self.seq_kv_lens_present = True
-        if self.seq_kv_lens_present:
-            self.mask_flags |= MASK_PADDED
         # Dense padded-Q trim backstops (engines.lower_dsl_prefill never sets
         # these combinations; a direct caller could).
         self._value_error_if(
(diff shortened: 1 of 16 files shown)


vedaanta:
@cudnn-ci-bot run frost

vedaanta:
Pushed `0171127` fixing the 26 `disallowed mismatches` failures reported on the previous revision.

**Root cause**: develop's new `d192x128` kernel (DSv3 geometry) is registered through the same `_sm100_spec` factory this PR modifies, so it inherited the `right_band_widening` capability claim (and lost the THD bottom-right gate this PR removes) — but the kernel itself, added after this PR was written, had neither the `BAND_RIGHT` diagonal offset nor the per-sequence `seqlen_q` resolve. Graphs with a widened band or THD bottom-right that now route to `sdpa_fwd_prefill_sm100_d192_d128` computed a plain-causal / global-S_q mask instead. (On develop these graphs simply never routed to FROST, which is why they passed there.)

**Fix**: same plumbing as the other five SM100 kernels — `make_cfg_d192` passes `BAND_RIGHT`; the kernel resolves the per-sequence Q length at all 8 `eff_seqlen_kv` sites and feeds it to the KV-loop bounds and the BR diagonal; its two specialized causal mask chunks (top-left / bottom-right fast paths) and its general `apply_mask_chunk` call apply the `BAND_RIGHT` widening.

**Validation on the updated revision**:
- `test_mhas_v2` full run: 2028 pass / 0 related failures, routing **581/3050 graphs on FROST (19.0%)** — the d192 engine now serves 81 graphs. (`test_sdpa_mxfp8_bwd_L0[test82]` failed its determinism check once under 4-way xdist GPU contention and passes serially — unrelated concurrency flake.)
- frost sm100 + analyzer suites: 462 pass.

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
Rebased onto develop head `3f17f5b` (12 commits, including #505's SM120 bwd SWA) and added a second commit unifying the mask parameterization.

**`c3721ad` — one canonical mask model: the diagonal band.** The stack had four spellings of the same mask (sm100 config: `mask_flags` bitmask + `swa_window` + `band_right` + `causal_bottom_right`; sm120 config: `is_causal` + `causal_bottom_right` + `window_size_left`; the adapter: `mask_flags` + `swa_window_runtime`; the analyzer facts: `window_left` + `right_bound` + `bottom_right`). Everything above the kernels now carries only the band — the model the analyzer, FlashAttention, and CUTLASS FMHA (`SlidingWindowMask(window_left, window_right)`) already share:

```python
window_left:  Optional[int]  # None = no lower bound; W keeps kv >= q-W
window_right: Optional[int]  # None = no upper bound; 0 = causal; R widens
bottom_right: bool           # anchors the band's diagonal
```

- The kernels' `MASK_FLAGS` bits are **derived** in `make_cfg_*`, so the trace-time `const_expr` branch folding is byte-for-byte unchanged; the CFG value fields rename `SWA_WINDOW → WINDOW_LEFT`, `BAND_RIGHT → WINDOW_RIGHT`, `CAUSAL_BOTTOM_RIGHT → BOTTOM_RIGHT` (46 sites) along with the shared bounds/mask helper vocabulary.
- The canonical band is resolved once, in the shared adapter base `__init__` — `is_causal` means "right bound 0", `window_size_right` widens it.
- SM120 fwd config joins the same model (`window_right in {None, 0}` validated — widening not plumbed there); its kernel internals keep local names.
- Not touched: the bwd sm120 config (`causal_top_left`, fresh from #505) — suggest aligning it in a follow-up.

Validation on the rebased + refactored branch: analyzer + sm100 + sm120 fwd suites **464 pass / 0 fail** (49 sm120-only skips on the SM100 box), black-clean.

vedaanta:
@cudnn-ci-bot run frost

YangXu1990uiuc (approved):

vedaanta:
Root-caused and fixed the `frost_tests:sdpa: [Blackwell]` failure (98 mismatches at `c3721ad`, mostly `fwd_ragged`).

**Root cause** — a rename the band-model refactor could not see: the KV-loop bounds feed read the right-band bound via a *string* attribute name, `getattr(CFG, "BAND_RIGHT", 0)`. After the CFG field renamed to `WINDOW_RIGHT`, the getattr silently fell back to 0 — so the per-element mask widened but `compute_kv_loop_bounds` did not, and the KV tile loop stopped at the plain-causal bound, never visiting the widened columns. Small-S shapes were masked by the coarse per-cluster slack (`cga_tile_m = 512` rows share one bound), which is why the frost suite stayed green while test_mhas_v2's large-S graphs failed on CI.

**Fix** (`908c5ad`, amended into the refactor commit):
- `window_right=int(CFG.WINDOW_RIGHT)` — plain attribute access, fails loud instead of silently defaulting.
- New `test_dsl_sm100_band_right_multi_cluster` regression (S=1280 widened band spanning multiple Q-tile clusters, all three flavors) — verified both ways: fails 3/3 with the broken getattr, passes with the fix.
- Plus the partial-KV-tile band tests from the review follow-up (`test_dsl_sm100_band_right_partial_kv_tile`, `..._uncovered_tail_rejected`).

**Validation**: local fwd_ragged back to 128/128; frost sm100 + analyzer suites green. The full local mhas_v2 run also surfaces `bwd_ragged` workspace-canary failures and a paged-test IMA — those reproduce identically on pristine develop in this environment (pip cuDNN 9.23) and are unrelated to this PR.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -1309,6 +1319,7 @@ def _make_cache_key(
     is_causal: bool = False,
     causal_bottom_right: bool = False,
     window_size_left: Optional[int] = None,
+    window_size_right: Optional[int] = None,
```
Aneureka, line 1379: nit: do we want to unify the naming by changing `window_size_*` to `window_*`?

Aneureka (approved):

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
@cudnn-ci-bot run frost

PR #476, opened 2026-08-04
Add FROST engines

Anerudhan:
cudnn frost engines 

Open source cudnn engines written using cutlass-primitives.

## What's included

- FROST / cuTile engine implementations, kernel templates, and engine routing
- GDN cuTile path
- GEMM and linear-attention benchmarks
- Accompanying Python tests

With contributions from cudnn team

## Summary by CodeRabbit

* **New Features**
  * Added FROST acceleration for GEMM, grouped MoE matmul, fused epilogues, linear attention, and SDPA workloads.
  * Added support for FP4/FP8 and MXFP8 data formats, variable-length sequences, recurrent states, masking, and quantized outputs.
  * Added unified plan discovery, ranking, selection, fallback, workspace handling, and execution reporting.
  * Added standalone benchmarks and examples for performance and correctness validation.
* **Documentation**
  * Expanded guidance for FROST engines, routing, configuration, linear attention, and GEMM integration.
* **Breaking Changes**
  * Removed FFT-based Causal Conv1d APIs and documentation.

PR #502, opened 2026-08-06
Make a family a kind of graph, and stop paying for the DSL to decline

YangXu1990uiuc:
## Before submitting

- [x] I added GitHub labels: `cat-cleanup`, `mod-frontend`, `mod-frost`, `orig-nv-eng`.

## Affected area

Python API or bindings

## Summary

Cleanup of the python engine dispatch that landed with #476. Each change deletes a declaration that duplicated a judgment made elsewhere — and every one of those duplicates was already wrong.

### An sm100 kernel now says it serves the sm100 *line*

`Capabilities.arches` was an exact set, `frozenset({(10, 0), (10, 3)})`, checked with `device_cc not in arches`. **Rubin (sm107) and Thor (sm110) are meant to reuse these kernels and both were excluded.** It is a range now — `SM100-119` — and the decline message says so instead of listing device families.

The manifest carried a *second*, narrower copy, which is deleted:

| family | manifest said | the implementation says |
|---|---|---|
| `frost_gemm` | SM100-**103** | `PIPELINE_ARCH_RANGES["sm100"] = ((100, 120),)`, commented "family-portable Blackwell instructions" |
| `frost_sdpa_bwd` | SM120-**121** | `Capabilities` 120-129 |

On Rubin the gemm family would not have been offered at all, so its engine never got asked. Whether an engine suits a device is said once now, by the engine.

### A family is a kind of graph, not a group of engines

Eight manifest entries were two different things wearing one name: three owned a 100-wide id block and returned several engines, five owned a single id and returned one. GDN and KDA had *two entries each declaring identical anchors*, so those graphs were claimed by two families.

Six families now, and classification is a lookup (`_ANCHOR_NODE_TO_FAMILY`) rather than N families each declaring a claim that then has to be proven disjoint. `family_for(graph)` is a pure property of the graph — no `sm`, no environment — because what kind of graph something is cannot depend on which machine is asking. Availability is a separate question.

### The manifest assigns engine ids

Ids were assigned in four places (two `_ID_OFFSETS` tables, two class attributes) and the manifest only validated containment afterwards. `instantiate()` hands each factory the ids its engines are to use; engines declare none. `opt_in` moves with them, per engine — the half-precision SDPA engines can graduate while the fp8 one matures.

`register_backend()` is now only the out-of-tree escape hatch its docstring always claimed. Eleven tests called it on in-tree engines, all redundant since the manifest discovers them; with them go the "in-tree owner may register itself" exemption and its test.

### `closed_under` is deleted, and the RESHAPE bug is fixed at its source

It promised gemm served RESHAPE when nothing did, so `matmul -> reshape` was claimed and then failed in `execute` demanding a buffer the caller never bound. The real defect was in the gemm analyzer: `_node_to_recorded_op` returned `None` for an unrecognized node and **the caller skipped it**, so any unhandled node type silently compiled a subgraph. It declines now.

### Deciding costs no framework

That deletion is only safe because declining stopped costing an import. Three package `__init__`s are lazy (PEP 562) and the DSL adapter resolves at build time:

| | before | after |
|---|---|---|
| `import cudnn.sdpa.graph_analyzer` | 1059 ms, +381 modules, pulls cutlass | **9.4 ms, +2 modules** |
| support-check module | +1387 modules | **+7 modules** |

Facts and capabilities speak `cudnn.data_type` rather than `torch.dtype`, and the device comes from `cudnn.create_device_properties()` — the backend's own descriptor, the same object the C++ deviceless-AoT path serializes. `graph_analyzer` imports no torch at all now. `test_import_boundaries.py` holds the line, in a fresh interpreter, measuring the delta against an empty one.

A missing or too-old CuTe DSL is a **decline at check_support**, via probes that never execute the module (`find_spec` 7 ms, `importlib.metadata` 5 ms, against ~4.3 s for the real import). `ImportError` also joins `decline_types()` as a second line of defence — without it, deferring the import had turned "extra not installed" from "the family is absent" into an exception that skipped past the backend sitting in the same ranked list.

## Why

Every declaration this deletes duplicated a judgment made elsewhere, and every one of the duplicates was already wrong: `closed_under` promised RESHAPE support nothing implemented, the manifest's arch range capped `frost_gemm` at SM103 against templates declaring `[100, 120)`, and `Capabilities.arches` enumerated device families rather than the line, excluding Rubin and Thor. The shared shape is **allow-by-default over an open set** — a coarse copy that says yes where the real check would say no.

## Related issues

Follows #476 (FROST engines). Related to #506, which fixes a CUDA-graph-capture bug in `buffers.probe()` that this investigation surfaced.

**None for callers.** `EngineFamily`, the manifest, `_attach_facts` / `_facts_for` are internal; lazy exports (`from cudnn.sdpa import SdpafwdSm100D256`) are preserved by PEP 562. Python engine ids are renumbered, which `engine_ids.py` documents as permitted while pre-release — no API could return one before `create_execution_plan()` learned about them.

`EngineRow` is gone rather than deprecated: it arrived with #476 two days before this PR, has never been in a release, and carried `anchors`/`closed_under`, both deleted by design — a shim preserving them would be a new place to lie.

`pyproject`'s cutedsl extra stays at `>=4.5.0`. The engines need 4.7.0, but pinning that here would make cudnn-frontend incompatible with anything holding the DSL back — `quack-kernels` pins `==4.6.0`. The floor lives at runtime instead, where it costs only the engines that need it: `check_support()` declines an older DSL and the graph goes to the backend, and the test gate uses the same predicate so the suite skips rather than fails.

Behavioural: `matmul -> reshape` declines at build instead of failing inside `execute`. That path could not have worked before.

## Testing

Blackwell sm100:

```
sdpa/frost + gemm/frost + test_mhas_v2  6678 passed, 2826 skipped   (-n 16)
test/python/test_engine_router.py         66 passed
test/python/test_import_boundaries.py      5 passed
```

The 4500/2119 split is identical to the counts before this series began.

`test_mhas_v2` matters separately: it is the only suite that goes through **routing** rather than pinning an engine and exercising the kernel, and it reports which engine served each graph. FROST serves **413/3201 graphs across four engines** (d128 182, d256 122, d128_fp8 55, d192_d128 54), so consolidating families, moving id assignment into the manifest, widening the arch declaration and deferring the DSL leave routing intact.

It also caught a bug the other suites structurally could not: deferring `torch` had put the import inside one branch of the execute path while a later, independent branch used it too, so an fp8 graph hit `UnboundLocalError`. Thirty tests. `sdpa/frost` stayed green throughout because it never reaches that path.

**On the public wheel**, which nothing had been verified against — this box ships `nvidia-cutlass-dsl-internal` while `pyproject` declares `nvidia-cutlass-dsl`. Swapped internal 0.3.0 for public 4.7.0: **4500 passed / 2119 skipped, identical**.

**Across architectures**, which matters because deleting the manifest's arch filter opened a path sm100 cannot exercise — the engines are now imported and must decline themselves:

```
sm80 (A100)   sm89 (L40S)   sm90 (H100)     engines decline cleanly, no DSL pulled
```

While there: twelve gemm registry tests had no arch gate at all and fail on any non-sm100 host **on develop today**; they now carry `requires_sm100`. The SDPA suite's five hand-copied `_is_sm100` gates, each pinned to exactly `(10, 0)` and so skipping on sm103, are one marker aligned with what the engines declare.

`test/python/linear_attention` fails 286 / passes 67 here — **the same 286/67/1769 on pristine `gh/develop`**, verified in a separate worktree. CUDA graph capture, untouched by this PR.

## Design doc

`docs/python_graph_and_execution_backends.md` is updated in this PR — it has been checked in since #336 but described the shape this replaces. Four sections added: what the manifest decides and what it leaves to the engine, how facts are attached and why after the freeze, how handle/stream/device resolve, and which imports each dispatch stage may pay for.

## Known gaps

- `create_execution_plan(engine_id, knobs)`, the explicit replay entry point, does not go through finalize/freeze/attach; only `create_execution_plans()` does.
- A custom Router calling `key()` can retry backend lowering after facts were attached.
- `_freeze()` is shallow: `MappingProxyType(dict(node.params))` freezes the mapping, not list-valued params.
- A failed planning attempt leaves the graph frozen while `register_backend()` / `set_router()` still gate on `_planning_done`, so an engine registered afterwards is ignored by the cached candidate set.
- `heuristics_sort` still does not read facts. Deliberate, and its docstring says what will: order a family's engines on its facts, then merge against the backend on predicted time.


YangXu1990uiuc:
Earlier comments on this PR were deleted rather than left to mislead: they described the pre-rescope design, and one claim in them is now **backwards** — I wrote that fp8 SDPA could not be split into its own family because its ids had shipped. `engine_ids.py` states the opposite (pre-release, nothing persisted them), which is why this version re-cuts the blocks freely.

Two things from those comments are worth keeping, both about measuring these suites:

**These suites randomize shapes from an unseeded `random`.** `s_q`, `s_kv`, `d_qk`, `d_v`, `h_k/h_v` and the mask bounds all come from `random.choice`; `@torch_fork_set_rng(seed=0)` forks only torch's RNG, i.e. tensor *values*. Two runs therefore compare different graphs — I measured roughly 49 tests of drift per run before noticing. Any before/after comparison on `test_mhas*` needs `random.seed()` pinned first, e.g. via a `-p` plugin outside the tree so both sides load the identical one.

**`-k` is a substring match.** `-k "test1 or test2 or test3"` selects `test1`, `test10`–`test19`, `test100`–`test199`, … — 182 tests, not 3. I misread my own filter and published a "cost is severely sublinear, so compilation dominates" conclusion from it. The real cost is close to linear at ~1.45 s/test with ~75 s of one-time per-process warmup, and the parallelism figures in the description supersede the ones I gave then.

YangXu1990uiuc:
Thanks — verified each against HEAD rather than the commit it was filed on, since three were already addressed by later commits in the series.

**Fixed (`d6e2b01f`)**

- **`to_torch_dtype` unguarded on an unmapped type.** Correct and the most useful of the six. Only Q's dtype is capability-checked, and `tensor_desc_from_ir()` runs on every bound tensor, so O / Stats / a side output can reach it with a type outside the map — `KeyError` deep in lowering instead of a decline. It raises `NotImplementedError` now, which the router treats as "this engine does not serve this graph".
- **Docstrings still naming `validate()` as where facts are attached.** Two were left (`manifest.py`, `graph_analyzer.py` module docstring). Planning attaches them, after `_finalize_backend_layout()` and `_freeze()`.

**Already fixed before the review landed** — these were filed against `b5160e860` / `bc68048dd`:

- The `_facts` comment describing a `(node count, facts)` tuple: it is `# analyzer callable -> its record; see _facts_for()` as of `a277c141`, which also changed the key from a `module.qualname` string to the callable itself.
- The manifest module docstring describing `anchors` / `closed_under` matching and competing family claims: rewritten in `b5160e86`. The one surviving `closed_under` mention is a deliberate note about why it was deleted.
- Test-level markers on the new tests: both files carry a module-level `pytestmark = pytest.mark.L0`, which the new tests inherit.

**Not doing: the `EngineRow` compatibility export.**

`EngineRow` is not a compatibility surface. It arrived with #476, which merged into `develop` on 2026-08-04 — two days before this PR — and has never been in a release, so nothing downstream can be importing it. It also has no use outside the library: it describes an *in-tree* engine family for the manifest to classify graphs with, while the out-of-tree surface is `BaseEngine` / `Router` / `register_backend()`, all unchanged.

More to the point, a compatibility adapter could not preserve the old behaviour even if we wanted it to. `EngineRow` carried `anchors` and `closed_under`, and both are gone by design: classification is a lookup table now (a graph belongs to one family or none, rather than N families each declaring a claim), and `closed_under` was a coarser duplicate of `check_support()` that promised RESHAPE support nothing implemented. A shim returning an object with those fields would be a fresh place to lie.

The PR description records this under API impact.

YangXu1990uiuc:
Latest round — two fixed, two not applicable.

**Fixed**

- **`pytest.skip` hid a failing probe.** Good catch: a probe that cannot `import cudnn` is exactly the regression that file exists to catch, and skipping reported it as "not checked". `pytest.fail` now.
- **Ruff E741 `l`** → `ln`.

**Not applicable**

- **Black on the `EngineFamily` fixtures.** Neither construct exceeds 160 chars, and `black --line-length 160 --check` reports both files unchanged. `pre-commit` runs black on every commit in this branch.
- **"callable the SDPA family names" grammar.** `names` is the verb, not a noun: *the callable [that] the SDPA family **names** in engines/manifest.py* — a reduced relative clause. The proposed "callable **for** the SDPA family names" reparses `names` as a noun and changes the meaning.

Also worth flagging, since it came out of a question about this PR rather than from review: deferring the DSL import past `check_support` had turned "cutedsl extra not installed" from *the family is absent* into an `ImportError` out of `build_plan()` — which is in neither the engine's except clause nor `decline_types()`, so it propagated past the backend sitting in the same ranked list. The engine now declines at check time using probes that never execute the module, and `ImportError` joins the decline types as a second line. Regression test added.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ classifiers = [
 
 [project.optional-dependencies]
 cutedsl = [
-    "nvidia-cutlass-dsl[cu13]>=4.5.0",
+    "nvidia-cutlass-dsl[cu13]>=4.7.0",  # the CuTe primitives these engines lower through
```
vedaanta: do we need this?

vedaanta (approved):

pyproject.toml
YangXu1990uiuc: No — you're right, and it's reverted in the next push.

Pinning the extra to `>=4.7.0` would make cudnn-frontend incompatible with anything holding the DSL back: `quack-kernels` pins `==4.6.0`, and vLLM and friends carry their own constraints. The extra stays `>=4.5.0`.

The 4.7.0 floor is real — it's what the CuTe primitives these engines lower through need — but it belongs at runtime, not in the dependency:

- **engine**: `check_support()` declines when the installed DSL is older, so the graph goes to the backend and only those engines are lost.
- **tests**: the same predicate, so an older DSL makes the suite skip rather than fail.

Both read `CUTEDSL_MIN_VERSION` from `cudnn/frost/buffers.py`, so there's one number rather than two that can drift.

The probe doesn't cost anything either — `importlib.util.find_spec` (7 ms) and `importlib.metadata.version` (5 ms), neither of which executes the module, against ~4.3 s and 1410 modules for a real `import cutlass`.

Anerudhan:
@cudnn-ci-bot run frost

Anerudhan (changes requested):
Few things:

1. Can you check-in a design doc so it is easier for agents in future..
2. Can you update your agents to create labels for your PR..
3. Use the PR template for license.

Aside on the code:

1. I am still vague about how the stream/current_device works. Is it a fallback if user does not provide the device?.

YangXu1990uiuc:
All four — thanks.

**Design doc.** There is one, `docs/python_graph_and_execution_backends.md`, checked in since #336 — it just described the shape this PR replaces, so following it you'd have gone looking for `anchors` and `closed_under` that no longer exist. Updated here (`cee6a0a3`) rather than starting a second one, since two design docs is exactly the duplicate-that-can-lie problem the rest of this PR is deleting.

Four sections added: what the manifest decides and what it deliberately leaves to the engine; how facts are attached and why after the freeze; how handle/stream/device resolve; and which imports each dispatch stage may pay for. Existing sections corrected where this PR moved them.

**Labels.** Added: `cat-cleanup`, `mod-frontend`, `mod-frost`, `orig-nv-eng` (and on #506: `cat-bug`, `mod-frost`, `mod-cutedsl`, `orig-nv-eng`). I'll make this part of opening a PR rather than a follow-up.

**PR template.** My fault — I wrote the body from scratch and dropped the checklist. Restored, license included.

**Stream / current_device.** Three separate things, and the short answer to your question is: it's a fallback for having *no handle*, not for a handle we failed to read.

- **handle** — `execute(..., handle=h)` if given, else the graph's own. The handle is what carries the stream.
- **stream** — `_resolve_stream(handle)` is `cudnn.get_stream(handle)`. No handle → `None`, and kernel wrappers then use the default stream. But a failed query on a *supplied* handle **raises** rather than quietly falling back: running on the wrong stream is a correctness bug, not a degradation, so it must not be silently absorbed.
- **device** — this PR changes it. The analyzer used to read compute capability and SM count from `torch.cuda.current_device()`, i.e. ambient state that a graph carries around and that goes stale if you switch device or pass a handle for another one. It now reads `cudnn.create_device_properties()` — the backend's own serialisable descriptor, the same object the C++ deviceless-AoT path uses. Engines re-check the arch in `check_support()` regardless.

That last one has a second payoff worth flagging: facts computed from a serialisable descriptor need no live device, which is the precondition for a python engine to participate in deviceless AoT at all. It cannot today.

Relatedly, #506 (separate PR, off `develop`) fixes a stream bug this investigation turned up: `buffers.probe()` called `__dlpack__()` with the default stream argument, so torch did a `record_stream` that is illegal inside a CUDA graph capture. 286 linear-attention tests, red on `develop` today.

Anerudhan (approved):

PR #506, opened 2026-08-07
Tell DLPack we are only reading, so probe() survives a graph capture

YangXu1990uiuc:
## Before submitting

- [x] I added GitHub labels: `cat-bug`, `mod-frost`, `mod-cutedsl`, `orig-nv-eng`.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

`buffers.probe()` turns a caller's buffer into `(pointer, shape, strides, dtype, device)` for a kernel launch. It reads metadata and does nothing else — so it never needed synchronisation. Its DLPack fallback called `__dlpack__()` with the default stream argument, which torch reads as *prepare this tensor for a consumer on another stream* and answers with `record_stream`. That is illegal inside a CUDA graph capture.

```python
-    capsule = dl()
+    try:
+        capsule = dl(stream=-1)   # DLPack: caller handles sync, do no bookkeeping
+    except TypeError:             # producer predating the stream kwarg
+        capsule = dl()
```

## Why it looked like a GDN/KDA bug

`probe()` prefers `__cuda_array_interface__`, whose `typestr` is numpy-flavoured. `float16` is `"<f2"` and resolves there; **bfloat16 has no numpy spelling**, so torch reports raw `"<V2"`, the lookup misses, and it falls through to DLPack.

So the failure needed three things at once — bf16 buffers, a capture, and `probe()` on the execute path — and only the linear-attention suites have all three:

```
286 failed  test_gdn_bprop_kernel.py (118), test_gdn_prefill_kernel.py (97),
            test_kda_prefill_kernel.py (63), test_gdn2_*, ops/test_gdn2_op.py
```

every one reporting `operation failed due to a previous error during capture` — an error naming neither this file nor this function. SDPA and GEMM use fp16/fp8 and take the CAI path; **a bf16 SDPA capture would have hit the same thing.**

Minimal reproduction, no cudnn involved:

```
dl()               -> capture fails
dl(stream=None)    -> capture fails
dl(stream=-1)      -> capture OK
__cuda_array_interface__ -> capture OK
```

## Why

`probe()` reads metadata and launches nothing, so it has no business asking a producer to synchronise. Passing the default stream argument made torch do bookkeeping on its behalf — correct for a cross-stream handoff, illegal inside a capture, and unnecessary either way.

## Related issues

Found while working on #502 (python engine dispatch cleanup); independent of it and based on current `develop`.

**None.** `buffers.probe()` is internal and its return value is unchanged. `stream=-1` is the DLPack-specified way to request no synchronisation; the `TypeError` fallback keeps producers whose `__dlpack__` predates the argument working.

## Testing

`test/python/test_buffer_probe.py` pins it on `probe()` directly rather than leaving the linear-attention suites to notice again — they do notice, but a hundred lines into unrelated setup and pointing at torch's stream bookkeeping. **Reverting the fix turns `test_probe_is_capture_safe[bfloat16]` red; that was checked.**

On sm100:

```
test/python/linear_attention          353 passed, 1769 skipped
                                      (286 failed / 67 passed on this same tree before)
+ sdpa/frost + gemm/frost + new test  4859 passed, 3892 skipped
```

## Not addressed here

`probe()` has no mapping for fp8 at all — `TypeError: unsupported buffer dtype (code=10, bits=8, lanes=1)` — so fp8 is absent from the test. Separate gap, and I have no fp8-under-capture case to verify a fix against.

Worth knowing for anyone weighing the two protocols: CAI is *not* strictly cheaper-and-better. It cannot express bf16 or fp8 (fp8 raises outright), it is CUDA-only (CPU tensors and JAX arrays expose only DLPack), it carries no ownership or lifetime semantics, and torch reports version 2, which has no `stream` field at all. Its capture-safety here comes precisely from knowing nothing about streams. The existing CAI-first-then-DLPack order is right; only the DLPack call was under-specified.

/cc @Anerudhan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA buffer inspection to avoid unnecessary synchronization.
  * Added compatibility for data producers that do not support stream options.
  * Ensured buffer inspection works safely during CUDA graph capture.

* **Tests**
  * Added coverage for float16, float32, and bfloat16 buffers.
  * Verified pointer, shape, strides, data type, and device reporting.

### DIFF
--- a/python/cudnn/frost/buffers.py
+++ b/python/cudnn/frost/buffers.py
@@ -235,7 +235,14 @@ def probe(buf):
     dl = getattr(buf, "__dlpack__", None)
     if dl is None:
         raise TypeError(f"buffer of type {type(buf).__name__} exposes neither __cuda_array_interface__ nor __dlpack__")
-    capsule = dl()
+    # stream=-1 is DLPack's "the caller handles synchronisation; do no
+    # bookkeeping". probe() only reads metadata, so it never needed any --
+    # and the default makes torch call record_stream, which is illegal inside
+    # a CUDA graph capture.
+    try:
+        capsule = dl(stream=-1)
+    except TypeError:  # a producer whose __dlpack__ predates the stream kwarg
+        capsule = dl()
     raw = _PyCapsule_GetPointer(capsule, b"dltensor")
     mt = ctypes.cast(raw, ctypes.POINTER(_DLManagedTensor)).contents
     t = mt.dl_tensor
--- a/test/python/test_buffer_probe.py
+++ b/test/python/test_buffer_probe.py
@@ -24,13 +24,20 @@ from cudnn.frost import buffers
 pytestmark = pytest.mark.L0
 
 requires_cuda = pytest.mark.skipif(not torch.cuda.is_available(), reason="needs a CUDA device")
+# bfloat16 needs SM80+. Reported per dtype so float16/float32 still run on older
+# cards -- they take the CAI path, which is the control for the DLPack one.
+_BF16_OK = torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 8
 
 # float16/float32 resolve through __cuda_array_interface__; bfloat16 has no
 # numpy typestr (torch reports raw "<V2"), so it is the dtype that reaches the
 # DLPack path -- the one that has to stay capture-safe. fp8 is absent because
 # probe() does not map it at all ("unsupported buffer dtype (code=10, bits=8)"),
 # which is a separate gap and not this fix's business.
-_DTYPES = [torch.float16, torch.float32, torch.bfloat16]
+_DTYPES = [
+    torch.float16,
+    torch.float32,
+    pytest.param(torch.bfloat16, marks=pytest.mark.skipif(not _BF16_OK, reason="bfloat16 needs SM80+")),
+]
 
 
 @requires_cuda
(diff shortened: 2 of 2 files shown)


YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc:
Both valid, both fixed in `ea103ef4`.

**BF16 gating.** Right — the module was gated on CUDA presence alone, so on a pre-SM80 card the bf16 cases would have *failed* rather than skipped. Now per-dtype:

```python
_BF16_OK = torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 8
_DTYPES = [torch.float16, torch.float32,
           pytest.param(torch.bfloat16, marks=pytest.mark.skipif(not _BF16_OK, ...))]
```

Per-dtype rather than module-level on purpose: float16 and float32 resolve through `__cuda_array_interface__`, so they are the control for the DLPack path bf16 takes. Losing them on an older card would leave the test proving nothing there. And agreed on no `cudnn.backend_version()` gate — this test never calls cuDNN.

**Stream ordering.** Also right, and a real hole rather than a style point: `torch.zeros` runs on the current stream, `t.add_(1)` on `side`, and nothing ordered them. `torch.cuda.synchronize()` afterwards waits for both to *finish* but imposes no order *between* them, so the warmup could have read the tensor before it was zeroed. `side.wait_stream(torch.cuda.current_stream())` added.

Re-checked that the test still does its job: reverting the `stream=-1` fix turns `test_probe_is_capture_safe[bfloat16]` red.

Anerudhan:
@cudnn-ci-bot frost

Anerudhan (approved):

PR #508, opened 2026-08-07
frost(spda): support ragged stats for SM120 frost sdpa forward engine

Aneureka:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Enable `thd_stats=True` for the SM120 FROST SDPA forward engine. The ragged stats output is written directly in the layout the graph declares — both packed layouts in use are supported:

declared layout | strides (dims (B, H, S, 1)) | ragged offsets | kernel store
-- | -- | -- | --
token-major [t, h] | stride_h == 1, stride_s == H | cu_q * H | lse[cu_q[b] + row, head]
head-major [h, t] | stride_s == 1, stride_h == head_stride | cu_q | lse[head, cu_q[b] + row]

## Why

To enhance the SM120 FROST SDPA forward engine.

## Related issues

Related to #381.

## API and compatibility impact

`SdpaFwdDslSm120` now accepts THD graphs with a stats output (both declared layouts); previously rejected with `NotImplementedError`. Dense stats behavior unchanged.

## Testing

* Test command: `pytest sdpa/frost/test_sdpa_fwd_dsl_sm120.py -m "L0 or L1"`
* Test result: 52 passed, 0 failed

## Summary by CodeRabbit

- **New Features**
  - Added ragged THD attention statistics (LSE) output with token-major and head-major layouts.
  - Added statistics support for causal attention, sliding windows, grouped-query attention, sinks, zero-length inputs, and mixed head dimensions.
  - Expanded support for eligible ragged SDPA graph configurations.

- **Bug Fixes**
  - Added validation for statistics buffer shapes, strides, layouts, and presence.
  - Improved zero-key/value handling with zero outputs and appropriate statistics values.
  - Reduced workspace requirements when statistics output is enabled.

PR #509, opened 2026-08-07
Add SM120 per-tensor FP8 (e4m3) SDPA-forward engine

YangXu1990uiuc:
## Before submitting

- [x] I ran `pre-commit run` and committed any formatting changes. (black 26.3.1 -l 160 on all touched python files)
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)). *(token lacks label permission on this repo — suggested: `cat-feature`, `mod-python-fe`, `mod-kernels`, `orig-community`)*

## Affected area

FE OSS kernels or CuTeDSL

## Summary

New opt-in engine **`sdpa_fwd_prefill_sm120_fp8`** (manifest slot 7): a per-tensor FP8 (e4m3) SDPA-forward prefill kernel for consumer Blackwell (SM120/SM121), reachable through the ordinary `graph.sdpa_fp8(...)` op with `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`.

- **Kernel** `python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py` — the e4m3 sibling of `prefill_f16_sm120.py`, with the MMA lowered to `mma.sync.aligned.m16n8k32.row.col.f32.e4m3.e4m3.f32`. e4m3 travels as Uint8 storage end to end (TMA → ldmatrix → MMA consume bit patterns). K B-fragments via byte-preserving `ldmatrix.m8n8.x4.b16`; V B-fragments via the hardware 8-bit transposed `ldmatrix.m16n16.x2.trans.b8` (SASS `LDSM.8.MT1616`). FP16 O; fp32 softmax denominator.
- **P reaches the PV MMA in registers.** The QK C-fragment owns columns `2*(t%4)+{0,1}` while the k32 A-fragment wants four consecutive bytes; the two differ only by an exchange inside each thread quad, so `pack_f8x2_pairs` + two `shfl.sync.idx` + one `prmt.b32` build each A operand from the `cvt.rn.satfinite.e4m3x2` results the softmax leaves in registers. No SMEM round trip.
- **SDPA_FP8 node convention** as on SM100: `descale_q*descale_k` folds into the softmax scale, `descale_v*scale_o` into an `o_scale_fused` epilogue scalar; `Amax_S` and `Amax_O` via bitcast-int32 atomic max into host-pre-zeroed buffers.
- **Dispatch**: manifest slot, `_sm120_fp8_spec` capability row (including a new `Capabilities.out_dtypes`, since `facts.dtype_o` previously never reached `mismatch()`), `SdpaFwdDslSm120` fp8 compile/execute path, shared PTX helpers in `tile_dsl.mma`.
- **Tile plans**: `propose_plans` now offers the capability row's whole `tile_ms x tile_ns` domain instead of one entry, so a caller can pin a point with `create_execution_plan(engine_id, SdpaFwdKnobs(...))`. Entry 0 carries no knobs and lets `config_sm120.fp8_tile_choice` decide from shape and SM count.

v1 envelope: E4M3 in / FP16 out, exact d128, causal / bottom-right (+SWA, +rect) / SWA / KV-padding masks, GQA/MQA, stats + `lse_optional`; no sink, no THD, no E5M2 — all gated in the capability row and negative-tested.

## Why

SM120 had only the f16/bf16 FROST SDPA cell; the quantized-attention alternatives on this arch are third-party INT8/FP4 kernels with materially worse numerics.

All numbers below: RTX PRO 6000 Blackwell Server Edition (188 SMs, SM120), cuDNN 9.25.0.15, CUPTI kernel time, 6 s device warm-up before the first configuration, interleaved repeats. **Three different baselines, which answer three different questions:**

| against | question | result |
|---|---|---|
| the backend's **native fp8 fprop** | should this merge? | **ahead on all 22 held-out shapes, 1.15x–1.93x** |
| the **bf16 SM120 cell** | what does fp8 buy? | **1.65x–1.89x** |
| the sibling internal SM120 fp8 kernel | how good is the implementation? | **1.05x–1.08x behind**, best tile vs best tile |

Accuracy is unchanged from v1: O max abs err 1e-3..3e-2 against an fp32 reference over dequantized inputs, LSE ~1e-6.

A note on the RTX 5080 table this PR previously carried (1.74–1.79x vs bf16): those were the **pre-shfl** kernel on a **different part**, and the ratio does not transfer. The same old kernel measured here gives only 1.27x–1.38x; the shfl path is what takes it to 1.65x–1.89x. The doc's claim that "the kernel is arch-identical so ratios are expected to transfer" was wrong and has been removed.

## Measured and rejected

Recorded because they are the obvious next things to try, and each cost a measurement to rule out. The kernel is issue-bound after the shfl change (ncu: Compute 74%, DRAM 3.5%, L1/TEX 49%), which is why reordering work does not help — only removing instructions does.

- **Prefetching K or V fragments one step ahead** (both present in the sibling kernel): within ±0.5%, no consistent sign.
- **Loading Q coalesced with a shfl transpose** (also in the sibling): reproducibly **0.5–2.3% slower** at s=512, the shape it should help most. Q is small enough to sit in L2, so the eight shuffles per d-fragment cost more than the saved transactions.
- **FAv4's rescale threshold** (hold the running max, as FORT's `softmax_op::rescale_threshold` and the SM100 fp8 cell do): **2.7–3.2x worse** O error in the regime where 98% of P falls below e4m3's smallest normal. It trades accuracy for speed, and the speed half needs a branch this kernel's structure does not give for free.
- **Dropping Amax_S / Amax_O / LSE** to match what the sibling computes: 0.39% of executed instructions. Not the source of the residual gap.

Measuring any of this needed a noise floor first: run-to-run spread is ~1% at s>=2048 but was **12% at s=512** until the probe warmed the whole device, which is larger than every effect above.

Additive. The engine is opt-in behind `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1` and declines anything outside its envelope, so graphs that do not match fall through to the backend as before. `Capabilities.out_dtypes` is a new field defaulting to empty, which only the quantized rows read.

`test_mhas_v2` run serially with `random` seeded: identical routing tally and identical failure set against `develop`.

## Related issues

- #510 — the torch glue around these kernels runs on torch's current stream rather than the handle's. The two sites on this PR's path are fixed here; the SM100 cells and the THD staging are not.
- #520 — the DSL fp8/mxfp8 engines drop `descale_s` and `scale_s` before the adapter is reached. Not SM120-specific and not fixed here.

## Summary by CodeRabbit

* **New Features**
  * Added optional SM120/SM121 support for per-tensor FP8 E4M3 scaled dot-product attention.
  * Supports FP16 outputs, causal, bottom-right, and sliding-window masking, grouped-query attention, variable-length layouts, padding, and optional LSE/Amax statistics.
  * Added automatic tile selection for supported FP8 workloads, including exact 128-dimensional heads.

* **Documentation**
  * Added performance, accuracy, compatibility, limitations, and optimization guidance for the new FP8 attention path.

* **Bug Fixes**
  * Improved validation and runtime handling for quantized attention scaling parameters.


YangXu1990uiuc:
Rebased onto `develop` (the conflict was with #508: it gave
`SdpaFwdDslSm120._execute_thd` an `lse_tensor` parameter at the spot this
branch inserts `_execute_fp8`; kept both, with #508's signature since the
shared docstring and the call site are already ragged-stats aware), and pushed
two commits.

### Offer the tile domain as plans, and pick by shape

`propose_plans` returned one entry, so the capability row's `tile_ms`/`tile_ns`
domain was unreachable: `SdpaFwdKnobs` validation, `PlanConfig.knobs` and
`create_execution_plan` replay were all present with nothing enumerating the
points. The adapter fell back to `SEQ_*_TILES[0]` — the largest tile that
fits, not the fastest.

`knob_candidates()` reads the domain off `Capabilities` (no second table to
drift from it), filters through `mismatch()`, and orders it via a new per-spec
`EngineSpec.knob_order`. Entry 0 keeps `knobs=None`: a delegation and a named
knob set are different requests — one may improve with the library, the other
must be honored verbatim — so existing name-pinning is untouched, and routing
is unchanged.

`config_sm120.fp8_tile_choice` picks from shape, causality and SM count:

- `kv_tile=64` unconditionally. The kernel is L1-bound on the P restage —
  ncu puts L1/TEX throughput at 72–77% against the backend kernel's 45–51%,
  with DRAM at 5–13% on both — and halving the KV tile halves that traffic
  per tile. Faster in 47 of 48 shapes; the exception by 0.15%.
- `q_tile=64` while the grid cannot fill the machine, over a wider window
  under a causal mask (the last Q tile does several times the work of the
  first, so finer tiles even out the tail).

FP8 only — the f16 cell wants `kv_tile=128` at long sequences and the two
share `SEQ_*_TILES`.

### Measurements

RTX PRO 6000 Blackwell (sm120, 188 SMs), cuDNN 9.25.0.15, CUPTI kernel time,
2 s clock warm-up per configuration, both measurement orders agreeing within
0.5%.

| | |
|---|---|
| regret vs the best of the enumerated domain, 22 held-out shapes | **1.005x mean, 1.045x worst** (1.0 = the delegation picked the domain's fastest point) |
| vs the backend's native fp8 fprop | faster on 19 of 22; **9.8 us vs 14.8 us** at the smallest grid (1.51x), behind by 4-7% on the other three |
| the previous default against its own domain | **11–25% slower, at every shape measured** |

`test_mhas_v2`, run serially and with `random` seeded (the tally is per-worker
state summarised on the controller, so it is only meaningful without `-n`):
identical routing tally and identical failure set against the committed files.

### One note on the baseline in the first commit message

The 1.74–1.79x there is against the **bf16 SM120 kernel**, which is the right
number for "what does fp8 buy". It is not a comparison against what already
serves these graphs. Against the backend's native fp8 fprop, this engine was
*behind* at the shipped default tile — the win above only exists once the tile
domain is enumerated. Worth keeping the two apart in the docs page.

Also, the two paths turn out to overlap less than the "sibling" framing
suggests: this engine emits FP16 O only, and the backend's sm120 fp8 fprop
declines FP16 O, so most fp8 graphs cannot move either way. Adding fp8 O is
not the store-path-only change the design doc claims — the QK C-fragment gives
each lane columns `2*(t%4)+{0,1}`, and fragments A and B cover columns 0–7 and
8–15, so packing four e4m3 into one 16-bit `stmatrix` slot would put
non-adjacent columns in one word. It needs either a different SMEM layout or
an accepted fp32→fp16→e4m3 double rounding. ncu says it buys no performance
(DRAM is nowhere near the limit) — it is a coverage change.

Unrelated and pre-existing, filed separately as #510: the torch glue around
these kernels runs on torch's current stream rather than the handle's.

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "审计 Frost Python DSL 引擎调用流程"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

YangXu1990uiuc:
Went through the review findings against the current code. Four fixed, one
verified as not-a-defect, one real but rooted a layer up from where it was
anchored — details below, and the branch has been rebased onto `develop` and
force-pushed with a further kernel change (see the commit).

### Fixed

**`amax_o_buf.div_` outside the launch stream** (Critical). Correct, and it
applies to the `O` copy-back on the same path. Both now sit in the
`_torch_stream_context(current_stream, device)` next to the resets that were
already there. This is also the subject of #510, which covers the same seam in
the SM100 cells and in the THD pre-kernel staging — those are untouched here.

**Docstring promised negative tests that did not exist** (Minor). Added
`test_fp8_sm120_fp8_output_not_offered` and
`test_fp8_sm120_non_128_head_dim_not_offered`, and corrected the docstring to
name what is actually covered.

Writing the first of those turned up something the review did not ask about:
`facts.dtype_o` never reached `mismatch()`, so the capability row had no O-dtype
domain at all. A graph asking for fp8 O was *claimed* by this engine and then
declined at build. `Capabilities.out_dtypes` now closes that, defaulting to
empty rather than unconstrained — a quantized row that forgets to declare
serves nothing, which fails loudly instead of silently over-claiming.

**E741 `O`** (Minor). Renamed to `out`/`o_ref`.

**`store-path` split across a line break** (Trivial). That paragraph was
rewritten: adding fp8 O is not the store-path-only change it claimed. The QK
C-fragment gives each lane columns `2*(t%4)+{0,1}` and fragments A and B cover
columns 0-7 and 8-15, so packing four e4m3 into one 16-bit `stmatrix` slot puts
non-adjacent columns in one word. ncu also puts DRAM at 5-13% on this kernel,
so halving the O write buys no time — it is coverage, not speed.

### Not a defect

**"Apply `fmul2` pairwise"** (Major). `o_scaled = fmul2(o_regs[o_off:8],
row_sum_inv_vec)` is correct as written. The proven f16 sibling uses the
identical eight-element form at `prefill_f16_sm120.py:1145`, and it is covered
by `test_sdpa_fwd_dsl_sm120.py`. Independently: if six of the eight elements
missed the `1/row_sum` normalization, O would be wrong by orders of magnitude,
not the ≤5e-2 the tests assert against an fp32 dequantized reference across
four mask kinds, GQA, KV padding and all four tiles.

### Real, but not fixed here

**`_execute_fp8` discards `descale_s` and `scale_s`** (Major). The behaviour is
real — a graph supplying non-unit values builds, runs, and is silently wrong.
Two corrections to the diagnosis:

- The drop is not in `_execute_fp8`. The engine's `execute_kwargs` never
  forwards those two operands (`engines.py`, the `is_mxfp8 or is_fp8` block),
  and `graph_analyzer` does not carry them either. They are gone before the
  adapter is called, and **the SM100 fp8 and MXFP8 cells lose them the same
  way** — this is not SM120-specific.
- The `seq_q_lens` half is already handled: `api_dsl.py` rejects
  `seq_q_lens_present and self._fp8` with "not plumbed for the FP8/MXFP8
  kernels".

Closing it means either folding both scalars (`scale_s` can ride in the `exp2`
addend and `descale_s` in `o_scale_fused`, but that needs a reference that
exercises non-unit values to validate) or forwarding them and rejecting
non-unit at execute. Either touches the execute signature shared with SM100,
which I cannot exercise on this part. Filed separately rather than half-fixed
inside an SM120 kernel change.

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "审计 Frost Python DSL 引擎调用流程"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

YangXu1990uiuc:
Rebuilt on the merged `develop` (which now carries #485 and #528) and re-verified. Force-pushed.

## What changed structurally

Rebuilt rather than rebased, because two things moved underneath this branch:

- **#485** unified the mask parameterization onto one band model (`window_left`/`window_right`/`bottom_right`). The kernel is ported the way #485 ported its f16 sibling; the adapter needed nothing, since #485 kept the public `is_causal`/`window_size_left` arguments and resolves the band once in the shared base.
- **#528** moved plan ranking out of the engines. The `propose_plans` + `knob_candidates` + `_sm120_fp8_knob_order` + `fp8_tile_choice` trio (~140 lines) is replaced by one line: the cell joins `_TILE_RULE_CELLS`.

Sharing that rule is a measurement, not an assumption — 30 causal and non-causal shapes on a 188-SM part give **regret 1.0058 geomean / 1.155 worst**, against the f16 rule’s own 1.009 / 1.054. The worst cell is a limit of the *features*: `B1xH16xS2048` and `B2xH8xS2048` causal reach the rule with identical inputs (grid 128, 16 KV tiles) and have opposite optima, so no threshold separates them. Recorded in the docstring rather than papered over with a second rule.

## Testing

| target | result |
|---|---|
| SM120 (RTX PRO 6000 Blackwell, 188 SM) fp8 | 16 passed |
| SM120 all `sdpa` suites | 177 passed |
| SM120 guards + tile rule | 80 passed |
| SM120 `test_mhas_v2` | 245 passed — routing unchanged, frost 87 / native 158 |
| SM100 `sdpa` + `gemm` | 4684 passed |
| SM100 `linear_attention` | 353 passed |
| CPU dispatch suites | 150 passed |

## ⚠️ One open decision for you

**S quantization is not implemented, and never was.** The kernel converts S to e4m3 for the PV MMA *unscaled*, so `scale_s`/`descale_s` reach no math. Previously they were silently dropped; they are now threaded through and a non-unit pair is **declined** instead of ignored.

But the standard SDPA_FP8 contract *does* supply non-unit S scales — `test/python/sdpa/fp8.py` computes `s_scale = get_fp8_scale_factor(s_amax, ...)` — and a decline at execute has **no plan fallback**, so such a graph now fails rather than silently returning a wrong answer.

Three options, and which one this PR is is a scope call:
1. **Implement it** — scale P before the `cvt` (keeping the `row_sum` denominator on unscaled P) and fold `descale_s` into `o_scale_fused`. Roughly five lines plus a new kernel runtime scalar, and it needs numerics validation.
2. **Ship as v1** — unit S scales only, opt-in (where it is today), limitation documented on the engine row.
3. Revert to silently ignoring them — not acceptable.

Currently at (2). Say the word and I will do (1).

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
```diff
@@ -0,0 +1,1568 @@
...
+                amax_q_idx = q_seq_idx + q_warp_row0 + (lane // 4) + row_half * 8
+                row_valid[row_half] = cutlass.Float32(1.0) if amax_q_idx < seqlen_q else cutlass.Float32(0.0)
+            lane_amax_o = cutlass.Float32(0.0)
+            for d_frag_pair in cutlass.range_constexpr(self.pv_d_frags // 2):
+                o_off = (d_frag_pair * 2) * 4
+                o_scaled = fmul2(o_regs[o_off:8], row_sum_inv_vec)
+                for i in cutlass.range_constexpr(8):
+                    # row_sum_inv_vec order: halves alternate 0,0,1,1,0,0,1,1.
+                    half = (i // 2) % 2
+                    lane_amax_o = cute.arch.fmax(lane_amax_o, cute.math.abs(o_scaled[i]) * row_valid[half])
```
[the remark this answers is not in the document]
YangXu1990uiuc: Verified on hardware and pushing back: this is a false positive.

`cutlass.Vector` vectorizes here despite the `Tuple[Float32, Float32]` annotation. The SM120 fp8 suite validates O, `amax_s` **and** `amax_o` against fp32 references across every mask, GQA, bottom-right, padding and all four enumerated tiles — 16/16 pass. If the 8-element `fmul2` computed only the first pair, O would be wrong for every `d_frag_pair > 0`, and `amax_o` (a max over all 8 lanes of `o_scaled`) would be wrong too. Both are checked.

I left the line as-is rather than "fixing" a kernel that measures correct.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -1777,6 +1837,96 @@ def execute(
...
+        o_view, o_needs_copy_back, o_scratch = self._to_bshd_writable(o_tensor)
+        o = o_scratch if o_needs_copy_back else o_view
+
+        # amax_s / amax_o: the kernel atomicMax'es into these buffers, so they
+        # MUST start at 0, reset on the LAUNCH stream (ordering vs the kernel).
+        amax_s_buf = amax_s.reshape(-1)[:1] if amax_s is not None else self._dummy("amax_s", device, lambda: torch.zeros(1, dtype=torch.float32, device=device))
+        amax_o_buf = amax_o.reshape(-1)[:1] if amax_o is not None else self._dummy("amax_o", device, lambda: torch.zeros(1, dtype=torch.float32, device=device))
+        with _torch_stream_context(current_stream, device):
+            amax_s_buf.zero_()
+            amax_o_buf.zero_()
```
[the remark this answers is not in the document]
YangXu1990uiuc: Both halves addressed, one of them was already fixed in a later commit than the one reviewed.

**Stream ordering:** `o_view.copy_()` and `amax_o_buf.div_()` are both inside `_torch_stream_context(current_stream, device)` on the branch head — that landed after this review.

**`reshape` copying:** still live, and correct. `reshape(-1)` returns a **copy** for a non-contiguous input, so the kernel would `atomicMax` into a temporary and the caller would read back the zeros it was reset to. Now goes through a `_amax_slot()` helper that uses `view()` and raises with a clear message instead. Applied to the SM100 fp8 path too, which had the identical bug pre-existing on `develop`.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -1777,6 +1837,96 @@ def execute(
...
+        patterns (see the kernel docstring).
+        """
+        import cutlass
+
+        def _scalar(t, default=1.0):
+            return float(t.reshape(-1)[0].item()) if t is not None else default
+
+        dq, dk, dv, so = _scalar(descale_q), _scalar(descale_k), _scalar(descale_v), _scalar(scale_o)
+        scale_softmax_log2 = scale_val * dq * dk * math.log2(math.e)
+        o_scale_fused = dv * so
```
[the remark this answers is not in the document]
YangXu1990uiuc: Split verdict after checking each half against the code.

**`descale_s`/`scale_s`: confirmed, and worse than described.** They reached no code at all — the analyzer never recorded them, so they were dropped before the adapter. Now threaded through facts → `SdpaBinding` → `bound_tensors()` → execute. (My first attempt at this was itself incomplete: I threaded them but forgot `bound_tensors()`, so `resolve_variant_pack()` never resolved the buffers and the validation could never fire. Caught by a later review pass.)

Note the semantics are stronger than a round-trip check: S **is** converted to e4m3 for the PV MMA, just unscaled, so a reciprocal pair like `(448, 1/448)` is *not* equivalent to `(1, 1)` — it asks for a different quantization range. Only the exact unit pair is honoured. See the open scope question in the PR comment.

**`seq_q_lens`: confirmed.** I initially concluded this was already rejected in `check_support` — that was wrong. `lower_dsl_prefill` clears `seq_q_lens_present` for fp8/mxfp8 *before* the adapter guard can see it (the code comment calls it a "known gap"), so a padded graph with short query lengths was admitted and wrote O and a finite LSE past the valid length. Now checked at execute, where the device value is readable — a plan-time decline was my first attempt and it wrongly rejected the `seq_len_q == S_q` case that works today (4 SM100 tests caught it).

python/cudnn/sdpa/fwd/engines.py
```diff
@@ -495,6 +499,92 @@ def _sm120_spec() -> EngineSpec:
...
+            tile_ms=frozenset({64, 128}),
+            tile_ns=frozenset({64, 128}),
+            cgas=frozenset({1}),
+        ),
+        lower=partial(lower_dsl_prefill, api_type=_SM120),
+        knob_order=_sm120_fp8_knob_order,
+    )
+
+
+def knob_candidates(spec: EngineSpec, graph) -> List[Optional[SdpaFwdKnobs]]:
```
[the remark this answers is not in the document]
YangXu1990uiuc: Moot — `knob_candidates` is deleted. #528 moved ranking out of the engines, so the whole `propose_plans`/`knob_order`/`knob_candidates` trio it belonged to is gone.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -0,0 +1,271 @@
...
+    _select_engine(g, engine_name(arch="sm120", fp8=True), tiles=tiles)
+    g.check_support()
+    g.build_plans()
+    vp.update({o: Ob, stats: lse, amx_s: amax_s, amx_o: amax_o})
+    g.execute(vp, torch.empty(max(g.get_workspace_size(), 1), device=dev, dtype=torch.uint8))
+    torch.cuda.synchronize()
+
+    ref_kw = _ref_kwargs(sdpa_kwargs)
+    o_ref, amax_s_ref = _ref(Q8.float() * dq, K8.float() * dk, V8.float() * dv, scale=scale, seq_lens_kv=seq_lens_kv, **ref_kw)
+    return Ob, o_ref, amax_s.item(), amax_s_ref, amax_o.item(), o_ref.abs().max().item()
```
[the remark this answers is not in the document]
YangXu1990uiuc: Fixed. `_ref` now returns `torch.logsumexp(scores, dim=-1)`, threaded through `_run` and compared in `_check` — including the `-inf` pattern for fully-masked rows, which is the part a tolerance-only check would miss.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -0,0 +1,250 @@
...
+"""End-to-end tests for the FROST SM120 DSL per-tensor FP8 SDPA-forward engine.
+
+Drives ``graph.sdpa_fp8`` (FP8 E4M3 Q/K/V + scalar per-tensor descales) routed to
+the ``sdpa_fwd_prefill_sm120_fp8`` engine, and validates O against an fp32-dequant
+reference. ``Amax_S`` and ``Amax_O`` are both produced in-kernel (bitcast-int32
+atomicMax over the pre-cast fp32 values); both are checked.
+
+SM120 v1 envelope (see engines._sm120_fp8_spec): E4M3 in / FP16 out only, exact
+d128, causal / bottom-right / SWA / KV-padding masks; no sink (Amax_S semantics),
+no THD. E5M2 and fp8 outputs are covered by negative tests.
```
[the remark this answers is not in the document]
YangXu1990uiuc: Fixed — corrected the docstring rather than adding tests it was describing. It now points at the actual negative coverage instead of claiming FP8-output and `D != 128` tests that do not exist.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -0,0 +1,250 @@
...
+    if lb is not None:
+        out["swa_window"] = lb - 1
+    return out
+
+
+def _check(O, O_ref, amax_s, amax_s_ref, amax_o, amax_o_ref):
+    diff = (O.float() - O_ref).abs().max().item()
+    assert diff <= 5e-2, f"max|O-ref|={diff:.4f} > 0.05"
+    assert abs(amax_s - amax_s_ref) <= 0.03, f"amax_s {amax_s:.4f} vs ref {amax_s_ref:.4f}"
+    assert abs(amax_o - amax_o_ref) <= 0.03, f"amax_o {amax_o:.4f} vs ref {amax_o_ref:.4f}"
```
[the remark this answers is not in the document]
YangXu1990uiuc: Fixed — `O` renamed to `out` at every site Ruff flagged.

docs/fe-oss-apis/attention/sdpa-fp8-sm120.md
```diff
@@ -0,0 +1,132 @@
...
+(dummy buffers when the graph doesn't request them) instead of specializing
+the template on amax presence. Tradeoff: one warp-REDUX + one global RED per
+warp of pure overhead in the no-amax case — measured noise — versus 2× fewer
+template specializations.
+
+**FP16 O only.** An fp8 O needs a `scale_o`-quantizing epilogue store
+(`F2FP.SATFINITE` pack + byte stmatrix path). Straightforward follow-up; cut
+from v1 to keep the epilogue identical to the proven f16 one. Note the fp16
+epilogue already does the `o_scale_fused` multiply, so adding fp8-O is a store
+-path change only.
```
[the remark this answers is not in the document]
YangXu1990uiuc: Stale — that text no longer exists in the file, and no hyphenated word is split across lines anywhere in it now.

YangXu1990uiuc:
## P quantization implemented, per the FORT convention

Following up on the open scope question — implemented rather than shipped as v1.

cuDNN’s `Scale_S`/`Descale_S` quantize **P, the softmax output**, not the scores: the frontend graph applies `Scale_S` *after* softmax and *after* `Amax_S`, then hands `Descale_S` to bmm2 ([scaled_dot_product_flash_attention.h:983-995](../blob/develop/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h#L983-L995)). The API name is a misnomer.

The kernel now matches the backend’s FORT ordering — amax on the unscaled softmax result, then scale, then cast:

- P is multiplied by `scale_s` immediately before the `cvt.rn.satfinite.e4m3x2`.
- `descale_s` folds into `o_scale_fused = descale_s · descale_v · scale_o`.
- `tile_sum` keeps consuming the **unscaled** P, so the softmax denominator — and therefore `Amax_S`, which is derived from `1/row_sum` — are unaffected. This mirrors FORT running `generate_amax_reduction_ops_fort` before `generate_scale_ops_fort`.

SM100 deliberately keeps the decline: its kernel has no `Scale_S` plumbing.

### The test that can actually fail

`Scale_S` and `Descale_S` are reciprocal in normal use, so a kernel that applies both and one that ignores both produce **identical O** — every existing test passes either way, and simply switching them to non-unit values would have "confirmed" nothing. `test_fp8_sm120_s_scales_are_actually_applied` breaks the reciprocity: with `Descale_S = k/Scale_S` the output must scale by exactly `k`. One-sided application gives 448× or 1/448×; ignoring both gives 1×.

### Cost

**~0.7–0.9%** at large shapes — six of six 8192-cells positive in a 0.65–0.92% band. The 30-cell geomean (1.0089) sits at the ~1% run-to-run floor and is not separately informative.

A free version exists and is deliberately left for its own change: fold `log2(scale_s)` into the `exp2` addend (already an `fma2` operand) and compensate `row_sum`/`Amax_S`/LSE once per row rather than per element.

### Correction to my earlier numbers

The tile-rule regret I posted before (`1.0058 geomean / 1.155 worst`) came from an **unseeded** sweep. Re-running the same code seeded gives **1.0046 / 1.039**, and the 1.155 cell turns out to be a tie — it was noise at a shape where both tiles are equal. I had built a claim about the rule’s feature set on that outlier; it does not survive repetition. What does survive is that the misses cluster on causal shapes, which the 22-shape fit in the design doc reports independently. The docstring and doc now say so, and the sweep is seeded.

### Final verification

| target | result |
|---|---|
| SM120 guards + tile rule | 139 passed |
| SM120 all `sdpa` | 179 passed |
| SM120 fp8, run twice | 18 passed, identical |
| SM120 `test_mhas_v2` | 245 passed — routing unchanged, frost 87 / native 158 |
| SM100 `sdpa` + `gemm` | 4684 passed |
| SM100 `linear_attention` | 353 passed |
| CPU dispatch | 150 passed |

`docs/fe-oss-apis/attention/sdpa-fp8-sm120.md` updated: the “Tile options” section described `propose_plans` and `fp8_tile_choice`, both deleted by #528, and there is now a design entry for P quantization with its cost.

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc:
### Scale_S scope: SM120 applies it, SM100 declines a non-reciprocal pair

Following up on the P-quantization question. Implementing `Scale_S` in the SM100 d=128 FP8 kernel was tried and reverted — it fails, and it turns out there is nothing to win.

**Why it fails.** That kernel uses a lazy rescale: the running max is only refreshed when a tile exceeds it by `RESCALE_THRESHOLD = 8` (log2), so its `P` is bounded by `2^8 = 256`, not by 1. e4m3 tops out at 448, so the headroom above 1.0 is *already spent* on the max-refresh skip — only `scale_s <= 448/256 = 1.75` is provably safe. With `scale_s = 448` the lazily-skipped tiles saturate: 45/46 fp8 cases passed and `test_fp8_masks[swa-e4m3]` failed at `max|O-ref| = 0.0686 > 0.05`. SWA is the worst case because a sliding window shifts the running max the most, which is exactly when the skip leaves the most slack.

This is the same constraint that makes the cuDNN backend's SM100 path ignore the pair.

**Why there is nothing to win.** Sweeping the scale on B2xH8xS256 e4m3:

| scale_s | 1 | 1.75 | 2 | 4 | 8 | 32 | 64 | 448 |
|---|---|---|---|---|---|---|---|---|
| swa | .0239 | .0270 | .0239 | .0239 | .0239 | .0239 | .0239 | **.0807** |
| causal | .0239 | .0270 | .0239 | .0239 | .0239 | .0239 | .0239 | **.0270** |
| none | .0044 | .0040 | .0044 | .0044 | .0044 | .0044 | .0044 | **.0090** |

Flat to the digit from 1 to 64. e4m3 is floating point, so relative precision does not move with scale, and subtracting the row max is already a **per-row** placement — which strictly dominates a per-tensor scale. The accuracy gain predicted in the design doc does not exist; the doc has been corrected.

**So:**
- **SM120** applies `Scale_S` before the e4m3 cast (`Descale_S` folded into `o_scale_fused`, FORT ordering). That kernel has no lazy rescale, so `P <= 1` and the full range is available.
- **SM100** honours a reciprocal pair analytically — nothing applied, so nothing owed back, which is *exact* — and declines a non-reciprocal one rather than silently returning a different O.

Validated at 48e70cab6 on both architectures: SM100 (parley Blackwell) 50 passed fp8+mxfp8 / 567 passed all sdpa; SM120 (RTX PRO 6000 Blackwell Server Edition) 18 passed fp8-sm120 / 179 passed all frost sdpa; 80 passed guards+tile-rule on each.

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "cudnn-FE #509 SM120 FP8 engine + Scale_S scope"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

YangXu1990uiuc:
@cudnn-ci-bot run frost

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
```diff
@@ -0,0 +1,1584 @@
...
+  ``descale_v * scale_o`` into the ``o_scale_fused`` scalar (cuDNN SDPA_FP8
+  node convention; see the SM100 fp8 adapter).
+- K B-fragments: classic byte-preserving ``ldmatrix.m8n8.x4.b16`` (sm_120a has
+  no non-transposed 8-bit ldmatrix; the b16 form gathers the right bytes when
+  each lane points at one 16-byte K-segment).
+- V B-fragments: hardware 8-bit transposed ``ldmatrix.m16n16.x2.trans.b8``
+  (SASS ``LDSM.8.MT1616``) — one issue covers a 32(kv) x 16(d-bytes) tile and
+  feeds two MMAs, register map (0, 2, 1, 3).
+- P: fp32 softmax output packed to e4m3 with ``cvt.rn.satfinite.e4m3x2.f32``
+  and staged through a per-warp SMEM tile (the k32 C->A fragment-column
```
Aneureka: I think `P` exists only in registers and is synchronized across threads using `shfl_sync`, rather than being staged in shared memory and read back through `ldmatrix`. This comment seems stale, and the other comments should also be updated to reflect the current kernel implementation.

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
```diff
@@ -0,0 +1,1584 @@
...
+            )
+            return prims.ldmatrix(
+                sV_ptr,
+                4,
+                prims.MMALayout.COL,
+                shape=prims.LoadShape.M16N16,
+                src_format=prims.LoadSrcFormat.B8,
+            )
+
+        # An ldmatrix is always in flight behind the tensor cores: the next
```
Aneureka: Seems stale comment. No prefetch or wrapping index exists in the loop; `V` loads are issued in-loop right before their MMAs.

python/cudnn/sdpa/graph_analyzer.py
```diff
@@ -267,6 +267,11 @@ class SdpaGraphFacts:
     descale_k_t: Any = None
     descale_v_t: Any = None
     scale_o_t: Any = None
+    # cuDNN's Scale_S/Descale_S: they quantize P, the softmax OUTPUT. The FROST
+    # kernels convert P unscaled, so execute declines a non-unit pair rather
+    # than dropping it (api_dsl._require_unit_s_scales).
```
Aneureka: `_require_unit_s_scales` -> `_require_reciprocal_s_scales`.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -1638,18 +1701,36 @@ def check_support(self) -> bool:
...
                 desc.device != self.q_desc.device,
                 f"{desc.name} must be on device {self.q_desc.device}, got {desc.device}",
             )
+        if self._fp8:
+            self._value_error_if(
+                not self._pertensor,
+                "SM120 fp8 serves the per-tensor SDPA_FP8 op only (no MXFP8 cell)",
+            )
+            self._not_implemented_error_if(self.thd, "SM120 fp8: THD/varlen is not wired yet (dense only)")
+            self._value_error_if(self.has_sink, "SM120 fp8 does not support attention sinks (Amax_S semantics)")
```
Aneureka, line 1734: Seems like we implement THD and sink related logics in the fp8 kernel, but not enable it?

docs/fe-oss-apis/attention/sdpa-fp8-sm120.md
```diff
@@ -0,0 +1,184 @@
...
+`prmt` transposes — this instruction is why the port is clean.
+
+**P reaches the PV MMA through shfl, not SMEM.** The QK C-fragment owns columns
+`2*(t%4)+{0,1}` while the k32 A-fragment wants 4 consecutive bytes, so the f16
+kernel's in-register repack does not apply. The two layouts differ only by an
+exchange inside each thread quad, so `pack_f8x2_pairs` + two `shfl.sync.idx` +
+one `prmt.b32` produce each A register directly from the `cvt.rn.satfinite.e4m3x2`
+results left in registers by the softmax.
+
+v1 of this kernel staged P through a per-warp `16 × kv_tile` SMEM tile instead
```
Aneureka: Do we want to let the document include the design decisions behind the kernels, or only reflect their current state? It seems the other FE OSS API documents focus on kernel interfaces and high-level design.

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
YangXu1990uiuc: Confirmed, and the code says so in its own words — `mma_pv` notes that two `shfl.sync` and one `prmt` "replace the SMEM round trip (and the 16 KB it needed)". The header was describing a design that got removed. Rewritten to say P stays in registers and the k32 C->A fragment-column mismatch is a thread-quad exchange.

Thanks for the "other comments" nudge — `P scale is fixed 1.0` on the next line went stale in this very PR, since Scale_S now multiplies P before the cast. Fixed in the same pass.

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
YangXu1990uiuc: Confirmed. `load_v_frags` is called inside the inner loop immediately before the MMAs that consume it; there is no wrapping index.

I nearly replaced it with "the extra live registers cost more than the latency it hid", then checked the measurement: V prefetch landed **within ±0.5%** against a ~1% run-to-run floor, i.e. no effect — because keeping P in registers leaves little ldmatrix latency to hide in the first place. The comment now says that instead.

python/cudnn/sdpa/graph_analyzer.py
YangXu1990uiuc: Fixed. The rest of that comment was wrong in two further ways once I looked: it claimed "the FROST kernels convert P unscaled", but SM120 now applies the scales, and the criterion is non-**reciprocal**, not non-unit (a reciprocal pair is served exactly, since nothing is applied and nothing is owed back). Rewritten to name both rows.

python/cudnn/sdpa/fwd/api_dsl.py
YangXu1990uiuc: Good catch, and the two halves turned out to differ.

**Sink** is not implemented at all here — there is no sink math in the kernel, only a rejection (`sinks` is always `None`, and both the kernel ctor and the adapter refuse `has_sink`). Amax_S semantics with a sink column are undefined for this cell. Now documented as "no sink math exists here" rather than implying dormant support.

**THD** was real, and is now wired and tested. Four layers were blocking it, each hiding the next:

1. the engine row declared `thd=False` ("deferred, dense execute only for v1");
2. `sdpa_support_surface.h` rejected `(prop_major == 12 && is_ragged)` outright — removed, and verified safe: with the FROST opt-in off, a ragged fp8 sm120 graph now declines at plan time via the backend's own engine-config check, so that guard was a redundant early-out;
3. the execute-time `seq_len_q` guard rejected any per-batch length `< S_q`, which under THD is the definition of ragged;
4. the ragged LSE is head-major `(H, head_stride)` — the kernel writes `lse[head, q_row_base + row]` — and its extent was pinned to the packed total, so a padded token capacity was inexpressible. The fake tensor is 2-D now, like the f16 cell's.

`_thd_pack` is shared with the f16 THD execute rather than duplicated; the f16 ragged suite is unchanged. New coverage: `test_fp8_sm120_thd{,_cross,_stats,_gqa}`.

docs/fe-oss-apis/attention/sdpa-fp8-sm120.md
YangXu1990uiuc: Agreed — dropped the file. It had become kernel design rationale (rejected alternatives, measured costs, tile-rule fits), which is not what the `fe-oss-apis` docs are for, and this change has no interface surface to document. The rationale worth keeping now lives at the code it explains or in the commit messages.

YangXu1990uiuc:
### THD (ragged) now served on the SM120 FP8 engine

Pushed `a49889988`, answering @Aneureka's review. All four code observations were correct; details are in the threads. The THD one turned out to be the substantial half, so summarising here.

`sinks` and `thd_varlen` were not the same situation. Sink has **no math in this kernel at all** — only a rejection, with `sinks` always `None` — so it is now documented as absent rather than dormant. THD was genuinely implemented in the kernel and merely unreachable. Four layers were blocking it, each hiding the next:

1. the engine row declared `thd=False` ("deferred, dense execute only for v1");
2. `sdpa_support_surface.h` rejected `(prop_major == 12 && is_ragged)` outright;
3. the execute-time `seq_len_q` guard rejected any per-batch length `< S_q`;
4. the ragged LSE layout.

**On (2)** — that guard also governs the native backend, so I did not simply delete it. With it removed and the FROST opt-in **off**, a ragged fp8 sm120 graph declines at plan time through the backend's own engine-config check (`No valid engine configs for {"engineId":8,"smVersion":1200,...}`). So it was a redundant early-out, not a capability statement, and removing it does not open a hole in the backend's support surface.

**On (3)** — that guard was mine, added earlier in this PR for the dense path, where a short `seq_len_q` would write O and a finite LSE past the valid length. Under THD, lengths shorter than `S_q` *are* the point, and the packed layout gives each sequence its own extent. Exempted rather than loosened.

**On (4)** — the ragged LSE is head-major `(H, head_stride)`: the kernel writes `lse[head, q_row_base + row]`, tokens contiguous within a head row. I initially had this backwards and declined head-major. Its extent was also pinned to the packed token total, so a caller's padded capacity was inexpressible; the fake tensor is 2-D now, matching the f16 cell, with the compile signature and the host rank check following.

`_thd_pack` is shared with the f16 THD execute rather than duplicated. New coverage: `test_fp8_sm120_thd{,_cross,_stats,_gqa}`. The cross-attention case deliberately uses Q shorter than KV — with bottom-right alignment a longer Q leaves leading rows with no valid column, where the reference's all-`-inf` softmax is NaN while the kernel writes the dead-row `O=0`; dead rows have their own coverage elsewhere.

Also parameterises the grouped-query test over `H_kv ∈ {1, 2}` so MQA is covered (CodeRabbit), and drops `docs/fe-oss-apis/attention/sdpa-fp8-sm120.md`.

| target | result |
|---|---|
| sm120 (RTX PRO 6000 Blackwell) fp8, incl. THD + MQA | 24 passed |
| sm120 all frost `sdpa` | 183 passed |
| sm120 native ragged fp8 (opt-in off) | declines cleanly |
| sm100 (parley Blackwell) fp8 + mxfp8 | 50 passed |
| sm100 all `sdpa` | 567 passed |
| dispatch + tile rule (no GPU) | 80 passed |

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "cudnn-FE #509 SM120 FP8 engine + Scale_S scope"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc:
### Rebased onto develop (#531) — conflicts resolved

`c98a35209`. #531 landed on the same sm120 frost sdpa_fwd surface, so two files conflicted. Both were semantic rather than textual, so noting how they were resolved:

**`config_sm120.validate_params`** — #531 relaxed `window_right` because the f16 kernel now does right-band widening; this branch had generalised the dtype check to `allowed_dtypes` for FP8. Taking either side alone was wrong: #531's version silently lets a widened band reach the FP8 template, which does not plumb it. Kept both and added an explicit `allow_right_band=False` that the FP8 template passes, so the limitation is stated rather than inherited. (The FP8 engine row also declares no `right_band_widening`, so such a graph is declined at dispatch too.)

**`api_dsl` SMEM sizing** — #531 sizes tiles at the envelope-padded head dims; this branch sizes the output term separately, since FP8 stages a byte per KV element but writes O in half. Both needed, combined into one call.

Validated at `c98a35209` on an sm120 board (RTX PRO 6000 Blackwell), after asserting both resolutions were actually installed:

| target | result |
|---|---|
| fp8 sm120, incl. THD + MQA | 24 passed |
| all frost `sdpa` (f16 + fp8 — #531's features included) | 192 passed, 480 skipped |
| dispatch + tile rule (no GPU) | 80 passed |
| sm100 (parley Blackwell) fp8 + mxfp8 / all `sdpa` | 50 passed / 569 passed |

---

### On the earlier `frost_tests:sm120` failure (job 391137376) — not from this PR

It reads like OOM but is not: it is `cudaErrorLaunchTimeout`, the RTX 5080's display watchdog. There is exactly **one** real failure — `test_sdpa_random_fwd_L0[test3]` — and the other 128 "ERROR at setup" entries are one sticky-context cascade, all at the same timestamp, surfacing through `conftest`'s `mem_get_info()` gate.

The failing graph is **fp16** on `sdpa_fwd_prefill_sm120` (the f16 row): `d_qk=d_v=240`, `s=616`, GQA 3:1, BOTTOM_RIGHT + padding. Attribution, checked rather than assumed:

- Every functional change this PR makes to shared dispatch code is gated on `is_fp8` / `is_mxfp8`; the f16 row is neither. The shared tile-rule change is a no-op for f16 (`(2,2)` is the previous default).
- A/B on one sm120 board, develop `b51a732b2` vs this branch: **21 passed on both**, identical routing, no hang (60.6s vs 54.4s).
- Not #531 either — that merged at 07:59 UTC, two hours *after* this job ran, and `d=240` was already inside the pre-#531 `d_qk=frozenset(range(16, 257, 16))`.

So it is a slow kernel meeting a consumer GPU's watchdog, not a correctness bug. Worth a look separately: the job runs pytest with 4 workers (`gw0`–`gw3`) sharing one RTX 5080, so contention plausibly pushes an already-heavy `d=240` case past the ~2s watchdog — which would make it flaky rather than deterministic. Caveat on my repro: the server board has no watchdog and a different SM count (which feeds the tile rule), so it cannot reproduce CI's exact kill; it does establish that this branch and develop behave the same.

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "cudnn-FE #509 SM120 FP8 engine + Scale_S scope"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

YangXu1990uiuc:
@cudnn-ci-bot run frost

Aneureka (approved):

PR #517, opened 2026-08-07
sdpa: add torch custom ops cudnn::sdpa_fwd / cudnn::sdpa_bwd

vedaanta:
## What

PyTorch custom ops — **`torch.ops.cudnn.sdpa_fwd` / `sdpa_bwd`** — exposing the cuDNN SDPA feature surface that `torch.nn.functional.scaled_dot_product_attention`'s aten contract cannot express:

- **attention sinks** (per-Q-head logits in the softmax denominator)
- **sliding window** (`window_left`, cuDNN visible-tokens convention)
- **bottom-right causal alignment**
- **padded batches** (`seq_len_q` / `seq_len_kv`)
- **THD / varlen packing** (FA-style `(T, H, D)` + `cu_seqlens`; non-contiguous K/V views of a fused KV projection are declared with their true strides)

The ops build pygraph `sdpa` / `sdpa_backward` nodes; the engine Router picks the serving plan (FROST OSS kernels or cuDNN-backend engines) per configuration. One new file + tests + docs; no existing behavior changes (the experimental dense module's backward is renamed `cudnn::sdpa_bwd_legacy` so both modules can coexist in one process until it is removed).

## Contract

| Aspect | Guarantee |
|---|---|
| `torch.compile` | `register_fake` meta kernels mirror the real output strides; `torch.library.opcheck` passes on dense and THD paths **including dynamic-shape AOT dispatch**, locked in by a test |
| autograd | `sdpa_fwd` is differentiable on the varlen path (`register_autograd`); the glue converts packed TH1 stats to the padded LSE layout **device-side** (no host reads — tracing/capture safe). Dense + sink backward raise `NotImplementedError` until their engine contracts land |
| threading | thread-local cuDNN handles, serialized graph builds, bounded (FIFO) graph cache |
| validation | one io dtype per call; k/o/grad_out shape checks; `h_k != h_v` GQA head splits validated (independent K/V head counts); int32 ragged-offset overflow guards; inert `causal_bottom_right` rejected; base-pointer realignment uses `clone()` (`contiguous()` cannot fix a misaligned base) |
| backward LSE | padded `(B, H, max_seqlen_q, 1)` fp32 — backend restriction (bprop THD rejects ragged LSE on SM8X/SM12X) |

## Tests (19, all L0 — `test/python/test_cudnn_sdpa_torch_ops.py`)

Dense: sinks (±causal) with **LSE value checks**, `h_k != h_v` GQA, **query-layout adoption** (4 permutations) vs an fp32 reference, sliding windows, bottom-right cross-seqlen, sinks+window, padded per-batch lengths. THD: fwd with LSE value checks, fwd+bwd (MHA + GQA), **end-to-end autograd**, kv-interleaved `(T,2,H,D)` views. Contract: opcheck (dense + THD/dynamic). All green on SM100 (cc 10.0) + previously SM120, both engine routes.

Docs: `docs/fe-oss-apis/sdpa-torch-ops.md`.

## Consumers

`torch.nn.attention.varlen` and `F.sdpa` integration builds on these ops in #554 (the `cudnn.torch` provider); TE/vLLM/Megatron-style frameworks can call them directly.

## Summary by CodeRabbit

* **New Features**
  * Added an extended scaled dot-product attention interface for dense and variable-length inputs.
  * Added support for attention sinks, sliding-window masking, causal alignment, padded sequences, grouped-query attention, and packed layouts.
  * Added optional log-sum-exp output and forward/backward operations.

* **Bug Fixes**
  * Renamed the legacy attention backward operator to avoid conflicts with the new implementation.

* **Documentation**
  * Added usage and capability documentation for the PyTorch attention operators.

* **Tests**
  * Added CUDA coverage for extended attention behavior, layouts, masking modes, and gradient results.

### DIFF
--- a/docs/fe-oss-apis/sdpa-torch-ops.md
+++ b/docs/fe-oss-apis/sdpa-torch-ops.md
@@ -1,0 +1,69 @@
+# SDPA torch custom ops: `cudnn::sdpa_fwd` / `cudnn::sdpa_bwd`
+
+PyTorch custom ops (`torch.library`) exposing the full cuDNN SDPA feature
+surface — the features `torch.nn.functional.scaled_dot_product_attention`'s
+aten contract cannot express:
+
+- **attention sinks** — per-Q-head logits folded into the softmax denominator
+- **sliding window** — `window_left` (cuDNN convention: visible tokens
+  *including* self; FA2's `(w, 0)` maps to `window_left = w + 1`)
+- **bottom-right causal alignment** — inference-style diagonals
+- **padded batches** — per-batch actual lengths via `seq_len_q` / `seq_len_kv`
+- **THD / varlen packing** — FlashAttention-style `(T, H, D)` + `cu_seqlens`
+
+The ops build cuDNN pygraph `sdpa` / `sdpa_backward` nodes; the engine Router
+picks the best serving plan (FROST OSS kernels or cuDNN-backend engines) per
+configuration. Graphs are cached per configuration (bounded, thread-safe;
+cuDNN handles are thread-local).
+
+## Usage
+
+```python
+import torch
+import cudnn
+
+_ = cudnn.sdpa_torch  # lazy public export: importing registers cudnn::sdpa_fwd / cudnn::sdpa_bwd
+
+# Dense BHSD with sinks + sliding window
+o, lse = torch.ops.cudnn.sdpa_fwd(q, k, v, scale, is_causal=True,
+                                  window_left=128, sinks=sinks, return_lse=True)
+
+# THD / varlen (FA-style packed (T, H, D) + cu_seqlens), differentiable:
+q, k, v = (t.requires_grad_(True) for t in (q_thd, k_thd, v_thd))
+o, lse = torch.ops.cudnn.sdpa_fwd(q, k, v, scale, is_causal=True,
+                                  cu_seqlens_q=cu, cu_seqlens_kv=cu,
+                                  max_seqlen_q=mx, max_seqlen_kv=mx,
+                                  return_lse=True)
+o.backward(grad)  # routes through cudnn::sdpa_bwd via register_autograd
+
+# Or through the python wrapper (same op underneath):
+o = cudnn.sdpa_torch(q, k, v, is_causal=True, cu_seqlens_q=cu, cu_seqlens_kv=cu,
+                     max_seqlen_q=mx, max_seqlen_kv=mx)
+```
+
+## Contracts and limits
+
+- Dense tensors are BHSD `(B, H, S, D)` (any strides; the graph declares the
+  actual layout). Varlen tensors are packed `(T, H, D)`; non-contiguous views
+  (e.g. K/V slices of a fused `(T, 2, H, D)` KV projection) are declared with
+  their true strides. On the varlen path, a non-dense innermost dim or a
+  misaligned base pointer is repaired by one copy (warned as slow path); the
+  dense path declares the given strides as-is.
+- One io dtype per call (`fp16` or `bf16`); mixed-dtype inputs are rejected.
+- `sdpa_bwd` serves the **THD/varlen** path. Dense backward and sink backward
+  (dSink) are follow-ups and raise `NotImplementedError`. It consumes a
+  **padded** `(B, H, max_seqlen_q, 1)` fp32 LSE (backend restriction: bprop
+  THD rejects ragged LSE on SM8X/SM12X).
+- Autograd (`register_autograd`) requires `return_lse=True` on the forward;
+  the glue converts the packed TH1 stats to the padded layout device-side.
+- Both ops ship `register_fake` meta kernels. `cudnn::sdpa_fwd` passes
+  `torch.library.opcheck` on the dense and varlen paths, including
+  dynamic-shape AOT dispatch (`torch.compile`-ready); the opcheck autograd
+  case exercises `cudnn::sdpa_bwd` through the registered backward.
+
+## Requirements
+
+- `nvidia-cudnn-frontend[cutedsl]`, cuDNN backend ≥ 9.6 (THD token-major
+  stats), sm80+.
+
+Tests: `test/python/test_cudnn_sdpa_torch_ops.py`.
--- a/python/cudnn/experimental/ops/sdpa.py
+++ b/python/cudnn/experimental/ops/sdpa.py
@@ -488,7 +488,10 @@ _lib.define(
 )
 
 _lib.define(
-    "sdpa_bwd(Tensor dO, Tensor q, Tensor k, Tensor v, Tensor o, Tensor stats, "
+    # Renamed from cudnn::sdpa_bwd: the canonical name now belongs to the
+    # consolidated op family in cudnn.sdpa.fwd.torch_op (this experimental
+    # module is slated to fold into it).
+    "sdpa_bwd_legacy(Tensor dO, Tensor q, Tensor k, Tensor v, Tensor o, Tensor stats, "
     "float attn_scale, bool is_causal=False, int diagonal_alignment=0, "
     "int left_bound=-1, int right_bound=-1, "
     "Tensor? seq_len_q=None, Tensor? seq_len_kv=None, "
@@ -625,7 +628,7 @@ def _sdpa_fake(
     return O, Stats
 
 
-def _sdpa_bwd_impl(
+def _sdpa_bwd_legacy_impl(
     dO: torch.Tensor,
     q: torch.Tensor,
     k: torch.Tensor,
@@ -723,11 +726,11 @@ def _sdpa_bwd_impl(
     return dQ_gpu, dK_gpu, dV_gpu
 
 
-_lib.impl("sdpa_bwd", _sdpa_bwd_impl, "CUDA")
+_lib.impl("sdpa_bwd_legacy", _sdpa_bwd_legacy_impl, "CUDA")
 
 
-@torch.library.register_fake("cudnn::sdpa_bwd")
-def _sdpa_bwd_fake(
+@torch.library.register_fake("cudnn::sdpa_bwd_legacy")
+def _sdpa_bwd_legacy_fake(
     dO: torch.Tensor,
     q: torch.Tensor,
     k: torch.Tensor,
@@ -800,7 +803,7 @@ def _sdpa_backward(ctx, dO, dStats):
         idx += 1
     cum_kv = saved[idx] if ctx.has_cum_kv else None
 
-    dQ, dK, dV = torch.ops.cudnn.sdpa_bwd(
+    dQ, dK, dV = torch.ops.cudnn.sdpa_bwd_legacy(
         dO,
         q,
         k,
--- a/python/cudnn/sdpa/fwd/torch_op.py
+++ b/python/cudnn/sdpa/fwd/torch_op.py
@@ -1,0 +1,973 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""PyTorch custom ops exposing the FULL cuDNN SDPA feature surface.
+
+``torch.ops.cudnn.sdpa_fwd`` / ``sdpa_bwd`` — the family-local torch contract
+for features ``torch.nn.functional.scaled_dot_product_attention`` cannot
+express:
+
+- **attention sinks** (per-Q-head logits folded into the softmax denominator)
+- **sliding window** (``window_left``)
+- **bottom-right causal alignment** (inference-style diagonals)
+- **padded batches** (per-batch actual sequence lengths)
+- **THD / varlen packing** (FlashAttention-style ``(T, H, D)`` + ``cu_seqlens``)
+
+Layout: dense tensors are BHSD ``(B, H, S, D)``; varlen tensors are packed
+``(T, H, D)`` with ``cu_seqlens_q/kv`` (``(B+1,)`` int32 token prefix sums, as
+in FA's ``flash_attn_varlen_func``).
+
+The ops build cuDNN pygraph ``sdpa`` / ``sdpa_backward`` nodes; the engine
+Router then picks the best serving plan (FROST OSS kernels or cuDNN-backend
+engines) per config.
+
+Backward contract: ``sdpa_bwd`` serves the THD/varlen path (dense backward and
+sink backward are follow-ups and raise ``NotImplementedError``). It consumes a
+PADDED ``(B, H, max_seqlen_q, 1)`` fp32 LSE — a backend restriction (bprop THD
+rejects ragged LSE on SM8X/SM12X). ``sdpa_fwd`` is differentiable on the
+varlen path via ``torch.library.register_autograd`` when called with
+``return_lse=True``; the autograd glue converts the packed TH1 stats to the
+padded layout the backward needs.
+
+Public entry point: ``cudnn.sdpa_torch`` (lazy export — accessing it imports
+this module, which registers the ops; ``torch`` is not imported before then).
+``import cudnn.sdpa.fwd.torch_op`` works too.
+"""
+
+import logging
+import threading
+from enum import IntEnum
+from typing import Dict, Optional, Tuple
+
+import torch
+
+import cudnn
+
+_logger = logging.getLogger(__name__)
+
+_TORCH_DTYPE_TO_CUDNN = {
+    torch.float16: cudnn.data_type.HALF,
+    torch.bfloat16: cudnn.data_type.BFLOAT16,
+}
+
+# cuDNN handles are NOT thread-safe (simultaneous use of one handle from two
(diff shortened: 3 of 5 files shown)


vedaanta:
All CodeRabbit findings triaged and addressed in `98331fe18` (single atomic commit; ops renamed to the canonical `cudnn::sdpa_fwd` / `sdpa_bwd` per maintainer direction). Point by point:

**Fixed — correctness:**
- **`contiguous()` cannot repair a misaligned base (Critical)** → `_normalize_thd` and the LSE path use `clone(memory_format=contiguous_format)`, which always allocates; the LSE guard also checks `data_ptr() % 16` and normalizes *before* `reshape`.
- **Fake kernels reported strides the real kernels don't produce** → the fwd fake mirrors the real O-stride selection (BSHD-physical detection); the bwd fake returns fresh contiguous empties in `q.dtype` instead of `empty_like` over possibly-non-contiguous inputs. Locked in by a new `opcheck` test (dense + THD, including dynamic-shape AOT dispatch).
- **Backward silently wrong after a sink forward** → `sdpa_bwd` now takes `sinks` and raises `NotImplementedError` when set (dSink is a follow-up), so the failure is loud.
- **`sdpa_fwd` not differentiable** → `register_autograd` wired for the varlen path; the packed-TH1→padded-LSE conversion is fully device-side (no `.item()`/host reads — this is what makes the dynamic-shape opcheck pass). Dense/sink autograd raise clearly. End-to-end autograd test added.
- **`k` never shape-validated** → shape checks on both paths (plus `o`/`grad_out` in bwd).
- **Gradient dtypes vs single `io_dtype`** → one io dtype per call enforced (`_check_io_dtypes`, friendly error for unsupported dtypes); `dk`/`dv` allocate in `q.dtype`.
- **Handle thread-safety** → thread-local handles (one cuDNN handle must not be used from two threads), graph builds serialized under a lock, cached execution lock-free.
- **CPU `torch.empty(0)` placeholders** → pinned to `q.device`, fp32, in real and fake kernels.
- **Inert `causal_bottom_right`** → rejected when no diagonal band is active.
- **int32 ragged-offset overflow** → host-side guards (`tokens x token-stride <= INT32_MAX`) on both paths, no device sync.

**Fixed — hygiene/tests:**
- docstring documents the backward + autograd contract; E741 `O` suppressed inline (renaming would reduce clarity, as suggested); RUF059 unused bindings fixed; the unreachable dense branch of `_build_bwd_graph` removed; graph cache bounded (FIFO, 128); tests gated on sm80+ and cuDNN ≥ 9.6; **LSE value comparisons** added vs the fp32 reference (dense incl. sinks-in-denominator, and packed THD); **padded dense-batch test** added; docs page added (`docs/fe-oss-apis/sdpa-torch-ops.md`).

**Skipped, with reasons:**
- *`max_total_seq_len_q/kv` should be actual token counts* — they already are: the actual packed totals, rounded up to the backend's 64-token accumulator granularity (now commented).
- *Stale `torch.ops.cudnn.sdpa_bwd` call in the test* — not stale: the test targets THIS module's `cudnn::sdpa_bwd`; it is the experimental module's op that was renamed (`sdpa_bwd_legacy`) to free the canonical name.
- *APIBase subclass + fe_api/ test location* — deliberate deviation: these are `torch.library` custom ops, not an OSS eager-API; `torch.library` (schema/fake/autograd/opcheck) *is* the wrapper layer here, and the tests exercise torch dispatch. Docs + gating + reference-comparison guidelines are all now met.

vedaanta:
Addendum — two more contract fixes landed in the same commit after re-running the upstream PyTorch suites against these ops (they were previously masked by the experimental dense module, which the provider no longer uses):

- **`h_k != h_v` support**: cuDNN allows K and V to carry independent head counts (each dividing `H_q`); the ops previously borrowed `H_kv` from `v` for the K descriptor. K/V/dK/dV now use separate `H_k`/`H_v` end to end, with divisibility validation. Covered by `test_gqa_hk_ne_hv` (mirrors upstream `test_cudnn_attention_gqa`, which passes again).
- **Output adopts the query's layout permutation**: the dense O-stride selection generalized from a binary BSHD/BHSD choice to Q's full dim-permutation (any B/H/S order, D innermost; broadcast inputs fall back to contiguous), with the fake kernel mirroring it — upstream `test_cudnn_attention_preserves_query_layout` (all 6 permutations × eager/compile) passes. Covered by `test_output_adopts_query_layout`.

Upstream verification on the final tip: `test_transformers.py -k cudnn` 23 pass / 0 fail; `test_varlen_attention.py` through the #554 provider 140 pass / 29 fail (all impl-identity spy asserts or paged cross-impl comparisons). Op suite: 19/19 incl. opcheck (dense + THD dynamic-shape).

python/cudnn/experimental/ops/sdpa.py
```diff
@@ -1065,10 +1068,10 @@ def _sdpa_bwd_impl(
     return dQ_gpu, dK_gpu, dV_gpu
 
 
-_lib.impl("sdpa_bwd", _sdpa_bwd_impl, "CUDA")
+_lib.impl("sdpa_bwd_legacy", _sdpa_bwd_impl, "CUDA")
 
 
-@torch.library.register_fake("cudnn::sdpa_bwd")
+@torch.library.register_fake("cudnn::sdpa_bwd_legacy")
```
Anerudhan, line 732: Can you fix the name about legacy ?

python/cudnn/sdpa/fwd/torch_op.py
```diff
@@ -0,0 +1,936 @@
...
+    DV = 203
+
+
+def _get_handle(device: torch.device):
+    """This thread's cuDNN handle for ``device``, bound to torch's current stream."""
+    handles = getattr(_tls, "handles", None)
+    if handles is None:
+        handles = _tls.handles = {}
+    if device not in handles:
+        handles[device] = cudnn.create_handle()
```
Anerudhan: May be we should guard it with the device contenxt. It silently creates on current GPU

python/cudnn/sdpa/fwd/torch_op.py
```diff
@@ -0,0 +1,936 @@
...
+    variant = {int(_UIDs.Q): q, int(_UIDs.K): k, int(_UIDs.V): v, int(_UIDs.O): o}
+    if return_lse:
+        variant[int(_UIDs.STATS)] = stats
+    if has_sinks:
+        variant[int(_UIDs.SINKS)] = sinks.to(torch.float32).reshape(1, H_q, 1, 1)
+    if is_thd:
+        # cuDNN ragged offsets are ELEMENT offsets per tensor, so each scales
+        # its token prefix sums by that tensor's OWN token stride. Small
+        # on-stream int ops — CUDA-graph-capture safe.
+        variant[int(_UIDs.RAGGED_Q)] = _int32_col(cu_seqlens_q * q.stride(0))
```
Anerudhan: We should confirm cu_seqlen_q is on device ? And _int64 as pointer ?

python/cudnn/sdpa/fwd/torch_op.py
```diff
@@ -0,0 +1,936 @@
...
+        H_k = k.shape[1]
+        # cuDNN supports h_k != h_v (each must divide h_q).
+        if k.shape != (T_kv, H_k, D_qk):
+            raise ValueError(f"k shape {tuple(k.shape)} must be (T_kv={T_kv}, H_k, D_qk={D_qk}) to match q and v")
+        if H_q % H_k or H_q % H_v:
+            raise ValueError(f"GQA head counts must divide H_q={H_q}; got H_k={H_k}, H_v={H_v}")
+        S_q, S_kv = max_seqlen_q, max_seqlen_kv
+        # cuDNN ragged offsets are int32 ELEMENT offsets: the largest offset
+        # (total tokens x token stride) must fit.
+        for name, t, total in (("q", q, T_q), ("k", k, T_kv), ("v", v, T_kv)):
```
Anerudhan: We fixed it for int64? ?

Anerudhan (changes requested):

python/cudnn/sdpa/fwd/torch_op.py
vedaanta: Done in be0cf7505 — ragged offsets are now **int64** end to end: the graph declares all 12 ragged-offset tensors as INT64 and the variant binds `_int64_col` columns (matching the test_mhas harness convention). The int32 overflow guards are gone — int64 element offsets cannot realistically overflow. Verified on both engine routes (FROST + backend), 19/19.

python/cudnn/sdpa/fwd/torch_op.py
vedaanta: Both done in be0cf7505: `_check_cu_seqlens` validates that `cu_seqlens_q/kv` live on q's device before they are bound (a CPU tensor would hand cuDNN a host address), and the offsets are int64 pointers now (INT64 graph declarations + `_int64_col` binding).

python/cudnn/sdpa/fwd/torch_op.py
vedaanta: Done in be0cf7505 — handle creation is pinned with `with torch.cuda.device(device):` so a tensor on cuda:1 never gets a handle silently created against the caller's current device.

python/cudnn/experimental/ops/sdpa.py
vedaanta: Done in be0cf7505 — the implementation identifiers now match the registered name: `_sdpa_bwd_legacy_impl` / `_sdpa_bwd_legacy_fake`. (The whole experimental module is slated for removal once dense backward lands in `cudnn::sdpa_bwd`.)

python/cudnn/sdpa/fwd/torch_op.py
```diff
@@ -0,0 +1,910 @@
...
+        tuple(k.stride()),
+        tuple(v.stride()),
+        tuple(o.stride()),
+        attn_scale,
+        is_causal,
+        causal_bottom_right,
+        window_left,
+        is_deterministic,
+        q.device,
+    )
```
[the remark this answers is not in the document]
vedaanta: Real bug — fixed in be0cf7505. The bwd cache key now includes `_round64(T_q)` and `_round64(T_kv)`: the packed token totals are baked into the graph (they size the dq accumulator via `max_total_seq_len_*`), so a plan built for smaller totals must never serve a call with larger ones. Rounding to the graph's own 64-token granularity keeps the cache hitting across calls that share an accumulator size. Thanks — this was the sharpest catch of the round.

docs/fe-oss-apis/sdpa-torch-ops.md
```diff
@@ -0,0 +1,60 @@
...
+o, lse = torch.ops.cudnn.sdpa_fwd(q, k, v, scale, is_causal=True,
+                                  window_left=128, sinks=sinks, return_lse=True)
+
+# THD / varlen (FA-style packed (T, H, D) + cu_seqlens), differentiable:
+q, k, v = (t.requires_grad_(True) for t in (q_thd, k_thd, v_thd))
+o, lse = torch.ops.cudnn.sdpa_fwd(q, k, v, scale, is_causal=True,
+                                  cu_seqlens_q=cu, cu_seqlens_kv=cu,
+                                  max_seqlen_q=mx, max_seqlen_kv=mx,
+                                  return_lse=True)
+o.backward(grad)  # routes through cudnn::sdpa_bwd via register_autograd
```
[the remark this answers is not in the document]
vedaanta: Done in be0cf7505 — `cudnn.sdpa_torch` is a lazy export in `python/cudnn/__init__.py` (accessing it imports the module, which registers the ops; `torch` stays un-imported until then — verified by an import-boundary check). The docs example now uses the public entry point.

python/cudnn/sdpa/fwd/torch_op.py
```diff
@@ -0,0 +1,956 @@
...
+        int(_UIDs.DQ): dq,
+        int(_UIDs.DK): dk,
+        int(_UIDs.DV): dv,
+        int(_UIDs.RAGGED_Q): _int64_col(cu_seqlens_q * q.stride(0)),
+        int(_UIDs.RAGGED_KV): _int64_col(cu_seqlens_kv * k.stride(0)),
+        int(_UIDs.RAGGED_V): _int64_col(cu_seqlens_kv * v.stride(0)),
+        int(_UIDs.RAGGED_O): _int64_col(cu_seqlens_q * o.stride(0)),
+        int(_UIDs.RAGGED_DQ): _int64_col(cu_seqlens_q * (H_q * D_qk)),
+        int(_UIDs.RAGGED_DK): _int64_col(cu_seqlens_kv * (H_k * D_qk)),
+        int(_UIDs.RAGGED_DV): _int64_col(cu_seqlens_kv * (H_v * D_v)),
```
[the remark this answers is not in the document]
vedaanta: Fixed in 506f693a1 — the cu prefix sums are widened to int64 BEFORE the stride multiply on BOTH paths (an int32 product would wrap before `_int64_col` ever saw it). Good catch.

vedaanta:
Rebased onto current `develop` (`8acc777cd`). One conflict, in `python/cudnn/__init__.py` — develop's new `gnn` lazy export landed on the same line as this PR's `sdpa_torch`; both kept.

Also **dropped the `#613` xfail on `test_thd_kv_packed_views`**. That fix merged in `3631ecb44` (#740), so K/V bound as views of a kv-interleaved `[T, 2, H, D]` buffer — the layout `torch.nn.attention.varlen` users produce by slicing a fused KV projection — are now served correctly instead of silently truncated. The test XPASSed, so it is unmarked and asserting for real.

Suite after the rebase: **19 passed**, no xfails (was 18 + 1 xfail).

@Anerudhan the CHANGES_REQUESTED on this is stale — every comment from that round was addressed a while back, and the PR is now rebased and green. Would appreciate a re-review when you get a chance.

vedaanta:
Rebased again onto `develop` `9ef93bf48` — clean, no conflicts this time. Both checks green (`pre-commit` + `merge-requirements`), MERGEABLE, suite still **19 passed**.

Note on the earlier `pre-commit` red: it was not from this PR. `develop`'s own Style check was broken from `781e3383b` (#693) through `b2445f03a` — black flagged `test/python/fe_api/grouped_gemm/test_grouped_gemm_glu_hadamard_quant.py`, a file this PR does not touch — and it went green again at `9ef93bf48` with that file byte-identical, so it was a hook-environment difference rather than content. Rebasing onto the fixed base cleared it.

@Anerudhan on the one review thread still open (`experimental/ops/sdpa.py`, "fix the name about legacy"): that was addressed back in `be0cf7505` — the identifiers are `_sdpa_bwd_legacy_impl` / `_sdpa_bwd_legacy_fake`, matching the registered `cudnn::sdpa_bwd_legacy`. Verified still in place after the rebase. The other 7 open threads are all marked outdated by GitHub (the code moved in the atomic rewrite). Happy to resolve them if you'd prefer, but wanted to leave that to you.

vedaanta:
Rebased onto latest `develop` `4fcc51334` (picked up ~10 commits, including #716 SM80 workspace carving, #768 split-KV ports and #757 d192 quantized-O guard, which all touch the SDPA paths this PR sits on). **No conflicts.**

Rebuilt the extension and re-ran on the new base rather than assuming the rebase was inert: **19 passed**, no xfails. Both checks green (`pre-commit` + `merge-requirements`), MERGEABLE.

Status is unchanged otherwise — the only thing outstanding is the re-review. @Anerudhan the CHANGES_REQUESTED here is from **Aug 11**, before the atomic rewrite that addressed it; of the 8 open threads GitHub marks 7 as outdated, and the 1 live one (`experimental/ops/sdpa.py`, "fix the name about legacy") was answered in `be0cf7505` — `_sdpa_bwd_legacy_impl` / `_sdpa_bwd_legacy_fake`, still in place after this rebase.

python/cudnn/experimental/ops/sdpa.py
```diff
@@ -625,7 +628,7 @@ def _sdpa_fake(
     return O, Stats
 
 
-def _sdpa_bwd_impl(
+def _sdpa_bwd_legacy_impl(
```
Anerudhan, line 631: What is the need for this?

vedaanta:
Follow-up opened as #780: **remove the experimental SDPA torch op** entirely, rather than keeping the `cudnn::sdpa_bwd` → `cudnn::sdpa_bwd_legacy` rename this PR carries.

That rename only exists to dodge a collision with the canonical `cudnn::sdpa_bwd` defined here. Deleting `cudnn.experimental.ops.sdpa` removes the collision at the source, so the rename hunk in this PR goes away once #780 lands — which also closes the last open review thread here (r3769537715).

#780 is stacked on this PR, because the three e2e benchmarks that use the experimental op can only move to `torch.ops.cudnn.sdpa_fwd` once it exists. So the order is #517 → #780.

python/cudnn/experimental/ops/sdpa.py
vedaanta: oh, I will remove in future PR

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
closing in favour of 780

PR #527, opened 2026-08-08
Rank SDPA plans in one place, and give SM120 one measured tile rule

YangXu1990uiuc:
## Before submitting

- [x] I ran `pre-commit run` and committed any formatting changes. (black 26.3.1 -l 160 on all touched python files)
- [ ] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*`. *(token lacks label permission — suggested: `cat-feature`, `mod-python-fe`, `orig-nv-eng`)*

## Affected area

FE OSS kernels or CuTeDSL — python dispatch (`cudnn.pygraph` plan ranking) and the SM120 SDPA-forward tile rule.

**Draft, stacked on #509.** GitHub will not take a fork branch as a base, so this targets `develop` and the diff carries #509's six commits too. **Review only the six commits from `ce8b3e10e` onward** — everything at or below `d7dd85ee5` belongs to #509 and disappears from this diff once it merges.

## Summary

An engine cannot rank plans. It sees neither its siblings nor the backend's entries, so `propose_plans` could only order its own knobs — and something downstream still had to merge the two sides, which `heuristics_sort` did by concatenating and calling it ranking. This moves the whole decision to one function per engine family and deletes what that makes redundant.

```
create_execution_plans([heur_mode.A, ...])
├─ _finalize_backend_layout() / _freeze() / _attach_facts()
└─ heuristics.rank(graph, engines, backend_plans, modes)     # engines/heuristics.py
   └─ <family>.recommend(modes, facts, offered, backend_plans)
        for mode in modes:
          OPENSOURCE -> our candidates only (no backend recommendation)
          A | B      -> ours + theirs, order decided by measurement
          FALLBACK   -> our safe config + the backend's fallbacks
```

What comes back **is** `graph.plans`, position for position. An engine answers two questions only: can I serve this graph (`check_support`) and compile me this config (`build_plan`).

### The backend's plans now arrive tagged by mode

Ranking the two sides against each other needs to know which backend entries are mode-A recommendations and which are fallbacks. That needed **no C++ change**: C++ appends each query to the same plan list and `get_execution_plan_count()` already exists, so asking one mode at a time gives the boundaries. Measured on a 512³ bf16 matmul (sm90, cuDNN 9.25):

| mode | plans | shape |
|---|---|---|
| `heur_mode.A` | `[0:15]` | all knob-bearing |
| `heur_mode.FALLBACK` | `[15:17]` | bare `eng0`/`eng7`, no knobs |

The two segments do not overlap.

### One SM120 tile rule instead of two, with a causal term

The f16/bf16 cell shared the fp8 cell's knob domain but had no rule, so it took whatever fit — 128x128 everywhere on this part. Measuring it produced the same rule plus one term: a causal mask halves the work per CTA, so the machine empties sooner and the finer Q tile keeps paying further out. Regret against the best of the enumerated `{64,128}²` domain:

| dataset | without the causal term | with it |
|---|---|---|
| RTX PRO 6000, 24 bf16 cells | 1.023x mean / 1.175x worst | **1.009x / 1.054x** |
| RTX PRO 6000, 28 fp8 cells | 1.0078x / 1.107x | 1.0066x / 1.107x (2 fixed, 2 broken) |
| RTX 5090 (170 SM), 14 fp8 cells | 1.0082x / 1.062x | **1.0007x / 1.006x** |
| 22 held-out fp8 shapes | 1.009x / 1.089x | flips the 3 worst cells to optimal |

On the part the rule was trained on the term is a wash; it earns its place on the second part and the held-out set, where it fixes exactly the causal cells that were the previous misses. So `fp8_tile_choice` and `f16_tile_choice` collapse into one `tile_choice`.

### Which side leads is a measurement

`_MEASURED_BEHIND` names cells timed **slower** than the backend; those follow it. sm120 fp8 is not there because it measures 1.20–1.83x the backend's native fp8 fprop across 28 shapes. A cell that has not been timed keeps the historical order and says so in a comment — moving one is an experiment, not an edit.

Coverage no longer rides on that ordering: `heur_mode.OPENSOURCE` puts everything FROST can build first whatever the measurements say, so a test asks for it and any graph still landing on a backend plan is one FROST cannot serve.

## Deleted

| | why |
|---|---|
| `BaseEngine.propose_plans` + its 3 implementations | 2 were verbatim copies of the default; nothing ever used the hook to decide anything |
| `BaseEngine.default_knobs` | only fed `propose_plans` |
| `EngineSpec.knob_order`, `knob_candidates` | ranking, in the wrong place |
| `heuristics_sort` | merging is part of ranking, not a step after it |
| `Router` class, `default_router`, `set_router`, `pygraph(router=)`, `engines/router.py` | policy has one home; `decline_types` moved to `base.py` where the contract is |
| `graph.engine` | pure alias of `selected_engine`, zero callers |
| `graph.from_serialized` | zero callers; `serialize`/`deserialize` are the pybind-era API and stay |
| `api_dsl`'s shape-driven tile choice | the tile rule ran twice — here and when ranking |

`knobs=None` no longer means "engine, pick for me". A None field survives only on an axis whose capability row declares no domain.

## Testing

| | |
|---|---|
| sdpa suites (RTX PRO 6000, sm120) | 171 passed, **4 failed** — see below |
| `test_mhas_v2` routing | 302 passed, tally unchanged: frost sm120 x87, fp8 x4, native x158/x53 |
| dispatch guards (opt-in OFF) | 69 passed |
| router / graph_native / gemm-frost (CPU) | 125 passed |

**Known, not yet fixed:** `test_dsl_sm120_representative_head_dimensions[208..256]`. At those head dims the adapter used to drop to a KV tile that fits; the heuristics now name 128 and the test pins the first entry. Traceback pending.

**Pre-existing, not from this change:** `test_a_replayed_plan_reports_its_own_notes` fails here and also fails at the base commit `d7dd85ee5` — C++ on 9.25 no longer raises for an index one past the plan count.

## Still to come on this branch

- Remove the out-of-tree engine concept: `register_backend` / `pygraph(backends=)` / `OUT_OF_TREE_ID_BASE` go, `_owners_for_id` decodes ids straight from `MANIFEST`, and test fakes inject a fake family instead
- Time the sm120 f16 cell against the backend and put the result in `_MEASURED_BEHIND` (today it is empty, i.e. "unmeasured keeps the old order")
- FALLBACK is a placeholder (smallest admissible tile) until someone picks the configs that cover the plane

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "审计 Frost Python DSL 引擎调用流程"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>


YangXu1990uiuc:
For a diff without #509's commits (GitHub will not take a fork branch as a base, so this PR targets `develop`): https://github.com/YangXu1990uiuc/cudnn-frontend/pull/1 — same six commits, based on `fp8-sm120`. Once #509 merges, this PR's diff narrows to those six on its own and that mirror can be closed.

YangXu1990uiuc:
Superseded by #528, which is the same framework change based on `develop` instead of stacked on #509. The dependency on #509 turned out to be scope, not code: the tile rule belongs in #509, and `knob_candidates`/`knob_order` are deleted by this work rather than modified by it. #528 carries no tuning rules at all -- just the ranking frame and the cleanup.

PR #529, opened 2026-08-09
Type-erase the gemm/cutedsl APIs for torch + JAX, with eager and jax.jit entry points

Anerudhan:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL (Python API)

## Summary

Two commits:

**Commit 1** type-erases the `gemm_amax` API so it accepts **JAX arrays alongside torch tensors** (torch imported only when torch tensors/dtypes are passed, jax only when JAX arrays are passed), and adds a **`jax.jit`-compatible XLA custom-call entry point** (`gemm_amax_jax_sm100`, via jax-tvm-ffi).

**Commit 2** extends this to **all APIs under `python/cudnn/gemm/cutedsl/`** (~35 files). Every API is now torch-lazy (modules import and public symbols resolve without torch installed); JAX support depends on each kernel's data path:

### Per-API JAX support matrix

| API | torch eager | JAX eager | `jax.jit` (XLA custom call) |
|---|---|---|---|
| `gemm_amax` | ✅ | ✅ | ✅ `gemm_amax_jax_sm100` |
| `gemm_swiglu` | ✅ | ✅ (incl. blockscaled MXFP8) | ✅ `gemm_swiglu_jax_sm100` (non-quantized kernel only¹) |
| `gemm_srelu` | ✅ | ✅ | —¹ |
| `gemm_dsrelu` | ✅ | ✅ | —¹ |
| `gemm_proj_rope_mxfp8` | ✅ | ✖ clear error² | — |
| `grouped_gemm` (unfused) | ✅ | ✖ clear error² | — |
| `grouped_gemm_{swiglu, dswiglu, srelu, dsrelu, glu, dglu, glu_hadamard, quant, wgrad}` | ✅ | ✖ clear error² | — |
| `discrete_grouped_gemm_{swiglu, dswiglu}` | ✅ | ✖ clear error² | — |

¹ Kernels whose compiled signature carries optional None-typed parameters (srelu/dsrelu, quantized swiglu) cannot cross the jax-tvm-ffi bridge yet — it has no way to supply `None` (a missing-attr probe hard-crashes). These raise `NotImplementedError` pointing at the eager wrappers; unlocking them is an upstream jax-tvm-ffi feature request.
² Torch-only data paths (device pointer arrays from `.data_ptr()`, torch workspaces, torch stream contexts; proj_rope additionally uses `from_dlpack(x.detach())`). Their public entry points reject non-torch tensors up front with `ValueError("... currently supports torch tensors only; JAX support is not yet implemented for this API")` instead of failing deep inside pointer/workspace machinery. Real JAX support there requires redesigning that machinery (framework-neutral workspace allocation, a pointer-extraction story for JAX buffers) and is future work.

JAX dtype note (dense fusions): fp8 flavors (`float8_e4m3fn/e5m2/e8m0fnu` via ml_dtypes) are fully supported; packed fp4 is not reachable from JAX (no packed-fp4 JAX dtype, and the `uint8` container path fails at kernel construction for torch and JAX alike — xfailed tests document it).

### Eager vs jit trade-off (JAX-supported APIs)

| | Eager (type-erased wrappers/classes) | `gemm_amax_jax_sm100` / `gemm_swiglu_jax_sm100` (XLA custom call) |
|---|---|---|
| Model | imperative, destination-passing: caller owns buffers and stream | functional: XLA owns buffers/stream, returns fresh arrays |
| Stream | CUDA legacy default stream (XLA does not track it) → sync before reading outputs | XLA's compute stream — correctly ordered, no manual sync |
| Under `jax.jit` | not usable (tracers have no buffers) | designed for it; registered `allow_cuda_graph=True` |
| CPU cost/launch (SM100, fp8 512×256×256, GPU kernel ~5.7 µs) | **15 µs** JAX / 4.9 µs torch | ~48 µs standalone (XLA dispatch dominated); **~5.3 µs/kernel e2e amortized** in a jitted graph |
| Best for | eager hot loops, benchmarking, interop | jitted programs, composing with other XLA ops |

Shared infrastructure: `cudnn/tensor_adapter.py` (framework detection via `sys.modules` probing — never imports a framework on another's behalf; shape/stride/device readers; torch-free compute-capability query) and `cudnn/gemm/cutedsl/_jax_ffi.py` (row-major descriptors from tracer avals, env-stream compile + XLA target registry; scalar params like swiglu's `alpha` pass as XLA call attributes via `arg_spec=["args", "attrs.alpha"]`).

## Why

The CuTeDSL data path is already framework-neutral: kernels compile with `--enable-tvm-ffi` against fake cute tensors built from metadata, and real tensors reach the compiled callable via DLPack. Only metadata reading, the dtype vocabulary, stream defaulting, and output allocation were torch-bound. Removing that coupling lets JAX users call the same kernels without importing torch, and the jax.ffi entry points give jitted programs correct XLA stream ordering and buffer management. Where real JAX support isn't feasible yet (grouped pointer-array/workspace machinery), failing fast with a clear error beats failing deep inside `.data_ptr()` calls.

## Related issues

None.

- **torch users: zero behavior change** across all APIs (verified suite-wide, see Testing). Two deliberate strictly-wider acceptances on the dense fusions: (1) unit-dim stride canonicalization means C-contiguous `l=1` torch tensors that previously failed stride checks now pass (the tvm-ffi ABI ignores strides of extent-1 dims); (2) `check_support` error messages on dense fusions spell dtypes as cutlass types — accept/reject behavior unchanged.
- **New public symbols**: `gemm_amax_jax_sm100`, `gemm_swiglu_jax_sm100` (lazy exports; require jax + jax-tvm-ffi, Python ≥ 3.11). Dense-fusion dtype parameters accept torch dtypes, numpy/ml_dtypes dtypes, dtype name strings, or cutlass types.
- **JAX contract** (dense fusions; JAX arrays are row-major, immutable-by-contract): A/B k-major `(M, K, 1)`/`(N, K, 1)`; outputs n-major only; batch `L == 1` only; SF tensors accepted in the torch logical atom view *or* the byte-identical physical C-contiguous `(L, MN', K', 32, 4, 4)` shape (both forms, either framework — safe because the kernels rebuild SF layouts from A/B shapes and use only base pointers).
- Grouped/discrete/proj_rope entry points raise `ValueError("... currently supports torch tensors only ...")` for non-torch tensors — previously they would fail with framework-specific attribute errors deep in the call.
- Packaging: `torch` and `jax` PEP 735 dependency groups (`pip install --group jax`); the `cutedsl` extra is unchanged.
- GPU/CUDA requirements unchanged (SM100+); no C++ or backend changes.

## Testing

On a B200-class SM100 (CC 10.0), Python 3.12, torch 2.13+cu130, jax 0.11 + jax-tvm-ffi 0.1.3, nvidia-cutlass-dsl 4.6. All comparisons are against a pristine baseline worktree at the parent commit, same environment:

```bash
cd test/python
# Dense suite: failure list byte-identical to baseline (64 known env-numerics failures);
# 783 passed (+20 = exactly the new JAX/guard tests), 518 skipped, 2 xfailed:
pytest fe_api/gemm/ -q
# Grouped/discrete suite: identical counts both trees --
# 634 passed, 509 skipped, 2 pre-existing collection errors
# (test_grouped_gemm_{glu,dglu}.py import util modules that do not exist upstream):
pytest fe_api/grouped_gemm/ -q --continue-on-collection-errors
# New JAX tests: 25 passed, 2 xfailed
pytest fe_api/gemm/test_gemm_amax_jax.py fe_api/gemm/test_gemm_swiglu_jax.py \
       fe_api/gemm/test_gemm_srelu_dsrelu_jax.py fe_api/gemm/test_cutedsl_jax_guards.py -q
```

New test coverage: swiglu eager bf16 + blockscaled MXFP8 wrapper paths and the jitted entry point (donation safety, static alpha); srelu/dsrelu **bit-identical cross-framework runs** (JAX and torch wrappers on identical input bytes must agree exactly — they share one compiled kernel); all 13 torch-only public wrappers reject JAX arrays with the clear error (one guard test per API). Import hygiene: all 39 `gemm.cutedsl` lazy exports resolve with torch poisoned out of `sys.modules`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,20 @@ dev = [
     "black==26.3.1",
     "clang-format==21.1.6",
 ]
+# Per-framework groups for the type-erased CuTeDSL APIs (currently
+# gemm.cutedsl.dense.amax): install with `pip install --group torch` /
+# `pip install --group jax`. GPU wheels (torch cuXX, jax[cuda12]/[cuda13])
+# are left to the user.
+torch = [
+    "torch",
+    "torch-c-dlpack-ext",
+]
+# jax-tvm-ffi provides the jax.jit-compatible XLA custom-call entry points
+# (e.g. gemm_amax_jax_sm100) and requires Python >= 3.11.
+jax = [
+    "jax>=0.4.35",
+    "jax-tvm-ffi>=0.1.3",
+]
 
 [build-system]
 requires = ["setuptools>=64", "cmake>=3.18", "ninja==1.11.1.1", "pybind11[global]>=2.13,<3"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,10 +80,10 @@ dev = [
     "black==26.3.1",
     "clang-format==21.1.6",
 ]
-# Per-framework groups for the type-erased CuTeDSL APIs (currently
-# gemm.cutedsl.dense.amax): install with `pip install --group torch` /
-# `pip install --group jax`. GPU wheels (torch cuXX, jax[cuda12]/[cuda13])
-# are left to the user.
+# Per-framework groups for the type-erased CuTeDSL GEMM APIs (JAX is supported
+# by the dense fusions: amax, swiglu, srelu, dsrelu): install with
+# `pip install --group torch` / `pip install --group jax`. GPU wheels
+# (torch cuXX, jax[cuda12]/[cuda13]) are left to the user.
 torch = [
     "torch",
     "torch-c-dlpack-ext",
--- a/python/cudnn/gemm/cutedsl/dense/swiglu/jax_api.py
+++ b/python/cudnn/gemm/cutedsl/dense/swiglu/jax_api.py
@@ -1,0 +1,120 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""JAX-native (XLA custom call) entry point for the GEMM + SwiGLU kernels.
+
+See dense/amax/jax_api.py for the integration pattern: the kernel variant is compiled
+with the TVM-FFI environment stream (runs on XLA's compute stream), all outputs are
+donated pre-initialized operands, and the call composes with jax.jit.
+"""
+
+from typing import Any, Optional, Tuple
+
+import jax
+import jax.numpy as jnp
+
+import cutlass
+
+from cudnn.datatypes import _convert_to_cutlass_data_type
+from cudnn.tensor_adapter import framework_dtype
+from cudnn.gemm.cutedsl._jax_ffi import get_or_register_env_stream_target, make_row_major_desc as _make_desc
+from .api import GemmSwigluSm100
+
+_registered_targets = {}
+
+
+def gemm_swiglu_jax_sm100(
+    a_tensor: Any,
+    b_tensor: Any,
+    alpha: float = 1.0,
+    ab12_dtype: Any = cutlass.Float32,
+    c_dtype: Any = cutlass.Float16,
+    acc_dtype: Any = cutlass.Float32,
+    mma_tiler_mn: Tuple[int, int] = (128, 128),
+    cluster_shape_mn: Optional[Tuple[int, int]] = None,
+    ### Quantize only arguments
+    sfa_tensor: Optional[Any] = None,
+    sfb_tensor: Optional[Any] = None,
+    sf_vec_size: int = 16,
+    vector_f32: bool = False,
+    ab12_stages: int = 4,
+) -> Tuple[Any, Any]:
+    """GEMM + SwiGLU as an XLA custom call; usable eagerly or under jax.jit.
+
+    A (M, K, 1) and B (N, K, 1) are k-major C-contiguous JAX arrays (or tracers).
+    Returns a plain ``(ab12_tensor, c_tensor)`` tuple of fresh n-major JAX arrays.
+    ``alpha`` is a static (trace-time) parameter.
+
+    Supports the non-quantized kernel only: the quantized kernel's compiled signature
+    carries None-typed parameters the XLA FFI bridge cannot supply -- use
+    ``gemm_swiglu_wrapper_sm100`` (eager) for blockscaled MXFP8 inputs from JAX.
+    """
+    ab12_dtype = _convert_to_cutlass_data_type(ab12_dtype)
+    c_dtype = _convert_to_cutlass_data_type(c_dtype)
+    acc_dtype = _convert_to_cutlass_data_type(acc_dtype)
+
+    m, _, l = a_tensor.shape
+    n, _, _ = b_tensor.shape
+    if l != 1:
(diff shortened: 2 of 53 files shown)


Anerudhan:
Pushed phase 2: **`gemm_amax_jax_sm100`** — a `jax.jit`-compatible XLA custom-call entry point via [jax-tvm-ffi], complementing the eager DLPack path.

- Kernel variant compiled with the TVM-FFI **environment stream** (`use_tvm_ffi_env_stream=True`) → runs on XLA's compute stream, correctly ordered with surrounding ops; no manual sync or `block_until_ready` caveats.
- Outputs are XLA-managed: both results are donated pre-initialized operands (`input_output_aliases` + `arg_spec=["args"]`), matching the kernel's destination-passing signature. Zero-init is a valid amax identity (the kernel accumulates `max(|c|) ≥ 0` via signed-int atomic max).
- Works under tracing (validation descriptors built from aval shape/dtype); compiled targets cached per problem/config; `allow_cuda_graph=True`.
- Lazily exported — torch-only users never import jax/jax_tvm_ffi. `jax` extra now includes `jax-tvm-ffi>=0.1.3` (py3.11+).
- Eager-path cleanup: dropped the redundant `tvm_ffi.from_dlpack` pre-wrap (tvm-ffi auto-converts DLPack tensors) — ~15% lower CPU launch overhead for JAX eager calls.

**Measured CPU launch overhead** (SM100, fp8 512×256×256, GPU kernel ~5.7 µs): torch eager `execute` 4.9 µs; JAX eager `execute` 15.0 µs (DLPack export cost); single jitted dispatch 47.6 µs e2e (XLA dispatch dominated); **8 chained calls in one jitted graph: 5.3 µs/kernel e2e** — inside a real jitted program the custom call is effectively GPU-bound.

Torch tests remain zero-diff (206 passed / 50 skipped); JAX suite now 6 passed / 2 xfailed.

python/cudnn/gemm/cutedsl/dense/swiglu/jax_api.py
```diff
@@ -0,0 +1,120 @@
...
+            "gemm_swiglu_jax_sm100 currently supports the non-quantized kernel only; "
+            "use gemm_swiglu_wrapper_sm100 (eager) for blockscaled MXFP8 inputs from JAX"
+        )
+
+    cache_key = (
+        tuple(a_tensor.shape),
+        tuple(b_tensor.shape),
+        _convert_to_cutlass_data_type(a_tensor.dtype),
+        _convert_to_cutlass_data_type(b_tensor.dtype),
+        alpha,
```
hwanseoc, line 74: `alpha` in the cache key forces a redundant recompilation on every call with different alpha

pyproject.toml
```diff
@@ -80,6 +80,20 @@ dev = [
...
+torch = [
+    "torch",
+    "torch-c-dlpack-ext",
+]
+# jax-tvm-ffi provides the jax.jit-compatible XLA custom-call entry points
+# (e.g. gemm_amax_jax_sm100) and requires Python >= 3.11.
+jax = [
+    "jax>=0.4.35",
+    "jax-tvm-ffi>=0.1.3",
+]
```
hwanseoc, line 96: requires revision per coderabbit
use dev, dev-torch, dev-jax

hwanseoc (changes requested):

hwanseoc (approved):

PR #522, opened 2026-08-08
frost(sdpa): accept the cu_seq_len (prefix-sum) length form for THD

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Accept the `cu_seq_len_q` / `cu_seq_len_kv` length form ((B+1,) prefix sums, cuDNN 9.24+) in the FROST SDPA forward engines for THD (ragged) graphs — SM100 (all f16 flavors) and SM120.

Prefix sums are the length currency every major varlen consumer natively holds, and feeding cuDNN's per-batch `SEQ_LEN` form costs them a device conversion on every call today: TransformerEngine launches `cu_seqlens_to_actual_seqlens` each fwd+bwd, and PyTorch's cuDNN SDPA glue runs two `at::diff` ops per call (its public `varlen_attn` API takes `cu_seq_q/k`). The FROST THD lowering already derives both forms host-side from its inherent tolist round-trip (the packed totals are runtime values that size the per-execute compile and grid), so accepting cu directly is free — no kernel changes:

- **graph_analyzer**: `cu_seq_q_t` / `cu_seq_kv_t` captured in facts (previously only a boolean); THD/padded graphs may carry either length form per side. Both-forms-on-one-side stays analyzer-VALID (`invalid` means malformed-for-everyone; the backend accepts it with its own precedence) and is declined by the engine gate instead.
- **engines**: a dedicated cu gate replaces the generic capability row — serving rows (SM100 f16, SM120) take THD cu graphs; dense cu graphs stay declined with a precise reason (the dense hot path is sync-free, so cu there needs an in-kernel `len = cu[b+1] - cu[b]` read mode — follow-up PR); ambiguous both-forms graphs decline explicitly. The binding routes the cu buffer through the same seq-lens execute argument.
[...]

PR #543, opened 2026-08-10
frost(sdpa): bind THD host prep to the launch stream; plan-time-only THD compile keys (#552)

vedaanta:
Three fixes to the FROST THD (varlen) execute hot path, one commit each, plus the AGENTS.md rules that keep them fixed.

## 1. THD host prep raced the kernel on a non-default handle stream

`execute()` launches the kernel on the stream carried by the execute-time handle (`ExecutionContext.stream`). But the THD host prep before that launch — the `[seq_kv | cu_q(B+1) | cu_k(B+1)]` metadata build and its one-shot H2D upload, the O-descriptor and dummy-sink scratch (SM100), the cached seq-q dummy's first-use allocation (SM120) — enqueued on **torch's current stream**. When the two streams differ, the prep and the kernel that consumes it race. Pre-existing since the FROST engines landed (#476).

**Fix:** resolve the launch stream first and run every piece of THD host prep inside `_torch_stream_context` — the same helper the fp8/mxfp8 amax paths already use. Allocations happen inside the context too, so caching-allocator blocks are stream-tagged to the stream that actually uses them. Exposure is limited to direct graph-API users with an explicit handle stream (the PyTorch integration runs on torch's current stream, where the context is a no-op-equivalent) — which is exactly why no test had caught it.

## 2. A dead per-execute fill kernel (SM100 THD)

Every SM100 THD execute zero-filled the per-sequence O-descriptor scratch. The kernel's builder pass copies every qword of each sequence's slot from the base descriptor (then patches address/extent) before the fence and before any consumer read, so stale bytes never survive; the `+16`-qword tail is never read at all. The fill was #476 belt-and-braces and is a wasted kernel launch per execute (AGENTS.md Rule 1).

**Fix:** allocate without the fill. The suites pass with deliberately uninitialized scratch, confirming nothing reads it. (The matching dummy-sinks no-fill is split into #573.)

## 3. The THD compile was keyed on the packed token totals (issue #552)

The per-shape kernel compile was `lru_cache`-keyed on `sq=t_q, skv=t_kv` (SM100/SM120) and `max_sq` (SM120) — values read out of the seq-length tensors at execute. Under continuous batching the packed totals change every step, so the cache never hits and every step pays a fresh multi-second `cute.compile`. This is the compile-side half of the plan Rule 3 (#570) records for the THD host round-trip: make `T` a runtime value instead of a compile-time constant.

**Fix, kernel side:** under THD the fake tensors' token extents are `cute.sym_int` symbols — one per ragged group (Q/O/LSE share `t_q`; K/V share `t_kv`) — so one compiled artifact re-binds any total. The batch stride, which is `t * token_stride` (a runtime value that never steps, batch extent is 1), is rebuilt symbolically from the dynamic extent rather than keyed. SM100's `_host` reads the runtime totals from the dynamic tensor shapes; SM120's `max_sq` becomes a runtime `__call__` argument that sizes the per-sequence grid, and its trace-time shape checks now compare only statically-known modes.

**Fix, adapter side:** with every key component plan-time-derivable, the THD compile moves to `compile()` — plan time, like every dense specialization — and the execute paths' cached calls are guaranteed hits (shared `_thd_compile_kwargs()`; the all-KV-zero clamp's swapped K/V strides mint their own entry on first hit). New regression tests on both arches execute twice with different packed totals and assert **zero new cache misses**, with numerics checked per total.

**Still open in #552** (unchanged here): the D2H `.tolist()` itself — it feeds the metadata upload, the ragged views' extents and the exact launch grid — and with it CUDA-graph capture. That needs the plan-time-max grid + in-kernel dead-tile exit + device `cu_seqlens` read evaluated there. The SM80 engines' `_compile_cached` (#493) has the same total-keyed pattern and is flagged as follow-up cleanup in Rule 4.

## 4. AGENTS.md Hard Rules 4 and 5

- **Rule 4 — compile keys are plan-time-only.** Rule 3 (#570) bans the D2H read; Rule 4 bans the key: a runtime value that arrives legally still must not become a compile key. Runtime extents compile dynamic, runtime launch scalars are call arguments, and derived values (a batch stride computed from a total) count as leaks. Rule 3's THD known-violation entry is updated to record the compile-side half as done.
- **Rule 5 — every torch op on the execute path is ordered on the launch stream** via `_torch_stream_context`, producers and consumers both — never implicitly on torch's current stream.

## Verification (all on SM100, B200-class)

| Suite | Result |
|---|---|
| f16 dense+THD (all flavors), fp8, mxfp8, graph-level THD (incl. #522's cu_seq_len forms), `sdpa_fwd_ex` op tests, and #570's `test_sdpa_execute_is_async.py` (D2H-ban + CUDA-graph capture/replay) | **498 passed / 4 skipped** |
| New `test_dsl_sm{100,120}_thd_compile_key_plan_time_only` | pass |

The SM120 kernel changes are structurally identical to the SM100 ones but need an SM120 machine to execute; the existing SM120 THD suites plus the new cache test cover them on SM120 CI.

## Related

- #552 — this PR closes its recompile leg; the D2H/capture legs stay open there.
- #573 — the split-out dummy-sinks no-fill; touches the same `else`-branch this PR stream-binds (trivial one-hunk rebase for whichever lands second).
- #570 — landed Rule 3 and the async-execute plumbing this PR builds on.

## Summary by CodeRabbit

- **Performance**
  - Improved variable-length (THD) attention execution by reusing compiled kernels across different packed sequence lengths.
  - Reduced unnecessary recompilation during repeated launches.

- **Bug Fixes**
  - Improved runtime validation for dynamic sequence sizes and LSE layouts.
  - Improved stream-aware handling of temporary metadata and launch resources.

- **API Changes**
  - Moved `thd_max_sq` from initialization/configuration to the execution call for SM120 attention kernels.

### DIFF
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -97,11 +97,14 @@ Known violations, all pre-existing and each needing a kernel-side change, so
 none is precedent:
 
 - THD `cu_seqlens` host cumsum (`sdpa/fwd/api_dsl.py`, `_execute_thd` on both
-  SM100 and SM120). `t_q`/`t_kv` reach the host only because `T` is a
-  compile-time constant. Compile on the `b * s_q_max` envelope and pass `T` as a
-  runtime argument, as the f16 kernels already do for head dims;
-  `sdpa_fwd_wrapper_sm80` shows the other half — it requires `max_s_q` from the
-  caller rather than deriving it.
+  SM100 and SM120). The compile-side half is DONE — the THD kernels compile
+  with dynamic token extents (Rule 4), so `T` is no longer a compile-time
+  constant and no compile is keyed on it. `t_q`/`t_kv` still reach the host
+  for the metadata upload, the ragged views' extents, and the launch
+  grid; removing that needs the plan-time-max (`b * s_q_max`) grid with
+  in-kernel dead-tile exit and the device-side metadata read (issue #552).
+  `sdpa_fwd_wrapper_sm80` shows the other half — it requires `max_s_q` from
+  the caller rather than deriving it.
 - Per-tensor FP8 descale readback (`_scalar` in the same file): fold on device,
   passing the pointers, as the backend FP8 sdpa does.
 - The FP8/MXFP8 `seq_len_q` guard in `sdpa/fwd/engines.py`. This one cannot be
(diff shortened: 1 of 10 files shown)


YangXu1990uiuc:
Numbering collision to resolve — #570 is now in `develop` (`4932443f`) and it also adds a hard rule to `python/cudnn/AGENTS.md`:

> **Rule 3 — `execute()` never reads device memory to the host.**

It lands at the same anchor this branch uses (immediately after Rule 2's "acceptance is a promise about the execute path" bullet), so expect a textual conflict there, and after resolving it the file would carry two Rule 3s.

Suggested resolution — this branch's rules shift by one:

| this branch | becomes |
|---|---|
| Rule 3 — compile keys are PLAN-TIME-ONLY | **Rule 4** |
| Rule 4 — execute-path torch ops ordered on the LAUNCH stream | **Rule 5** |

Three notes while you are in there, none of them blocking:

- **The THD round-trip is described twice.** This branch's "remaining THD execute host round-trip" bullet and develop's Rule 3 first known-violation are the same call — `_checked_cu_seq_lens(...).tolist()` in `_execute_thd` (`sdpa/fwd/api_dsl.py:590`, and the SM120 copy). develop's entry carries the fix sketch (compile on the `b * s_q_max` envelope and pass `T` as a runtime argument, the way the f16 kernels already do for head dims; `sdpa_fwd_wrapper_sm80` is the other half, it requires `max_s_q` from the caller). Pointing at it rather than restating keeps one description of the item; the #552 pointer is worth keeping either way.

- **Rule 4 and Rule 3 compose, they do not overlap.** Rule 3 also bans *blocking* — `torch.cuda.synchronize()`, stream/event `synchronize()` — which is the other way an execute path can get the stream question wrong. Ordering prep on the launch stream is the fix; synchronizing to paper over the race is the thing Rule 3 forbids.

- **The new fallback at `_thd_pack` can be ~4 µs cheaper.** The added
  ```python
  if current_stream is None:
      current_stream = cuda.CUstream(torch.cuda.current_stream(q_buf.device).cuda_stream)
  ```
  is on the execute path, and `torch.cuda.current_stream()` measures 4.28 µs against 0.098 µs for `torch._C._cuda_getCurrentRawStream(dev)` — same value, ~43x. #570 put that behind `_raw_stream()` in `api_base.py:47`; since this call site wants a specific device rather than the current one, the direct form keeps that: `cuda.CUstream(_raw_stream(torch)(q_buf.device.index))`. (`self._get_default_stream(None)` is the same helper but resolves against `torch.cuda.current_device()`.)

<sub>note to self: claude::774e8e99-23ad-4a94-be0d-53ed5ee4def9 — "审计前端Python API和Frost引擎代码复杂度" · cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_gpu/fe_pr1</sub>

jhjpark (approved):

vedaanta:
@cudnn-ci-bot run frost

PR #553, opened 2026-08-11
Add cudnn.jax.call on CuTeDSL's native JAX bridge; jit entry points across the GEMM CuTeDSL APIs (stacked on #534)

Anerudhan:
> **Stacked on #534** (→ #530) — this branch contains those PRs' commits plus one new commit (`2ff2239a6`). **Review only the top commit.** Will be rebased as the base PRs merge.

## Before submitting

## Affected area

FE OSS kernels or CuTeDSL (Python API)

## Summary

Replaces the jax-tvm-ffi backend of the `jax.jit` entry points with **CuTeDSL's native JAX integration** (`cutlass.jax.cutlass_call`), wrapped as a new **`cudnn.jax.call`**, and extends jit coverage to **every JAX-reachable GEMM API in `python/cudnn/gemm/cutedsl`**.

### Per-API status

**Dense fusions** — full JAX eager support (from #529), now all with jit entry points:

| API | JAX eager | `jax.jit` entry point |
|---|---|---|
| gemm_amax | ✅ | ✅ `gemm_amax_jax_sm100` (backend swap, contract unchanged) |
| gemm_swiglu | ✅ | ✅ `gemm_swiglu_jax_sm100` — **now incl. blockscaled MXFP8 quantized** (was `NotImplementedError`) |
| gemm_srelu | ✅ | ✅ `gemm_srelu_jax_sm100` — **new** (was impossible over jax-tvm-ffi) |
| gemm_dsrelu | ✅ | ✅ `gemm_dsrelu_jax_sm100` — **new** (dprob = bridge-managed zero-init atomic-add accumulator) |
| gemm_proj_rope_mxfp8 | ✅ `w_out_in=True` (from #530) | ✅ `gemm_proj_rope_mxfp8_jax_sm100` — **new**, both input paths (BF16 and MXFP8), dispatch on `x.dtype` |

**Grouped / discrete-grouped** — JAX eager support in discrete (pointer-array) modes (from #530), now all with jit entry points:

| API | JAX eager | `jax.jit` entry point |
|---|---|---|
| grouped unfused (BF16) | ✅ discrete | ✅ `grouped_gemm_jax_sm100` — **new** |
| grouped glu (BF16) | ✅ discrete | ✅ `grouped_gemm_glu_jax_sm100` — **new** (no bias; `b_major="k"`; `linear_offset` is a compile-time constant) |
| grouped dglu (BF16) | ✅ discrete | ✅ `grouped_gemm_dglu_jax_sm100` — **new** (dprob/dbias are bridge-managed zero-init accumulators, no caller-zeroed buffers) |
| grouped dsrelu (FP8) | ✅ discrete | ✅ `grouped_gemm_dsrelu_jax_sm100` — **new** (full d/SFD/dprob/dbias output set) |
| grouped wgrad (BF16) | ✅ | ✅ `grouped_gemm_wgrad_jax_sm100` — **new** (discrete output pointers; per-expert dgrad buffers are caller-owned external memory, so the entry returns a completion **token** to `block_until_ready`) |
| discrete-grouped swiglu (FP8) | ✅ | ✅ `discrete_grouped_gemm_swiglu_jax_sm100` — **new** (k-major weights; c/d/d_col/SFD/amax all XLA-managed) |
| discrete-grouped dswiglu (FP8) | ✅ | ✅ `discrete_grouped_gemm_dswiglu_jax_sm100` — **new** (relu/srelu epilogues; d_row/d_col/SFD/amax/dprob/dbias) |

**Not reachable from JAX** (eager or jit; unchanged from #530, clear `ValueError`s):

| API | Why |
|---|---|
| grouped swiglu / srelu / quant (legacy + unified) | SFA is an **MMA-permuted tensor argument** in all modes (the kernel consumes the full SF layout rather than rebuilding it from GEMM shapes). `TensorSpec.mode` on this bridge is the future path. |
| grouped dswiglu (contiguous/legacy) | dense-only weight layout (expert-outermost strides) |
| glu_hadamard | MMA-permuted SF tensor arguments |
| block-scaled glu / dglu / wgrad backends | MMA-interleaved scale-factor layouts with no row-major equivalent |
| fp8-D srelu/dsrelu configs; packed-fp4 (uint8) containers | sfd/norm_const outputs unreachable; fp4 has no JAX dtype (and the uint8-as-fp4 container is broken upstream for torch too) |

### `cudnn.jax.call`

`cutlass_call` plus the conveniences the cuDNN kernels need:
- `initialized_outputs`: pre-initialized donated buffers for accumulator outputs (amax/dprob/dbias) **and** for internal workspaces the kernels write (per-expert TMA descriptor tables — XLA inputs are immutable, so grouped workspaces are modeled as donated zero-init outputs that are never surfaced). The bridge drops aliased inputs from the kernel argument list, so the `@cute.jit` adapters keep the kernels' exact destination-passing signatures.
- `TensorSpec` presets: `gemm_operand_spec()` (explicit stride ranks for `(MN, K, 1)` operands — trailing-unit-dim buffers defeat the bridge's leading-dim inference, on inputs and pure results alike, hence all outputs are donated) and `sf_atom_spec()` (presents physical C-contiguous SF buffers in the logical MMA atom view via `TensorSpec.mode`, no transpose materialized).

**Why srelu/dsrelu/quantized-swiglu (and the optional-output grouped kernels) are now possible:** their kernels carry optional None-typed parameters that the jax-tvm-ffi ABI could not supply. With per-kernel `@cute.jit` adapters, those `None`s are compile-time constants inside the adapter — the bridge never sees them. Kernel instance / `max_active_clusters` / scalar knobs travel as constexpr kwargs (participating in the bridge's compile cache; the instances are cached per config on our side).

**Grouped-specific jit mechanics:** per-expert pointer arrays (`b_ptrs`, `sfb_ptrs`, `wgrad_ptrs`) pass as regular packed-uint8 (or x64 int64) input arrays and are recast to `Int64` pointers inside the adapters via `cute.recast_ptr`; rows at/past `padded_offsets[-1]` come back zero-filled (donated zero-init outputs). Two contract caveats, documented per API: `padded_offsets` *values* cannot be host-validated under tracing (shapes/dtypes still are), and the weight buffers behind pointer arrays must stay alive and unmoved across every execution of the traced computation.

## Why

The previous backend needed an extra dependency (jax-tvm-ffi) and hit a hard ABI limit on optional kernel parameters. `cutlass.jax` ships inside `nvidia-cutlass-dsl` (already provided by the `cutedsl` extra), declares CUDA-graph compatibility to XLA, and the constexpr-`None` adapter pattern unlocked every remaining JAX-reachable kernel — jit coverage now matches eager JAX coverage exactly.

## Related issues

Stacked on #534 (→ #530).

- Pre-existing jit entry-point contracts unchanged (same names, same tuple returns); swiglu's quantized configs go from `NotImplementedError` to working; **ten new entry points** (srelu, dsrelu, proj_rope, unfused, glu, dglu, grouped dsrelu, wgrad, discrete swiglu, discrete dswiglu). Eager paths and torch behavior untouched.
- New public module `cudnn.jax` (`call`, `TensorSpec` re-export, spec presets, initializers) — lazily imported, requires jax ≥ 0.5.
- Packaging: the `jax` dependency group drops `jax-tvm-ffi` and moves the floor to `jax>=0.5` (`cutlass.jax` requirement). `gemm/cutedsl/_jax_ffi.py` is removed.
- Dispatch overhead (SM100, fp8 512×256×256): 69 µs e2e single jitted dispatch, **9.0 µs/kernel amortized in a chain of 8** — modestly above the jax-tvm-ffi backend (48 / 5.3 µs), traded for a native dependency-free bridge, CUDA-graph declaration, and the newly unlocked kernels.
- Narrowings vs the eager wrappers, called out in docstrings/docs: glu jit fixes `b_major="k"` and treats `linear_offset` as a compile-time constant; discrete swiglu/dswiglu jit require k-major weights; wgrad jit offers discrete output mode only.

## Testing

On a B200-class SM100 (CC 10.0), Python 3.12, jax 0.11, nvidia-cutlass-dsl 4.6:

```bash
cd test/python
pytest fe_api/gemm/test_gemm_amax_jax.py fe_api/gemm/test_gemm_swiglu_jax.py \
       fe_api/gemm/test_gemm_srelu_dsrelu_jax.py fe_api/gemm/test_gemm_proj_rope_mxfp8_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_jax.py fe_api/grouped_gemm/test_grouped_gemm_glu_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_dglu_jax.py fe_api/grouped_gemm/test_grouped_gemm_dsrelu_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_wgrad_jax.py \
       fe_api/grouped_gemm/test_discrete_grouped_gemm_swiglu_jax.py \
       fe_api/grouped_gemm/test_discrete_grouped_gemm_dswiglu_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_swiglu_jax.py fe_api/grouped_gemm/test_grouped_gemm_srelu_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_quant_jax.py fe_api/grouped_gemm/test_grouped_gemm_dswiglu_jax.py \
       fe_api/grouped_gemm/test_grouped_gemm_glu_hadamard_jax.py -q
# 55 passed, 2 xfailed (pre-existing fp4-container xfails)
pytest fe_api/gemm/test_gemm_amax.py -q   # torch spot-check: 206 passed, 50 skipped (unchanged)
```

- Every new jit entry point has a **jit-vs-eager bit-identical test** (`.view(np.uint8)` comparison on identical input bytes — both paths run the same kernel): once eagerly, then `jax.jit`'d twice (donation safety + compile/registration cache). Atomic-add accumulators (dprob) compare at the existing tight fp32 tolerance.
- amax/swiglu jit tests pass **unchanged** on the new backend; swiglu's quantized-jit test now asserts numerics against the dequantized reference.
- Import hygiene: `import cudnn` succeeds and all jit entry points raise a clear `ImportError` with torch **and** jax absent (`sys.modules` nulled); eager symbols unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added JAX support for compatible GEMM, grouped GEMM, fusion, quantization, and gradient APIs, including eager and `jax.jit` execution.
  * Added native JAX custom-call support with donated outputs, initialization options, and CUDA graph compatibility.
  * Added framework-neutral Torch/JAX tensor handling and support for additional layouts and data types.

* **Documentation**
  * Expanded JAX usage guidance, limitations, synchronization, layouts, buffer lifetimes, and installation options.

* **Tests**
  * Added JAX coverage, Torch parity checks, JIT validation, and unsupported-input error handling.


python/cudnn/gemm/cutedsl/dense/srelu/jax_api.py
```diff
@@ -0,0 +1,281 @@
...
+    cache_key = (
+        tuple(a_tensor.shape),
+        tuple(b_tensor.shape),
+        tuple(sfa_tensor.shape),
+        tuple(sfb_tensor.shape),
+        _convert_to_cutlass_data_type(a_tensor.dtype),
+        _convert_to_cutlass_data_type(b_tensor.dtype),
+        _convert_to_cutlass_data_type(sfa_tensor.dtype),
+        _convert_to_cutlass_data_type(sfb_tensor.dtype),
+        alpha,
```
hwanseoc: another occurance of alpha in cache key

python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py
```diff
@@ -0,0 +1,223 @@
...
+        "k",
+        "n",
+        MoEGroupedGemmBf16Kernel.FIX_PAD_SIZE,
+    ):
+        raise ValueError("Unsupported BF16 grouped GEMM tile, cluster, alignment, or layout configuration")
+
+    cache_key = (
+        m,
+        n,
+        k,
```
hwanseoc: python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py:168-170
python/cudnn/gemm/cutedsl/grouped/glu/jax_api.py:165-167
python/cudnn/gemm/cutedsl/grouped/dglu/jax_api.py:202-204
python/cudnn/gemm/cutedsl/grouped/dsrelu/jax_api.py:325-327

Claude flags these MNK cache keys as pure waste

the outputs of the compilation does not depend on kernel cache, cache value is created from

```
kernel = MoEGroupedGemmBf16Kernel(acc_dtype, mma_tiler_mn, cluster_shape_mn,
                                  expert_cnt, generate_c, use_dynamic_sched, ...)
mac    = f(cluster_shape_mn)
ws     = kernel.get_workspace_bytes()
```

mma_tiler_mn, cluster_shape_mn does not vary with mnk dims

python/cudnn/__init__.py
hwanseoc: cudnn.jax unreachable from a bare import, and the error is a bare `AttributeError: jax` with no hint that import cudnn.jax is the fix.

```
# works
import cudnn.jax
# doesn't work
import cudnn
cudnn.jax.TensorSpec (attributeError jax)
```

hwanseoc (changes requested):

python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py
hwanseoc: correction: the outputs of the compilation does not VARY with kernel cache MNK

hwanseoc (changes requested):
There is a functional bug if a scale-factor tensor is non-contiguous. It does not reproduce today, because every SF tensor in-tree is contiguous.

`_sf_desc_is_physical` (`grouped/dsrelu/api.py:285`) decides between the physical and atom layouts by shape alone, and `_check_sf_shape` (`:297`) validates shape alone. The physical branches (`:1057`, `:1082`, `:1103`) then hardcode row-major strides for the kernel (`:1065`, `:1087`, `:1109`, `:1116`). Both of these tensors get the same verdict:

```
contiguous : stride=(4096,1024,512,16,4,1)  contig=True   -> "physical"
permuted   : stride=(4096, 512,2048,16,4,1) contig=False  -> "physical"
```

JAX cannot reach this, since its arrays are always row-major. A torch caller can, and the kernel then reads at the wrong addresses and returns wrong numbers with no error.

The same pattern appears in `discrete_grouped/swiglu/api.py:202` and `dswiglu/api.py:203`, of which CodeRabbit flagged only dswiglu.

The fix is a `_check_tensor_stride` call in the physical branch.

^ Coderabbit also flags this as a medium-nice-to have bug, but seems like a high functionality bug to me for Torch. Please change

hwanseoc (approved):

python/cudnn/gemm/cutedsl/dense/srelu/jax_api.py
Anerudhan: Fixed in 3b5ab5f9c — `alpha` dropped from both cache keys in this file (srelu and dsrelu). The cached `(kernel, mac)` never depended on it: alpha reaches the kernel only as a constexpr kwarg of the custom call (where it participates in the bridge's own compile cache), and was otherwise only forwarded to `check_support`, which doesn't validate it.

python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py
Anerudhan: Fixed in 3b5ab5f9c — `m`/`n`/`k` dropped from all four cache keys (unfused, glu, dglu, dsrelu). You're right that the cached `(kernel, mac, workspace_bytes)` triple depends only on the constructor/cluster config; `can_implement` (the only shape-dependent step) already runs before the cache lookup on every call, so shape validation is unaffected.

python/cudnn/__init__.py
Anerudhan: Fixed in 3b5ab5f9c — `cudnn/__init__.py` now lazily imports the submodule on attribute access, so `import cudnn; cudnn.jax.TensorSpec` works without a separate `import cudnn.jax` (asserted in `test_gemm_amax_jax_jit_sm100`). jax itself stays unimported until first access, and the fallback `AttributeError` now names the module (`module 'cudnn' has no attribute ...`).

hwanseoc:
There seems to be a lot of redundant cache key vectors,
I think these will come back to us as "compilation overhead bugs"
Could you have Claude audit the cache keys to see if they're compilation cache results actually vary by each key vectors

Anerudhan:
**Re: non-contiguous scale-factor tensors (review [#pullrequestreview-4910023689](https://github.com/NVIDIA/cudnn-frontend/pull/553#pullrequestreview-4910023689))** — fixed in `2ff2239a6`, and slightly wider than asked:

- `_check_sf_shape` now validates **strides in both branches** in `grouped/dsrelu/api.py`, `discrete_grouped/swiglu/api.py`, and `discrete_grouped/dswiglu/api.py` — the physical form must be C-contiguous, and the atom view must be exactly the `(3, 4, 1, 5, 2, 0)` permutation of a C-contiguous physical allocation (a C-contiguous tensor allocated directly *in the atom shape* had the same silent-wrong-results failure mode as your permuted-physical repro).
- **`dense/amax/api.py` had the identical pattern** (its `_check_sf_shape` was shape-only and the kernel likewise consumes only the SF base pointer, per its own docstring) — same both-branch stride validation added there. Expected-stride literals go through `canonicalize_unit_dim_strides` to match the canonical descriptors' unit-dim convention.
- New negative test `test_gemm_amax_rejects_noncontiguous_scale_factors` covers your repro (shape-matching non-contiguous physical) plus the C-contiguous-atom-shape case; both valid forms still accepted. Full torch amax suite (206) and the grouped-dsrelu / discrete swiglu/dswiglu torch+JAX suites pass unchanged.

Dense srelu/dsrelu/swiglu-quantized build their compile-time SF layouts from the descriptor's actual strides rather than hardcoding row-major, so they were left as-is.

Anerudhan:
**Re: cache-key audit ([comment](https://github.com/NVIDIA/cudnn-frontend/pull/553#issuecomment-5259063669))** — audited every `_kernel_cache` in the stack for components the cached compilation artifact doesn't vary with; fixes in `2ff2239a6`:

| Entry point(s) | Verdict |
|---|---|
| dense **amax, swiglu, srelu, dsrelu** | ❌ full problem shapes were in the key, but the cached `(kernel, mac)` is shape-independent → every new M/N/K minted a fresh instance and (since the bridge's FunctionSpec keys on the instance) a redundant kernel compilation. **Fixed**: split into a config-only instance cache + a separate validated-signatures set (shapes stay only where they belong — validation caching). |
| grouped **unfused, glu, dglu, dsrelu** | ❌ M/N/K — removed in the previous round; `can_implement` (the only shape-dependent step) runs before the cache lookup. |
| dense srelu/dsrelu | ❌ `alpha` — removed in the previous round (constexpr kwarg of the call; participates in the bridge's own cache, not ours). |
| **proj_rope** (bf16in + mxfp8in) | ✅ shapes in the key are correct: the cached entry (grid_m, swizzle, t2r_x8, k_scale_words, …) genuinely derives from the shapes. |
| grouped **wgrad**, discrete **swiglu/dswiglu** | ✅ already config-only; every component is a kernel-constructor argument. |

Residual dtype components (e.g. `c_dtype`/`d_dtype` where they aren't constructor args) were kept: distinct output dtypes force distinct bridge compilations regardless (different output avals), so they can't cause redundant kernel compiles and the key stays self-describing. Verified: 26 passed / 2 xfailed across the dense + grouped JAX suites after the restructure.

PR #564, opened 2026-08-12
fix : serialize Tensor_attributes::alignment

tp5uiuc:
## Before submitting

## Affected area

C++ frontend API or graph construction

## Summary

`Tensor_attributes::alignment` is serialized by neither `to_json` nor `from_json`, so `set_alignment(n)` is lost across any JSON round trip.

- `to_json` now emits `"alignment"`, and only when it differs from the default.
- `from_json` reads it with `json::value()`, falling back to the default.
- New `Tensor_attributes::default_alignment` replaces the literal `16` that was duplicated in the member initializer.
- New round-trip and graph-payload cases in `test/cpp/serialize.cpp`.

## Why

Alignment is part of what makes a tensor selectable by a given engine, so a graph rebuilt from a payload can pick a kernel that assumes more alignment than the pointer actually has. The omission also reaches `j["tensors"]` in the structural payload — embedded in the plan blob when `serialize_structure` is true, and what `Graph::deserialize` rebuilds tensors from.

## Related issues

None.

- **New public constant:** `Tensor_attributes::default_alignment` (`static constexpr int64_t`, 16). No existing declarations change.
- **Wire format:** `"alignment"` is a new *optional* key. Older payloads carry no such key and must not throw, hence `json::value()` rather than `j.at()`; newer payloads are ignored by older frontends. `GRAPH_JSON_VERSION` is unchanged, since the key is optional in both directions.
- **`Graph::key()`** hashes the serialized graph. Emitting only when non-default keeps it byte-identical for every graph that never calls `set_alignment`. Graphs that do call it now hash distinctly, which is the intent — they previously collided with their differently-aligned counterparts.
- No performance impact.

## Testing

Release build, x86_64 Ubuntu 24.04, CUDA 13.4, cuDNN 9.20, A100.

- Full build (library, `tests`, `samples`, `legacy_samples`) clean under `-Werror -Wall -Wextra -Wpedantic`.
- `./build/bin/tests "[serialize]"` → `All tests passed (72 assertions in 13 test cases)`.
- Reverting only `utils/serialize.h` and rebuilding → `13 | 11 passed | 2 failed`, with `REQUIRE( alignment_of(j, "image") == 4 )` expanding to `16 == 4`, confirming the new cases cover the defect.
- Python suite not run; no Python or binding code is touched.

## Summary by CodeRabbit

* **New Features**
  * Tensor alignment now has a documented default of 16 bytes.
  * Serialization omits default alignment values while preserving custom alignment settings.
  * Older serialized data remains compatible by restoring the default alignment when unspecified.

* **Bug Fixes**
  * Preserved tensor and convolution graph alignment through serialization and deserialization.

### DIFF
--- a/include/cudnn_frontend/graph_properties.h
+++ b/include/cudnn_frontend/graph_properties.h
@@ -131,7 +131,7 @@ class Tensor_attributes {
 
     std::shared_ptr<Tensor_attributes> ragged_offset;
     int64_t ragged_offset_multiplier = 1;
-    int64_t alignment                = 16;  // Default to 16 bytes
+    int64_t alignment                = default_alignment;
     int64_t vector_count             = 1;   // Default to 1 (no vectorization)
     int64_t vector_dimension         = -1;  // Default to -1 (not set)
 
@@ -412,6 +412,9 @@ class Tensor_attributes {
         reordering_type = value;
         return *this;
     }
+
+    // Pointer alignment assumed by the backend when selecting engines, in bytes.
+    static constexpr int64_t default_alignment = 16;
 
     int64_t
     get_alignment() const {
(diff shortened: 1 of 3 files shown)


tp5uiuc:
@cudnn-ci-bot run backend

hwanseoc (approved):
LGTM

PR #570, opened 2026-08-13
Make the FROST SDPA execute path async where it can be, and stop it re-deriving build-time facts

YangXu1990uiuc:
## Before submitting

## Affected area

Python API or bindings (FROST SDPA engines + the shared `APIBase` helper).

## Summary

`execute()` is an async launch API. The FROST SDPA path had grown host work that
contradicts that in two ways; this fixes the second and documents the first.

A follow-up commit answers review: it **freezes tensor names on a planned graph**, which is what makes the cached name index sound. `_rename_tensor` was the one setter exempt from the freeze; the exemption existed for a typo in sample 24 (`graph.tensor_like(...)` where every sibling line says `bwd_graph`, two cells after that graph was built, against the comment directly above it). The sample is fixed here. A name reaches the lowered graph and is a variant-pack key, so it is identity, like the uid that has been guarded all along.

**A rule for the first.** A device-to-host read makes `execute()` synchronous.
Rules 1 and 2 in `python/cudnn/AGENTS.md` both cite CUDA-graph capture as the
reason for what they ban, but neither names the D2H read itself — which is how
these landed, one of them with a comment calling the syncs "inherent". **Rule 3**
now names it, and lists the reads that remain with what each needs to go. No
kernel changes here, so nothing is removed yet; the list is the plan, not a
claim.

**The second, fixed.** Per-call rework of facts fixed at build:

| | |
|---|---|
| `SdpaBinding` rebuilt its name/uid indices on every execute — three passes over the bound list, five dict constructions | builds the index once |
| the plan wrapper re-read every bound uid per call | reads them at plan construction |
| `_get_default_stream` used `torch.cuda.current_stream()`, whose `Stream` wrapper costs 4.3 µs | `torch._C._cuda_getCurrentRawStream`, 0.1 µs, same value |

## Why

Measured on SM100, dense f16 `(b,h,s,d) = (2,8,256,256)` causal, min over 25 reps
of bursts of 64 calls from a drained queue:

| | before | after |
|---|--:|--:|
| `graph.execute` | 35.5 µs | **24.8 µs** |
| ↳ `resolve_variant_pack` | 5.2 | **0.85** |
| ↳ `api.execute` | 19.8 | **15.2** |
| ↳ compiled-kernel launch (floor) | 6.6 | 6.6 |

`resolve_variant_pack` cost ~1.3 µs per bound operand, so the saving grows with
operand count: 4 for dense f16 fwd, 11 for bwd, 13 for fp8 fwd. Its cost was
three passes over the bound list and five dict constructions — `get_uid` /
`get_name` are plain attribute reads on the IR tensor, not pybind round trips.

Scope, since it is uneven across the 11 registered FROST SDPA specs (9 fwd, 2
bwd):

| | reaches |
|---|---|
| binding index, plan uid list | **all 11** |
| raw default stream | the **6** that share `SdpaFwdDslSm100.execute` — f16 d128 / d192×128 / d256 / d512, FP8, MXFP8 — and only when the caller passes no explicit stream |

The helper runs before the dense/THD/quant branch, so it is not a dense-only
win: four of those six declare `thd=True`, and FP8/MXFP8 enter the same way. The
other five (SM120 fwd ×2, SM80 fwd, SM120 bwd, SM80 bwd) each inline their own
`torch.cuda.current_stream(dev)` and are untouched here — unifying them is a
clean follow-up.

Outside SDPA the same helper is called by NSA Compression / Selection / TopK and
by the block-scaled GLU / DGLU / Wgrad grouped-GEMM APIs.

The layered table above is for attribution only — adjacent rows are not
subtractable, since some layers are handed a raw stream and some resolve it from
None. The end-to-end base/head numbers are the ones to read.

On why the D2H reads matter beyond speed: a blocking D2H during stream capture is
illegal, so a path that does one cannot be CUDA-graph captured — how inference
stacks run decode. And its cost is the queue, not the transfer: one `.tolist()`
measures 11 µs against a drained queue and **2.6 ms** behind 16 queued 4096³
matmuls.

One note for reviewers, since it is the obvious first suggestion: the FP8
`seq_len_q` guard **cannot** move to `check_support()`. `use_padding_mask=True`
requires a `seq_len_q` tensor even when only KV is padded, so no static rule
separates "declares per-batch Q lengths" from "the lengths are actually short".
I tried it; it declines the KV-only-padding population these kernels serve
correctly (`test_fp8_padding`, 4 failures). The read also is not hiding a wrong
answer — it raises. It goes when the FP8 kernels get the epilogue trim.

## Related issues

None.

**One deliberate tightening: `set_name()` on a planned graph now raises**, as
`set_uid()` already did. A no-op rename (same name) stays legal. There is no
valid caller in-tree — the one that existed was a typo in sample 24, fixed here
— but this is a public-surface restriction and should be read as one, not as an
internal cleanup.

Otherwise none: no kernel change, and no change to which engine serves a graph.
`_get_default_stream` returns the same stream by a cheaper accessor, with a
fallback to the documented one if a torch build ever drops the private symbol.

## Testing

New `test/python/sdpa/frost/test_sdpa_execute_is_async.py`, two angles:

- **the cause** — bans the known common D2H and blocking entry points for one execute
  (`item`/`tolist`/`cpu`/`numpy`/`__float__`/`__int__`/`__bool__`/`__index__`,
  `to` with a host destination, `torch.is_nonzero`, `torch.equal`, and the
  `synchronize` family), so a regression names the accessor instead of surfacing
  later as a capture failure.
- **the symptom, where nothing else looks** — capture/replay of `graph.execute()`
  with **no handle**. That is the path that resolves the stream itself;
  `test_sdpa_stream_respect.py` only covers the handle-carrying one.

It is a blacklist and does not pretend to be exhaustive — `.to(some_cpu_tensor)`,
a CPU-target `copy_` and driver-level synchronization all get through it; the
capture test is the backstop for those. Scoped to the dense f16 rows, which are
clean; widen it as the listed violations land.

All on SM100 (Blackwell), `CUDA_VISIBLE_DEVICES` pinned, no
`CUDNN_FRONTEND_ENABLE_FROST_ENGINES` (CI does not set it):

| suite | result |
|---|---|
| `test/python/sdpa/frost` | **579 passed**, 165 skipped |
| `test/python/test_graph_native.py` (owns the freeze contract) | 54 passed |
| `test_dispatch` + `test_variant_pack_normalization` + `test_sdpa_with_caching` + `test_mhas_v2` | 2247 passed, 707 skipped |
| `test/python/gemm` | 5768 passed, 2860 skipped |
| sdpa thd / edge / flexible / caching / chunked | 43 passed, 1 skipped |
| `test/python/fe_api` | 87 failed, 2404 passed — **all inherited** |

`sdpa/frost` is the suite that actually covers this change; `test_mhas_v2` routes
0/3201 graphs to FROST, so its green means "no backend regression", not coverage.

The `fe_api` failures were A/B'd against the PR base on the same box, GPU, venv
and `.so`, swapping only `python/cudnn`: **87 failed on both sides, identical
sets, zero unique to the branch** (see the comment thread for the run).

<sub>note to self: claude::774e8e99-23ad-4a94-be0d-53ed5ee4def9 — "FROST SDPA host overhead audit"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_gpu/fe_pr1</sub>

## Summary by CodeRabbit

* **Performance**
  * Improved SDPA execution by reusing prepared buffer and workspace information, reducing repeated setup work.
  * Accelerated tensor binding and buffer resolution for repeated operations.

* **Bug Fixes**
  * Improved default CUDA stream handling across supported execution environments.
  * Prevented SDPA execution from reading device data back to the host, preserving asynchronous behavior and CUDA graph capture.
  * Prevented tensor renaming after graph execution planning, while preserving valid no-op renames.
  * Corrected tensor creation in the LayerNorm training sample.

* **Documentation**
  * Added guidance for avoiding host-side device-memory reads and supporting CUDA graph capture.

### DIFF
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -67,6 +67,50 @@ unsupported input runnable.
   descriptors) — acceptance is a promise about the execute path, not about
   what the adapter can patch up.
 
+**Rule 3 — `execute()` never reads device memory to the host.**
+
+Rules 1 and 2 both cite CUDA-graph capture as the reason for what they ban, but
+neither names the thing that breaks it most directly: a device-to-host read.
+
+- **No `.item()` / `.tolist()` / `.cpu()` / `.numpy()` / `float(tensor)` /
+  `int(tensor)`**, and no branch or f-string that forces one, on an execute
+  argument or anything derived from one. A D2H read makes `execute()`
+  synchronous — the whole point of an async launch API is gone.
+- **It is a functional gap, not a slow path.** A blocking D2H during stream
+  capture is illegal, so a path that does one **cannot be CUDA-graph captured
+  at all** — which is how every inference stack runs decode.
+- **Its cost is the queue, not the transfer.** Measured on SM100: one
+  `.tolist()` costs 11 µs against a drained queue, 2.6 ms behind 16 queued
+  matmuls. Any figure you measure in a microbenchmark is the floor.
+- **If a device value must shape the launch**, pass its pointer and dereference
+  in-kernel, or compile on an envelope and let the kernel read the real extent
+  from device metadata (the f16 prefill kernels already do this for head dims).
+- **A validation that needs a device read is not a validation.** Decline the
+  declaration in `check_support()` — per Rule 2, the graph says what it will
+  hand you — or assert in-kernel. Reading lengths back to decide whether to
+  raise buys nothing: the Router had to choose an engine before any buffer
+  existed.
+
+Known violations, all pre-existing and each needing a kernel-side change, so
+none is precedent:
+
+- THD `cu_seqlens` host cumsum (`sdpa/fwd/api_dsl.py`, `_execute_thd` on both
+  SM100 and SM120). `t_q`/`t_kv` reach the host only because `T` is a
+  compile-time constant. Compile on the `b * s_q_max` envelope and pass `T` as a
+  runtime argument, as the f16 kernels already do for head dims;
+  `sdpa_fwd_wrapper_sm80` shows the other half — it requires `max_s_q` from the
+  caller rather than deriving it.
+- Per-tensor FP8 descale readback (`_scalar` in the same file): fold on device,
+  passing the pointers, as the backend FP8 sdpa does.
+- The FP8/MXFP8 `seq_len_q` guard in `sdpa/fwd/engines.py`. This one cannot be
+  lifted to `check_support()`: `use_padding_mask=True` requires a `seq_len_q`
+  tensor even when only KV is padded, so no static rule separates "declares
+  per-batch Q lengths" from "the lengths are actually short" — declining the
+  declaration would drop the KV-only-padding population these kernels serve
+  correctly. It goes away when the FP8 kernels get the epilogue trim; until
+  then the read is what keeps a short length from being silently ignored.
+- The ragged cache-key `max()` in `sdpa/{fwd,bwd}/api.py`.
+
 ## Frontend-only kernel package layout
 
 ```
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -72,10 +72,12 @@ unsupported input runnable.
 Rules 1 and 2 both cite CUDA-graph capture as the reason for what they ban, but
 neither names the thing that breaks it most directly: a device-to-host read.
 
-- **No `.item()` / `.tolist()` / `.cpu()` / `.numpy()` / `float(tensor)` /
-  `int(tensor)`**, and no branch or f-string that forces one, on an execute
-  argument or anything derived from one. A D2H read makes `execute()`
-  synchronous — the whole point of an async launch API is gone.
+- **No `.item()` / `.tolist()` / `.cpu()` / `.to("cpu")` / `.numpy()` /
+  `float(tensor)` / `int(tensor)` / `torch.is_nonzero`**, and no branch or
+  f-string that forces one, on an execute argument or anything derived from one.
+  A D2H read makes `execute()` synchronous — the whole point of an async launch
+  API is gone. **Nor may it block**: no `torch.cuda.synchronize()`, no
+  stream/event `synchronize()`. A sync reads nothing but costs the same.
 - **It is a functional gap, not a slow path.** A blocking D2H during stream
   capture is illegal, so a path that does one **cannot be CUDA-graph captured
   at all** — which is how every inference stack runs decode.
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -112,6 +112,15 @@ none is precedent:
   correctly. It goes away when the FP8 kernels get the epilogue trim; until
   then the read is what keeps a short length from being silently ignored.
 - The ragged cache-key `max()` in `sdpa/{fwd,bwd}/api.py`.
+- `cu_seqlens_{q,k}.to(dtype=..., device="cpu")` in the SM80 packed-THD backward
+  (`sdpa/bwd/kernels/bprop_f16_sm80.py`). Reachable only through the standalone
+  wrapper: the registered `sdpa_bwd_sm80` spec declares `thd=False`, so
+  `graph.execute()` does not route here. Still a violation, and it is the one to
+  fix first if that spec ever gains THD.
+
+When auditing this list, grep for the ARGUMENT, not the call shape:
+`device="cpu"` finds `to(dtype=..., device="cpu")`, which `to(device="cpu")`
+misses.
 
 ## Frontend-only kernel package layout
 
--- a/python/cudnn/_pygraph.py
+++ b/python/cudnn/_pygraph.py
@@ -433,12 +433,16 @@ class pygraph:
         becomes ambiguous and leaves the unique-name index."""
         if name == t.name:
             return
-        # NOT freeze-guarded: names are labels (classic allows renaming after
-        # build — the lowered graph already carries the old label, and labels
-        # have no execution semantics).
+        # Freeze-guarded like every other setter. A name is a label, but it is
+        # also a variant-pack key: a compiled plan may be holding the name it
+        # was built with, and the lowered graph keeps the old one, so a rename
+        # after planning leaves two answers to "which tensor is 'q'" -- and
+        # swapping two names would silently rebind buffers. Nothing needs to
+        # rename a planned graph; build another one.
+        self._check_mutable("rename a tensor")
         if self._tensors.get(t.name) is t:
             del self._tensors[t.name]
-        object.__setattr__(t, "name", name)  # label write is exempt from the freeze
+        object.__setattr__(t, "name", name)
         if name in self._tensors or name in self._ambiguous_names:
             self._tensors.pop(name, None)
             self._ambiguous_names.add(name)
--- a/python/cudnn/api_base.py
+++ b/python/cudnn/api_base.py
@@ -41,6 +41,21 @@ def _torch():
     return sys.modules.get("torch")
 
 
+_RAW_STREAM = None
+
+
+def _raw_stream(torch):
+    """``device_index -> raw CUstream int``, resolved once.
+
+    ``torch._C._cuda_getCurrentRawStream`` is private, so fall back to the
+    documented accessor if a torch build ever drops it.
+    """
+    global _RAW_STREAM
+    if _RAW_STREAM is None:
+        _RAW_STREAM = getattr(torch._C, "_cuda_getCurrentRawStream", None) or (lambda _dev: torch.cuda.current_stream().cuda_stream)
+    return _RAW_STREAM
+
+
 def _is_framework_tensor(obj: Any) -> bool:
     """True for framework tensors (torch/jax/numpy/...) as opposed to dtypes or shape/stride tuples."""
     return hasattr(obj, "__dlpack__")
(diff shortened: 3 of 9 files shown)


YangXu1990uiuc:
Test matrix complete. All on SM100 (Blackwell), `CUDA_VISIBLE_DEVICES` pinned, no `CUDNN_FRONTEND_ENABLE_FROST_ENGINES` (CI does not set it, and setting it turns opt-in rows into fake failures).

| suite | result |
|---|---|
| `test/python/sdpa/frost` | **578 passed**, 165 skipped |
| `test/python/test_mhas_v2.py` | 2179 passed, 707 skipped |
| `test/python/test_dispatch.py` | 58 passed |
| `test/python/gemm` | 5768 passed, 2860 skipped |
| sdpa thd / edge / flexible / caching / chunked / cudnn_sdpa_op | 43 passed, 1 skipped |
| `test/python/fe_api` | 87 failed, 2404 passed — **all 87 inherited, see below** |

`sdpa/frost` is the suite that actually covers this change; `test_mhas_v2` routes 0/3201 graphs to FROST, so its green means "no backend regression", not coverage.

**On the 87 `fe_api` failures.** I ran the identical suite against the PR base (`bf40fb5ce`) and against this branch — same box, same GPU, same venv, same `.so`, only `python/cudnn` swapped between runs:

```
base    87 failed, 2404 passed, 1651 skipped, 56573 warnings
branch  87 failed, 2404 passed, 1651 skipped, 56573 warnings

only on the BRANCH (would be mine):   (none)
only on the BASE (fixed by branch):   (none)
```

Failure sets are identical, so none is from this PR. They span `test_rubin_kernel_dispatch.py`, `test_gemm_swiglu.py`, `test_DSA_indexer_top_k.py` and `fe_api/sdpa/test_sdpa_bwd.py` — pre-existing on develop and out of scope here.

**Rule 3 test**, verbose, both parametrizations:

```
test_sdpa_execute_is_async.py::test_execute_reads_no_device_memory_to_the_host[256] PASSED
test_sdpa_execute_is_async.py::test_execute_reads_no_device_memory_to_the_host[512] PASSED
```

**Re-measured on the final commit** (dense f16 (2,8,256,256) causal, min over 25 reps of bursts of 64 from a drained queue): `graph.execute` 25.7 µs against 35.5 on the base — `resolve_variant_pack` 0.84, `api.execute` 15.6, launch floor 6.7. Dense CUDA-graph capture still replays bit-identical to eager.

YangXu1990uiuc:
Thanks — pushed `9a3d1b1` in response. Verified every point against the exact head first; two landed, two did not.

### Name cache staleness — fixed, but by closing the rename rather than teaching the cache

You were right that it can go stale, and I went looking for why `_rename_tensor` was exempt. The exemption was added so that this line in sample 24 would stop raising:

```python
one_bwd = graph.tensor_like(one_cpu).set_name("one")
```

That is a typo. `graph` is the **forward** graph, built two cells earlier; every sibling line in the block says `bwd_graph`, and the comment directly above it reads *"DO NOT reuse tensor handles from the forward graph because tensors are not shared across graphs."* It is also the only line there using `.set_name()` rather than `name=`, and the only post-build rename in the entire notebook set.

So instead of caching less, this freezes the rename — a name reaches the lowered graph (the same commit that added the exemption also made `push_output_attrs` push it) and is a variant-pack key, which makes it identity, like the uid that has been freeze-guarded all along. A no-op rename stays legal, mirroring a no-op re-uid. The sample is fixed here. `test_identity_mutation_frozen_after_planning` now asserts both.

That makes the cache sound because the graph cannot change underneath it, rather than because the cache is careful — which also keeps the hot path free of a rebuild.

One related hole this does **not** close, worth knowing: `tensor()` on a frozen graph is still allowed, which is what let the typo reach a rename at all instead of erroring at the `tensor_like` call. Bigger blast radius; separate change.

`bound_tensors()` returning the internal list is a fair point — noted, though every caller copies it (`list(...)`) today.

### Test scope — fixed

Added `torch.is_nonzero`, `torch.equal`, `Tensor.__index__`, `Tensor.to` with a host-destination check (device-to-device casts still pass), and `synchronize` on `torch.cuda` / `Stream` / `Event`. Rule 3 gained the blocking half: a sync reads nothing and costs the same.

Also added the no-handle capture/replay test you suggested — agreed that was the real gap, since that is the path the raw-stream branch actually serves. Passing.

### Wording — fixed

- `get_uid`/`get_name` are plain attribute reads (`return self.uid` in `graph_types.py`). My commit message and two comments said pybind round trips; corrected to what actually cost the time — three passes over the bound list and five dict constructions.
- Scope corrected in the body. You had the right conclusion by a different route: it is not that SM120 misses out, it is that **only `SdpaFwdDslSm100.execute` calls `_get_default_stream` at all**. SM120 fwd fp8/thd, SM120 bwd and three SM80 kernel files each inline their own `cuda.CUstream(torch.cuda.current_stream(dev).cuda_stream)`. Unifying those is a clean follow-up.
- Agreed the layered rows are not subtractable (raw stream vs None); the body now says so and points at the end-to-end A/B.

### Not reproduced

The SM80 THD backward `.to(device="cpu")` — at `fcf932c2d`, `grep` for `to(device="cpu")`, `to("cpu")`, `.cpu()` and `is_nonzero` across all of `python/cudnn/` returns nothing. If you have a path for it, point me at the file and I will add it to the list.

### Agreed, not claimed

Right that this is not the end state — it is still UID map → Tensor-key dict → id-key dict. The ordered-slots binding is the destination; this PR is the part that needed no ABI change.

YangXu1990uiuc:
Pushed — remote head is now **`87f7e45`** (3 commits on `bf40fb5`).

| | |
|---|---|
| `fcf932c2` | the async/build-time-facts change |
| `9a3d1b1` | rename freeze + widened D2H ban + no-handle capture test + sample typo |
| `87f7e45` | scope corrections from this round |

`87f7e45` contains:

- `bound_tensors()` returns a **tuple** — it is the binding's own record and the index it comes from is cached, so a caller editing that list in place would be editing the cache.
- **The SM80 standalone THD `.to(cpu)` added to the Rule 3 list.** I was wrong to say it did not exist: `bprop_f16_sm80.py:1768` does `cu_seqlens_q.to(dtype=torch.int32, device="cpu")`, and my grep pattern `to(device="cpu")` required `device=` first, so it missed it. The list now says to grep the *argument*, not the call shape. Scoping it accurately cuts both ways though — the registered `sdpa_bwd_sm80` spec declares `thd=False`, so `graph.execute()` cannot route there. Pre-existing, standalone-only, first thing to fix if that spec gains THD.
- **The test no longer claims to ban "every" D2H accessor.** It bans the ones that keep showing up; `.to(some_cpu_tensor)`, a CPU-target `copy_` and driver-level synchronization get through, and the capture test is the backstop. A blacklist that admits its holes is worth more than one that implies it has none.
- The binding comment no longer says the cache is sound by construction — what makes it safe is the graph lifecycle (built after the freeze, and a frozen graph can no longer rename or re-uid). `SdpaBinding` is still an ordinary mutable dataclass; ordered slots would remove the question.

PR body updated for the two things it was overstating:

- **Stream scope is 6 of 11 registered specs**, not "every flavor" and not "SM100 dense". The helper runs before the dense/THD/quant branch, so the six that share `SdpaFwdDslSm100.execute` (f16 d128 / d192×128 / d256 / d512, FP8, MXFP8 — four of them `thd=True`) all get it, and only when the caller passes no explicit stream. The other five inline their own lookup. Outside SDPA: NSA Compression/Selection/TopK and the block-scaled GLU/DGLU/Wgrad grouped-GEMM APIs.
- **API impact no longer says "None".** The rename freeze is a deliberate public tightening and is written up as one.

Tests: `graph_native` 54 passed, async/capture 3 passed, `sdpa/frost` 579 passed / 165 skipped, dispatch+variant-pack+caching+mhas_v2 2247 passed / 707 skipped — all on the previous head. A re-run over `sdpa/frost` + `graph_native` + `gemm/frost` (the other `bound_tensors()` caller, which is what the tuple change could break) is in flight against `87f7e45`; I will post it here.

YangXu1990uiuc:
Re-run against `87f7e45` is green. Clean tree at the tested commit (`git status` empty), so this is the pushed code.

```
sdpa/frost + test_graph_native.py + gemm/frost
6401 passed, 3025 skipped, 0 failed   (15:18)
```

`gemm/frost` is in there deliberately: it is the other `bound_tensors()` caller, so it is what the tuple change could have broken.

Full matrix for this head:

| suite | result |
|---|---|
| `test/python/sdpa/frost` + `test_graph_native.py` + `test/python/gemm/frost` | **6401 passed**, 3025 skipped |
| `test_dispatch` + `test_variant_pack_normalization` + `test_sdpa_with_caching` + `test_mhas_v2` | 2247 passed, 707 skipped |
| `test/python/gemm` | 5768 passed, 2860 skipped |
| sdpa thd / edge / flexible / caching / chunked | 43 passed, 1 skipped |
| `test/python/fe_api` | 87 failed, 2404 passed — all inherited, A/B'd against the base earlier in this thread (identical failure sets, zero branch-only) |

All on SM100 (Blackwell), `CUDA_VISIBLE_DEVICES` pinned, no `CUDNN_FRONTEND_ENABLE_FROST_ENGINES` (CI does not set it, and setting it turns opt-in rows into fake failures).

Ready for a confirmation pass against `87f7e45`.

vedaanta (approved):

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
@cudnn-ci-bot run frost

YangXu1990uiuc:
@cudnn-ci-bot run frost

PR #512, opened 2026-08-07
frost(sdpa): support ragged stats for SM100 frost sdpa forward engine

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Enable ragged THD Stats for the SM100 FROST SDPA forward engine (all four f16/bf16 flavors: d128, d192/d128, d256, d512) — the SM100 counterpart of #508 — and bring the SM100 f16 kernels to SM120 parity on `has_lse`:

- **Ragged Stats, written in the declared layout.** The THD epilogue stores the caller's ragged Stats buffer directly — no extra kernels on the execute path:

  declared layout | strides (dims (B, H, S, 1)) | ragged offsets | kernel store
  -- | -- | -- | --
  token-major [t, h] | stride_h == 1, stride_s == H | cu_q * H | lse[cu_q[b] + row, head]
  head-major [h, t] | stride_s == 1, stride_h == head_stride | cu_q | lse[head, cu_q[b] + row]

  No template parameter is added (matching how SM120 keys this per compile): the layout is a per-shape `compile()` specialization, fully encoded in the LSE fake tensor's static layout — token-major binds its natural packed rank-2 `(T, H)` view, head-major keeps the kernels' native rank-3 `(1, QH, head_stride)` packing, and the epilogue branches on the static rank.

- **`has_lse` specialization (SM120 parity).** The f16 kernels now None-specialize the LSE argument: a stats-less graph compiles the LSE store out, so the workspace dummy-LSE chunks disappear at every level — plain dense inference graphs report `get_workspace_size() == 0`, and THD keeps only its metadata/o-desc chunks. The `lse_tensor` execute contract becomes strict in both directions (required when compiled with a Stats output, rejected when compiled without — previously an unrequested dense `lse_tensor` was silently written). The SM100 f16 spec advertises `lse_optional`; the FP8/MXFP8 kernels still write an LSE unconditionally and keep the engine-carved dummy.

[...]

PR #575, opened 2026-08-13
sdpa: fix SM100 frost zero-KV cluster deadlock + enable zero-length seq_len_kv sweeps

vedaanta:
## Before submitting

## Affected area

Python FE OSS kernels (frost SM100 d192 SDPA) + test infrastructure

## Summary

Two commits (fix first so the branch stays bisect-green):

**1. `frost(sdpa)`: fix an SM100 cluster deadlock on zero-KV varlen tiles.** Zero-length `seq_len_kv` sequences produce EMPTY tiles (kv-loop bounds collapse). On empty tiles the correction→TMA-STG chain that arrives the d192 Q∪O alias gate has **no dependency on the TMA-LDG warp**, so a delayed LDG loses the race and its parity bootstrap credit is consumed by a real arrive — mbarrier parity waits cannot recover once a producer leads by two phases, so LDG parks forever and the whole cluster wedges (MMA starves on `mb_q_full`, softmax on `mb_bmm1_done`). Manifests as a warm-GPU, test-order-dependent hang; root-caused via cuda-gdb shared-memory dumps of a hung CTA (alias gate at phase 1 with LDG never past its first wait; a per-warp round trace splitting exactly at the two starved warps).

Fix: make the alias gate a proper FULL/EMPTY pair — new `mb_qo_slab_free` return edge (LDG arrives after consuming each alias phase; STG waits it before each alias arrive), bounding either side's lead to one phase by construction. d192 is the only `QO_ALIAS` kernel; this mirrors the throttle the d256 flavor already has (`mb_tmastg_go`). Diff is minimal: `_common_sm100.py` + the d192 kernel, +43/−1.

**2. `tests`: re-enable the ~10% zero-length `seq_len_kv` variant** in `test/python/sdpa/random_config.py` (commented out since the test_mhas_v2 refactor; its original form also had a latent `randint` empty-range crash on `s_q > s_kv` configs, fixed by decoupling the lower bound). This is the regression coverage for the fix.

CLC-scheduler hardening found during the same investigation was split out to keep this PR minimal — see #610 (implementation preserved there for a head start).

## Why

Zero-length KV sequences are valid inputs (cuDNN semantics: O := 0, LSE := -inf for dead rows) that frameworks do produce. The sweeps never exercised them, and the frost SM100 d192 kernel deadlocked on them under real scheduling pressure.

## Related issues

Related to #512. Hardening follow-up: #610. The same deadlock signature was hit by the PR #606 Blackwell pipeline (envelope-grid empty tiles); that branch carries the equivalent fix until one of the two merges.

None user-visible. One extra smem mbarrier array + one wait/arrive pair per tile in the aliased (d192) kernel.

## Testing

On SM100 (B200), cuDNN 9.26 nightly + 9.24 wheel, frost engines enabled: the previously-hanging ragged pair passes repeatedly (was ~90% hang warm); fwd+bwd ragged suites, lean_attn, and the frost SM100 fwd L0+L1 suite pass; full `test_mhas_v2.py` L0 completes with only pre-existing 9.26-nightly backend rejections of mixed seq-len forms (reproducible with frost disabled). The combined content also passed `frost_tests:sdpa [Blackwell]` (2830/2830) on GitLab pipeline 62846289 via PR #606's branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing attention workloads with empty or variable-length sequences.
  * Prevented synchronization issues that could occur during delayed data transfers or empty-tile processing.
  * Improved coordination when reusing temporary memory during attention computation.

* **Tests**
  * Expanded randomized testing to include zero-length key/value sequences, improving coverage for edge cases.

### DIFF
--- a/python/cudnn/sdpa/fwd/kernels/_common_sm100.py
+++ b/python/cudnn/sdpa/fwd/kernels/_common_sm100.py
@@ -45,6 +45,10 @@ class Bars(NamedTuple):
     mb_empty_mainloop: object
 
     mb_q_o_alias: object
+    # Return edge of the Q∪O alias gate (see the soundness note in
+    # make_classic_bars): TMA-LDG arrives after consuming each alias-gate
+    # phase; TMA-STG waits it before its next alias arrive.
+    mb_qo_slab_free: object
 
 
 class D256Bars(NamedTuple):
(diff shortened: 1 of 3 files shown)


vedaanta:
@cudnn-ci-bot run frost

jhjpark (approved):

vedaanta:
@cudnn-ci-bot run frost

PR #579, opened 2026-08-13
sdpa fp8 sm107: port the has_lse specialization; add a static SMEM guard

vedaanta:
## Summary

Two follow-ups to #576, one urgent:

1. **Signature-parity fix (Rubin fp8 is broken at develop tip without it).** #574 merged between #577 and #576, giving the SM100 fp8 kernel `compile(has_lse=...)` — which the **shared** adapter call site now passes — while the SM107 sibling (copied pre-#574) kept the old signature: `TypeError` on any Rubin fp8 compile. This ports #574's fp8-kernel hunks onto the sibling verbatim. Exactly the sibling-refresh coordination flagged in #576's description, realized within hours of the merges.

2. **SMEM accounting guard**, responding to [this review thread](https://github.com/NVIDIA/cudnn-frontend/pull/576#discussion_r3773357812): the reviewer's arithmetic is right (BF16/FP16-O at 9 stages ≈ 242 KiB > the standard 227 KiB opt-in) but the conclusion isn't — the configuration is legal on GR100 through the sm107 **oversized-SMEM launch mode** (function attribute 16), which the required internal cutlass-dsl toolchain enables for sm_107a kernels, and it has been board-validated e4m3→bf16 across the full suite. Instead of rejecting working dtypes, an import-time guard now accounts the geometry against the GR100 hardware budget with a clear failure message for future geometry bumps.

## Validation
| part | result |
|---|---|
| SM100 (cc10.0) | 34 passed |
| SM107 (cc10.7 board) | 34 passed — full fp8 e2e through the sibling; `generate_stats` matrix exercises both has_lse populations |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added optional log-sum-exp (LSE) output support for FP8 attention workloads.
  * LSE generation can now be disabled when not needed, while amax outputs remain available.
  * Added support for Rubin SM107 shared-memory sizing and validation.

* **Bug Fixes**
  * Improved output tensor handling for THD-formatted workloads, helping ensure correct results across supported configurations.

PR #580, opened 2026-08-13
sdpa fp8 sm107: fused LDTM row-max + row-sum-in-MMA

vedaanta:
## Summary

Two performance levers baked into the SM107 sibling kernel (Blackwell untouched; the only shared-file change is a **guarded** `is_exclusive` parameter on `tile_dsl.tmem_alloc` that ≤512-column callers never trace):

1. **Fused LDTM row-max** — the unmasked softmax path does the S load + row-max in one `tcgen05.ld.red.f32.max` per 64-col chunk (the MXFP8 kernel's cc10.3 mechanism); masked iters keep the software reduction.
2. **Row-sum in MMA** — an N=16 ones-BMM (P × ones) lands the softmax denominator in 16 Σ TMEM columns beside O, where the correction warp's α-rescale keeps it online-softmax-consistent for free; the softmax warps stop pulling P back through the RF to sum it. Σ sums the fp8-*quantized* P, exactly matching O's numerator. TMEM uses the full 576-col sm107 exclusive grant (dealloc must return the grant — smaller deallocs leak as `cudaErrorTensorMemoryLeak`).

## Perf (same-session A/B vs #579, w2u1g cc10.7, B8 H16 D128 e4m3→bf16 dense)

| shape | #579 baseline | this PR | Δ |
|---|---|---|---|
| 16k nomask | 4.017 ms | **3.471 ms** | −13.6% |
| 16k causal | 2.387 ms | **2.070 ms** | −13.3% |
| 65k nomask | 60.780 ms | **51.842 ms** (5430 TFLOPS eff) | −14.7% |
| 65k causal | 31.777 ms | **27.013 ms** | −15.0% |

## Validation
SM107 board 34/34 (full fp8 e2e: all masks / sink / stats / both fp8 dtypes / all output dtypes vs the fp32 dequant reference — the amax_s/LSE checks cover the Σ path); SM100 box 34/34.

**Stacked on #579** (contains its commit; rebases clean once it merges). The f16x2-MUFU softmax lever is deliberately excluded — it ships separately with its own numerics evidence.

## Summary by CodeRabbit

* **New Features**
  * Added configurable exclusive tensor-memory allocation.
  * Enhanced FP8 scaled dot-product attention on Rubin SM107 hardware with improved softmax-denominator accumulation.
  * Added optional log-sum-exp output, which can be disabled when unnecessary.
  * Added support for empty key/value tiles and improved attention accumulation.
  * Updated compilation options to select LSE output while retaining amax reductions.

**Update (review hardening):** the all-ones B tile and its init loop are now parametrized on `CTA_MMA` (`_ONES_ROWS = 16 // CTA_MMA`). At today's cga2 this is binary-identical (8 rows); it future-proofs a cga1 config, where the fixed 8-row tile would have had the descriptor walk 1 KiB past the tile into neighboring SMEM — functionally masked only because the epilogue consumes Σ column 0 alone, which is not an invariant worth leaning on. Re-gated: board 34/34, SM100 quick pass.

### DIFF
--- a/python/cudnn/frost/tile_dsl/tmem.py
+++ b/python/cudnn/frost/tile_dsl/tmem.py
@@ -8,16 +8,25 @@ import cutlass.cute as cute
 
 
 @cute.jit
-def tmem_alloc(tmem_ptr_i32, num_cols: int, cta_group_kind):
-    # No is_exclusive: it only applies to >512-column allocations on sm_107/sm_109
-    # (cute.arch.tmem.is_tmem_allocation_exclusive), so it is a no-op on the
-    # <=512-column sm100/sm103 allocations here, and it is fenced out of the public
-    # cutlass-dsl tcgen05_alloc wrapper.
-    nvvm.tcgen05_alloc(
-        tmem_ptr_i32,
-        cutlass.Int32(num_cols),
-        group=cta_group_kind,
-    )
+def tmem_alloc(tmem_ptr_i32, num_cols: int, cta_group_kind, is_exclusive: cutlass.Constexpr[bool] = False):
+    # is_exclusive is the sm_107/sm_109 >512-column mechanism (GR100's 576-col
+    # TMEM; cute.arch.tmem.is_tmem_allocation_exclusive) — pass True only
+    # there.  It is fenced out of the PUBLIC cutlass-dsl tcgen05_alloc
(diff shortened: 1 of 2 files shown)


vedaanta:
@cudnn-ci-bot run frost

jhjpark (approved):

PR #584, opened 2026-08-13
frost(sdpa): bottom-right diagonal + sliding window on the SM100 rows

vedaanta:
Closes the `bottom_right_with_swa` kernel gap on the SM100 rows — the next lever from the routing census (~157 graphs blocked solely by "bottom-right causal combined with a sliding window").

**The change**: the band shifts wholesale with its diagonal. Under `BOTTOM_RIGHT` the SWA lower limit is `q + (S_kv − S_q) − W` — the same `causal_diag` offset the upper (causal) limit has always applied. That offset now enters the two SWA terms:
- `compute_kv_loop_bounds`: the SWA left tile bound and the unmasked-lo anchor;
- `apply_mask_chunk`: the per-element SWA limit.

Both fold to the plain top-left math when the diagonal is zero. THD gets the per-sequence diagonal for free through `resolve_seqlen_q`, and the widened band composes — `BR + SWA + window_right` all work together. Gates lifted in the config validation, the adapter backstop, and `bottom_right_with_swa=True` on the f16 and per-tensor FP8 rows.

**Deliberately not flipped**: the MXFP8 row — the single mhas graph BR+SWA admits there trips the executor's pre-existing SF-size mismatch (an SF-layout issue independent of the mask); it flips once that plumbing is fixed.

**Validation** (B200, pip cuDNN 9.23):
- Inline A/B: BR+SWA at skv>sq / equal / d192, BR+SWA+widening, and a TL-SWA regression — all reference-exact on FROST. (BR with `s_q > s_kv` is rejected by cuDNN graph validation itself, so it cannot reach any engine.)
- Suite: `swa_br` joins the dense and THD combo axes — 48 new combos pass; the torch reference's SWA term is diagonal-aware; probe tests updated (the old rejection test superseded by accepts-tests).
- Full run: 2072 pass; the only failures are the two known pre-existing `cu_seq_len` tests (9.23 backend) — the earlier mxfp8 red disappeared with the row gated.
- mhas routing (paged/bwd_ragged excluded for the known env IMA): **949/2274 (41.7%)**, up from 38.6% — d128 +29, d192x128 +5, d256 +17, fp8 +20 graphs.

## Summary by CodeRabbit

* **New Features**
  * Added support for bottom-right causal masking combined with sliding-window attention on supported hardware.
  * Improved attention alignment when query and key/value sequence lengths differ.
  * Expanded support across relevant data types and execution configurations.

* **Bug Fixes**
  * Corrected sliding-window boundaries for bottom-right causal attention, ensuring masking bands are consistently positioned.

* **Tests**
  * Added coverage for dense and variable-length attention shapes and hardware eligibility scenarios.

### DIFF
--- a/python/cudnn/frost/tile_dsl/mask.py
+++ b/python/cudnn/frost/tile_dsl/mask.py
@@ -31,7 +31,13 @@ def apply_mask_chunk(
         return reg_S
 
     neg_inf = cutlass.Float32(mask_value)
-    q_minus_w = q_abs - cutlass.Int32(window_left) if (mask_flags & MASK_SWA) else None
+    # The whole band shifts with the diagonal: under BOTTOM_RIGHT the SWA
+    # lower limit is q + (S_kv - S_q) - W — the same causal_diag offset the
+    # upper (causal) limit uses below. Top-left keeps the plain q - W.
+    q_minus_w = None
+    if mask_flags & MASK_SWA:
+        swa_base = (q_abs + causal_diag) if bottom_right else q_abs
+        q_minus_w = swa_base - cutlass.Int32(window_left)
     # window_right is the compile-time diagonal-band right bound (cuDNN
     # diagonal_band_right_bound): kv columns up to q + window_right (plus the
     # bottom-right diagonal offset) stay unmasked. 0 = plain causal.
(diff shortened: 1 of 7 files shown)


vedaanta:
@cudnn-ci-bot run frost

jhjpark (approved):

PR #581, opened 2026-08-13
frost(sdpa): serve ragged S_kv tails on the f16 rows via synthesized padding

vedaanta:
One-capability flip, no kernel or lowering changes: the f16/bf16 SM100 rows (d128 / d192x128 / d256 / d512) now opt into `skv_tail_via_padding`, the mechanism the FP8 row has always used for a ragged `S_kv` (not a multiple of the 128 KV tile) whose tail no mask covers — the lowering synthesizes full-length per-batch KV lengths (a workspace-carved `(B,)` int32) and the kernel's padded path masks the tail. Mathematically identical output; costs only the padded-path overhead.

This was the top config-level item from the post-#485/#512 routing census (~207 graphs blocked solely by the tail rule, plus combos).

**Routing impact** (test_mhas_v2, paged/bwd_ragged excluded for the known local-env native IMA, pip cuDNN 9.23, B200): **877/2274 graphs on FROST (38.6%)**, up from 581/2273 (25.6%) at the parent commit — ~296 graphs.

**Tests**
- New `test_dsl_sm100_ragged_skv_tail_via_padding` (no-mask ragged tail, per flavor) and `test_dsl_sm100_band_right_uncovered_tail_via_padding` (widened band whose last unmasked column reaches past `S_kv` — previously rejected, now served through the synthesized padding), both reference-exact on GPU.
- The two tests that pinned the old rejection behavior flip to pin the new served behavior.
- Full run: 2010 pass / 0 fail (frost sm100 + analyzer suites + mhas_v2 minus the env-excluded native suites).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for handling ragged key-value sequence tails through synthesized padding.
  * Supports non-128-aligned KV lengths, including cases without a mask and with widened causal masking.

* **Bug Fixes**
  * Improved attention execution for ragged KV sequences whose tails are not covered by masking.

* **Tests**
  * Added end-to-end coverage confirming successful execution and eligibility for these cases.

PR #587, opened 2026-08-13
sdpa fp8: dense head-dim envelope + one engine per arch x dtype family

vedaanta:
Two coupled changes, in two commits: per-tensor FP8 gains the dense head-dim ENVELOPE, and the engine table collapses to **one engine per arch × dtype family** — head dims are now a lowering concern (kernel-flavor pick), not an engine identity.

## Commit 1 — fp8: serve the dense head-dim envelope

Per-tensor FP8 runs head dims below its native tiles through the same TMA zero-padding ENVELOPE the f16/bf16 flavors use: `compile()` takes the actual `(d_qk, d_v)` so the TMA descriptors carry the real extents (OOB loads zero-fill — **exact in FP8** — and O stores clip at `d_v`). Head dims must be multiples of 16 (the 16-byte TMA global-stride rule at 1 byte/elem); the descales are scalars, so the envelope is arch-independent (Blackwell + Rubin). Both d128 siblings (SM100/SM107) **and the d192×128 kernel** change in lockstep, so the whole per-tensor family is uniformly enveloped; MXFP8 stays exact-native (SF plumbing not audited for zero-padding), and THD stays native-tile (the packed THD compile key carries no head-dim entries).

This is the landing zone for the ViT d=72-in-80 contract (e.g. Qwen3-VL vision encoders) without caller-side re-padding to 128.

## Commit 2 — engines: one row per arch × dtype family

The 12-row table (4 f16 + 2 mxfp8 + 3 fp8 + 2 sm120 + sm80) becomes **7 rows**: `sdpa_fwd_prefill_sm100`, `_sm100_mxfp8`, `_sm100_fp8`, `_sm107_fp8`, `_sm120`, `_sm120_fp8`, `_sm80`. Head-dim vocabulary on `Capabilities` shrinks to two fields:

- `d_shapes` — the NATIVE kernel-flavor shapes the lowering picks among (`api_dsl._pick_flavor`, smallest covering flavor: the tightest tile, least padded work).
- `d_pad_multiple` — envelope alignment (8 = f16, 16 = fp8, 1 = SM80 host-side padding, **0 = exact native shapes only** — MXFP8). `d_envelope` and the `thd_d_envelope` notch are gone; `thd_d_shapes` / `split_d_shapes` express the flavors that carry the THD leg / SplitHelpers (the quantized families wire both in d128 only).

Honesty fixes that fall out as row data:
- The f16 and MXFP8 rows shrink to SM 100–106 (no Rubin lowering exists) — a Rubin f16/mxfp8 graph is now ineligible at probe time instead of erroring at build.
- **Fixes `test_dispatch.py::test_every_engine_spec_has_a_manifest_slot`, red on develop since #651**: the sm107 row had no manifest slot, so `FrostSdpaFwdEngines` silently never built it — Rubin per-tensor FP8 was unreachable on the graph route. The consolidated rows get fresh append-only slots (11–14; the per-head-dim slots 0–4/6/9/10 are retired, never reused).
- The adapter accepts an explicit `softmax_precision=FLOAT` on every per-tensor flavor (it is the pipeline each of them runs; `place()` hands it out from the row domain); HALF stays d128 + cc10.7.

What did NOT change: ranked-plan knob sets (sched/tile/split/softmax axes), the sibling-kernel files and their parity model, and the per-arch-line fp8 row split from #651. Cross-flavor plan ranking (running a small-d graph on a larger flavor via `select_plan`) is retired with the per-d rows; if flavor A/B testing is ever wanted it should return as a knob, not as engine identity.

## Rubin CI enablement

The sm107 CI lane's narrow `FROST_TEST_PATHS` exists because the f16/MXFP8 suites *failed* (not skipped) on cc10.7 — the lane's config note asks to widen it "when those modules learn to skip". They now do: `requires_pre_rubin_blackwell` (SM 100–106) gates the f16/MXFP8/stream/async/split suites, while the fp8 e2e suite runs on Rubin and pins `sdpa_fwd_prefill_sm107_fp8`. The lane's path list can then include the whole frost directory plus `test_dispatch.py` (GitLab-side follow-up).

## Tests

- Row/mismatch shape tests rewritten for the family rows (envelope covers, d%16, exact-only MXFP8, THD/split flavor pins, Rubin arch lanes).
- New e2e: d80/d80 and d96/d64 (± causal), the padded ViT shape, and **(160, 96) through the d192×128 flavor's envelope**.
- `test_dispatch.py` green again; heuristics propose/place suites green on the family rows.

### DIFF
--- a/python/cudnn/engines/manifest.py
+++ b/python/cudnn/engines/manifest.py
@@ -174,18 +174,19 @@ MANIFEST: Tuple[EngineFamily, ...] = (
         "cudnn.sdpa.fwd.engine",
         "FrostSdpaFwdEngines",
         # Slots are FIXED FOREVER; append the next free one, never reorder.
+        # RETIRED (one engine per arch x dtype family absorbed the per-head-dim
+        # rows; kernel-flavor choice moved into the lowering — never reuse):
+        #   0 sm100_d128, 1 sm100_d256, 2 sm100_d512, 3 sm100_d128_mxfp8,
+        #   4 sm100_d128_fp8, 6 sm100_d192_d128, 9 sm100_d192_d128_fp8,
+        #   10 sm100_d192_d128_mxfp8
         slots={
-            "sdpa_fwd_prefill_sm100_d128": EngineSlot(0, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d256": EngineSlot(1, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d512": EngineSlot(2, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d128_mxfp8": EngineSlot(3, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d128_fp8": EngineSlot(4, opt_in=True),
             "sdpa_fwd_prefill_sm120": EngineSlot(5, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d192_d128": EngineSlot(6, opt_in=True),
             "sdpa_fwd_prefill_sm120_fp8": EngineSlot(7, opt_in=True),
             "sdpa_fwd_prefill_sm80": EngineSlot(8, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d192_d128_fp8": EngineSlot(9, opt_in=True),
-            "sdpa_fwd_prefill_sm100_d192_d128_mxfp8": EngineSlot(10, opt_in=True),
+            "sdpa_fwd_prefill_sm100": EngineSlot(11, opt_in=True),
+            "sdpa_fwd_prefill_sm100_mxfp8": EngineSlot(12, opt_in=True),
+            "sdpa_fwd_prefill_sm100_fp8": EngineSlot(13, opt_in=True),
+            "sdpa_fwd_prefill_sm107_fp8": EngineSlot(14, opt_in=True),
         },
         analyzer=("cudnn.sdpa.graph_analyzer", "analyze"),
         heuristics=("cudnn.sdpa.fwd.heuristics", "recommend"),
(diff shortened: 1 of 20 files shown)


vedaanta:
Pushed 4b6503c: the d<=128 envelope is no longer gated to cc10.7 — per-tensor FP8 zero-fill is exact and the descales are scalars, so the argument is arch-independent (the restriction was caution, not a constraint). MXFP8 stays exact-d128 (SF plumbing not audited for padding). The d=80 envelope e2e now runs on every fp8-admitted part and passes on SM100 (cc10.0) silicon: 4/4 in test_sdpa_fp8_sm107.py.

vedaanta:
Rebased onto develop tip (8ed6be06a, still two commits). What the rebase absorbed:

- **#602 (Amax_S dropped from the FP8 kernels)**: `_execute_fp8_thd` adapted to the single-`amax_o` ABI — the textual merge was clean but semantically stale, caught by the hardware gate.
- #606/#608 (zero-host-read THD), #575, #581 underneath — no interaction with this PR's paths; the fp8 THD execute keeps its host-metadata form for now (adopting #606's device-built metadata on the fp8 path is natural follow-up material).

Hardware gate on B300 (cc 10.3, backend 9.25, DSL 4.7.0, FE source-built from this branch): **36 passed** across `test_sdpa_fwd_fp8_sm100.py` + `test_sdpa_fp8_sm107.py` — including the d=80 envelope e2e and the irregular-varlen THD e2e on sm103 silicon.

yanqinz2 (approved):

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
Stale-thread cleanup: every CodeRabbit Major/Critical above targets the pre-rebase THD-leg version of this PR (`_execute_fp8_thd`, `amax_s`, the THD LSE-layout binds) — that code was dropped in the envelope re-scope and does not exist on the current head. The three residual nits are resolved on the current diff: the `# noqa: A001` sits on the `def compile(` line in all three fp8 kernels, the docstring example is the 16-aligned `d=72-in-80` contract, and `ruff --select E741,A001` is clean on the touched files.

vedaanta:
@cudnn-ci-bot run frost

PR #589, opened 2026-08-13
Memoize the CuTeDSL hot-path dtype/stride conversions (~20% off per-call host)

YangXu1990uiuc:
## Summary

Every CuTeDSL GEMM API rebuilds a per-shape cache key on **every** call. Building that key re-runs `_convert_to_cutlass_data_type` ~10×/call and `canonicalize_unit_dim_strides` ~4×/call — even on a cache hit, and even though both are pure functions whose results are invariant for a given shape/dtype. This memoizes them.

Both live in the shared `datatypes.py` / `tensor_adapter.py`, so the win applies to **all** cutedsl kernels (amax, swiglu, srelu, dsrelu, grouped, …), not just amax.

## Measurement (SM100, torch wrapper host cost / call, async dispatch)

Same compiled kernel; only the host path is timed. fp8, c=bf16, sf_vec=32.

| shape | before | after |
|---|---|---|
| 512×256×256 | 33.8 µs | 27.5 µs |
| 2048×2048×2048 | 33.3 µs | **26.8 µs** |
| 8192×4096×2048 | 33.0 µs | 27.1 µs |

~**20% off the per-call host overhead**, shape-independent (it's fixed host work). At small kernels this is the difference between host-bound and not; at large kernels it's hidden under the GPU anyway.

Cache behaves as expected after warmup (50 calls, one shape):
```
_convert_to_cutlass_data_type : hits=368 misses=15 currsize=15   # finite dtype key space
canonicalize_unit_dim_strides : hits=205 misses=3  currsize=3
```

## Why it's safe

- Both are **pure functions of hashable args** (dtype objects / shape+stride tuples). Same input → same output for the process lifetime (cutlass availability is fixed at import).
- `_convert_to_cutlass_data_type` raises `ValueError` for unmappable dtypes — `lru_cache` does not cache exceptions, so that path is unchanged.
- `_convert` → `maxsize=None` (finite dtype space, currsize stays ~15). `canonicalize` → `maxsize=4096` (keys are shape-derived; bounded for hygiene).
- No behavior change; correctness verified against the gemm_amax reference (`check_ref_gemm_amax`, amax bit-exact vs max|c|).

## Context

Part of a host-overhead investigation on the imperative cutedsl APIs (TE/JAX MLPerf ask). This is the cheapest, zero-risk slice; the larger levers (per-launch bare-launch of the cute dispatch, C++/AOT execute to remove Python entirely, folding the amax `-inf` fill into the kernel) are separate.

## Test plan

- [x] Correctness vs reference on SM100 (gemm_amax fp8), unchanged.
- [x] `pre-commit run` (black) clean.
- [ ] CI test suites (not run locally).

---
<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) · note to self: claude session — "cutedsl host-overhead: hot-path memoization" · cwd /home/scratch.yanxu_libs/cudnn_frontend</sub>

## Summary by CodeRabbit

* **Performance**
  * Improved performance for repeated data type conversions.
  * Improved performance when repeatedly processing tensor dimension strides.

### DIFF
--- a/python/cudnn/datatypes.py
+++ b/python/cudnn/datatypes.py
@@ -3,6 +3,7 @@
 
 import sys
 import importlib
+import functools
 
 
 def is_windows():
@@ -246,6 +247,9 @@ def _torch_to_cutlass_data_type(data_type, interpret_uint8_as_fp4x2: bool = Fals
     return None
 
 
+# Memoized: pure map from a (hashable) dtype to its cutlass type, called O(10x) per
+# CuTeDSL launch to build the per-shape cache key. Finite dtype key space.
+@functools.lru_cache(maxsize=None)
 def _convert_to_cutlass_data_type(data_type, interpret_uint8_as_fp4x2: bool = False):
     if is_cutlass_available():
         import cutlass
--- a/python/cudnn/tensor_adapter.py
+++ b/python/cudnn/tensor_adapter.py
@@ -13,6 +13,7 @@ already in sys.modules, so probing sys.modules never triggers an import.
 
 from __future__ import annotations
 
+import functools
 import sys
 from dataclasses import dataclass
 from typing import Any, Optional, Tuple
@@ -152,6 +153,9 @@ def get_compute_capability() -> Tuple[int, int]:
     return major, minor
 
 
+# Memoized: pure function of (shape, stride), called per input tensor on every
+# CuTeDSL launch to build the cache key; bounded since keys are shape-derived.
+@functools.lru_cache(maxsize=4096)
 def canonicalize_unit_dim_strides(shape: Tuple[int, ...], stride: Tuple[int, ...]) -> Tuple[int, ...]:
     """Give extent-1 dims the dense "outermost" stride (numel) so that layouts that differ
     only in unit-dim strides -- which the kernels cannot observe -- compare and compile equal."""


Anerudhan:
Moving to 1.29.0
(CI/conflict etc.)

YangXu1990uiuc (commented):

Reviewed `c046cabf0247` against base `c61a5e4eb5c5`.

Memoizing this host work is worthwhile, and the branch's CPU improvement reproduces. On current develop, #627 already memoizes dtype conversion, so I would retain that implementation and land only the bounded stride cache. That also preserves the fallback below. No P0 found; this is a comment-only review because the PR belongs to this account.

**[P2] Preserve the uncached fallback for unhashable dtype spellings** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/c046cabf02475daebd37b0b75f0483091afdca7d/python/cudnn/datatypes.py#L252))

The decorator hashes arguments before the existing converter can handle them. For example, an unhashable object with dtype=np.dtype('float32') previously resolves to cutlass.Float32, but now raises TypeError. Likewise, _convert_to_cutlass_data_type_or_none([('field', 'f4')]) previously returns None for the unsupported structured dtype and now leaks TypeError into capability/routing checks. Preserve an uncached fallback around the memoized implementation. When resolving the current rebase conflict, the simplest solution is to keep the dtype memo already added by #627, which has this fallback, and retain only this PR's stride-cache change.

Evidence: Reproduced on branch point dd4de54dd786 and head c046cabf0247 with the real NumPy/CUTLASS converter. A private current-develop overlay retaining #627's dtype memo preserves both behaviors. Attribution: Introduced by placing lru_cache directly on the dtype-conversion entry point; current develop already contains a compatible memo from #627.

Validation: CPU probes passed 10,201 assertions per version across torch/NumPy/ml_dtypes/CUTLASS dtype spellings, alternating FP4 interpretation flags, invalid hashable dtypes, lazy imports, 5,000 shape/stride cases and the 4,096-entry stride-cache bound. Unhashable cases were checked separately and produced the finding above. Nine focused SM100 GEMM tests passed: FP4/FP8, compile/execute and warm wrapper paths, both output layouts, and rejection of noncontiguous scale factors. Reused a matching built extension after verifying identical binding sources and headers. SM100 baseline/head/head/baseline measurement of the actual torch GEMM-amax wrapper, E5M2 inputs, E8M0 scales and BF16 output: CPU medians for 512x256x256 were 23.32–24.51 -> 18.27–18.42 us, and for 2048x2048x2048 were 23.89–24.87 -> 18.52–18.54 us. GPU captured work stayed approximately 8.19–8.20 and 26.64–26.65 us respectively. Reference checks, poisoned-output replay and changed-A zero-output controls passed in every arm. For current develop, tested a private stride-only overlay that retains #627's dtype memo. After interleaving cached/uncached stride calls within the same process and compiled GEMM, CPU medians improved 20.99 -> 18.02 us (small shape) and 21.11 -> 18.25 us (large shape). Reference/replay controls passed, and GPU work remained about 8.20/26.64 us. Separate-process integration CPU timings were noisy, so these paired measurements support the marginal benefit.

Limitations: The original-branch timing compares against dd4de54d, before #627; it is not the remaining gain over today's develop. CPU timing includes the wrapper's output allocation and amax initialization/enqueue. GPU timing includes the captured amax initialization plus GEMM, excludes compilation and host overhead, and covers only these two shapes. Current develop has a conflict in datatypes.py. The private integration keeps develop's dtype memo unchanged and adds only the bounded stride cache; no change was pushed to this PR. There is no pending fork test workflow to approve or existing test pipeline in the PR thread. Resolve the rebase first, then request test CI with a standalone `@cudnn-ci-bot run oss` comment.

vedaanta (approved):

Reviewed `c046cabf0247` against base `9b9f32781d0c`.

Thank you for your contribution! Independently reviewed both memoization changes and their callers. No P0 or high-risk P1 found; the existing unhashable-dtype P2 remains reproducible. Current develop already has #627's compatible dtype memo, so retain that implementation and the bounded stride-cache improvement when resolving the datatypes.py merge conflict. Approval under risk-v2 does not establish merge readiness.

- **[P2] Still open:** [Preserve the uncached fallback for unhashable dtype spellings](https://github.com/NVIDIA/cudnn-frontend/pull/589#pullrequestreview-5204549623).

Validation: Fresh CPU comparison executed the actual source modules from branch point, PR head and live develop with real NumPy 2.2.6, PyTorch 2.13.0+cu130 and CUTLASS dependencies. Each revision passed 5,174 ordinary checks: supported dtype spellings, alternating packed-FP4 interpretation flags, unsupported hashable inputs, and 5,000 shape/stride cases. The head's stride cache remained bounded at 4,096 entries; module loading preserved lazy torch/CUTLASS imports. Both previously reported unhashable cases still raise TypeError on the PR head only. This independently confirms the linked P2 rather than reporting it as a new discovery. Nine interleaved helper-only timing rounds gave median 7.699 microseconds at the original branch point, 1.073 on the head, and 3.424 on current develop for a fixed batch of five dtype conversions and four stride canonicalizations. This measures only host helper work, not the complete wrapper or GPU execution, and the head/develop difference combines different dtype-cache implementations with stride caching. Reused the linked same-head review's nine focused SM100 GEMM passes and controlled wrapper/GPU measurements; these were not rerun here. That review's paired current-develop stride-only overlay improved CPU medians 20.99 to 18.02 and 21.11 to 18.25 microseconds for its two shapes, with GPU work approximately unchanged at 8.20 and 26.64 microseconds. An isolated merge into live develop reproduced a datatypes.py conflict. Current-head GitHub Actions show successful merge-requirements runs and no pending test workflow. No broad test CI was triggered while rebase is required.

Limitations: Fresh validation targets the changed pure-Python helpers; the installed native extension supplied enum definitions only. No new GPU, JAX execution, or merged-tree validation is claimed. The original branch's performance gain is not the incremental benefit over current develop, which already memoizes dtype conversion. Resolve the conflict before treating the branch as integration-ready.

PR #591, opened 2026-08-13
Drive the grouped GEMM host overhead to the dispatch floor (wrapper 84->~22us; execute() trusts by default)

YangXu1990uiuc:
## Summary

Drives the host overhead of the SM100 grouped GEMM (the ~94 µs "torch wrapper" case) down to the compiled-kernel dispatch floor, so FE's cost **above the tvm-ffi dispatch is ~3 µs** — the floor a caller pays invoking the raw tvm-ffi module directly. Pure Python, no kernel/DSL change. This is the **reference template** for the other imperative CuTeDSL APIs.

1. **`fastsig()`** (in `tensor_adapter`) — a cheap, torch-only, M-invariant operand signature: read the torch attributes once and zero the extent-1 dims' strides; no adapter dispatch, no stride-order sort. **~2.4 µs** for the operand set vs ~15 µs through the canonical adapters. The shared primitive behind the imperative APIs' fast paths.
2. **Wrapper fast path** — key the compiled op by operand **metadata** (`shape/stride/dtype/device`, M masked) and scalar config, **not** object identity, so a training loop passing *fresh* tensors of the same shape each step hits it. M-invariant (extent-1 strides zeroed, M dim masked) → a loop that only grows M reuses one compiled op. On a hit: allocate the outputs and dispatch. Torch-only (JAX falls to the slow path).
3. **`execute()` trusts by default** — the hot path does no per-launch validation, just the pointer-stream record and the dispatch. This is an expert API: matching the compiled shapes is the caller's contract. Pass `_trusted=False` for a fully checked launch; the wrapper does exactly that on the **first call of each configuration**, so wrapper users still get their inputs validated once before the trusted hot path. One `_dispatch()` launch site shared by both paths.

### Why metadata, not object identity

An `id(tensor)` key only hits when the *same* objects are re-passed — a benchmark that reuses its tensors. A real training loop passes fresh activations each step, so an id key misses (→ full ~94 µs path). The metadata key hits for fresh objects of the same shape. Verified: 3 distinct M values → **1** compiled op.

### Design notes (what this converged from)

- The earlier `_trusted` flag + a per-operand descriptor cache are gone. Measurement drove it: a validated-set cache (validate-once-then-trust) still cost ~5 µs/call of key-building on the hot path for a first-call safety net that only helped *direct* `execute()` callers — the wrapper already validates first-call separately. Trusting by default removes that concept entirely and makes direct `execute()` fast (~71 → ~22 µs) too.
- `fastsig` is the *only* shared primitive; the ~25-line fast-path body stays per-API (each API's operands/outputs differ). Not extracted into a callback driver — per the repo's "engine autonomy over dedup" preference, the duplication is cheap and legible and the concept count stays at one.

## Measurement (SM100, bf16, m=n=k=2048, 8 experts; host cost / call, min-over-reps)

| | before | after |
|---|---|---|
| `grouped_gemm_wrapper_sm100()` — same objects | **84 µs** | **~22 µs** |
| `grouped_gemm_wrapper_sm100()` — **fresh objects each step** | 84–94 µs | **~24 µs** |
| direct `op.execute()` (default trust) | 72 µs | **~22 µs** |
| raw tvm-ffi dispatch floor (a raw-module caller pays this too) | | 19.3 µs |

Wrapper is **within ~3 µs of the dispatch floor**, same for fresh inputs as reused. The residual ~19 µs is the tvm-ffi DLTensor marshal + `cuLaunchKernelEx`; going below needs a true bare-launch (separate track).

## Correctness

- Fast-path result **bit-matches** the checked path (`max_abs_diff = 0`).
- Dynamic M collapses to a single compiled op (M-invariant key).
- `execute(_trusted=False)` (the wrapper's first call, and any checked-mode caller) still rejects a mismatching operand (wrong K / transposed layout / wrong dtype) with the original `ValueError`.
- `test_grouped_gemm_bf16.py` (torch) and `test_grouped_gemm_jax.py` (jax eager + jit) pass on SM100.

## Scope / follow-ups

- Lands on `grouped/unfused` (bf16) as the template. The same pattern applies **verbatim** to the other imperative CuTeDSL APIs (glu/dglu/swiglu/amax/…) — a mechanical fast-follow using the shared `fastsig`.
- **JAX**: the fast path is torch-only; JAX takes the (validating) slow path each call — eager JAX is not the hot path, and under `jax.jit` Python is traced out entirely (host cost is XLA's, amortized). The jit primitives already exist for 11 APIs on the shared `cudnn.jax.call` transport.

## Test plan

- [x] SM100 correctness: bit-identical output; fresh objects hit; dynamic-M → 1 op; wrong-shape rejected in checked mode.
- [x] `test_grouped_gemm_bf16.py` (torch) + `test_grouped_gemm_jax.py` (jax) pass.
- [x] `pre-commit` (black, line-length 160) clean.
- [ ] CI suites.

---
<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) · note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 — "支持JAX调用frontend kernel" · cwd /home/scratch.yanxu_libs/cudnn_frontend</sub>

## Summary by CodeRabbit

- **Performance Improvements**
  - Accelerated repeated grouped matrix multiplication calls through cached dispatch information.
  - Added a Torch-optimized path for eligible output-allocating operations.
  - Reduced per-call overhead for trusted execution scenarios.

- **API Updates**
  - Execution methods now support an internal trusted-execution option while retaining validation when needed.
  - Improved handling of pointer-based launches and CUDA stream resolution.

### DIFF
--- a/python/cudnn/gemm/cutedsl/grouped/unfused/_bf16_api.py
+++ b/python/cudnn/gemm/cutedsl/grouped/unfused/_bf16_api.py
@@ -254,14 +254,20 @@ class GroupedGemmBf16API(APIBase):
         import torch
 
         handle = int(current_stream)
-        torch_current = torch.cuda.current_stream(b_ptrs.device)
-        torch_default = torch.cuda.default_stream(b_ptrs.device)
-        if handle == torch_current.cuda_stream:
-            launch_stream = torch_current
-        elif handle == torch_default.cuda_stream:
-            launch_stream = torch_default
-        else:
-            launch_stream = torch.cuda.ExternalStream(handle, device=b_ptrs.device)
+        # Resolving the handle to a torch.cuda.Stream queries current+default stream every
+        # launch; the mapping is stable, so cache it per (device, handle).
+        cache = self.__dict__.setdefault("_launch_stream_cache", {})
+        launch_stream = cache.get(handle)
+        if launch_stream is None:
+            torch_current = torch.cuda.current_stream(b_ptrs.device)
+            torch_default = torch.cuda.default_stream(b_ptrs.device)
+            if handle == torch_current.cuda_stream:
+                launch_stream = torch_current
+            elif handle == torch_default.cuda_stream:
+                launch_stream = torch_default
+            else:
+                launch_stream = torch.cuda.ExternalStream(handle, device=b_ptrs.device)
+            cache[handle] = launch_stream
         b_ptrs.record_stream(launch_stream)
 
     def check_support(self) -> bool:
@@ -533,6 +539,9 @@ class GroupedGemmBf16API(APIBase):
         *,
         dynamic_m: bool = False,
     ) -> TensorDesc:
+        # Only reached on a checked launch (execute(_trusted=False), e.g. the wrapper's
+        # first call of a configuration): build and check the canonical descriptor. The
+        # trusted hot path never gets here.
         desc = self._make_tensor_desc(tensor, name=name, canonical=True)
         if desc.dtype != sample.dtype:
             raise ValueError(f"{name} dtype mismatch: expected {sample.dtype}, got {desc.dtype}")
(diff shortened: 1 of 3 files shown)


hwanseoc:
closing in favor of 627

PR #592, opened 2026-08-14
Avoid zero-initializing JAX grouped GEMM outputs

mgoldfarb-nvidia:
## Before submitting

- [ ] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*`.

## Affected area

Python API or bindings

## Summary

Stop zero-initializing the D, C, and workspace outputs of `grouped_gemm_jax_sm100`. The output specs and CuTeDSL kernel are unchanged. Document that rows at or beyond the final padded offset are unspecified.

## Why

Passing `initialized_outputs` makes JAX zero-fill three output buffers before every grouped GEMM dispatch. The grouped GEMM writes the addressed D/C rows, and its helper writes the workspace. Requiring zero-filled padding adds large device operations to the JAX path without benefiting full-coverage calls, while the equivalent PyTorch API allocates outputs with `empty_strided`.

Same-node A/B results on an isolated ComputeLab B200 using
`benchmark/grouped_gemm_dispatch/bench_dispatch.py` from
[MR 2332](https://gitlab-master.nvidia.com/cudnn/cudnn_frontend/-/merge_requests/2332)
at head `92b46b39099cec904d62aec0de074b8c0c712201`:

| M x N x K | Parent (us) | This PR (us) | Latency reduction |
| --- | ---: | ---: | ---: |
| 2048 x 2048 x 2048 | 37.135 | 30.751 | 17.2% |
| 4096 x 2048 x 2048 | 53.145 | 37.618 | 29.2% |
| 8192 x 4096 x 2048 | 132.452 | 101.144 | 23.6% |
| 16384 x 4096 x 4096 | 412.191 | 367.296 | 10.9% |

Environment: one B200 (SM100), cuDNN 9.24.0, CUDA 13.3, JAX 0.11.0, and NVIDIA CUTLASS DSL 4.6.2. Each value is the median of five complete harness invocations, alternating parent/head run order. Each invocation uses five warmups followed by 30 calls and one final `block_until_ready()`. Median kernel and Torch-execute controls differed by at most 1.5% between variants.

## Related issues

None.

Rows before `padded_offsets[-1]` are unchanged. Rows at or beyond that offset are now explicitly unspecified instead of guaranteed to be zero. No kernel or output-layout changes are included.

## Testing

- `uvx pre-commit run --files python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py` (passed)
- `cd test/python && timeout 120 python -m pytest -q -m L0 fe_api/grouped_gemm/test_grouped_gemm_jax.py::test_grouped_gemm_jax_jit_matches_eager` on ComputeLab B200 (`1 passed`)
- Five alternating parent/head runs of `timeout 900 python benchmark/grouped_gemm_dispatch/bench_dispatch.py --api unfused --json <output>` on the same ComputeLab B200

### DIFF
--- a/python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py
+++ b/python/cudnn/gemm/cutedsl/grouped/unfused/jax_api.py
@@ -28,7 +28,7 @@ from cutlass.cute.nvgpu import OperandMajorMode
 
 from cudnn.datatypes import _convert_to_cutlass_data_type
 from cudnn.tensor_adapter import framework_dtype
-from cudnn.jax import TensorSpec, call, gemm_operand_spec, zeros_init
+from cudnn.jax import TensorSpec, call, gemm_operand_spec
 from ..moe_utils import MoEWeightMode
 from .moe_grouped_gemm import MoEGroupedGemmBf16Kernel
 
@@ -108,9 +108,8 @@ def grouped_gemm_jax_sm100(
     row offsets, ``alpha (experts,)`` float32, ``prob (m, 1, 1)`` float32, and
     ``b_ptrs`` holding per-expert ``(n, k)`` k-major bfloat16 weight base addresses
     (packed little-endian uint8, 8 bytes per pointer — or int64 with x64 mode).
-    Returns ``(d_tensor, c_tensor)`` with ``c_tensor`` None unless ``generate_c``;
-    rows at/past ``padded_offsets[-1]`` come back zero-filled (the outputs are
-    donated zero-initialized buffers).
+    Returns ``(d_tensor, c_tensor)`` with ``c_tensor`` None unless ``generate_c``.
+    Rows at/past ``padded_offsets[-1]`` are unspecified.
     """
     c_dtype = _convert_to_cutlass_data_type(c_dtype)
     d_dtype = _convert_to_cutlass_data_type(d_dtype)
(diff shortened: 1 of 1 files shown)


hwanseoc:
Closing in favor of #631, which removes zero init for more than just unfused

hwanseoc:
Closing in favour of #631, which folds this change in and extends the same audit to the rest of the JAX entry points. The mechanism, the measurement protocol and the framing here are yours — #631 credits this PR as its basis.

For the record, I reproduced this independently on a separate B200 before building on it. `jax.jit` overhead above kernel went from +26.7 / +21.1 / +29.0 / +52.3 µs to +17.1 / +5.4 / +2.9 / +2.6 across the four shapes. The diagnosis was right and the fix is right.

Two things #631 adds that are worth knowing about:

**The tail change is not observable in a clean test.** On fresh device memory the rows past `padded_offsets[-1]` read back as zeros whether or not the fill is present, so a test that does not dirty the allocator first passes either way. #631 adds `test_grouped_gemm_jax_addressed_rows_survive_a_dirty_allocator`, which poisons the allocator before dispatch — with poisoning, every tail element comes back holding the poison value.

**It does not generalize.** I tried removing `initialized_outputs` from all nine remaining sites; six had to be reverted — `dense/amax`, `dense/srelu`, `grouped/dglu`, `grouped/dsrelu`, `discrete_grouped/swiglu`, `discrete_grouped/dswiglu`. The comments there citing the bridge's leading-dim inference rejecting trailing-unit-dim buffers on pure results are still accurate, not stale. All six fail the jit-vs-eager test without the donation, and `grouped/dglu` returns results differing from torch by 1.125 where it matched exactly.

Happy to hand #631 back or split it if you would rather land this one yourself — no attachment to who carries it.

PR #595, opened 2026-08-14
frost(sdpa): add more features to sm120 frost fp8 sdpa_fwd kernel

Aneureka:
## Before submitting

## Affected area

- FE OSS kernels or CuTeDSL

## Summary

Widens the SM120 FP8 SDPA forward engine (`sdpa_fwd_prefill_sm120_fp8`) from its initial envelope (e4m3 in → fp16 out, head dims %32, BSHD dense, causal/SWA/padded/THD) toward parity with the f16 SM120 engine.

**New capabilities** (each with a declared `Capabilities` bit and tests):

- **E5M2 inputs**: Q/K/V may be e4m3 or e5m2.
- **Independent output dtype**: O may be fp16, bf16, e4m3, or e5m2.
- **Attention sinks**: align with SM100 fp8 cell semantics. The sink logit joins the softmax denominator and LSE, P carries no sink column, and Amax_S becomes the sink-extended `alpha / new_sum`. Rows with zero visible keys produce finite LSE (= sink logit) and O = 0.
- **Head-dim envelope**: actual head dims may be any multiple of 16 (TMA 16-byte global-stride rule at 1 byte/elem; f16 is %8 at 2 bytes/elem). Head tiles round up to the 32-element fp8 granule and TMA zero-fills the padded columns, so S and O stay bit-exact: zero K columns don't change QK^T, zero V columns produce zero O columns, and neither Amax is affected. Native tiles stay any multiple of 32 up to 256 with QK^T / P@V sides independent.
- **Per-batch `seq_len_q` trim (dense padded)**: padded Q rows write O := 0, LSE := −inf, and are excluded from the amaxes; zero-length Q batches included.
- **Right-band widening**: `window_size_right > 0` on both top-left and bottom-right diagonals.
- **Single-KV-tile shapes**: `skv_tile=0`: s_kv ≤ 128 graphs are served.
- **THD token-major Stats**: (T, H) LSE layout in addition to head-major, matching the f16 engine.
- **`dense_flex` layouts**: any dense layout with the head dim innermost-contiguous is normalized to the kernel's compact BSHD by the shared adapter (one gather copy in, one scatter copy back for O; zero-copy when already BSHD-physical).

## Why

To enhance the SM120 SDPA forward FP8 engine.

## Related issues

Related to #381.

- **Public API**: none. No graph-API or Python-API surface changes; the engine's declared `Capabilities` widen, so graphs previously declined by this row are now served by it (previously-served graphs select the same specializations).
- **Behavior/correctness**: THD graphs binding raw rank-3 (T, H, D) ragged buffers with h > 1 previously produced corrupted O; they are now correct.
- **Supported platforms**: unchanged — SM120/SM121 for this engine; cuDNN 9.x backend, `nvidia-cutlass-dsl` 4.7.0, Python/torch requirements as before.
- **Internal surface**: `frost/tile_dsl` helper renames (`mma_m16n8k16_f32`, `mma_m16n8k32_f32`, `fp32_to_fp8x2`, `pack_fp8x2_pairs`); all in-tree call sites (f16 fwd/bwd kernels) updated in this PR.
- **Performance**: none. All new features are `const_expr`-gated compile-time specializations.

## Testing

```bash
cd test/python
pytest sdpa/frost/test_sdpa_fwd_fp8_sm120.py sdpa/frost/test_sdpa_fwd_dsl_sm120.py \
       sdpa/frost/test_sm120_tile_rule.py sdpa/frost/test_sdpa_bwd_dsl_sm120.py -m "L0 or L1"
# 235 passed, 8 skipped (pre-existing backend-version gates)
```

## Summary by CodeRabbit

- **New Features**
  - Expanded SM120 FP8 attention support for E4M3 and E5M2 inputs.
  - Added FP16, BF16, E4M3, and E5M2 output options.
  - Added attention sinks, right-side causal windows, sequence trimming, ragged inputs, and flexible head dimensions.
  - Added token-major and head-major attention-statistics layouts, dense-flex support, and zero-length sequence handling.
  - Improved FP8 conversion and support for padded or non-aligned head dimensions.

- **Tests**
  - Expanded coverage for mixed FP8 types, layouts, masking, trimming, quantized outputs, and edge cases.

### DIFF
--- a/python/cudnn/frost/tile_dsl/pointwise.py
+++ b/python/cudnn/frost/tile_dsl/pointwise.py
@@ -3,6 +3,7 @@
 
 
 import inspect
+from typing import Type
 
 import cutlass
 from cutlass.cute.arch.nvvm_wrappers import inline_ptx
@@ -146,6 +147,29 @@ def fp32_to_fp8_pack(values, *, dtype_tag: str):
         read_only_args=list(values),
     )
     return cutlass.Vector.from_elements((u0, u1, u2, u3), cutlass.Int32)
+
+
+@cute.jit
+def fp32_to_fp8x2(lo: cutlass.Float32, hi: cutlass.Float32, *, dtype: cutlass.Constexpr[Type[cutlass.Numeric]] = cutlass.Float8E4M3FN) -> cutlass.Uint16:
+    """Pack two fp32 into fp8 bytes: low byte = fp8(lo), byte 1 = fp8(hi)."""
+    if cutlass.const_expr(dtype != cutlass.Float8E4M3FN and dtype != cutlass.Float8E5M2):
+        raise TypeError(f"Invalid FP8 dtype: {dtype}")
+    cvt_tag = "e4m3x2" if cutlass.const_expr(dtype == cutlass.Float8E4M3FN) else "e5m2x2"
+    return cute.arch.inline_ptx(
+        "{ .reg .f32 fa, fb; mov.b32 fa, {$r0}; mov.b32 fb, {$r1}; " + f"cvt.rn.satfinite.{cvt_tag}.f32 " + "{$w0}, fa, fb; }",
+        write_only_types=[cutlass.Uint16],
+        read_only_args=[hi.bitcast(cutlass.Int32), lo.bitcast(cutlass.Int32)],
+    )
+
+
+@cute.jit
+def pack_fp8x2_pairs(pair0: cutlass.Uint16, pair1: cutlass.Uint16) -> cutlass.Int32:
+    """Two fp8x2 halves into one 32-bit MMA A/B operand (pair0 = low half)."""
+    return cute.arch.inline_ptx(
+        "mov.b32 $0, {$1, $2};",
+        write_only_types=[cutlass.Int32],
+        read_only_args=[pair0, pair1],
+    )
 
 
 def vec_scale_pair(vec, scalar, N):
--- a/python/cudnn/frost/tile_dsl/pointwise.py
+++ b/python/cudnn/frost/tile_dsl/pointwise.py
@@ -125,9 +125,14 @@ def fp32_to_fp16(lo, hi, *, dtype=cutlass.Float16):
     )
 
 
-def fp32_to_fp8_pack(values, *, dtype_tag: str):
+def fp32_to_fp8_pack(values, *, dtype: Type[cutlass.Numeric]):
     assert len(values) == 16, f"fp32_to_fp8_pack: expected 16 input values, got {len(values)}"
-    assert dtype_tag in ("e4m3", "e5m2"), f"fp32_to_fp8_pack: dtype_tag must be 'e4m3' or 'e5m2', got {dtype_tag!r}"
+    if dtype == cutlass.Float8E4M3FN:
+        dtype_tag = "e4m3"
+    elif dtype == cutlass.Float8E5M2:
+        dtype_tag = "e5m2"
+    else:
+        raise TypeError(f"fp32_to_fp8_pack: dtype must be Float8E4M3FN or Float8E5M2, got {dtype}")
 
     u0, u1, u2, u3 = inline_ptx(
         "{ .reg .b16 lo, hi;\n"
--- a/python/cudnn/sdpa/fwd/config_sm120.py
+++ b/python/cudnn/sdpa/fwd/config_sm120.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Optional
 
-from cudnn.frost.tile_dsl.constants import DTYPE_BF16, DTYPE_E4M3, DTYPE_FP16  # noqa: F401  (DTYPE_E4M3 re-exported for the FP8 template)
+from cudnn.frost.tile_dsl.constants import DTYPE_BF16, DTYPE_E4M3, DTYPE_E5M2, DTYPE_FP16  # noqa: F401  (DTYPE_E4M3 re-exported for the FP8 template)
 
 SEQ_Q_TILES = (128, 64)
 SEQ_KV_TILES = (128, 64)
@@ -51,6 +51,10 @@ class TemplateParams:
     """
 
     dtype_qkv: int = DTYPE_FP16
+    # O dtype (frost tile_dsl codes, same table as dtype_qkv). The f16
+    # template stores O at the input dtype and ignores this; the FP8 template
+    # selects its quantizing-store epilogue from it (E4M3/E5M2/BF16/FP16).
+    dtype_o: int = DTYPE_FP16
     # The mask is ONE diagonal band (same model as config_sm100 / the analyzer
     # facts): per-side offsets from the diagonal, None = unbounded on that
     # side. window_right = 0 is plain causal; window_right > 0 widens the
@@ -77,12 +81,14 @@ def validate_params(
     Reachable failures should already have been rejected by the engine
     capabilities or adapter support checks; this validation is a backstop for
     direct template use. ``allowed_dtypes`` defaults to the FP16/BF16 template's
-    set; the FP8 template passes its own. ``allow_right_band=False`` also
-    rejects a widened right band, which the FP8 template does not plumb.
+    set; the FP8 template passes its own. ``allow_right_band=False`` rejects a
+    widened right band for templates that do not plumb one.
     """
 
     if params.dtype_qkv not in allowed_dtypes:
         raise ValueError(f"SM120 SDPA: dtype_qkv must be one of {allowed_dtypes}; got {params.dtype_qkv}")
+    if params.dtype_o not in (DTYPE_E4M3, DTYPE_E5M2, DTYPE_BF16, DTYPE_FP16):
+        raise ValueError(f"SM120 SDPA: dtype_o must be a frost tile_dsl dtype code (E4M3/E5M2/BF16/FP16); got {params.dtype_o}")
     if params.window_right is not None and params.window_right < 0:
         raise ValueError(f"SM120 SDPA: window_right must be None (unbounded) or >= 0 (0 = plain causal); got {params.window_right}")
     if not allow_right_band and params.window_right not in (None, 0):
--- a/python/cudnn/sdpa/fwd/config_sm120.py
+++ b/python/cudnn/sdpa/fwd/config_sm120.py
@@ -74,21 +74,24 @@ class TemplateParams:
 def validate_params(
     params: TemplateParams,
     allowed_dtypes: tuple[int, ...] = (DTYPE_BF16, DTYPE_FP16),
+    allowed_o_dtypes: tuple[int, ...] = (DTYPE_BF16, DTYPE_FP16),
     allow_right_band: bool = True,
 ) -> None:
     """Validate the SM120 template specialization.
 
     Reachable failures should already have been rejected by the engine
     capabilities or adapter support checks; this validation is a backstop for
-    direct template use. ``allowed_dtypes`` defaults to the FP16/BF16 template's
-    set; the FP8 template passes its own. ``allow_right_band=False`` rejects a
+    direct template use. ``allowed_dtypes``/``allowed_o_dtypes`` default to the
+    FP16/BF16 template's sets (which stores O at the input dtype and plumbs no
+    quantizing epilogue); the FP8 template passes its own — FP8 in, any of its
+    four quantizing-store epilogues out. ``allow_right_band=False`` rejects a
     widened right band for templates that do not plumb one.
     """
 
     if params.dtype_qkv not in allowed_dtypes:
         raise ValueError(f"SM120 SDPA: dtype_qkv must be one of {allowed_dtypes}; got {params.dtype_qkv}")
-    if params.dtype_o not in (DTYPE_E4M3, DTYPE_E5M2, DTYPE_BF16, DTYPE_FP16):
-        raise ValueError(f"SM120 SDPA: dtype_o must be a frost tile_dsl dtype code (E4M3/E5M2/BF16/FP16); got {params.dtype_o}")
+    if params.dtype_o not in allowed_o_dtypes:
+        raise ValueError(f"SM120 SDPA: dtype_o must be one of {allowed_o_dtypes}; got {params.dtype_o}")
     if params.window_right is not None and params.window_right < 0:
         raise ValueError(f"SM120 SDPA: window_right must be None (unbounded) or >= 0 (0 = plain causal); got {params.window_right}")
     if not allow_right_band and params.window_right not in (None, 0):
--- a/python/cudnn/sdpa/fwd/engines.py
+++ b/python/cudnn/sdpa/fwd/engines.py
@@ -109,7 +109,7 @@ class Capabilities:
     # global-stride rule at 2 bytes/elem) via TMA zero-padding — the kernel's
     # descriptors carry the ACTUAL extents, so padded contraction columns load
     # as exact zeros (S/softmax unchanged) and O stores past d_v are
-    # OOB-clipped. False = only the native dims above are eligible (FP8/MXFP8,
+    # OOB-clipped. False = only the native dims above are eligible (MXFP8,
     # whose SF plumbing is not audited for zero-padding).
     d_envelope: bool = False
     # Alignment rule for envelope rows: graph head dims must be multiples of
@@ -171,8 +171,9 @@ class Capabilities:
     cu_seq_len: bool = False
     # Dense padded + stats needs the per-batch seq_len_q LSE trim (padded
     # q-rows write LSE=-inf / O=0, cuDNN >= 9.14). Plumbed for the half
-    # kernels via SEQ_Q_LENS_PRESENT; the FP8/MXFP8 kernels lack the epilogue
-    # trim, so their specs keep this False.
+    # kernels and the SM120 fp8 kernel via SEQ_Q_LENS_PRESENT; the SM100
+    # FP8/MXFP8 kernels lack the epilogue trim, so their specs keep this
+    # False.
     padded_stats: bool = False
     # s_q == 1 (decode-shaped) graphs; the SM80 prefill kernels are gated off.
     decode: bool = True
@@ -278,7 +279,7 @@ def mismatch(capabilities: Capabilities, facts: "ga.SdpaGraphFacts", knobs: Opti
         if m > 1 and (facts.d_qk % m != 0 or facts.d_v % m != 0):
             return (
                 f"envelope zero-padding requires D_QK/D_V multiples of {m} (TMA 16-byte "
-                f"global-stride constraint at 2 bytes/elem); graph has D_QK={facts.d_qk}/D_V={facts.d_v}"
+                f"global-stride constraint); graph has D_QK={facts.d_qk}/D_V={facts.d_v}"
             )
     elif facts.d_qk not in capabilities.d_qk or facts.d_v not in capabilities.d_v:
         return f"serves D_QK in {sorted(capabilities.d_qk)}/D_V in {sorted(capabilities.d_v)}; graph has D_QK={facts.d_qk}/D_V={facts.d_v}"
@@ -736,7 +737,7 @@ def lower_dsl_prefill(
     # buffer the FP8/MXFP8 kernels can't honor (dense padded-Q trim is not
     # plumbed there — known gap) is dropped here rather than erroring at
     # execute.
-    seq_q_lens_present = facts.padded and not facts.thd and facts.seq_q_t is not None and not (facts.is_mxfp8 or facts.is_fp8)
+    seq_q_lens_present = facts.padded and not facts.thd and facts.seq_q_t is not None and not ((facts.is_mxfp8 or facts.is_fp8) and api_type == _SM100)
     api = _adapter(api_type)(
         sample_q=ga.tensor_desc_from_ir(facts.q_t, name="q"),
         sample_k=ga.tensor_desc_from_ir(facts.k_t, name="k"),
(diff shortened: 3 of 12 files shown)


Aneureka:
@cudnn-ci-bot run frost

Aneureka:
@cudnn-ci-bot run frost

YangXu1990uiuc:
### Pre-existing bug this PR brushes against: THD all-zero-KV binds O's storage as V

Not introduced here — it is on `develop` from #509, and I put it there. Flagging it because this PR makes O's dtype independent of Q/K/V, which is exactly the assumption the bug rests on, and it touches this code already. Small enough that folding it in is probably cheaper than a separate PR; happy to send one instead if you'd rather.

**What's wrong.** In the THD zero-KV clamp (`api_dsl.py:2521` for SM120, `:1358` for SM100):

```python
t_kv = 1
K = q_buf.as_strided(...)   # Q carries the input dtype -- fine
V = o_buf.as_strided(...)   # O's dtype is NOT the input dtype
```

`V` must carry the **input** element type. Borrowing `O`'s storage was only ever size-correct, never type-correct; it went unnoticed because the FP8 row was FP16-out only, so the mismatch was uniform.

**Reproduced** on an sm120 board at `11c16ffe5` (a THD fp8 graph with every `seq_kv_lens == 0`):

```
ValueError: Mismatched Tensor on argument #2 when calling:
  __call__(q: Tensor([1, 350, 8, 128], uint8), k: Tensor([1, 1, 8, 128], uint8),
           v: Tensor([1, 1, 8, 128], uint8), o: Tensor([1, 350, 8, 128], float16), ...)
```

Argument #2 is `v`: the kernel wants 1 byte/elem, the bound tensor is FP16.

**Fix.** Borrowing `q_buf` instead is one line, but not unconditionally sound — it needs `kh*d_v <= t_q*qh*d_qk`, and with independent `d_qk`/`d_v` (e.g. 64/256) that fails. The file's existing cached-dummy helper is always correct and off the hot path:

```python
V = self._dummy(
    f"thd_v_stub_{d_v}", dev, lambda: torch.zeros(kh * d_v, dtype=q_buf.dtype, device=dev)
).as_strided((1, 1, kh, d_v), (kh * d_v, kh * d_v, d_v, 1), 0)
```

**Why no test caught it.** The f16 suite has `test_dsl_sm120_thd_all_kv_zero_stats`; the fp8 suite has no equivalent. Worse, its helper cannot express the case at all — `_run_thd_fp8`'s `_quant_seqs` does `s.abs().amax()` per sequence, which raises on a zero-token tensor:

```python
# per-tensor FP8: ONE descale for the whole packed batch; a zero-token
# sequence contributes no elements, and amax() on an empty tensor raises
amax = max((s.abs().amax().item() for s in seqs if s.numel()), default=1.0)
```

With that, `_run_thd_fp8(seq_q_lens=[200, 150], seq_kv_lens=[0, 0])` is a one-line regression test.

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "cudnn-FE #509 SM120 FP8 engine + Scale_S scope"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

python/cudnn/sdpa/fwd/config_sm120.py
```diff
@@ -77,12 +81,14 @@ def validate_params(
...
     direct template use. ``allowed_dtypes`` defaults to the FP16/BF16 template's
-    set; the FP8 template passes its own. ``allow_right_band=False`` also
-    rejects a widened right band, which the FP8 template does not plumb.
+    set; the FP8 template passes its own. ``allow_right_band=False`` rejects a
+    widened right band for templates that do not plumb one.
     """
 
     if params.dtype_qkv not in allowed_dtypes:
         raise ValueError(f"SM120 SDPA: dtype_qkv must be one of {allowed_dtypes}; got {params.dtype_qkv}")
+    if params.dtype_o not in (DTYPE_E4M3, DTYPE_E5M2, DTYPE_BF16, DTYPE_FP16):
```
YangXu1990uiuc: `allowed_dtypes` narrows the QKV dtype per template, but there is no `allowed_o_dtypes` counterpart, so this check only tests membership in the union.

The f16 template **ignores** `dtype_o` (it stores O at the input dtype), yet `TemplateParams(dtype_qkv=DTYPE_FP16, dtype_o=DTYPE_BF16)` passes validation. Direct template use — the case this function exists as a backstop for — then gets FP16 bytes in a buffer it believes is BF16, silently. The FP8 template is the only one that reads `dtype_o` at all.

Suggest mirroring the existing narrowing: `allowed_o_dtypes=(...)`, with the f16 template passing its own `dtype_qkv`-only set.

python/cudnn/sdpa/fwd/engines.py
```diff
@@ -736,7 +737,7 @@ def lower_dsl_prefill(
     # buffer the FP8/MXFP8 kernels can't honor (dense padded-Q trim is not
     # plumbed there — known gap) is dropped here rather than erroring at
     # execute.
-    seq_q_lens_present = facts.padded and not facts.thd and facts.seq_q_t is not None and not (facts.is_mxfp8 or facts.is_fp8)
+    seq_q_lens_present = facts.padded and not facts.thd and facts.seq_q_t is not None and not ((facts.is_mxfp8 or facts.is_fp8) and api_type == _SM100)
```
YangXu1990uiuc: Deciding whether the per-batch `seq_len_q` trim is plumbed by testing the adapter **class name** (`api_type == _SM100`) puts a per-kernel property somewhere `Capabilities` can't see it.

Every sibling gap in this file is a capability bit — `padded_stats`, `seq_q_trim`, `bottom_right_padded_seq_q`. Two concrete costs:

- A future arch adapter reusing `lower_dsl_prefill` for a quantized kernel without the trim (an SM107/Rubin fp8 row — `_load_sm100_kernel_module` already anticipates one via `_SM107_FP8_KERNEL_FILE`) silently gets `seq_q_lens_present=True` and writes O/LSE past `seq_len_q[b]`.
- The same condition is re-derived ~140 lines below (`engines.py:878`) to decide whether to pay a `seq_q_buf.min().item()` D2H sync, so the truth lives in two places that must agree.

A `seq_q_trim` bit on the spec rows collapses both.

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
```diff
@@ -145,6 +155,11 @@ def ceil_div(a: int, b: int) -> int:
     return (a + b - 1) // b
 
 
+def round_up_head_tile(d: int) -> int:
```
YangXu1990uiuc, line 163: `round_up_head_tile` is implemented here, and open-coded again in the adapter — `api_dsl.py` computes `granule = _SM120_FP8_HEAD_TILE_GRANULE if self._fp8 else _SM120_HEAD_TILE_GRANULE; d_qp = -(-d_q // granule) * granule` to size SMEM.

The two must agree or `__call__` raises `"runtime Q/K head dimensions must round up to the kernel head_tile_qk"` at trace time. Changing the FP8 granule (or adding a third) means editing two files.

This belongs in `config_sm120.py` next to `FP8_HEAD_TILE_GRANULE`, with both call sites importing it.

python/cudnn/sdpa/fwd/kernels/prefill_fp8_sm120.py
```diff
@@ -1520,24 +1627,27 @@ def compile(  # noqa: A001
...
 
     kernel = SM120FusedMultiHeadAttentionForward(
-        in_dtype=STORAGE_DTYPE,
-        out_dtype=OUT_STORAGE_DTYPE,
+        in_dtype=IN_DTYPE,
+        out_dtype=OUT_DTYPE,
         is_causal=PARAMS.window_right is not None,
         bottom_right=PARAMS.bottom_right,
         window_size_left=PARAMS.window_left,
+        window_size_right=(PARAMS.window_right if PARAMS.window_right else None),
```
YangXu1990uiuc: `window_size_right=(PARAMS.window_right if PARAMS.window_right else None)` collapses `0` and `None` through a **truthiness** test, and `TemplateParams` documents that distinction as load-bearing (`window_right = 0` is plain causal; `None` is unbounded).

The f16 twin passes `window_size_right=PARAMS.window_right` straight through. Both end at `self.window_right = 0` today so behaviour matches, but the divergence is the kind that only surfaces once the two stop being equivalent. Passing it through, as the sibling does, removes the question.

python/cudnn/frost/tile_dsl/pointwise.py
```diff
@@ -148,6 +149,29 @@ def fp32_to_fp8_pack(values, *, dtype_tag: str):
     return cutlass.Vector.from_elements((u0, u1, u2, u3), cutlass.Int32)
 
 
+@cute.jit
+def fp32_to_fp8x2(lo: cutlass.Float32, hi: cutlass.Float32, *, dtype: cutlass.Constexpr[Type[cutlass.Numeric]] = cutlass.Float8E4M3FN) -> cutlass.Uint16:
```
YangXu1990uiuc, line 158: This module now offers two conventions for selecting the same FP8 flavor: `fp32_to_fp8_pack` (line 128) takes `dtype_tag: str` and asserts `in ("e4m3", "e5m2")`, while the new `fp32_to_fp8x2` takes `dtype: Type[cutlass.Numeric]` and derives the tag internally.

A caller holding a cutlass dtype must hand-map it to a string for one and pass it directly to the other; passing `dtype="e4m3"` here fails at trace time with `Invalid FP8 dtype: e4m3` rather than a signature error. `fp32_to_fp16` directly above takes the cutlass type, so that seems like the convention to keep.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -238,46 +482,32 @@ def test_fp8_sm120_single_kv_tile_padded():
...
+    its own toolchain guard (repeated -- the guarded failure mode was
+    nondeterministic)."""
+    scale = 1.0 / math.sqrt(128)
+    for _ in range(3):
+        res = _run(1, 4, 4, 97, 97, scale=scale, sdpa_kwargs=dict(use_causal_mask=True), io_dtype=torch.float8_e5m2)
+        _check(*res, tol_o=1e-1)
 
-def _fp8_graph_offers_sm120(io_dtype, o_dtype, D=128, D_v=None, sink=False):
+
+def _fp8_graph_offers_sm120(io_dtype, o_dtype, D=128, D_v=None, sink=False, layout="bshd"):
```
YangXu1990uiuc, line 543: The new `layout` parameter has no caller — all eight call sites use the default `"bshd"`, so the `dense_flex` capability widening has no offer/decline coverage.

If `layouts=frozenset({"bshd", "dense_flex"})` were dropped from `_sm120_fp8_spec`, or `dense_layout_ok` regressed, `test_fp8_sm120_dense_flex_layout` would fail with an engine-selection error out of `_select_engine` rather than a clean "not offered". Either add `assert _fp8_graph_offers_sm120(..., layout="bhsd")` or drop the parameter.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -50,49 +58,101 @@ def _ref(qd, kd, vd, *, scale, is_causal=False, bottom_right=False, swa_window=N
...
+    out = torch.matmul(probs, v_e)
+    lse = torch.logsumexp(scores, dim=-1)
+    if seq_lens_q is not None:
+        # cuDNN dense padded-Q trim: rows at/past seq_len_q[b] write O := 0 /
+        # LSE := -inf and contribute nothing to the amaxes (softmax of a
+        # fully -inf row is NaN in torch; the trim replaces it).
+        slq = torch.as_tensor(seq_lens_q, device=dev, dtype=torch.long).view(b, 1, 1, 1)
+        row_dead = i >= slq
+        out = torch.where(row_dead, torch.zeros((), device=dev), out)
+        probs = torch.where(row_dead, torch.zeros((), device=dev), probs)
```
YangXu1990uiuc: Dead statement: `out` is computed from `probs` at line 70, and `probs` is never read again before `return`. This zeroing has no effect.

Either delete it, or — if it is standing in for an intent like a P-amax check the reference no longer performs — make it do that, because as written a reader will assume the trim is being validated.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -213,18 +297,178 @@ def test_fp8_sm120_padding(causal):
...
+    assert _fp8_graph_offers_sm120(cudnn.data_type.FP8_E4M3, cudnn.data_type.HALF, D=80)
+    assert not _fp8_graph_offers_sm120(cudnn.data_type.FP8_E4M3, cudnn.data_type.HALF, D=72)
+
+
+@pytest.mark.L0
+@pytest.mark.parametrize("mask", ["none", "causal"])
+@torch_fork_set_rng(seed=0)
+def test_fp8_sm120_sink(mask):
+    """Attention sink: per-Q-head logit in the softmax denominator (virtual
+    column with no V row). O rescales, the LSE extends, and Amax_S follows
```
YangXu1990uiuc: This docstring (and `test_fp8_sm120_e5m2` at :491, `test_fp8_sm120_sink_offered` at :565) describes **Amax_S** semantics, but there is no Amax_S output — the module docstring says so at line 9 ("graphs that declare one are declined and route to the native backend") and the kernel says "There is no Amax_S output — nothing consumes it".

None of the three tests requests or checks an Amax_S. Someone debugging sink numerics will go looking for a path that does not exist, and the wording implies a checked invariant that no assertion covers.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -213,18 +297,178 @@ def test_fp8_sm120_padding(causal):
...
+
+    "none" reaches the shape through the padded route (per-batch KV length);
+    "none_dense" through the dense route that skv_tile=0 opened -- both bound
+    the KV trip count at compile time, i.e. both are collapse-guard
+    populations."""
     scale = 1.0 / math.sqrt(128)
     seq_lens_kv = [s] if mask == "none" else None
     for _ in range(3):
-        res = _run(1, 4, 4, s, s, scale=scale, sdpa_kwargs=_MASKS[mask], seq_lens_kv=seq_lens_kv)
+        res = _run(1, 4, 4, s, s, scale=scale, sdpa_kwargs=_MASKS.get(mask, {}), seq_lens_kv=seq_lens_kv)
```
YangXu1990uiuc: `_MASKS.get(mask, {})` was loosened for the single id `"none_dense"`, but it now makes **any** unknown mask id a silent no-mask run — and `_ref_kwargs({})` also returns `{}`, so the case passes against an unmasked reference.

A future typo in the parametrize list (`"casual"`, `"swa "`) reports green while testing nothing. Adding `"none_dense": {}` to `_MASKS` and going back to `_MASKS[mask]` keeps typos fatal.

YangXu1990uiuc (commented):
Rest of the review — 10 inline below, plus two that have no diff line to hang on.

Context: I wrote the engine this extends (#509), so these are mostly "here is why that limit was there", not objections. The correctness item is in a separate comment above and is **my** pre-existing bug, not this PR's.

---

**`heuristics.py:82` — the tile heuristic and the adapter size SMEM differently, and this PR makes the FP8 row an envelope row.** (`heuristics.py` isn't touched here, so no inline anchor.)

`_sm120_tiles` / `_mode_a` call `smem_bytes(facts.d_qk, facts.d_v, m, n, ...)` on the graph's **actual** head dims, while `check_support` now rounds to the envelope granule (`32 if fp8 else 16`) and calls `_sm120_smem_bytes(d_qp, d_vp, ...)`. So mode A can offer a `tile_n` the adapter then rejects at build — and because the tile arrives as a *knob*, the adapter's "pick the largest that fits" fallback is skipped, so the plan declines instead of degrading. That is exactly what `_sm120_tiles`' docstring promises it avoids ("Configs the kernel cannot fit are not runners-up").

Concretely on the **f16** row, which shares this path: `D_QK=184, D_V=200` gives `128*(184+200)*2 = 98304 <= 101376`, so `tile_n=128` is offered; the adapter rounds to 192/208 and gets `128*(192+208)*2 + 16 = 102416 > 101376`. The FP8 row's own footprint always fits today (max 65536 B), so the new granule-32 gap is latent rather than firing — but the fix is to have the heuristic round with the same granule.

**`prefill_fp8_sm120.py:177` — `SUPPORTED_HEAD_TILES` is imported from the f16 16-granule list.** It binds `(16, 32, 48, ..., 256)` on a class whose head tiles must be multiples of `FP8_HEAD_TILE_GRANULE = 32` (`SUPPORTED_HEAD_TILES_FP8`); `qk_d_frags = head_tile_qk // MMA_TILER[2]` truncates silently, so a tile of 48 would contract 32 of 48 columns. Nothing reads the attribute today, so it is purely misleading for the next reader — and there is no `__init__` assertion backing the docstring's "Must be a multiple of 32 between 32 and 256".

<sub>note to self: claude::304e9e55-1db7-4285-967f-001cb21032f3 — "cudnn-FE #509 SM120 FP8 engine + Scale_S scope"<br>cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_libs/fe_sm120</sub>

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
```diff
@@ -213,18 +297,178 @@ def test_fp8_sm120_padding(causal):
...
+
+
+_O_TOL = {torch.bfloat16: 5e-2, torch.float8_e4m3fn: 1.5e-1, torch.float8_e5m2: 3.5e-1}
+
+
+@pytest.mark.L0
+@pytest.mark.parametrize("o_dtype", [torch.float8_e4m3fn, torch.float8_e5m2, torch.bfloat16])
+@pytest.mark.parametrize("mask", ["none", "causal"])
+@torch_fork_set_rng(seed=0)
+def test_fp8_sm120_o_dtype(o_dtype, mask):
```
YangXu1990uiuc, line 347: FP8 O is covered against causal/none, the head-dim envelope, the seq_q trim and THD — but never crossed with the two other new axes it interacts with:

- **`layout="bhsd"` / `"padded_s"`**, where `_to_bshd_writable` allocates an FP8 scratch and `o_view.copy_(o_scratch)` does a strided FP8 copy-back (a torch path with historically uneven FP8 support). `test_fp8_sm120_dense_flex_layout` is FP16-out only.
- **sinks**, where `row_sum_inv = alpha / new_sum` feeds the direct store and a sink-scaled value could saturate `cvt.rn.satfinite`. `test_fp8_sm120_sink` is FP16-out only.

Either combination could regress green today.

YangXu1990uiuc (approved):

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
Aneureka: Fixed.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
Aneureka: Fixed

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
Aneureka: Added tests.

test/python/sdpa/frost/test_sdpa_fwd_fp8_sm120.py
Aneureka: Added the assertion.

python/cudnn/sdpa/fwd/engines.py
Aneureka: Added `dense_seq_q_trim` to reflect the current SM100 SDPA FP8/MXFP8 kernel limitation to handle dense seq_q_trim.

python/cudnn/sdpa/fwd/config_sm120.py
Aneureka: Addressed.

Aneureka:
@cudnn-ci-bot run frost

Aneureka:
@cudnn-ci-bot run frost

python/cudnn/frost/tile_dsl/pointwise.py
Aneureka: Addressed.

python/cudnn/sdpa/fwd/heuristics.py
```diff
@@ -77,9 +77,15 @@ def _sm120_tiles(caps: Capabilities, facts) -> Tuple[int, int]:
...
+    # SMEM terms size differently -- see config_sm120.smem_bytes. The kernel
+    # stages ENVELOPE head tiles (actual dims round up to the granule), so
+    # the fit check must round the same way or a tile offered here declines
+    # at build -- and a knob-carried tile skips the adapter's fallback.
     qkv_itemsize, o_itemsize = (1, 2) if facts.is_fp8 else (2, 2)
-    fits = [n for n in sorted(caps.tile_ns, reverse=True) if smem_bytes(facts.d_qk, facts.d_v, tile_m, n, qkv_itemsize, o_itemsize) <= SMEM_CAPACITY_BYTES]
+    granule = FP8_HEAD_TILE_GRANULE if facts.is_fp8 else HEAD_TILE_GRANULE
+    d_qp = -(-facts.d_qk // granule) * granule
+    d_vp = -(-facts.d_v // granule) * granule
+    fits = [n for n in sorted(caps.tile_ns, reverse=True) if smem_bytes(d_qp, d_vp, tile_m, n, qkv_itemsize, o_itemsize) <= SMEM_CAPACITY_BYTES]
```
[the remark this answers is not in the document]
Aneureka: Not an issue for now and will fix this in later PRs.

Aneureka:
@cudnn-ci-bot run frost

Aneureka:
@cudnn-ci-bot run frost

PR #602, opened 2026-08-14
frost(sdpa): drop the Amax_S output from the FP8 kernels

vedaanta:
**Stacks on #584** (frost-br-swa) — the diff shows both commits until #584 merges; review only `9f4255b` here.

Nothing consumes `Amax_S`, and producing it cost every FP8 execute a stream-ordered zero-fill plus a per-row bitcast-int32 `atomicMax` in the epilogue. This removes it end to end:

- **Kernels** (SM100, SM107 sibling, SM120 FP8): epilogue reduction + atomicMax gone, ABI slot and fake-tensor compile args dropped. `Amax_O` untouched.
- **Adapters**: `amax_s` buffer plumbing and the launch-stream `.zero_()` removed from both execute paths.
- **Engines**: binding/resolve plumbing removed, plus an honesty gate — a graph that *declares* the `Amax_S` output is declined ("which the FROST engines do not produce") so it routes to an engine that writes it.
- **Analyzer subtlety**: `sdpa_fp8` returns the `Amax_S` port unconditionally, so the fact counts only a real (non-virtual, `set_output(True)`) tensor — pinned by a new probe test (unrequested port ⇒ eligible; requested output ⇒ ineligible).
- **Tests**: the mhas fp8 harness and both frost fp8 suites stop requesting/checking `Amax_S`; the sm120 sink-decline test keeps its behavior with an amax-free rationale.

**Validation** (B200, pip cuDNN 9.23, on top of #584): analyzer + fp8 + sm100 suites + mhas_v2 (paged/bwd_ragged env-excluded): **2105 pass**, fp8 routing unchanged (101 graphs), only the two known pre-existing `cu_seq_len` failures (9.23 backend).

## Summary by CodeRabbit

* **New Features**
  * SM100 and SM120 attention now support bottom-right causal masking with sliding windows where supported.
  * FP8 attention now reports and validates only the independently produced output scaling metric.

* **Bug Fixes**
  * Graphs requesting the unsupported softmax scaling metric are now rejected.
  * Removed obsolete FP8 output handling that could produce unsupported results.

* **Documentation**
  * Updated FP8 execution and output behavior documentation to reflect streamlined metric support.

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -666,8 +666,7 @@ class SdpaFwdDsl(APIBase):
         signature only with additional optional keyword arguments; an adapter
         whose engine capabilities accept FP8/MXFP8 graphs must also accept the
         FP8 operand set the lowering adds for those graphs (``sf_q/sf_k/sf_v``,
-        ``descale_q/descale_k/descale_v``, ``scale_o``, ``amax_o``,
-        ``amax_s`` — see :meth:`SdpaFwdDslSm100.execute`).
+        ``descale_q/descale_k/descale_v``, ``scale_o``, ``amax_o`` — see :meth:`SdpaFwdDslSm100.execute`).
         """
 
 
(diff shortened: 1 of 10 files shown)


jhjpark (approved):

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
Both CodeRabbit findings addressed in `4e22873`: the stale 3-value `_ref` unpack in `_run_template_tail` (a real break on the sm120-only template-tail path, which local SM100 runs skip — good catch) is now the 2-value form, and the E741 `O`/`O_ref` identifiers in the sm100 fp8 suite are renamed `out`/`o_ref` throughout (including the quantization-floor line). sm100 fp8 suite re-run green (32/32).

PR #611, opened 2026-08-15
FE execute-path host overhead: idempotent set_stream + skip the discarded per-call context

YangXu1990uiuc:
Two independent, self-contained cuts to the FE Python execute path's per-call host overhead. Both remove a redundant per-call `cudnn*` backend round-trip that a framework routing ops one-by-one through cuDNN pays every iteration. Neither changes API or results.

For context, the cuDNN backend execute itself is already at cuBLAS parity (`execute_plan_at_index`, pinned plan, ~8.4µs vs `torch.mm` ~7.6µs for a 256³ matmul); these two FE-side round-trips are a large part of the remaining per-op host gap.

## 1. Make `cudnn.set_stream` idempotent (skip `cudnnSetStream` when the stream is unchanged)

`cudnnSetStream` is not a cheap pointer store. For a non-null stream, `cudnn::ops::SetStream` (backend `src/graph/src/context.cpp`) issues **several CUDA driver queries on every call** — green-context detection (`cuStreamGetGreenCtx`), `cudaStreamGetPriority`, `cudaDeviceGetStreamPriorityRange`, plus a `cudaEventRecord` device check when the stream changes — to maintain cuDNN's internal per-priority / per-green-context stream pool. It does this **even when the stream has not changed** (no unchanged-stream early return).

Measured on B200 (host-bound µs/call): `cudnn.set_stream` is **~2.5µs regardless of stream kind** (legacy 2.49 / normal 2.63 / high-priority 2.66 / same-stream-repeated 2.63), and it is **not** the pybind layer (a trivial pybind→backend call is 0.11µs). A framework that calls `set_stream` before every `execute` pays this ~2.4µs every iteration.

Cache `{handle: last_stream}` in the Python layer; `set_stream` forwards to the backend only on a change, and `destroy_handle` forgets the entry so a reused handle address is not wrongly skipped. The bindings are renamed `set_stream`/`destroy_handle` → `_raw_set_stream`/`_raw_destroy_handle` (private, same pattern as `_execute`→`execute`) and the public names become thin caching wrappers; `get_stream` is unchanged and stays consistent with the cache. Assumes a handle is not driven from two streams concurrently (the normal single-stream case).

Regression test mocks the raw backend call (no GPU): verifies unchanged→skipped, changed/new-handle→forwarded, and destroy→re-armed.

## 2. Skip the discarded per-call context in `graph.execute()`

`execute()` built a caller `ExecutionContext` (a `cudnnGetStream` round-trip + an object alloc) at the top of **every** call, but only used it when the plan was not yet built. In steady state the plan is built, so the context was computed and thrown away every call — a ~2.9µs tax that made `execute()` slower than `execute_plan_at_index()` for the *identical* plan.

The fix moves the context build inside the `not _is_built` branch, its only user. Rigorously isolated on a single-plan graph (both calls run the identical kernel, so no plan confound), SM100, 256³ bf16 matmul:

| | before | after |
|---|---|---|
| `execute()` | 16.3 µs | **10.5 µs** |
| `execute_plan_at_index(0)` | 13.9 µs | 10.9 µs |
| gap | +2.4 µs | **~0** (execute() now matches/beats it) |

A surgical control (patch `_build_context` to a no-op) drops unpatched `execute()` by exactly 2.87µs and the patched build by 0.48µs — i.e. the fix removes precisely that discarded round-trip. `test_matmul_bias_relu` 34 passed on SM100 with the change; results bit-identical.

## Complementary backend fix

An unchanged-stream early return in `SetStream` (backend) would help **all** callers, including framework code that calls `cudnnSetStream` directly (e.g. before each op). Filed as a backend MR; this FE cache is the immediate, backend-version-independent half.

---
<sub>note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 — cuDNN per-call host-overhead. cwd /home/scratch.yanxu_libs/cudnn_frontend</sub>

## Summary by CodeRabbit

* **New Features**
  * Added stream management support for cuDNN handles.
  * Stream assignments are tracked independently for each handle.
  * Reassigning the same stream avoids unnecessary backend operations.

* **Bug Fixes**
  * Stream changes are applied correctly when a handle receives a different stream.
  * Cached stream settings are cleared when handles are destroyed, ensuring correct behavior when handles are reused.

* **Performance**
  * Stream and execution context setup is reduced during repeated graph executions.

### DIFF
--- a/python/cudnn/__init__.py
+++ b/python/cudnn/__init__.py
@@ -21,7 +21,6 @@ symbols_to_import = [
     "backend_version",
     "backend_version_string",
     "get_last_error_string",
-    "destroy_handle",
     "norm_forward_phase",
     "reduction_mode",
     "behavior_note",
@@ -31,7 +30,6 @@ symbols_to_import = [
     "create_device_properties",
     "get_stream",
     "numerical_note",
-    "set_stream",
     "build_plan_policy",
     "data_type",
     "tensor_reordering",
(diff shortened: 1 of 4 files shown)


vedaanta (approved):

YangXu1990uiuc:
Closing in favor of #612. #611 was the minimal fallback — the execute-path host-overhead fixes (idempotent `set_stream`, skipping the discarded per-call `ExecutionContext`) without the first-class `cudnn.Handle`. #612 lands the same host-overhead wins as part of the first-class Handle (which owns `{backend_handle, device, stream}` and gives the device its home for build-time scoping), so the two are alternatives and #612 is the one going forward.

PR #612, opened 2026-08-16
First-class cudnn.Handle (create_handle returns an object owning {backend handle, device, stream})

YangXu1990uiuc:
Full-suite alternative to #611 (first-class `cudnn.Handle`); #611 stays open as the minimal, low-risk fallback. **Exactly one merges.**

## What

Make `cudnn.create_handle()` return a first-class `Handle` object instead of a bare int, giving the handle a home for the per-handle state that had accreted as side tables and per-engine queries:
- **stream** — the module-global `_handle_to_stream` cache (#611) moves onto `Handle.stream`;
- **device** — the three parallel device stacks (backend handle; pygraph's `sm_count`/`sm_version`/`device_property` args; frost's own `current_device()` + driver introspection) unify behind `Handle.device`.

The naming anticipates the front end being "cudnn" and today's cuDNN becoming "cudnn backend": this object is the `handle`; the wrapped `cudnnHandle_t` is its `backend_handle`. It is **optional** by design — python engines (frost, cutedsl, linear-attention) need device+stream, not a `cudnnHandle_t`.

The C++ boundary mirrors the backend's own model: a device is carried as `CUDNN_BACKEND_DEVICEPROP_DESCRIPTOR` through the whole build/compile chain (engine / heuristic / execution-plan), and the handle is needed only to populate that descriptor or, at execute, to supply the stream. This PR is the Python analog: `Handle.device` (a `DeviceInfo`) is the device concept threaded into the python engines' build; the handle is consulted for the stream at execute.

Design + a full call-site inventory: [docs/handle_first_class_design.md](docs/handle_first_class_design.md).

## Backward compatibility — transparent for normal use

Every handle-taking cudnn API (`execute`, `set_stream`, `get_stream`, `destroy_handle`, all graph methods) is Handle-aware — the backend handle is extracted **explicitly** via `to_backend_handle(h)` at each named handoff. `Handle` has the dunders the real usage needs: truthy (`if handle:`), identity hash (dict key), identity eq (compares cleanly against `'auto'`/`None` sentinels). Both directions work: a Handle from `create_handle()`, and a foreign raw-int `cudnnHandle_t` a framework created via the C API (isinstance → registry + live `cudnnGetStream`).

The one deliberate change: `Handle` has **no `__index__`/`__int__`**, so code that used the handle as a *raw integer* (`int(handle)`, passing it to a non-FE C API expecting `intptr_t`, integer arithmetic) must read `handle.backend_handle`. This is fail-loud by design — a Handle reaching a binding unconverted raises rather than being silently coerced. A repo-wide grep (python + tests + samples) found **zero** such patterns.

## Landed (all validated)

**Handle core** (`_handle.py`, `__init__.py`): explicit `to_backend_handle(h)` at every handoff (grep `backend_handle` to trace it; an inventory confirmed the set is closed — the `__getattr__` delegation carries no handle). `Handle.stream` is the source of truth (absorbs the write-only `_handle_to_stream`); `get_stream()` reads it with no `cudnnGetStream` round-trip, removing the live query `_resolve_stream` did on every python-engine execute. Passthrough methods (`get_workspace_size{,_plan_at_index}`, `populate/update_cuda_graph`) name `handle`/overrides explicitly (no `*args`); `deserialize` stays a genuine data-or-handle passthrough. Carries #611 (set_stream idempotency + skip the discarded per-call `execute()` context).

**Common device layer** (`_device.py`): `DeviceInfo` is the FE's single owner of a GPU's facts — each fact a `@cached_property` queried once and cached on the instance, one instance per ordinal. `Handle.device` is it; `frost/device.py`'s fact functions become thin shims onto it (its ~24-file / 65-site surface unchanged), so frost consumes the same object instead of a parallel introspection stack.

**Environment layer** (`_env.py`): CUDA driver/runtime versions are process-global, not per-device — collected here (mirroring the backend's argument-less `cudnnGetVersion`/`cudnnGetCudartVersion`), replacing the inline `cuDriverGetVersion` gate in `DeviceInfo` and the duplicated `cudaRuntimeGetVersion` probes in the cutile GDN/KDA engines.

**Device-consumer adoption (frost / linear-attention)**: a `build_device(ordinal)` scope, set once at each engine's `build_plan` hinge from `ctx.handle.device.ordinal`, makes every device-derived build constant follow the handle's GPU. All three tiers now read through the one scope: the engine (`current_device()`), the tile/arch codegen, and the GDN/GDN2/KDA **kernels** (their baked persistent-grid `num_sm`/`max_active_clusters`, previously read off a scope-blind `cudaGetDevice`). `_check_plan_device` stays the execute-time launch guard. sdpa engines need nothing (no device constant baked at build).

## Execute hot-path overhead (measured, parley)

The Handle removes per-execute host work on the python-engine path (frost / LA / cutedsl):
- **stream resolve** `cudnnGetStream` (~1.5 µs) → `handle.stream` attribute;
- **discarded per-execute `_build_context` rebuild** removed (~2.3 µs; from #611);
- **ExecutionContext** frozen-dataclass → `NamedTuple` (still immutable) + `import cudnn` hoisted out of `_resolve_stream`: `_build_context` 647 → 427 ns.

Net: python-engine execute saves **~4.0 µs/execute** vs develop; classic backend execute saves **~2.3 µs** (the shared discarded-rebuild removal; `to_backend_handle` adds a negligible ~65 ns).

## Validation (SM100 parley, fe-jax test bed)

`test_la` 359 / 0 · frost gemm 5805 / 0 · `test_set_stream_cache` 4 · `test_device_info` 3 · `test_build_device` cross-device redirect 3 (scopes a build to L40S/H100/A100 and asserts every frost constant follows — proves device-follows-handle without two Blackwells) · matmul/conv/rope 41 · deviceless-AOT deserialize + workspace passthrough green. The 4 `test_block_scale_quantize_dynamic_shape` failures are pre-existing (override-shape needs cuDNN ≥ 9.21; local backend is 9.20 — identical on develop).

---
<sub>note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 — first-class cudnn.Handle. cwd /home/scratch.yanxu_libs/cudnn_frontend · worktree /home/scratch.yanxu_gpu/fe-handle</sub>

## Summary by CodeRabbit

* **New Features**
  * Added first-class `cudnn.Handle` objects with device and stream tracking.
  * Added device information and CUDA environment APIs.
  * Added device-scoped Frost builds with architecture-aware compilation.
  * Preserved compatibility with raw integer handles.
  * Added explicit handle support for workspace sizing and CUDA graph operations.
* **Bug Fixes**
  * Improved stream caching and handle cleanup.
  * Ensured plan construction uses the handle’s associated device when available.
* **Documentation**
  * Added a design specification covering handle, device, stream, and build behavior.

### DIFF
--- a/docs/handle_first_class_design.md
+++ b/docs/handle_first_class_design.md
@@ -37,7 +37,9 @@ becomes one consumer rather than the anchor.
    handle), so the set is closed. `Handle` deliberately has **no `__index__`**:
    the only path to the backend is those explicit calls, and a Handle that
    reaches a binding unconverted fails loudly instead of being silently coerced.
-   C++ never reads device off the handle. **No `.cpp` changes.**
+   C++ never reads device off the handle. **The C++ handle ABI is unchanged**;
+   the only `.cpp` change is renaming the `set_stream`/`destroy_handle` bindings
+   to `_raw_*` (`python/properties.cpp`) so the Python wrappers own those names.
 2. **The `create_handle` wrapper must be defined AFTER the `__init__.py` symbol-
    copy loop**, or the raw pybind symbol shadows it.
 3. **Dunder minimalism.** `Handle` defines **no** int-coercing dunders and leaves
--- a/docs/handle_first_class_design.md
+++ b/docs/handle_first_class_design.md
@@ -169,6 +169,31 @@ covers them all at once:
   the whole build bakes for the handle's GPU. `tile_config._sm_count()` is
   re-routed off `torch.cuda.current_device` onto `frost.device` so it honours the
   scope too (it was the one query that bypassed `current_device()`).
+- **Frost GEMM compile target (done):** the scope also had to reach the *cute
+  compile arch*, which the earlier constants did not. cutedsl derives the compile
+  target from the ambient CUDA device (`torch.cuda.get_device_capability`), so a
+  build for handle-GPU-A while GPU-B is current baked A's constants into a
+  B-targeted kernel. `_frost_compile_options()` (in `gemm/frost/compiler.py`) now
+  pins `--gpu-arch sm_<scope>` into the `cute.compile()` options string, so the
+  compile target follows the scope. The arch is part of the baked, content-hashed
+  source, so a cross-arch kernel can no longer collide in the JIT cache with a
+  same-source same-machine one. **Limitation:** the pin is honoured on the public
+  `nvidia-cutlass-dsl >= 4.7` (frost's `CUTEDSL_MIN_VERSION`) and on internal RCs;
+  only a public wheel below the floor never threads `--gpu-arch`, and frost already
+  declines those as too-old (`buffers.cutedsl_too_old`, which the support check
+  reuses so an internal RC's own `0.x` numbering is judged new, not old). On such a
+  wheel a handle-scoped build **fails loud** — it cannot pin the target and cutedsl
+  resolves it from an arch captured at *import* time, which we can neither set nor
+  reliably read (a live-device comparison would miss an import-on-B / build-on-A
+  process), so `_frost_compile_options` refuses any `build_device`-scoped build
+  rather than bake scope constants into a possibly-mis-targeted kernel. An unscoped
+  build makes no cross-device promise and is unchanged.
+- **Not yet scope-following (documented holes):** `check_support`/kernel-selection
+  gates read the ambient arch (`buffers.current_sm()`), and the linear-attention
+  kernels lazy-compile at first *execute* — after the build scope has closed — so
+  their compile target is the execute-time device. Same-GPU (scope == ambient, the
+  normal case) all of these agree; a handle-scoped build on a sub-floor wheel is
+  the only case that diverges, and it is the fail-loud path above.
 - `_check_plan_device` **stays unchanged** and correct: it is the EXECUTE-time
   launch guard and must check the *live* current device (where the launch is
   going) against the baked device — the override is a build-scope only, unset at
@@ -205,6 +230,22 @@ is correct (and the build scope is not active then).
   without needing two Blackwells (device queries only, no kernel launch).
   `_check_plan_device` remains the execute-time launch guard against baking on one
   GPU and launching on another.
+- **Compile-target follows scope** (SM100, cutedsl 4.7): the `--gpu-arch` pin is
+  non-regressive on the matching-arch path — `test_matmul.py` bf16 sweep 677
+  passed / 337 skipped end-to-end with `--gpu-arch sm_100a` baked in. That the pin
+  actually moves the target is shown by compiling one graph three ways: `sm_100a`
+  (machine) and `sm_103a` (a different Blackwell sibling) both compile, while
+  `sm_90a` fails in the arch-specific NVVM backend — impossible if the option were
+  ignored (all three would target sm_100 and pass), so on 4.7 the option reaches
+  the compiler. The sub-floor-wheel fail-loud is unit-checked by forcing the
+  support probe false and asserting a `build_device`-scoped build raises while an
+  unscoped one passes through.
+- **Forced through flashinfer's GEMM fuzzer** (SM100): this build dropped into
+  flashinfer's `.venv` (via a `PYTHONPATH` shim) and forced onto the `cudnn`
+  backend across the full unified GEMM fuzz cross-product (bf16 / fp8 / nvfp4 /
+  mxfp4 / mxfp8, mm + bmm) — 731 passed / 0 failed / 151 xfailed (the xfails are
+  flashinfer's pre-tracked, backend-agnostic findings). The first-class `Handle`,
+  `set_stream` idempotency and `destroy_handle` clear are exercised on every one.
 
 ## Relationship to PR #611
 
--- a/docs/handle_first_class_design.md
+++ b/docs/handle_first_class_design.md
@@ -1,0 +1,161 @@
+# First-class `cudnn.Handle` — design
+
+The backend `cudnnHandle_t` carries a device binding (compute capability, SM
+count, ...) and the current stream. On the FE side the handle is a **bare int**
+(`create_handle()` returns `reinterpret_cast<intptr_t>(handle)`), so it has no
+place to hang per-handle state. That state has instead accreted as side tables
+and per-engine queries:
+
+- stream: the module-global `_handle_to_stream` dict (PR #611);
+- device: **three** parallel stacks — the backend handle; pygraph's separate
+  `sm_count`/`sm_version`/`device_property` args (deviceless AOT); and frost's
+  own `current_device()` + driver introspection (`frost/device.py`), which every
+  python engine re-queries because nothing hands it the device.
+
+This makes `create_handle()` return a first-class `Handle` object that owns
+`{backend_handle, device, stream}`, collapsing those into one concept. The
+naming anticipates the front end BEING "cudnn" and today's cuDNN becoming "cudnn
+backend": this object is the `handle`; the wrapped `cudnnHandle_t` is the
+`backend_handle`. It is **optional** by design — python engines (frost, cutedsl,
+linear-attention) need device+stream, not a `cudnnHandle_t`, so the backend
+becomes one consumer rather than the anchor.
+
+`note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 — first-class Handle. cwd /home/scratch.yanxu_libs/cudnn_frontend`
+
+## Hard constraints (from a full call-site inventory)
+
+1. **The C++ boundary needs no change; the handoff is EXPLICIT in Python.** Every
+   handle-consuming binding takes `std::intptr_t` / `std::optional`. The backend
+   handle is extracted **explicitly** in our Python code — `to_backend_handle(h)`
+   at the named handoffs (`_execute*`, `backend_graph`) and `unwrap_handles(args,
+   kwargs)` at the opaque passthroughs (`get_workspace_size`, cuda-graph,
+   `deserialize`) — so a reader can grep `backend_handle` and trace the plumbing
+   top-to-bottom without an IDE. A full inventory confirmed **every** handle→C++
+   handoff is in `_pygraph`/`__init__` (the `__getattr__` delegation carries no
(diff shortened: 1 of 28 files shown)


docs/handle_first_class_design.md
```diff
@@ -0,0 +1,211 @@
...
+
+This makes `create_handle()` return a first-class `Handle` object that owns
+`{backend_handle, device, stream}`, collapsing those into one concept. The
+naming anticipates the front end BEING "cudnn" and today's cuDNN becoming "cudnn
+backend": this object is the `handle`; the wrapped `cudnnHandle_t` is the
+`backend_handle`. It is **optional** by design — python engines (frost, cutedsl,
+linear-attention) need device+stream, not a `cudnnHandle_t`, so the backend
+becomes one consumer rather than the anchor.
+
+`note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 — first-class Handle. cwd /home/scratch.yanxu_libs/cudnn_frontend`
```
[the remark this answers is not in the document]
YangXu1990uiuc: Removed (commit 56b21b7). The session-provenance note-to-self is gone from the published doc; provenance lives in the PR and git history.

python/cudnn/_handle.py
```diff
@@ -0,0 +1,72 @@
...
+    so there is no hidden path from a Handle to the backend: a Handle that reaches
+    a C++ binding unconverted fails loudly rather than being silently coerced. The dunders are therefore minimal on purpose —
+    NO ``__index__``/``__int__`` (no implicit int coercion) — and
+    ``__eq__``/``__hash__``/``__bool__`` are left at the object defaults
+    (identity equality, identity hash, always-truthy), which is what the
+    surrounding code needs: the handle stays a valid dict key, stays truthy in
+    ``if handle:`` guards, and does not raise on ``wrapper.py``'s ``== 'auto'``.
+    """
+
+    __slots__ = ("backend_handle", "_ordinal", "stream")
```
Anerudhan: Remove __slots__ as it is not required.

docs/handle_first_class_design.md
```diff
@@ -0,0 +1,252 @@
+# First-class `cudnn.Handle` — design
+
+The backend `cudnnHandle_t` carries a device binding (compute capability, SM
+count, ...) and the current stream. On the FE side the handle is a **bare int**
+(`create_handle()` returns `reinterpret_cast<intptr_t>(handle)`), so it has no
+place to hang per-handle state. That state has instead accreted as side tables
+and per-engine queries:
+
+- stream: the module-global `_handle_to_stream` dict (PR #611);
```
Anerudhan, line 9: Do you want to keep #611 or close it in favor of #612?

docs/handle_first_class_design.md
```diff
@@ -0,0 +1,252 @@
+# First-class `cudnn.Handle` — design
```
Anerudhan, line 1: Please add labels and trigger the CI

python/cudnn/engines/base.py
```diff
@@ -45,7 +45,7 @@ def execute(self, graph, uid_to_data, ctx):
 
 from abc import ABC
 from dataclasses import dataclass
```
Anerudhan, line 47: Remove dataclass import if not required.

python/cudnn/engines/base.py
```diff
@@ -90,14 +90,19 @@ class PlanConfig:
...
-class ExecutionContext:
+class ExecutionContext(NamedTuple):
     """Runtime context passed to a compiled plan at execute time.
 
     Everything an engine may need is explicit here — no engine should reach
     into private graph state, hard-code a stream, or allocate hidden workspace.
     ``stream`` is resolved from the handle when available (classic
     ``cudnn.set_stream(handle, ...)`` semantics).
+
+    A ``NamedTuple``, not a frozen dataclass: both are immutable (an engine
```
Anerudhan: Extra unwanted comment.

docs/handle_first_class_design.md
```diff
@@ -0,0 +1,252 @@
...
+argument-less globals (`cudnnGetVersion`, `cudnnGetCudartVersion`), never off a
+handle or the `DEVICEPROP` descriptor — cuDNN's own version stays there,
+`cudnn.backend_version()`. The CUDA queries had accreted as re-reads in each
+engine (the `DeviceInfo` oversized-SMEM gate, the cutile GDN/KDA `check_support`);
+`_env` collects them. ~100 ns and off the execute hot path — the cache is a single
+owner returning a constant, not a speed play.
+
+## Stream model
+
+`Handle.stream` is the single source of truth. Today the stream lives in two
```
Anerudhan, line 141: What is the user wants to change the stream at execute.. Should he have to reset the Handle with new stream in the plan and rebuild the plan ?  Sounds confusing with the current way of calling execute.

python/cudnn/_pygraph.py
Anerudhan: Suggest adding an execute with `Handle` class..

YangXu1990uiuc:

python/cudnn/_handle.py
YangXu1990uiuc: Done in 1b57953a — dropped `__slots__` (Handle is created once per (device, stream), not on a hot path, so it buys nothing worth the restriction).

docs/handle_first_class_design.md
YangXu1990uiuc: #612 is the full first-class Handle; #611 was the minimal fallback (the two are alternatives — exactly one lands). Intent is to land #612 and close #611 in its favor — @YangXu1990uiuc to make the call.

docs/handle_first_class_design.md
YangXu1990uiuc: Done — added `cat-feature` / `mod-frontend` / `mod-frost` / `orig-nv-eng` and triggered CI (`@cudnn-ci-bot run python_tests,frost`).

python/cudnn/engines/base.py
YangXu1990uiuc: `dataclass` is used just below by `PlanConfig` (`@dataclass(frozen=True)`, base.py:68), so the import stays. Only `ExecutionContext` is a `NamedTuple` — it's built on every execute, where the frozen-dataclass `__init__`'s `object.__setattr__` per field costs ~220 ns; `PlanConfig` is built rarely, so the frozen dataclass is fine there.

python/cudnn/engines/base.py
YangXu1990uiuc: Trimmed to one line in 1b57953a (kept the load-bearing "NamedTuple not dataclass: ~220 ns/execute cheaper" so a future refactor doesn't silently regress it).

docs/handle_first_class_design.md
YangXu1990uiuc: No rebuild needed for a stream change. The plan is **device**-scoped, not stream-scoped: the stream is read from the handle at execute (`ctx.stream` = the handle's current stream, classic `set_stream` semantics), so `cudnn.set_stream(handle, new_stream)` before the next `execute` repoints the launch stream with no rebuild. Only a **device** change needs a rebuild — and the execute-time device guard raises on a mismatch rather than launching on the wrong GPU. (A caller running two streams concurrently on one shape needs its own handle per stream, since the handle carries a single stream.)

python/cudnn/_pygraph.py
YangXu1990uiuc: Done in 27b8ab748 — `execute` (and every handle API: `set_stream` / `get_stream` / `destroy_handle` / `get_workspace_size`) now takes a `cudnn.Handle` (or `None`) only, and rejects a raw backend int. `to_backend_handle` is exactly the `isinstance(handle, Handle)` unwrap you sketched, applied at every C++ boundary; the stale `handle: int` annotation here is now `Optional[Handle]`. Since `cudnn.create_handle()` is the only way to make a handle in the Python API, no real caller passed a raw int (verified across flashinfer / sglang / FE code; torch uses the C++ frontend).

docs/handle_first_class_design.md
YangXu1990uiuc: yep, will close #611 in favor of this one

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:
### Local cross-arch validation (gh CI mirror is lagging, so running the handle/device suites locally)

Ran on **parley** (cudnn 9.20, torch 2.13+cu130) across four architectures, at HEAD `1b57953a`:

| arch | GPU | result |
|---|---|---|
| SM80 | A100-PCIE-40GB | **80 passed, 3 skipped** |
| SM89 | L40S | **80 passed, 3 skipped** |
| SM90 | H100 NVL | **80 passed, 3 skipped** |
| SM100 | Blackwell | **81 passed, 2 skipped** |

Suites (chosen to cover the first-class-Handle surface + a real end-to-end execute):
- `test/python/test_dispatch.py` — handle plumbing through dispatch (execute → build_plan reaches `ctx.handle`), decline/scope contracts.
- `test/python/test_set_stream_cache.py` — `set_stream`/`get_stream`/`destroy_handle` on a Handle (idempotent set_stream, cleared-after-destroy) + **raw backend int now rejected**.
- `test/python/test_device_info.py` — the `DeviceInfo` layer `Handle.device` resolves through.
- `test/python/test_native_backend_lowering.py` — real `create_handle()` → `pygraph(handle=)` → build → `execute(handle=)` on the backend path.
- `test/python/gemm/frost/test_build_device.py` — the `build_device` device-scoping the Handle feeds (SM100-gated; the extra pass on Blackwell / the skips elsewhere).

The skips are the SM100-only frost/`build_device` cases on the non-Blackwell GPUs (expected).

Also confirmed a real Handle-only end-to-end by hand on SM100: `create_handle()` → build → `execute(handle=Handle)` matches torch (rel-L2 1.6e-3), a raw int passed to `execute`/`set_stream`/`destroy_handle` raises `TypeError`, and `destroy_handle` clears the backend handle.

<sub>note to self: claude::11323ca1-07bc-4fc4-8ec7-ba95d8f061d8 · cwd /home/scratch.yanxu_libs/cudnn_frontend · worktree /home/scratch.yanxu_gpu/fe-handle</sub>

Anerudhan (approved):

PR #614, opened 2026-08-16
frost(sdpa): order every execute-path tensor prep on the launch stream (Rule 5)

vedaanta:
## Before submitting

## Affected area

FROST SM100/SM120 SDPA forward execute paths (`python/cudnn/sdpa/fwd/api_dsl.py`)

## Summary

The dense/fp8/mxfp8 execute paths resolved the launch stream but ran their tensor prep on torch's **current** stream: `_to_bshd`'s gather copy (non-compact layouts), the cached dummies' first-use zero-fill, `_reshape_sf`'s `.contiguous()`, and some O-scratch copy-backs / amax post-ops. With an explicit caller stream (the execute-time handle's), that work races the kernel launch — the same class of bug as the PR #543 THD-upload race, and flagged by review on #608.

Fix, uniformly across the five sites (SM100 dense f16 / mxfp8 / fp8, SM120 dense f16 / fp8): resolve `current_stream` FIRST, run the prep inside `_torch_stream_context(current_stream, device)`, and put the consumers (copy-backs, `amax_o.div_()`) in the same context — matching what the THD paths and the amax resets already did per AGENTS.md Rule 5.

Rebased over #619's device scale-fold: the fp8 paths' `_scale_view` calls are inside the wrap too — `None` binds a cached 1.0 dummy whose **first-use `torch.ones` fill is itself a launch**, so it needs the same ordering (and the post-kernel `amax_o.div_(scale_o_view)` is now a device op inside the consumer wrap).

The PyTorch-integration path launches on torch's current stream, where the context is a no-op; only direct graph-API users with an explicit stream were exposed.

## Validation

- SM100 (B200, cuDNN 9.26): `test_sdpa_fwd_dsl_sm100.py` + `test_sdpa_fwd_fp8_sm100.py` + `test_sdpa_fwd_mxfp8_sm100.py` + `test_sdpa_stream_respect.py` + `test_sdpa_stream_ordering.py` + `test_sdpa_execute_is_async.py`, L0+L1 — **546 passed, 0 failed**.
- SM120 (RTX 5080, cuDNN 9.24): `test_sdpa_fwd_dsl_sm120.py` + `test_sdpa_fwd_fp8_sm120.py` + stream/async suites, L0+L1 — **173 passed, 17 skipped, 0 failed** — the former "10 known head_dim_tail" failures are gone (they were stale direct-call-helper TypeErrors, fixed on develop via #619).

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -1149,30 +1149,33 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
             )
             return
 
-        Q = self._to_bshd(q_tensor)
-        K = self._to_bshd(k_tensor)
-        V = self._to_bshd(v_tensor)
-        O_view, o_needs_copy_back, O_scratch = self._to_bshd_writable(o_tensor)
-
         device = q_tensor.device
-        sinks_t = (
-            self._checked_sinks_1d(sinks)
-            if sinks is not None
-            else self._dummy("sinks", device, lambda: torch.zeros(self.h_q, dtype=torch.float32, device=device))
-        )
-        seq_kv_t = (
-            self._checked_seq_lens(seq_kv_lens, "seq_kv_lens")
-            if seq_kv_lens is not None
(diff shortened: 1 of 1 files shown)


Anerudhan:
Needs CI/conflict resolution

Moving to 1.29

PR #619, opened 2026-08-17
frost(sdpa): #608 follow-ups — stale THD docs; FP8 scales fold in-kernel (Rule 3, Scale_S gone below the graph); baked 2^4 P-cast bias

vedaanta:
## Before submitting

## Affected area

Per-tensor FP8 + MXFP8 kernels and adapter (SM100 / SM107 / SM120); the fp8 lowering; SM100 THD setup-launch comments; `python/cudnn/AGENTS.md`.

### 1. Docs cleanup

The d128/d192 SM100 THD setup-launch comments now describe the plan-time declared-S_q envelope (not the pre-#606 host-computed grid); the resolved THD `cu_seqlens` entry is dropped from AGENTS.md's "Known violations".

### 2. FP8 scales fold in-kernel; Scale_S/Descale_S expunged below the graph

Every fp8 graph execute paid a D2H sync: the adapter `.item()`-read the caller's device scale tensors to fold `descale_q·descale_k` / `descale_[s·]v·scale_o` host-side — the last big Rule 3 violation, and why fp8 execute was not CUDA-graph-capturable.

Now the kernels take `descale_q/k/v` + `scale_o` as **unconditional 1-element fp32 tensor parameters** and fold them in-kernel (1-elem loads → L2 broadcast); the scalar args carry only the bases. One compile form — no flag. The adapter binds the caller's tensors as-is (`None` binds a cached 1.0), `amax_o` divides by the *device* `scale_o`, and `_scalar` plus every execute-path `.item()` are deleted; the AGENTS.md violation entry is retired.

**Scale_S/Descale_S no longer exist below the graph.** The lowering neither resolves nor forwards them (the op's tensors stay bound in the variant pack, simply never read — no framework produces a meaningful non-reciprocal pair: vLLM and FlashInfer have no S-scale surface, and TE's pair is reciprocal by construction); the execute signatures dropped the parameters; SM100's execute-time reciprocal check (itself a Rule 3 readback) is deleted, and SM120's Scale_S kernel machinery is removed. `test_fp8_sm100_s_scales_ignored` pins that wild pair values change nothing, bitwise.

Also repairs the `_run_template_tail` direct-call helper for the current kernel ABI — the ten `test_fp8_sm120_head_dim_tail_direct` L1 tests had been failing with a positional-arg **TypeError since #608** grew the kernel signature under them (never numeric failures); they now pass 10/10.

### 3. Baked 2⁴ P→fp8 cast bias (fp8 SM100/SM107/SM120 + MXFP8 SM100)

P was quantized to fp8 at unit scale, using at most 2⁴ of e4m3's 448 range (the lazy-rescale skip bounds P by `2^RESCALE_THRESHOLD` = 2⁴) while flat-row entries (P ≈ 1/S) sat near the subnormal cliff (~2⁻⁹). Each kernel now bakes `P_CAST_LOG2_SCALE = 4.0`: **P enters BMM2 peaking at 2⁸ = 256 < 448** — no saturation — and flat rows stay in normal range out to S ≈ 2¹³. The invariant `RESCALE_THRESHOLD + P_CAST_LOG2_SCALE ≤ log2(448)` is documented at each constant. The bias is numerically free beyond the improved quantization: it rides the exp2 argument (EX2 is binade-shift-exact), the sums stay in the same 2⁴ units so the O normalization cancels it outright (SM120 de-scales `row_sum` by the exact 2⁻⁴ pre-finalize instead), and only the LSE subtracts the constant (sink denominator lifted to match). This is an internal quantization choice, not cuDNN's Scale_S — that knob no longer exists below the graph.

## Validation (per tree revision; final rebased-tree runs in the PR checks)

- **SM100 (B200, 9.26 nightly)**: fp8 file L0+L1 (incl. the sync-debug-pinned device-scale tests — graph execute under `torch.cuda.set_sync_debug_mode(2)`) + fp8 fwd/bwd + mxfp8 fwd/bwd sweeps + graph-analyzer probes — green at every step (349✓ descale, 293✓ P-cast fp8, full battery re-running post-rebase).
- **SM120 (RTX 5080, 9.24)**: fp8 file L0+L1 — **103 passed, 0 failed**, including all ten >128-head-dim tail accuracy tests and #595's feature set through the device path.

### DIFF
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -96,23 +96,6 @@ neither names the thing that breaks it most directly: a device-to-host read.
 Known violations, all pre-existing and each needing a kernel-side change, so
 none is precedent:
 
-- THD `cu_seqlens` host cumsum (`sdpa/fwd/api_dsl.py`). RESOLVED
-  (issue #552) on every forward THD engine — SM100 f16 (all families) and
-  SM120 f16/fp8 — and the reference for future ports: the kernels compile
-  with dynamic token extents (Rule 4), a setup kernel builds the metadata
-  buffer device-side from the caller's length tensors (cu prefixes
-  normalized), every ragged view binds its buffer's capacity, and the grid
-  is the plan-time declared-S_q envelope — SM100 with dead units exiting
-  via `_thd_decode`'s `batch == n_batch` sentinel, SM120 with its
-  per-sequence rectangular grid whose past-the-length tiles always drained
-  without stores. Zero host reads, pinned per arch by the sync-debug and
-  CUDA-graph capture tests; `_thd_host_lens` is gone from the adapter.
-  Remaining follow-up: the envelope's dead-tile tax under far-oversized
-  declarations (measured ~145 ns/dead unit exposed on SM100, <0.2% at
-  realistic declarations; a capped persistent grid reading a device-side
-  live-unit count would bound it by resident clusters).
-  `sdpa_fwd_wrapper_sm80` shows the caller-provided alternative — it
-  requires `max_s_q` rather than deriving it.
 - Per-tensor FP8 descale readback (`_scalar` in the same file): fold on device,
   passing the pointers, as the backend FP8 sdpa does.
 - The FP8/MXFP8 `seq_len_q` guard in `sdpa/fwd/engines.py`. This one cannot be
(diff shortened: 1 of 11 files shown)


jhjpark (approved):

vedaanta:
@cudnn-ci-bot run frost

PR #627, opened 2026-08-17
reduce cuteDSL grouped gemm CPU latency

hwanseoc:
## Description

Cuts CuTeDSL grouped-GEMM host overhead by memoizing work that is a pure function of
values the memo keys already hold. No check is skipped, and no operand is trusted
because of its identity.

Wall-clock microseconds per call, torch wrapper path, 2048x2048x2048 with 8 experts,
B200, 30 calls amortized behind one sync. Kernel time is 18.4 us.

| wrapper | develop | this PR |
|---|---:|---:|
| `grouped_gemm_wrapper_sm100` | 118.1 | 39.6 |
| `grouped_gemm_glu_wrapper_sm100` | 161.1 | 40.6 |
| `grouped_gemm_dglu_wrapper_sm100` | 185.0 | 51.1 |

The pre-allocated `execute()` path isolates the descriptor cache on its own:
68.2 to 24.7 us on unfused, 69.2 to 24.8 us on GLU.

Medians of 3 interleaved runs against base `d811df965` on an otherwise idle B200. Each
variant was guarded at import to confirm which tree had loaded before timing.
Benchmark harness:
https://gitlab-master.nvidia.com/cudnn/cudnn_frontend/-/merge_requests/2332

### Three changes, smallest first

1. Two process-wide lookup caches, 34 lines. `_convert_to_cutlass_data_type` and the
   `torch.Tensor` class probe in `is_torch_tensor` each get called tens of times per
   launch, and each answers from a fixed table. Every CuTeDSL op benefits.

2. Per-launch descriptor and stream caches, about 14 lines per op. `execute()` rebuilds
   a canonical `TensorDesc` for every operand on every launch, which is the largest
   single cost in `execute()`, and it re-resolves the launch stream. Both outcomes are
   decided entirely by values that now sit in the cache key. A miss runs the full check.

3. The wrapper memo, which is most of the diff. Everything between the wrapper entry and
   `op.execute()` is derivation: resolving dtypes, deriving `(m, n, experts)`, rebuilding
   the op cache key. All of it is a pure function of operand metadata plus the scalar
   config. A hit skips the derivation and reuses the result. It still calls
   `op.execute()`, which validates every operand.

### Why the key is metadata and not id()

An identity-keyed memo is unsound here. CPython recycles a tensor's address as soon as
it is freed, so a freshly allocated tensor routinely lands on an address the memo has
already seen. Measured on this branch's benchmark, 200 freshly allocated tensors
produced 1 distinct `id()`. That means 199 of 200 calls would take a fast path derived
from an object the memo never validated, silently returning a wrong-sized output once
the token count changed.

`_operand_meta` reads shape, strides, dtype and device. That is everything the
derivation consumes and nothing else. Data pointers stay out because they vary per call
and nothing derived depends on them. Their alignment gets re-checked inside `execute()`.

## What this PR deliberately does not do

Four things were built, measured, and then cut because the win did not justify the code.

The validation hoist, `_metadata_validated`, 141 lines. It let a memo hit skip the
metadata block inside `execute()`. It looked worth 53 us on dGLU, but only because dGLU
lacked change 2. Give dGLU the 14-line descriptor cache instead and the hoist drops to
5-10 us across the three ops. It also put a validation-skipping parameter on the public
`execute()` signature. Dropped.

The wgrad memo, 180 lines for roughly 15 us. Worst ratio in the set by 3-5x. Dropped.

A `_stride_order` rewrite. I originally claimed the tuple-sort form beat the `key=`
lambda it replaced by about 1.3 us per operand. Measured in isolation, it is not faster
at all: 10.30 vs 10.38 us for nine operands. Reverted, and the comment claiming
otherwise is gone.

A dsrelu memo, which broke 12 cache-behaviour tests. `use_full_dynamic` masks shapes
from the cache key on purpose, `valid_m == 0` returns before the derivation runs, and
`deterministic=None` resolves from a global. A metadata key cannot hold any of those
honestly. Reverted.

Of the 17 CuTeDSL GEMM ops, 3 are optimized here. Of the remaining 14, none has the
`_validate_live_tensor` layer that change 2 attaches to, so there is nothing there to
cache. Seven mask shapes via `use_full_dynamic` or return early on `valid_m == 0`, which
rules out the wrapper memo. The last six are tractable but unbenchmarked, and I would
not memoize one without building its benchmark first.

## Two traps worth knowing if you extend this

dGLU defaults `linear_offset` during normalization. The memo bypasses normalization, so
the hit path has to resolve the same default from `act_func` or it hands `None` to the
kernel. A single-call test cannot catch this, because only the second call takes the hit
path.

A parameter added to a wrapper but not to its memo key collides silently. This has now
happened twice on rebases: `sf_fp8_dtype_override` in GLU, then `situ_beta1` and
`situ_beta2` in both GLU and dGLU after the SiTU-GLU activation landed in #645. Both
feed the op cache key, so a hit would have served an op compiled for different betas.
Wrong numerics, no error. `test_memo_key_covers_every_wrapper_parameter` auto-discovers
every memoized wrapper and fails on an uncovered parameter. It caught both.

## Type of change

- [x] Performance improvement

## Testing

`test/python/fe_api/test_grouped_gemm_wrapper_memo.py`, 6 tests, pins that a hit matches
the cold path byte for byte, that alternating token counts with freshly built operands
each step still return correct shapes, that a fresh wrongly-typed tensor is still
rejected once the memo is warm, that a transposed operand takes a different key, that
the GLU hit matches its cold path, and the key-coverage guard above.

Scoped to what this diff can reach: `test/python/fe_api/grouped_gemm/` plus the memo
tests. `test_gemm_swiglu.py`, `test_sdpa_bwd.py` and `test_NSA_swa.py` were also run
against develop, because `datatypes.py` and `tensor_adapter.py` are process-wide.

Those three files fail 80 tests identically on develop with this branch reverted, same
64 / 12 / 4 split and same test IDs. They are tolerance misses in dense GEMM and
attention backward, plus an NSA gate that wants backend 9.26.0 against 9.24 here. This
PR adds no new failures.

## Checklist

- [x] Code follows the style guidelines, `black --line-length 160`
- [x] Self-review performed
- [x] Comments added where the reasoning is not obvious from the code
- [x] New tests added and passing

### DIFF
--- a/python/cudnn/datatypes.py
+++ b/python/cudnn/datatypes.py
@@ -246,7 +246,25 @@ def _torch_to_cutlass_data_type(data_type, interpret_uint8_as_fp4x2: bool = Fals
     return None
 
 
+_cutlass_data_type_memo: dict = {}
+
+
 def _convert_to_cutlass_data_type(data_type, interpret_uint8_as_fp4x2: bool = False):
+    # Resolution is process-constant and this runs ~20x per grouped-GEMM wrapper call,
+    # always over the same handful of dtypes. Unhashable spellings fall through uncached.
+    try:
+        key = (data_type, interpret_uint8_as_fp4x2)
+        cached = _cutlass_data_type_memo.get(key)
+    except TypeError:
+        return _convert_to_cutlass_data_type_uncached(data_type, interpret_uint8_as_fp4x2)
+    if cached is None:
+        cached = _convert_to_cutlass_data_type_uncached(data_type, interpret_uint8_as_fp4x2)
+        if cached is not None:
+            _cutlass_data_type_memo[key] = cached
+    return cached
+
+
+def _convert_to_cutlass_data_type_uncached(data_type, interpret_uint8_as_fp4x2: bool = False):
     if is_cutlass_available():
         import cutlass
 
--- a/python/cudnn/gemm/cutedsl/grouped/dglu/_bf16_api.py
+++ b/python/cudnn/gemm/cutedsl/grouped/dglu/_bf16_api.py
@@ -27,6 +27,8 @@ from cudnn.tensor_adapter import (
     get_compute_capability,
     get_data_ptr,
     get_device,
+    get_shape,
+    get_strides,
     get_version,
     is_torch_tensor,
     to_host_list,
@@ -82,6 +84,9 @@ class GroupedGemmDgluBf16API(APIBase):
         else:
             raise ValueError("Provide sample_b for dense mode or (num_experts, b_shape, b_dtype) " "for discrete mode, but not both")
 
+        # A canonical TensorDesc is rebuilt and re-checked for every operand on every
+        # launch, which is the largest single cost in execute(). See _validate_live_tensor.
+        self._live_desc_cache: dict = {}
         self.a_desc = self._make_tensor_desc(sample_a, name="sample_a", canonical=True)
         self.b_desc = self._make_tensor_desc(sample_b, name="sample_b", canonical=True)
         self.c_desc = self._make_tensor_desc(sample_c, name="sample_c", canonical=True)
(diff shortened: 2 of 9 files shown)


hwanseoc:
Test run on the final commit (`40d6a95`), B200:

```
pytest fe_api -q -m L0 -k grouped_gemm
1475 passed, 1300 skipped, 9 failed  (28:57)
```

The 9 failures are diffed against unmodified `develop` rather than assumed — **identical sets**, all in `test_rubin_kernel_dispatch.py` (Rubin/SM107 quantization validation running on a B200). Nothing introduced by this PR.

`uvx pre-commit run`: clang-format skipped, black and black-jupyter pass.

hwanseoc:
@coderabbitai ignore

hwanseoc:
@cudnn-ci-bot run oss

Anerudhan (approved):

PR #628, opened 2026-08-17
test: directed SDPA bwd test for rows with very negative LSE

vedaanta:
Guards the bug class behind nvbug 6591137 (internal; cuDNN dev MR !4225, 9.26 MR !4227): the SM90 bprop kernel walks KV in 2-warp-group (128-row) CTA tiles but decided the out-of-bounds mask from the 64-row tile granularity, so for `s_kv = 64 (mod 128)` the TMA-zero-filled KV tail entered `exp(0 - LSE)` unmasked, overflowing to inf — and grad_q to NaN — for query rows with `LSE < -ln(FLT_MAX) = -88.7`.

**Why the existing random suites never caught it:** randn-like inputs give `LSE ~ log(s_kv) > 0`, about 100 away from the overflow threshold; and with benign LSE the unmasked garbage only ever multiplies zero-filled K/V/dO operands, so every stored output stays bit-identical — no tolerance could flag it. The failure needs *adversarial* data, not tighter checks.

The new `test_sdpa_negative_lse_bwd_L0` engineers query rows anti-aligned with every key (all scaled logits ~ -100, so `LSE ~ -100 + log(s_kv)`), at `s in {64, 192, 1216, 1280, 1000} x {bf16, fp16}` — covering both halves of a 128-row KV group, a multiple of 128, and a non-multiple.

Verified on H100 NVL: on unfixed cuDNN (9.30 official) exactly the three `s = 64 (mod 128)` lengths fail per dtype; all 10 pass with the fixed library.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for scaled dot-product attention backward operations with very negative log-sum-exp values.
  * Expanded validation across multiple sequence lengths and FP16/BF16 data types.
  * Added coverage for directed L0 backward-attention scenarios.

### DIFF
--- a/test/python/test_mhas_v2.py
+++ b/test/python/test_mhas_v2.py
@@ -27,7 +27,7 @@ from sdpa.random_config import (
     RandomChoice,
     SlidingWindowMaskGenerator,
 )
-from sdpa.fp16 import exec_sdpa
+from sdpa.fp16 import exec_sdpa, TensorUid
 from sdpa.fp8 import exec_sdpa_fp8
 from sdpa.mxfp8 import exec_sdpa_mxfp8
 from sdpa.blocked import fetch_blocked_tests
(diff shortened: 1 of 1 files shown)


Anerudhan:
CI/labels/etc.
Moving to 1.29

PR #631, opened 2026-08-17
Avoid zero-initializing JAX outputs the kernel already writes

hwanseoc:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL (Python/JAX entry points)

## Summary

Stop passing `initialized_outputs` for JAX outputs the kernel already writes, in `grouped/unfused`, `grouped/glu`, `dense/swiglu`, and the workspace of `grouped/wgrad`. No kernel, output spec, or layout changes. Adds two JAX tests.

**Supersedes and closes #592** (@mgoldfarb-nvidia), which established this for `grouped/unfused` and is folded in here so the whole audit lands as one piece. The mechanism, measurement protocol and framing are Michael's.

## Why

`initialized_outputs` makes XLA zero-fill the buffer before every dispatch. That fill scales with the output, so it was the dominant host-visible cost of the JAX path — and the torch wrapper never paid it, since it allocates with `torch.empty_strided`.

| entry point | dropped |
|---|---|
| `grouped/unfused` | `d`, `c`, workspace — from #592 |
| `grouped/glu` | `d`, `c`, workspace |
| `dense/swiglu` | `ab12`, `c` |
| `grouped/wgrad` | workspace only; the template keeps its zeros, since the kernel writes through `wgrad_ptrs` and never touches that buffer |

### Deliberately unchanged — the part worth carrying forward

I tried removing it from every remaining site and reverted six: `dense/amax`, `dense/srelu`, `grouped/dglu`, `grouped/dsrelu`, `discrete_grouped/swiglu`, `discrete_grouped/dswiglu`.

Their comments cite the bridge's leading-dim inference rejecting trailing-unit-dim buffers on pure results, and that reason is **still accurate** — not stale documentation, which is what I had assumed. All six fail the JAX jit-vs-eager test without the donation, and `grouped/dglu` returns results differing from torch by 1.125 where it matched exactly.

So this is not mechanically applicable across the JAX entry points. An audit that assumes it is will silently corrupt six of them.

## Related issues

Supersedes #592.

For `grouped/unfused` and `grouped/glu`, rows at or past `padded_offsets[-1]` are now **unspecified** rather than zero — the same behavior the torch wrapper has always had via `empty_strided`. Documented in both docstrings. Rows before that offset are bit-identical to the torch path under full and partial coverage. No change to supported GPU/cuDNN/CUDA/Python versions.

Performance — B200, bf16, 8 experts, `jax.jit` wall time, median of 3 alternating base/head runs via `benchmark/grouped_gemm_dispatch/bench_dispatch.py`. Each invocation asserts which module tree it loaded before measuring, so the two columns cannot silently be the same build:

| API | M × N × K | kernel | base (µs) | this PR (µs) |
|---|---|---:|---:|---:|
| unfused | 2048 × 2048 × 2048 | 18.8 | 38.8 | **35.4** |
| unfused | 4096 × 2048 × 2048 | 32.0 | 52.6 | **37.2** |
| unfused | 8192 × 4096 × 2048 | 106.9 | 136.9 | **110.5** |
| unfused | 16384 × 4096 × 4096 | 388.0 | 443.2 | **392.2** |
| glu | 2048 × 2048 × 2048 | 18.4 | 41.5 | **34.5** |
| glu | 4096 × 2048 × 2048 | 31.4 | 50.1 | **36.1** |
| glu | 8192 × 4096 × 2048 | 105.0 | 122.9 | **108.7** |
| glu | 16384 × 4096 × 4096 | 384.0 | 416.3 | **387.1** |

At the largest shape that is +4.2 µs over the kernel for unfused and +3.1 for glu, against +55 and +32 before. The overhead no longer grows with output size, which was the original complaint.

## Testing

Two new JAX tests, one per changed grouped entry point:

- `fe_api/grouped_gemm/test_grouped_gemm_jax.py::test_grouped_gemm_jax_addressed_rows_survive_a_dirty_allocator`
- `fe_api/grouped_gemm/test_grouped_gemm_glu_jax.py::test_grouped_gemm_glu_jax_addressed_rows_survive_a_dirty_allocator`

Both run partial coverage (`padded_offsets[-1] < m`) with the allocator poisoned before dispatch, and assert the addressed rows match the torch wrapper bit-for-bit. The poisoning is the point: on fresh device memory the untouched tail reads back as zeros whether or not the fill is present, so a test that does not dirty the allocator first passes either way and proves nothing.

```
cd test/python && pytest fe_api -q -m L0 -k jax
57 passed, 4252 deselected, 2 xfailed
```

That suite is what caught the six reverts above.

```
uvx pre-commit run --files <changed>
clang-format....(no files to check)Skipped
black...........Passed
black-jupyter...Passed
```

### DIFF
--- a/python/cudnn/gemm/cutedsl/dense/swiglu/jax_api.py
+++ b/python/cudnn/gemm/cutedsl/dense/swiglu/jax_api.py
@@ -20,7 +20,7 @@ import cutlass.utils
 
 from cudnn.datatypes import _convert_to_cutlass_data_type
 from cudnn.tensor_adapter import framework_dtype
-from cudnn.jax import call, gemm_operand_spec, row_major_desc as _make_desc, sf_atom_spec, zeros_init
+from cudnn.jax import call, gemm_operand_spec, row_major_desc as _make_desc, sf_atom_spec
 from .api import GemmSwigluSm100
 
 # config_key -> (kernel instance, max_active_clusters). The instance does not vary
@@ -145,16 +145,14 @@ def gemm_swiglu_jax_sm100(
         jax.ShapeDtypeStruct((m, n, l), framework_dtype(ab12_dtype, "jax")),  # ab12
         jax.ShapeDtypeStruct((m, n // 2, l), framework_dtype(c_dtype, "jax")),  # c
     )
-    # Outputs are donated pre-initialized inputs: the bridge's leading-dim inference
-    # rejects trailing-unit-dim buffers on pure results, and the donated-input path
-    # carries the explicit (1, 0, 2) layout spec.
+    # Neither output needs zeroing: the kernel writes ab12 and c in full over (m, n, l)
+    # and (m, n // 2, l). Zero-filling them was a full-size device write per dispatch.
     if not is_quantized:
         ab12_tensor, c_tensor = call(
             _swiglu_adapter,
             output_shape_dtype=out_types,
             input_spec=(operand, operand),
             output_spec=(operand, operand),
-            initialized_outputs={0: zeros_init, 1: zeros_init},
             kernel=kernel,
             mac=mac,
             alpha=float(alpha),
@@ -166,7 +164,6 @@ def gemm_swiglu_jax_sm100(
             output_shape_dtype=(out_types[1], out_types[0]),
             input_spec=(operand, operand, sf, sf),
             output_spec=(operand, operand),
-            initialized_outputs={0: zeros_init, 1: zeros_init},
             kernel=kernel,
             mac=mac,
             alpha=float(alpha),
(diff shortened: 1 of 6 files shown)


hwanseoc:
@cudnn-ci-bot run oss

hwanseoc:
@cudnn-ci-bot run oss

Anerudhan (approved):

hwanseoc:
@coderabbit ignore

hwanseoc:
Thanks for the approval. Two follow-ups before this is merged, one of them yours (@coderabbitai):

1. **Availability gate** — CodeRabbit was right: both new tests called `skip_unless_sm100()` without first checking `cutlass.jax.is_available()`, unlike every other JAX test in the same two files. On an SM100 machine without the CuTeDSL JAX extensions they would have failed rather than skipped. Fixed in 41462e0.

2. **CI has not covered the final commit.** Both mirror pipeline runs so far were on `pr-631-99a334b`, which predates the `grouped/unfused` fold-in from #592 — i.e. the largest part of this change has not been through the internal pipeline. Re-triggering below.

@cudnn-ci-bot run oss

hwanseoc:
@cudnn-ci-bot run oss

PR #636, opened 2026-08-17
benchmark: attention inference suite (context/generation, cudnn vs cudnn_oss)

vedaanta:
## What

A new standalone perf harness, `benchmark/attention_inference/`, measuring attention **as served** — split into two phases mirroring fwd/bwd in `sdpa_benchmark_training`:

- **context** (TFLOPS): full prefill (`s_q == s_kv`) and chunked prefill (512/1024-token chunks against a 64k/128k cache, bottom-right causal), stacked per-kind in one chart.
- **generation** (GB/s + % of memory SOL): `q_tokens = 1 + MTP` for MTP 0–3 against a 128k cache, one chart subplot per MTP width.

Two backends are swept and charted:

- `cudnn` — the graph API on native backend engines (heur A + FALLBACK)
- `cudnn_oss` — the same graph planned with `heur_mode.OPENSOURCE` + `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`, so only the frontend's open-source engines may serve it; the winning plan is recorded per case

Reference backends (flashinfer / flash_mla / b12x / FA4) remain implemented in `benchmark_single_attention.py` for ad-hoc `--backend` runs but are out of the default sweep.

## Sweep design

- Models from official HF configs: llama3.1 (GQA d128), qwen3.5 (GQA d256), gpt-oss (d64, SWA-128 + sinks), deepseek-v4 (shared-K=V MQA d512), kimi-k3 (absorbed MLA 576/512), and an autoregressive video-DiT.
- Every model swept across **TP 1/2/4/8 shards**; charts render each shape as a cluster whose ticks run tp1 → tp8 with per-shard head counts.
- Generation batch axis is **[1, 128]**: measured on B300, batches in between are a smooth occupancy ramp (b32 vs b256 within ~10% at kv ≥ 32k), so the suite keeps the latency anchor and the bandwidth plateau. Same measurement showed kv 32k vs 128k identical at the plateau — one long cache suffices.
- fp8-e4m3 KV axis on llama/qwen/gpt_oss/dsv4 (full fp8 graph with amax outputs; unsupported combos — e.g. d>128 fp8 — chart as blanks).
- Every expanded case owns an x slot whether or not it ran: unsupported combinations render as blank slots, never dropped ticks.

## Results included

`results/<config>/b300/` carries a full B300 (sm_103) sweep. Highlights: the frost d512 prefill engine outruns the native sm103 path ~2.8× on dsv4 prefill and serves all shared-KV decode; native q=1 decode runs ~90% of memory SOL but MTP>0 widths drop to ~10%; b1/tp8 decode reaches only 19% of the batched plateau on llama. sm120 (RTX PRO 6000) results are queued as a follow-up commit on this branch.

## Test plan

- Harness smoke-tested per-case and end-to-end (runner → CSV → charts) on sm100 and sm103.
- `benchmark/` only; no library code touched except one `sdpa_benchmark_training/runner.py` line tolerating `flash_attn` without `__version__`.

## Summary by CodeRabbit

* **New Features**
  * Added an attention-inference benchmarking suite for context and generation workloads.
  * Added support for multiple attention backends, model architectures, tensor-parallel configurations, KV-cache formats, masking modes, and cache layouts.
  * Added benchmark result export, filtering, dry runs, automated execution, and performance charts.
  * Added configurations for Llama, Qwen, Kimi, DeepSeek, GPT-OSS, and video DiT workloads.
  * Added training benchmark configurations for Kimi K3 and DeepSeek-V4.

* **Documentation**
  * Added usage instructions, supported scenarios, measurement details, and configuration guidance.

* **Bug Fixes**
  * Improved environment logging when optional package version metadata is unavailable.


vedaanta:
Addressed the review in ad2a12b:

- **TP sharding now follows framework practice**: `with_tp_shards` requires q heads to divide evenly and kv heads to divide-or-replicate (`tp % kv == 0`), raising otherwise. The 9-head video DiT is therefore whole-model only — deployments sequence/context-parallelize such models rather than head-shard them.
- **fp8 graph**: sink presets are rejected (previously ran sinkless but recorded success); q=1 sliding-window decode is bottom-right anchored to match the bf16 path; Q/O bytes counted at 1B since the cudnn fp8 graph holds q/k/v/o in e4m3.
- **docs reconciled**: dsv4 states it benchmarks the unwindowed shared-KV core, kimi's context sweep documents the intentional blank slots, `kv_cache_dtypes` explains the full-fp8 realization vs the serving mixed mode.
- **tooling**: `load_config` re-raises missing deps from inside valid config modules, dry runs honor `--filter`/`--phase`, `run_all.sh` guards `cd`, validates config names, and creates the log dir.

The B300 fp8 rows will be refreshed by a re-run with the corrected byte accounting and sink rejection (gpt_oss fp8 becomes honest blanks).

brandonfzhang (approved):

PR #639, opened 2026-08-17
test(python): isolate crashing and hanging tests so one bad test cannot poison the run

hwanseoc:
@coderabbitai ignore

## Problem

Three failure modes are unrecoverable inside a pytest process, and each turns one bad test into a run full of lies:

- a **segfault / abort** kills the interpreter and the rest of the run with it;
- a **faulting kernel** (illegal memory access, device-side assert, ...) leaves the CUDA context sticky, so every later CUDA call -- fixture setup included -- fails with the same error;
- a **hung test** blocks the session forever.

## Design

Process isolation with a supervisor. Tests run inside a pytest-xdist worker; the worker is killed the moment its context is dead or a test overruns its deadline. The xdist controller reports the offending test, replaces the worker, and the rest of the session runs in a fresh process with fresh fixtures. All of it lives in one `# Crash isolation` block in `test/python/conftest.py`.

| piece | what it does |
|---|---|
| `pytest_cmdline_main` (tryfirst, ahead of xdist's) | sets `numprocesses=1` when no `-n` was passed; raises `--max-worker-restart` to 100000 (xdist's default is 4x the worker count; every crashing test costs one). Skipped under `-n<N>`, `-s`, `--pdb`, `--collect-only`, `CUDNN_TEST_NO_ISOLATION=1`, or when xdist is absent. Skipped in workers too: they re-enter this hook with `numprocesses` reset to `None`, and would otherwise spawn workers of their own. |
| `pytest_runtest_logfinish` | probes the context after **every** test with one `torch.cuda.synchronize()`. A sticky error is returned by every later CUDA call, so this catches faults wherever they surfaced -- including ones the test swallowed in a `try/except`. Kills the worker only after the test's report has gone out, so the failure keeps its real traceback rather than showing up only as "worker crashed". With no supervisor it calls `pytest.exit` instead of cascading. |
| `pytest_runtest_logstart` | arms `faulthandler.dump_traceback_later(CUDNN_TEST_TIMEOUT, exit=True)` (default 900 s, `0` disables). Stays armed through the probe, because the probe blocks on any kernel the test left running. On expiry: all thread stacks to the real stderr, then `_exit()`. |

**Why faulthandler and not a Python thread or `SIGALRM`:** a GPU hang parks the main thread inside a CUDA driver call *holding the GIL*. A Python watchdog thread is starved and a signal handler never runs. Measured, not guessed: the first version of this PR used a Python thread, and instrumentation showed its 5 s tick arriving 110 s late in exactly that situation. faulthandler's watchdog is a C thread that needs no GIL.

Also drops the dead `pytest_load_initial_conftests` arg injection: pluggy snapshots hook impls before conftests register, so it never ran (its `--tb=short --no-header` defaults already live in `pytest.ini`).

## Relationship to #791

#791 removed the old `synchronize` monkeypatch because it hard-exited the worker on *any* async CUDA error, which would take a whole per-GPU shard down. This PR's kill is different in two ways: it fires only when the context is actually dead (verified by the probe) or the test has hung -- states in which that worker could not have run anything further anyway -- and the controller replaces the worker and re-queues its pending tests, so the shard survives. A per-GPU pinning scheme should key off the `--tx` spec (which a replacement worker inherits), not the `gw<N>` id (which changes).

## Validation (B200, cuDNN 9.27)

A 24-test fault-mix probe (temporary, not committed) with a healthy test after every fault:

| injected fault | outcome |
|---|---|
| plain `assert` failure / Python exception | FAILED, worker keeps running |
| `pytest.raises` / `skip` / `xfail` | unaffected |
| device-side assert (OOB `index_select`) | real `AcceleratorError` traceback, worker killed + replaced |
| real IMA (compiled kernel writing 4 TiB past a tensor) | real `AcceleratorError` traceback, worker killed + replaced |
| CUDA fault swallowed by the test's `try/except` | caught by the post-test probe, worker killed + replaced |
| CPU segfault, `SIGABRT` | worker killed + replaced |
| CUDA fault in **fixture teardown** | teardown ERROR reported, worker killed + replaced |
| CPU hang (`sleep`) | `Timeout` + stacks naming the test frame, worker killed + replaced |
| GPU hang, sync inside the test | same |
| GPU hang, kernel left running, test returns without sync | probe blocks, watchdog fires *inside the probe*, worker killed + replaced -- this case was RED before the faulthandler rewrite |

Every "healthy after X" test passed. Single-process (`-s`) fallbacks: dead context -> `pytest.exit` with a one-line reason; hang -> `Timeout` dump and hard exit. Regression subset (`test_conv_fprop.py test_matmul_bias_relu.py test_rmsnorm.py`): `52 passed`, unchanged.

## Things to know

- **Killing a worker does not stop a kernel it left running.** The driver keeps that context alive until the kernel ends (a dead pid stayed visible in `nvidia-smi` at 98 % util), and the next worker can block behind it. The deadline bounds the damage to one test at a time, but only a GPU reset frees the SMs. Host-side hangs and bounded-but-slow kernels are fully handled.
- A crashing test appears twice in the summary: its own report plus xdist's "worker crashed while running ...".
- Default single-test runs now spawn one worker (~2 s startup), so live `print` / `--dryrun` output needs `-s`, which also turns isolation off.
- Reproducing a real IMA needs a compiled kernel: `cudaMemcpyAsync` from an unmapped or freed pointer is rejected up front (`cudaErrorInvalidValue`), and storage aliasing past an allocation stays inside the expandable-segment mapping.


hwanseoc:
@cudnn-ci-bot run help

hwanseoc:
@cudnn-ci-bot run python_samples,python_tests

hwanseoc:
Rebased onto `develop` (9ef93bf), and validated on a real B200 + cuDNN 9.27 backend instead of the earlier stub sandbox.

**Fault injection** (temporary probe file, not committed: 3 crashing tests interleaved with healthy ones):

| injected fault | result |
|---|---|
| out-of-bounds `index_select` -> device-side assert surfaced via `synchronize` | real traceback + `AcceleratorError` reported, worker killed, replaced |
| same fault swallowed by the test's own `try/except` | caught by the post-test context probe, worker killed, replaced |
| `ctypes.string_at(0)` (SIGSEGV) | worker killed, replaced |

Every healthy test before and after the three faults passed; the session ran to completion (`4 failed, 4 passed`) instead of dying at the first crash. With `CUDNN_TEST_NO_ISOLATION=1` the same file stops at the first fault -- and the session-scoped `cudnn_handle` teardown then fails with `cudnnHandle Destroy failed`, which is exactly the sticky-fixture symptom this PR removes.

**Regression check:** `test_conv_fprop.py test_matmul_bias_relu.py test_rmsnorm.py` -> `52 passed` under the injected `-n1`.

**Overhead:** ~2 s per session for worker startup (`test_rmsnorm.py`: 17.7 s with `-n1` vs 15.7 s in-process). The duplicated `bringing up nodes...` line is stock xdist output, present with a plain `-n1` too.

hwanseoc:
Dropped the `torch.cuda.synchronize` monkeypatch and the `synchronize_unsafe` alias entirely -- the crash guard supersedes it.

The guard's remaining job after this PR was printing a filtered traceback and re-raising; pytest already reports that failure with a better traceback, and the hard exit now happens in `pytest_runtest_logfinish`, where the report has already been emitted. Nothing outside `conftest.py` referenced `synchronize_unsafe`, so the probe just calls `torch.cuda.synchronize()`.

Re-validated on B200 after the removal: same fault-injection results (device-side assert surfaced through `synchronize`, the same fault swallowed by the test, and a SIGSEGV -- each kills only its worker, which is replaced; healthy tests before and after all pass), and the failing test still reports the real `torch.AcceleratorError` traceback. Regression subset unchanged: `52 passed`.

hwanseoc:
Built a 21-test mixed-fault probe (untracked, not for commit) and ran it on B200. It found one gap -- **hangs** -- which this push closes.

| injected fault | outcome |
|---|---|
| plain `assert` failure | FAILED, worker keeps running |
| ordinary Python exception | FAILED, worker keeps running |
| `pytest.raises` / `skip` / `xfail` | passed / skipped / xfailed, unaffected |
| device-side assert (OOB `index_select`) | real `AcceleratorError` traceback, worker killed + replaced |
| real IMA (kernel writing 4 TiB past a tensor) | real `AcceleratorError` traceback, worker killed + replaced |
| CUDA fault swallowed by the test's `try/except` | caught by the post-test probe, worker killed + replaced |
| CPU segfault (`ctypes.string_at(0)`) | worker killed + replaced |
| `SIGABRT` (`abort()`) | worker killed + replaced |
| CUDA fault in **fixture teardown** | teardown ERROR reported, worker killed + replaced |
| CPU hang (`sleep(3600)`) | **was: hung the session forever** -> now killed by the watchdog |
| GPU hang (spin kernel + `synchronize`) | **was: hung the session forever** -> now killed by the watchdog |

Final run: `12 failed, 12 passed, 1 skipped, 1 xfailed, 1 error in 81 s` with 8 worker restarts (gw0-gw7). Every "healthy after X" test passed, so fixtures recover fully after each crash. GPU left at 0% util / 0 MiB -- the spin kernel dies with its context. Regression subset unchanged: `52 passed`.

## New in this push: hang watchdog

A daemon thread armed per test kills the worker after `CUDNN_TEST_TIMEOUT` seconds (default 900, `0` disables), dumping all thread stacks first:

```
[watchdog] test_zz_crash_probe.py::test_705_hang_gpu: still running after 25s, killing the test process
  File ".../test_zz_crash_probe.py", line 166 in test_705_hang_gpu
  ...
```

Two details worth knowing:
- It is a **thread, not a signal**: a GPU hang parks the main thread inside a CUDA sync, where Python signal handlers never run. `SIGALRM` would never fire.
- It **suspends pytest's global capture before printing**. Mid-test, fd 2 is redirected into the per-test capture file, which dies with the process -- without the suspend the message and the stacks are silently lost (first version of this hit exactly that).

## Note on reproducing a real IMA

`cudaMemcpyAsync` from an unmapped or freed device pointer is rejected up front by the runtime (`cudaErrorInvalidValue`, no fault), and storage aliasing far past an allocation stays inside the expandable-segment mapping. Only a compiled kernel writing out of bounds produced a genuine `cudaErrorIllegalAddress`.

vedaanta (approved):

hwanseoc:

hwanseoc:

hwanseoc:
@cudnn-ci-bot run python_tests frost

PR #642, opened 2026-08-18
fix(sdpa): move the pre-9.26 Stats packed-BHSD check to post_validate_node

vedaanta:
## Problem

The Stats layout check added in 423767bf4 (#304) runs in `validate_sdpa_support_surface()`, which is called from `pre_validate_node()` — i.e. **before shape inference**. Users (and every C++ sample) that leave the Stats output dim/stride unset for inference, e.g.

```cpp
Stats->set_output(true).set_data_type(fe::DataType_t::FLOAT).set_uid(STATS_UID);
```

hit `stats_dim.size() == 4` == false at check time, so `graph->build()` fails with `GRAPH_NOT_SUPPORTED` on every cuDNN < 9.26. This broke the nightly `cpp_samples:cudnn_9.19.0.56` CI jobs (Ampere + Hopper) on develop starting 2026-08-15 — every stats-generating fwd sample fails. 9.26+ jobs skip the check entirely, which is why only the old-cuDNN jobs went red.

## Fix

Move the check to `SDPANodeBase::post_validate_node()`, which runs after `infer_properties_node()` has filled an unset Stats with packed BHSD (`{h*s_q, s_q, 1, 1}`) — alongside the existing O innermost-stride check. Inferred layouts now pass by construction; explicitly-set non-BHSD layouts are still rejected, and the error still surfaces from `validate()`/`build()`.

The backward-node twin of this check is left where it is: Stats is an *input* there, so its dims/strides are always user-set before validation.

## Verification

- Full `cudnn_frontend.h` TU compiles with `-Wall -Werror` (g++ 11, cuDNN 9.2x headers, CUDA 13.1).
- clang-format applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for scaled dot-product attention statistics outputs on older cuDNN versions.
  * Unsupported non-ragged statistics layouts are now rejected with a clear “graph not supported” result.
  * Valid ragged outputs and supported newer cuDNN versions remain unaffected.

### DIFF
--- a/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
+++ b/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
@@ -498,6 +498,25 @@ class SDPANodeBase : public NodeCRTP<DerivedT> {
 
         CUDNN_FE_VALIDATE_STRIDE(output_names::O, attributes.outputs);
 
+        // Non-ragged Stats layouts other than packed BHSD are not correctly supported prior to 9.26.0.
+        // Runs post shape inference so that an unset Stats layout (always inferred as packed BHSD)
+        // is not rejected.
+        auto const& stats_out = attributes.outputs.find(output_names::Stats);
+        bool const has_stats  = (stats_out != attributes.outputs.end()) && (stats_out->second != nullptr);
+        if (has_stats && !stats_out->second->get_ragged_offset() && detail::get_backend_version() < 92600) {
+            auto const& stats_dim           = stats_out->second->get_dim();
+            auto const& stats_stride        = stats_out->second->get_stride();
+            bool const stats_is_packed_bhsd = stats_dim.size() == 4 && stats_stride.size() == 4 &&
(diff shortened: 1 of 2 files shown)


vedaanta:
@cudnn-ci-bot run frost, backend

vedaanta:
Ran the eight CI-failing cpp samples locally on a Blackwell box against **cuDNN 9.25** (< 9.26, so the check path is active), built with this repo's CMake (`-Wall -Werror`):

- **At base (origin/develop, pre-fix):** `test cases: 8 | 7 failed | 1 skipped` — reproduces the nightly `cpp_samples` failure exactly (every stats-generating fwd sample fails `graph->build()`).
- **With this PR:** `test cases: 8 | 7 passed | 1 skipped, assertions: 391 | 391 passed`.

Tests: Toy sdpa forward, Cached sdpa, flexible graph, dropout, paged caches, paged decode+prefill, CUDA graph, sink.

Adnios (approved):

PR #649, opened 2026-08-18
style: fix black formatting in cutedsl grouped GEMM python files

hwanseoc:
@coderabbitai ignore

## Problem

The scheduled `analysis:clang-format` job on `develop` fails on the black check ([job 401881704](https://gitlab-master.nvidia.com/cudnn/cudnn_frontend/-/jobs/401881704)):

```
would reformat python/cudnn/gemm/cutedsl/grouped/glu/api.py
would reformat python/cudnn/gemm/cutedsl/grouped/wgrad/moe_blockscaled_grouped_gemm_wgrad_rubin.py
would reformat python/cudnn/gemm/cutedsl/grouped/dglu/api.py

Oh no! 3 files would be reformatted, 602 files would be left unchanged.
```

## Fix

Ran `black --line-length 160 --fast` (pinned 26.3.1, matching `.pre-commit-config.yaml`) on the three files.

Two manual follow-ups on top of the mechanical reformat:

- **Latent bug** in `GluCall` / `DgluCall`: `sf_fp8_dtype_override: Optional[Literal["e5m3"]] = None,` had a stray trailing comma, so the dataclass default was the tuple `(None,)`, not `None`. Black faithfully rewrote it to `= (None,)`; dropped the comma instead so the default matches the annotation and the public API's own `= None` default.
- Joined an implicitly-concatenated `ValueError` message that black collapsed onto one line.

## Verification

`black --line-length 160 --fast --check python test samples` → `606 files would be left unchanged`.

### DIFF
--- a/python/cudnn/gemm/cutedsl/grouped/dglu/api.py
+++ b/python/cudnn/gemm/cutedsl/grouped/dglu/api.py
@@ -107,7 +107,7 @@ class DgluCall:
     mma_tiler_mn: Tuple[int, int] = (256, 256)
     cluster_shape_mn: Optional[Tuple[int, int]] = None
     sf_vec_size: int = 16
-    sf_fp8_dtype_override: Optional[Literal["e5m3"]] = None,
+    sf_fp8_dtype_override: Optional[Literal["e5m3"]] = None
     vector_f32: bool = False
     m_aligned: int = 256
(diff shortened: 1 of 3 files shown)


hwanseoc:
@cudnn-ci-bot run none

PR #650, opened 2026-08-18
benchmark(sdpa): add the Hopper (sm90) row to the peak-MMA table

brandonfzhang:
`_FLOPS_PER_CLOCK_PER_SM` only had sm100 and sm12x entries, so H200 runs computed no `peak_mma_tflops` and their charts drew no MMA-throughput max line while every other GPU's charts have one.

H100/H200 (sm90): 989.5 dense BF16 TFLOPS with FP32 accumulate = 132 SMs x 1.83 GHz x **4096 FLOPs/clk/SM**; FP8 dense is 2x (8192). No mxfp8 entry — Hopper has no MXFP8 datapath, and those cases already record as unsupported.

## Summary by CodeRabbit

* **Performance**
  * Added Hopper SM90 peak-throughput benchmarks for BF16, FP16, and FP8 dense MMA operations.
  * Updated FP8 performance estimates to reflect twice the BF16 and FP16 rates.


brandonfzhang:
Folding this one-line fix into the upcoming benchmark artifact refresh PR (so the refreshed H200 charts and the table change land together in one review). Closing in favor of that PR.

PR #585, opened 2026-08-13
Add LPT_L2 scheduler

yanzhuo607:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Summary by CodeRabbit

* **New Features**
  * Added configurable tile scheduling for supported attention workloads, including natural, LPT, and L2-aware LPT strategies.
  * Improved causal-attention scheduling by selecting L2-aware grouping when beneficial.
  * Added support for grouped-query attention and variable key/value sequence lengths across execution paths.
  * Extended scheduler support across supported SM100 and SM120 execution paths.

* **Documentation**
  * Clarified reduction and quantization swizzle documentation.

PR #574, opened 2026-08-13
frost(sdpa): has_lse specialization for the FP8/MXFP8 SM100 flavors

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran the repo formatter (black -l 160) on the changed files.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

`has_lse` specialization for the FROST SM100 FP8 (per-tensor) and MXFP8 SDPA forward flavors — `lse_optional` parity with what #512 did for the f16 flavors.

- **Kernels** (`prefill_d128_fp8_sm100.py`, `prefill_d128_mxfp8_sm100.py`): the LSE argument is `Optional[cute.Tensor]` end to end (`_kernel`/`_host`/correction-warp-group), the epilogue Stats store is guarded with `cutlass.const_expr(lse_tensor is not None)` (compiled out entirely when absent), and `compile()` grows `has_lse: bool = True` that builds a `None` fake LSE when `False`. The `amax_s`/`amax_o` atomicMax writes are independent of the LSE and unchanged.
- **Adapter** (`api_dsl.py`): `SdpaFwdDslSm100.compile()` keys `has_lse` on `sample_lse` for the FP8/MXFP8 branch; `execute()` drops the cached dense dummy-LSE fallback (a stats-less compile binds `None`), and the strict `lse_tensor` presence contract now applies to FP8 too (both directions — a requested LSE must be bound, an unrequested one is rejected). `_execute_fp8`/`_execute_mxfp8` pass `None` through.
- **Engine specs** (`engines.py`): `_sm100_fp8_spec` / `_sm100_mxfp8_spec` flip `lse_optional=True`. Every `lower_dsl_prefill` row is now `lse_optional`, so the engine-level dummy-LSE carve (`dummy_lse_bytes` sizing + the `carver.take` in `_execute`) is dead and removed. The SM80 row keeps the `False` default but lowers through `lower_sm80_prefill`, which never read the flag.

## Why

[...]

PR #656, opened 2026-08-19
frost(sdpa): port #585 LPT schedulers to the SM107 fp8 sibling; defaults must obey declared knob domains

### LINKED ISSUE #653: frost(sdpa): SM107 fp8 sibling missing the #585 LPT_L2 scheduler port — all causal fp8 declines on Rubin
## Symptom

On cc10.7 at develop tip, every causal/masked per-tensor FP8 SDPA graph fails plan build:

```
NotImplementedError: sdpa_fwd_prefill_sm100_d128_fp8: SCHED_LPT_L2 decode requires qh_per_kh and seqlen_kv at every call site
```

`test_sdpa_fwd_fp8_sm100.py -m 'L0 or L1'`: **29 failed / 4 passed** on a Rubin board (the 4 survivors are the nomask cases). The same suite is 33/33 on SM100.

## Cause

#585 added the LPT_L2 tile scheduler to `prefill_d128_fp8_sm100.py` (and the adapter's auto causal sched policy now selects it), but the hunk-symmetric port to the SM107 sibling `prefill_d128_fp8_sm107.py` was not done — the sibling carries only the decode guard, which correctly refuses. 27 vs 7 `SCHED_LPT|qh_per_kh` references between the two files.

This is the sibling-lockstep hazard (see #579's history): kernel-shared changes must land in both files, and nothing gates it when no Rubin silicon is in the merge loop.

## Fix

Port #585's scheduler hunks to the sibling (mechanical, hunk-symmetric), validate causal/BR/SWA fp8 on cc10.7. Interim mitigation if needed: pin `sched_policy=SCHED_NATURAL` on the Rubin route until the port lands.

vedaanta:
Fixes #653.

Two changes that are one rule: **a scheduler the served kernel cannot decode must be unreachable — whether requested or defaulted.**

## 1. The #585 port (the bug fix)

#585 landed the LPT/LPT_L2 tile schedulers in `prefill_d128_fp8_sm100.py`, but the hunk-symmetric port to the SM107 sibling was never done. The adapter's auto causal sched policy picks `SCHED_LPT_L2`, the sibling's decode guard refuses, and every causal/masked per-tensor FP8 graph on cc10.7 failed plan build (29 of 33 suite cases). Ported position-for-position with the SM100 file: `qh_per_kh`/`seqlen_kv` threaded through the five warp groups and appended at all ten scheduler decode call sites.

## 2. Domain honesty end to end

- Both fp8 kernel files declare `SUPPORTED_SCHED_POLICIES` — decode truth at the source.
- The adapter gains `_sm100_sched_domain()`; `check_support` now admits explicit NATURAL/LPT/LPT_L2 requests (was NATURAL-only — under-declared while the lowering auto-picked LPT), and the causal defaulting heuristic **clamps its pick into the served route's domain**, falling back to NATURAL. A default may never be a value the route could not honor as an explicit request — the invariant that turns this bug class into a clean fallback.
- The three SM100-family engine rows declare the true request domain `{NATURAL, LPT, LPT_L2}` (previously `{NATURAL}` — dishonest in both directions). SM120 rows untouched.
- `frost/README` knob-channel section documents the rule.

## 3. Sibling-lockstep fences (`test_sdpa_fp8_sibling_parity.py`)

Third silent-drift incident for this kernel pair (#574 signature skew, #602 amax ABI, #585/#653). The fences: source-scan tests (no GPU/DSL — run in any CI) assert every decode call site carries the LPT args and the sched-domain constants match; module tests assert `compile()` signatures in lockstep, kernel domains == adapter table, and CFG divergence is exactly the intentional Rubin set `{TILE_K_HW_BMM1, TILE_K_HW_BMM2, STAGES_KV}`. Negative-tested: removing one call site's args turns the fence red.

## Testing

- SM100 (cc10.0): 40 passed — 5 parity + 2 routing + 33 fp8 e2e (unchanged behavior; LPT paths exercised as before).
- **Rubin (cc10.7, w2u1g board): 33/33 fp8 e2e — the suite that fails 29/33 at develop tip — plus 7/7 parity+routing.** Causal/BR/SWA per-tensor FP8 now runs LPT-scheduled on sm107 silicon.

## Summary by CodeRabbit

* **New Features**
  * Expanded scheduler support for SM100 and SM107 FP8 attention routes, including natural, LPT, and L2-grouped LPT policies.
  * Improved causal scheduling fallback when a selected policy is unavailable.
  * Added SM107 FP8 configuration support for dense processing and grouped-query attention.

* **Bug Fixes**
  * Prevented invalid route-specific defaults from causing undecodable plans.

* **Tests**
  * Added parity checks to keep sibling FP8 routes, scheduling options, and configurations synchronized.

### DIFF
--- a/python/cudnn/frost/README.md
+++ b/python/cudnn/frost/README.md
@@ -559,6 +559,17 @@ If a kernel cannot run the requested scheduler policy, the answer is "this
 engine cannot serve this plan", not "ran with a different policy". A knob
 object of the wrong operation's type is rejected outright.
 
+**Defaults obey the same domains as requests.** When no value is requested,
+the adapter's defaulting policy (`_causal_sched_policy` and friends) picks
+one -- and that pick must lie inside the served route's declared domain,
+falling back to the universal default otherwise. A default the kernel cannot
+decode is a plan-build failure, not a preference: issue #653 (the LPT
+scheduler landed in one FP8 kernel sibling but the auto causal policy chose
+it for both) is exactly the failure this rule closes. The kernel files
+declare their decode domain (`SUPPORTED_SCHED_POLICIES`), the adapter's
+routing table mirrors it, the engine rows must not declare beyond it, and
+`test_sdpa_fp8_sibling_parity` pins the three in lockstep.
+
 Generic discoverability survives without the enum: knob domains are ordinary
 dataclass fields on `Capabilities`, so "list every engine and the knobs it
 honors" is a `dataclasses.fields()` walk over the spec table.
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -128,6 +128,21 @@ def _torch_stream_context(current_stream: Optional[cuda.CUstream], device: torch
 _SCHED_L2_BUDGET_BYTES = 50 * 1024 * 1024
 
 
+def _sm100_sched_domain(rubin: bool) -> frozenset:
+    """Tile-scheduler policies the routed kernel file can DECODE.
+
+    One set per route (the SM107 sibling is hunk-symmetric with the SM100
+    kernel since the #585 port); kept as a function so a future divergence is
+    a one-line change HERE plus its kernel constant — the sibling-parity test
+    asserts this table equals each file's SUPPORTED_SCHED_POLICIES. Both the
+    explicit-request gate (check_support) and the defaulting clamp (compile)
+    read it: a default may never be a value the route could not honor as an
+    explicit request.
+    """
+    del rubin  # same domain on every SM100-family route today
+    return frozenset({SCHED_NATURAL, SCHED_LPT, SCHED_LPT_L2})
+
+
 def _causal_sched_policy(s_kv: int, d_qk: int, d_v: int, elem_bytes: int) -> int:
     """SCHED_LPT_L2 vs SCHED_LPT for a causal graph (see _SCHED_L2_BUDGET_BYTES)."""
     one_head_bytes = int(s_kv) * (int(d_qk) + int(d_v)) * int(elem_bytes)
@@ -801,8 +816,8 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
         )
         self.flavor = _pick_flavor(d_qk, d_v)
         self._value_error_if(
-            self.sched_policy != SCHED_NATURAL,
-            f"SM100 DSL SDPA only supports sched_policy={SCHED_NATURAL}",
+            self.sched_policy not in _sm100_sched_domain(self._device_cc == (10, 7)),
+            f"SM100 DSL SDPA serves sched_policy in {sorted(_sm100_sched_domain(self._device_cc == (10, 7)))}; got {self.sched_policy}",
         )
         for requested, supported, name in (
             (self.tile_m, 128, "tile_m"),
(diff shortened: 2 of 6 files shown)


Anerudhan:
Needs conflict resolution/CI etc.

YangXu1990uiuc (approved):

Reviewed `f418fd8251f6` against base `c61a5e4eb5c5`.

The SM107 FP8 decode-argument port fixes the reported LPT_L2 compilation failure, independently reproduced on Rubin. The broader domain declaration still needs a THD restriction, and the source-only tests should avoid importing the optional DSL adapter. No P0 found.

**[P2] Restrict the new scheduler domain for THD plans** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f418fd8251f6aef9fd7ff8b6ed1330084708cc0a/python/cudnn/sdpa/fwd/engines.py#L440))

This row also serves THD, so the expanded domain now accepts explicit LPT/LPT_L2 requests for ragged plans. But _dispatch_decode_initial/_payload in _common_sm100.py return _thd_decode whenever _thd_on is true; neither requested dense scheduler runs. check_support and compile both succeed, silently ignoring the requested scheduling policy. Make the effective domain depend on THD as well as the route, and reject unsupported non-NATURAL requests in both the engine probe and standalone adapter unless the THD scheduler actually implements them. The parity test only checks the two dense FP8 siblings and misses this combination.

Evidence: SM100 direct-API probe with B=2,H=4,S=256,D=128: the baseline rejects THD policies 1/2, while the PR accepts and compiles both. Source tracing confirms both THD dispatch functions bypass _decode_initial/_payload unconditionally. Attribution: Newly reachable relative to branch point 8a758b87: this PR expands both the row domain and adapter gate while leaving the THD decoder unchanged. Develop has also broadened scheduler eligibility independently.

**[P2] Locate source files without importing the DSL adapter** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f418fd8251f6aef9fd7ff8b6ed1330084708cc0a/test/python/sdpa/frost/test_sdpa_fp8_sibling_parity.py#L37))

The first two tests intentionally have no requires_dsl marker, but _kernel_dir imports api_dsl, which imports APIBase and then cutlass. In a core-only environment these new L0 tests fail instead of remaining source-only; the other three tests correctly skip. Derive the kernel directory from the lightweight cudnn package location (or the test checkout) without importing the adapter, and update the sibling paths for the current kernel-directory layout when rebasing.

Evidence: Dependency-isolation run with CUTLASS unavailable: 2 failed, 3 skipped. Both failures trace through _kernel_dir -> api_dsl -> api_base -> import cutlass; with DSL installed all five tests pass. Attribution: Introduced by the new unmarked source tests and their location helper.

Validation: Built a matching extension from the reviewed head. SM100: 38 parity/FP8 tests passed. Rubin baseline 8a758b87: 29 failed, 4 passed, with the failures reporting missing qh_per_kh/seqlen_kv for SCHED_LPT_L2. Rubin head: 38 passed. Both runs explicitly selected L0 and L1. Checked all 10 initial/payload decode calls in the modified Rubin sibling and the corresponding SM100 calls. Direct-API probes covered dense/THD, causal/noncausal and NATURAL/LPT/LPT_L2, separately from the reference suite. Additional Rubin persistent-handoff probe: B=4,Hq=8,Hkv=2,S=2048,D=128, causal GQA with KV lengths [2048,1536,1024,768]. The 512-CTA grid exceeds the GPU's 208 SMs. Both LPT and LPT_L2 passed the FP32-dequant reference and produced bit-identical outputs; the actual compiled scheduler policy was verified.

Limitations: Current develop has substantial integration conflicts and per-flavor scheduler restrictions that this older branch predates. Preserve its distinction between an unspecified policy and explicit NATURAL; the PR head still substitutes a causal policy for explicit NATURAL, which also happens on its baseline and is already addressed in current develop. The tentative merge pairs the old SM107 d128 file with kernels/sm107/prefill_d192_d128_fp8.py. Reapply the port to the actual d128 route deliberately, preserve current per-flavor restrictions and THD default handling, and update test paths; I have not validated a resolved current-develop integration. The previously failing Rubin LPT_L2 path has no valid before-fix timing baseline. No end-to-end performance claim is made here. There is no pending fork test workflow or test pipeline in the PR thread; after resolving the conflicts, request fresh CI with a standalone `@cudnn-ci-bot run frost` comment.

PR #658, opened 2026-08-19
sdpa fwd: KV split for the SM100/SM120 prefill kernels

yanzhuo607:
Add an optional split over the KV sequence: with split_kv > 1 each Q tile's KV range is cut into contiguous chunks, every chunk runs as its own persistent tile, and the per-chunk (O, LSE) partials are reduced by kernels/split_combine_sm100.py. At split_kv == 1 the added closures fold away and the traced code is unchanged.

Flavors: sm100 d128 (f16/bf16, fp8, mxfp8), d192/128, d256 and d512, plus sm120 f16/bf16. The knob is gated per flavor so a flavor that does not thread it cannot silently accept it, and the config backstop rejects THD, attention sink and the flattened scheduler grids.

Also adds an optional cga1 cluster width (cta_mma=1) for d128 and d192/128. cga1 has no collective MMA to halve per-CTA K/V, so d128 recovers the extra SMEM through the existing Q/O alias and the fp8 family scales the KV stage depth with the cluster width.

Measured 3-5x end to end at S_q=128 over a 32K KV run on B200, with numerics unchanged against both the unsplit kernel and an fp32 reference.

Tests: test/python/sdpa/frost/test_sdpa_fwd_split_kv_sm100.py -- 56 cases over even and uneven splits (including empty ones), dense / causal / SWA / bottom-right / padded masks, GQA, bf16, fp8 and mxfp8 at both cluster widths, and the recombined LSE.

## Before submitting

## Summary by CodeRabbit

* **New Features**
  * Added configurable KV splitting for SDPA forward execution across supported SM100 and SM120 kernels.
  * Added stable recombination of split partial outputs, with optional LSE and output-amax results.
  * Enabled additional d128 and d192 cluster configurations and configurable CTA MMA widths.
  * Improved support for empty, masked, padded, uneven, causal, and grouped-query KV splits.

* **Bug Fixes**
  * Prevented invalid configurations and NaN results from empty or inactive splits.
  * Added validation for required LSE output, scheduling, masking, and memory constraints.

### DIFF
--- a/python/cudnn/sdpa/fwd/config_sm100.py
+++ b/python/cudnn/sdpa/fwd/config_sm100.py
@@ -88,10 +88,37 @@ class TemplateParams:
     seq_q_lens_present: bool = False
     sched_policy: int = SCHED_NATURAL
     thd_varlen: bool = False
+    # KV split: each Q tile's KV loop range is cut into ``split_kv`` contiguous
+    # chunks, each run as its own persistent tile writing a partial (O, LSE)
+    # that kernels/split_combine_sm100.py reduces.  1 = off (byte-identical
+    # codegen to the single-pass kernel).
+    split_kv: int = 1
+    # MMA cluster width: 2 = cga2 collective tcgen05.mma.cta_group::2 (a CTA
+    # pair share one MMA, each holding half of every K/V tile); 1 = cga1, one
+    # independent CTA per tile.
+    #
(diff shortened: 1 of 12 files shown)


yanzhuo607:
@cudnn-ci-bot run frost

vedaanta (approved):

vedaanta (approved):

PR #671, opened 2026-08-19
test(grouped_gemm): skip the GLU sf_fp8 override test off SM100

hwanseoc:
OSS CI fix.

`test_grouped_gemm_glu_bf16_rejects_sf_fp8_dtype_override` calls `grouped_gemm_glu_wrapper_sm100` without the SM100 guard every other test in the file carries, so on pre-Blackwell GPUs it fails with

```
RuntimeError: GroupedGemmGluSm100 requires SM100+, found SM80
```

from the wrapper's own arch check, before reaching the `ValueError` it means to assert. Adds the same skip the siblings use.

@coderabbitai ignore


hwanseoc:
@cudnn-ci-bot run oss

PR #672, opened 2026-08-19
test(dsa): compare only the effective top-k slice

hwanseoc:
OSS CI fix.

`check_ref_indexer_top_k` compared whole output rows. When `top_k` exceeds a row's sequence length, only `min(top_k, seq_len)` entries are actually picked — past that the reference pads with zeros while the kernel leaves its sentinel init, so the comparison was between two different paddings:

```
AssertionError: Tensor-likes are not equal!
Mismatched elements: 431 / 705 (61.1%)
Greatest absolute difference: inf at index (0,)
```

Caps both sides at the effective `k`, which is what the surrounding comment already described but the code did not do.

@coderabbitai ignore


hwanseoc:
@cudnn-ci-bot run oss

PR #675, opened 2026-08-20
sdpa fwd: pick and run a KV split automatically

yanzhuo607:
KV split was reachable only by building TemplateParams by hand: nothing chose a split count and the adapters could not execute one, so every graph ran with split_kv=1.

choose_split_kv picks the count. A CTA holds its tile for the whole KV loop, so a launch costs whole waves; it minimises, over powers of two,

    waves(s) = ceil(base_ctas * s / sm_count)
    cost(s)  = waves(s) * (ceil(kv_tiles / s) + CTA_COST)

An under-full launch splits until the wave is full; an over-full one with a partial-wave tail splits finer to smooth it; an exactly balanced one never splits. Powers of two only, since split_kv is a TemplateParams field and so a kernel-module cache key.

Both adapters derive the launch geometry from their own config and the SM count from cudnn._device, then thread the result into TemplateParams; the config backstop stays the authority, so a split a flavor will not accept falls back to
1. Running one is two launches: scratch_workspace_bytes sizes the split-major O and LSE partials, the kernel writes those instead of the caller's O, and split_combine_sm100 reduces them into O and into Stats when asked. The per-chunk LSE is compiled in whenever split_kv > 1, since it is the weight the combine reduces with.

Covered: SM100 half precision (d128/d192/d256/d512), SM100 FP8 and MXFP8, and SM120 half precision. The FP8 family splits only when it stores O in half precision -- the combine reduces in half, and reducing quantized partials would lose what the split is meant to be neutral about -- and its amax_o comes from the combine over the recombined O, since a per-split epilogue sees only its own partial and over-reports. THD, sink and SM120 FP8 keep split_kv=1; the last has no kernel support.

Adds split_kv to the knob vocabulary (SdpaFwdKnobs.split_kv, a Capabilities.split_kvs domain, its mismatch row), advertised by the seven engines whose lowering can honor it, plus tests for the chooser and for each adapter path.

## Before submitting

## Summary by CodeRabbit

- **New Features**
  - Added automatic KV splitting for supported SDPA forward workloads on SM100 and SM120.
  - Added support for FP8 and MXFP8 execution, including recombined outputs, LSE values, and amax reporting.
  - Added automatic workspace sizing and support for caller-provided workspace buffers.
  - Added validation for split configuration compatibility.

- **Tests**
  - Added coverage for split selection, workspace modes, numerical accuracy, and hardware-specific behavior.

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -29,6 +29,29 @@ from cudnn.frost.tile_dsl.constants import (
     SCHED_NATURAL,
 )
 from cudnn.sdpa.fwd.config_sm100 import TemplateParams as Sm100TemplateParams
+from cudnn.sdpa.fwd.config_sm100 import make_cfg_d128, make_cfg_d192, make_cfg_d256, make_cfg_d512
+from cudnn.sdpa.fwd.heuristics import choose_split_kv
+from cudnn._device import device_info
+
+# Flavor (max D_QK, max D_V) -> its make_cfg, for the split decision. The config
+# backstop rejects what a flavor will not honor, which _decide_split_kv treats
+# as "no split", so this only has to be keyed the same way as
+# _SM100_KERNEL_FILES.
+_SM100_MAKE_CFG = {
+    (128, 128): make_cfg_d128,
+    (192, 128): make_cfg_d192,
+    (256, 256): make_cfg_d256,
+    (512, 512): make_cfg_d512,
+}
+
+
+def _split_combine_module():
+    """The KV-split reduction pass; imported lazily like the kernel templates."""
+    from cudnn.sdpa.fwd.kernels import split_combine_sm100
+
+    return split_combine_sm100
+
+
 from cudnn.sdpa.fwd.config_sm120 import (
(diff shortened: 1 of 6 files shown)


yanzhuo607:
@cudnn-ci-bot run frost

vedaanta (dismissed):

vedaanta (approved):

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
Heads-up: this PR's split-KV work has been folded into #692 as commit `fdd78da0d` with your authorship preserved — the `choose_split_kv` cost model, the FP8/MXFP8 split + recombined-amax machinery, the SM120 wiring, and the test suites all carried over. The one architectural change: the chooser now runs in the heuristic's `propose()` and the value travels as the explicit `split_kv` knob on the ranked plan (honored-or-ineligible), rather than the adapter deciding internally — that's the knob-set design #692 establishes, and it makes every split plan name-addressable and autotune-replayable. The conflicts this branch had against develop (from #648/#682/#689) don't need resolving anymore. Please review the fold in #692 — happy to adjust anything that got lost in translation.

vedaanta:
closing in favour of 692

Issue #604, opened 2026-08-15
frost(sdpa): SM80 _compile_cached keys the THD compile on the packed token totals (per-step recompile)


vedaanta:
## Context

The SM80 (A100) FROST SDPA engines (#493) compile through `_compile_cached` in `python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py` — an `lru_cache` whose key includes `SQ` and `SKV`. Under `THD_VARLEN` those are the **packed token totals**, read out of the seq-length tensors at execute time.

That is the same defect #543 fixed on SM100/SM120 (the recompile leg of #552): under continuous batching the packed totals change every step, so the cache never hits and each step pays a fresh multi-second `cute.compile`. AGENTS.md Hard Rule 4 (added in #543) now bans runtime-data compile keys and names this site as the known open cleanup.

## Proposal

Migrate the SM80 THD compile to the same shape as the SM100/SM120 fix in #543:

- token extents as `cute.sym_int` symbols in the fake tensors (one symbol per ragged group), so one compiled artifact re-binds any packed total;
- runtime launch scalars (grid extents, `max_s_q`) as `cutlass.Int32` call arguments, never compile parameters — note `sdpa_fwd_wrapper_sm80` already takes `max_s_q` from the caller;
- derived key components audited for leaks (e.g. strides computed from totals), per Rule 4;
- compile moved to plan time, with a cache-miss regression test mirroring `test_dsl_sm{100,120}_thd_compile_key_plan_time_only` (two executes with different packed totals, zero new `cute.compile` calls, numerics checked per total).

Also worth sweeping while in there: the SM80 packed-THD **backward** does `cu_seqlens_{q,k}.to(dtype=..., device="cpu")` (`sdpa/bwd/kernels/bprop_f16_sm80.py`) — listed in Rule 3's known violations; it is the first thing to fix if the `sdpa_bwd_sm80` spec ever gains THD.

## Related

- #543 — the SM100/SM120 fix this mirrors (dynamic token extents, plan-time-only compile keys).
- #552 — the umbrella THD execute host-round-trip issue (D2H sync / capture legs still open).
- #493 — where the SM80 engines landed.

PR #682, opened 2026-08-20
SDPA: drop the legacy standalone d=256 fwd/bwd stacks; port SM80 forward to the SdpaFwdDsl adapter path

vedaanta:
Two-part cleanup of the SDPA-forward OSS stack: delete the legacy standalone d=256 SM100 implementations (both passes), and move SM80 onto the same `SdpaFwdDsl` adapter path SM100/SM120 use, so one lowering function drives every forward cell.

## Commit 1 — remove the legacy standalone d=256 SM100 fwd/bwd stacks (−9,220 LOC)

- Delete `fmha_forward_sm100_d256.py` and the 3-kernel backward family (`fmha_backward_sm100_2kernel.py`, `fmha_dq_d256_sm100.py`, `fmha_dkdv_d256_sm100.py`).
- Remove `SdpafwdSm100D256` / `SdpabwdSm100D256`, their wrappers, every lazy export, the docs pages, and the wrapper tests.
- Experimental torch op: drop the `cudnn::sdpa_{fwd,bwd}_d256` custom ops and the pre-9.23-backend OSS routing — d=256 always takes the backend graph path now (the op tests skip d=256 on backends < 9.23 instead of exercising the OSS route).
- The graph-dispatched `sdpa_fwd_prefill_sm100_d256` FROST engine is unaffected and remains the OSS forward implementation for this cell.

**Coverage note:** the standalone SM100 OSS backward (never reachable from `graph.sdpa_backward()`) goes away with nothing OSS replacing it until `sdpa/bwd` grows an SM100 `ENGINE_SPECS` row (tracked separately).

## Commit 2 — port the SM80 forward onto the SdpaFwdDsl adapter path

`fwd/api.py` is deleted; SM80 now has the same shape as SM100/SM120:

- New `SdpaFwdDslSm80` adapter in `api_dsl.py` implementing the `SdpaFwdDsl` contract (descriptor-level `check_support` with flavor pick / mask / scheduler resolution, no-op `compile` — the kernels self-cache, execute that binds caller buffers). SM80-only features (bias, ALiBi, block_mask, RoPE, score stats) are extra optional `execute` keywords, as the contract permits.
- `_sm80_spec` lowers via `partial(lower_dsl_prefill, api_type=_SM80)`; the row now declares `dense_seq_q_trim` (the kernels are plumbed) and `lse_optional`. `lower_dsl_prefill` forwards the SM80 feature operands only to adapters that declare the keywords.
- Both SM80 kernels gain optional `out_o/out_lse/out_score_*` output binding; outputs that a feature path never writes (seq_len_q trim, zero-length padded batches, block-masked rows) keep their zero-init semantics via a conditional zero-fill. The engine path loses its per-execute output copy-backs.
- `_torch_stream_context` gains the NGC `ExternalStream(0/1/2)` silent-no-op workaround previously present only in the deleted `api.py` copy — THD execute paths now carry the fix too.
- `AGENTS.md` Rule-3 known-violation list updated: the ragged cache-key `.max().item()` sites died with the legacy wrappers.

**Deliberately unchanged:** dense GQA keeps the adapter-side K/V head expansion until the kernels' native dense-GQA path is qualified on A100 (see `graph_analyzer.expand_gqa_heads`); host-side V padding for off-flavor head dims stays; the wrapper THD path's compile-key issue (#604) is untouched.

**Breaking (experimental API):** the `SdpafwdSm100D256` / `SdpabwdSm100D256` classes and wrappers are removed; `SdpafwdSm80` is replaced by `SdpaFwdDslSm80` (constructor-compatible for the common kwargs). `sdpa_fwd_wrapper_sm80` keeps its public signature.

## Validation

On SM100 (B200, backend 9.24, torch 2.14 nightly):
- `test_cudnn_sdpa_op.py` basic/causal/varlen: 5/5 passed (d=256 through the backend path)
- `sdpa/frost/test_sdpa_frontend_integration.py`: 10 passed, 1 skipped
- `sdpa/frost/test_sdpa_fwd_dsl_sm100.py` + `test_sdpa_execute_is_async.py`: 470 passed
- import smoke: all 11 `ENGINE_SPECS`, manifest, exports; legacy names gone

SM80 validation on A100 in progress (results to follow in a comment); the `fe_api/sdpa/test_sdpa_{fwd,bwd}_sm80.py` suites are the gate.

## Summary by CodeRabbit

* **New Features**
  * Added an updated SM80 SDPA forward implementation for dense and packed inputs, supporting biases, RoPE, sinks, variable-length sequences, and LSE outputs.
  * Added support for caller-provided output and LSE buffers.

* **Changes**
  * SDPA operations with head dimension 256 now use the cuDNN backend directly.
  * Removed experimental SM100 D=256 SDPA APIs and documentation.
  * SM80 SDPA no longer supports ALiBi, block masks, or score-statistics outputs.

* **Tests**
  * Updated coverage for the revised SM80 implementation and backend requirements.


vedaanta:
## A100 validation complete

Ran on an A100-PCIE-40GB (the board the SM80 flavor table was swept on), cuDNN backend 9.24, torch 2.14 nightly cu132:

| Silicon | Suite | Result |
|---|---|---|
| **A100** (SM80) | `fe_api/sdpa/test_sdpa_{fwd,bwd}_sm80.py`, **all test levels** | **105 passed** |
| **A100** | graph-path end-to-end: `graph.sdpa()` → ranked plans → `sdpa_fwd_prefill_sm80` pinned → `lower_dsl_prefill` → `SdpaFwdDslSm80` → kernel with bound outputs, vs fp32 torch reference | dense causal + Stats (×2 executions), GQA h_kv=2, padded-KV + Stats (the zero-fill branch) — all matched; workspace size 0 as designed |
| B200 (SM100) | `test_cudnn_sdpa_op.py` basic/causal/varlen (incl. d=256 via the backend path) | 5 passed |
| B200 | `sdpa/frost/test_sdpa_frontend_integration.py` | 10 passed, 1 skipped |
| B200 | `sdpa/frost/test_sdpa_fwd_dsl_sm100.py` + `test_sdpa_execute_is_async.py` | 470 passed |
| H100 (SM90) | sm80 suites collect & skip cleanly on non-SM80 | 9 skipped, 0 errors |

The graph-path check caught one real gap before it ran: the graph API declares Stats/score outputs as `(B, H, S, 1)` while the kernels write `[B, H, SQ]` — fixed in cb4be77f4 (`squeeze(-1)` view rebind in `SdpaFwdDslSm80.execute`, zero-cost).

vedaanta:
## Scope trim: ALiBi / block_mask / score-stats dropped from the SM80 FROST path (e9ebf02d1)

Per review direction, the SM80 row no longer serves ALiBi, `block_mask`, or the `score_max`/`score_sum_exp` side outputs. Graphs requesting them decline at the capability row and are served by the cuDNN backend — the same mechanism as the other features the FROST rows deliberately don't carry (`dropout`, `paged_kv`, …). The `Capabilities` fields stay: they are the decline gates `mismatch()` compares every graph against.

−632 / +42 lines: capability flags, `lower_dsl_prefill` operand plumbing (bias remains the one SM80 extra), the adapter/wrapper/THD keywords, and every `const_expr`-gated device block in both kernels (~−520 lines there; `RESCALE_THRESHOLD` collapses to its flag-off value, the block-mask hybrid guard to the unconditional dense form — surviving traces byte-identical).

Re-validated on the A100:
- `fe_api/sdpa/test_sdpa_{fwd,bwd}_sm80.py`, all levels: **105/105 passed** (unchanged — the suites had no coverage of the removed features)
- graph-path end-to-end (dense causal + Stats ×2, GQA, padded KV): all matched the torch reference
- new negative check: an ALiBi graph lists **no** FROST entry and runs on the backend
- SM100 frontend integration re-run: 10 passed

vedaanta:
## Backward gets the same trim: ALiBi / block_mask dropped, dBias kept (e238518d1)

Mirrors the forward-row change: `sdpa_bwd_sm80` no longer declares `alibi`/`block_mask`, so such graphs decline at the capability row and the cuDNN backend serves them. **dBias stays fully served** — the bprop kernel's dbias code is byte-unchanged (verified: 50 dbias/has_bias references before and after).

−128/+15 lines across the capability row, `lower_sm80_bwd` binding, `SdpabwdSm80.execute` / `sdpa_bwd_wrapper_sm80` / `_thd_backward` / the d64 fast-path gate, and every `const_expr`-gated alibi/block-mask device block in `bprop_f16_sm80.py` (all pure block deletions — this kernel's gates needed no collapse rewrites).

Re-validated on the A100: `fe_api/sdpa/test_sdpa_{fwd,bwd}_sm80.py`, all levels — **105/105 passed**; graph-path end-to-end + the ALiBi negative-routing check pass.

Feature surface of the SM80 FROST rows after this pair of commits: fwd = masks/SWA/BR/padded/sink/stats/**bias**; bwd = the same plus **dBias**/dSink/deterministic. ALiBi, block_mask, and the score-stat outputs route to the backend on both passes.

vedaanta:
@cudnn-ci-bot run frost

test/python/test_cudnn_sdpa_op.py
```diff
@@ -151,20 +147,19 @@ def sdpa_reference_fwd_bwd(
     )
 
 
+# cuDNN 9.23.0 added d=256 SDPA fprop and bprop support in the backend; the
```
egilliam-nv, line 150: Now that only the backend supports D256 (and only in >= 9.23), should this knowledge instead be in `scaled_dot_product_attention.h` or `sdpa_support_surface.h`(to reject if D=256 and version < 9.23)? (Probably more likely in `scaled_dot_product_attention.h` if we don't always know D when sdpa_support_surface.h is called)

python/cudnn/sdpa/__init__.py
```diff
@@ -14,11 +14,7 @@
 from typing import Any
 
 _LAZY_EXPORTS = {
-    "SdpabwdSm100D256": (".bwd", "SdpabwdSm100D256"),
-    "sdpa_bwd_wrapper_sm100_d256": (".bwd", "sdpa_bwd_wrapper_sm100_d256"),
-    "SdpafwdSm100D256": (".fwd", "SdpafwdSm100D256"),
-    "sdpa_fwd_wrapper_sm100_d256": (".fwd", "sdpa_fwd_wrapper_sm100_d256"),
-    "SdpafwdSm80": (".fwd", "SdpafwdSm80"),
+    "SdpaFwdDslSm80": (".fwd", "SdpaFwdDslSm80"),
```
egilliam-nv: I don't quite understand this file :) but is there a reason Sm100, etc. aren't also in this list?

python/cudnn/sdpa/bwd/engines.py
```diff
@@ -479,8 +479,8 @@ def _sm80_spec() -> EngineSpec:
             dbias=True,
             dsink=True,
             bias=True,
-            alibi=True,
-            block_mask=True,
+            # ALiBi / block_mask graphs are deliberately NOT served — the row
```
egilliam-nv: This comment is probably not necessary, since it's just documenting a historical artifact and no other engine has it.

python/cudnn/sdpa/fwd/kernels/prefill_d256_f16_sm80.py
egilliam-nv, line 1: Are we still keeping the d256 sm80 kernel?

python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py
```diff
@@ -1117,21 +1040,10 @@ def _imin_i32(a, b):  # min(a, b) for Int32 (IMNMX) — bias OOB clamp
...
-            bm_kvblk = (kv_iter * cutlass.Int32(tile_n)) // cutlass.Int32(128)
-            bm_byte = bm_pp[bm_row_base + (bm_kvblk // cutlass.Int32(8))]
-            bm_bit = bm_kvblk % cutlass.Int32(8)
-            block_on = ((bm_byte.to(cutlass.Int32) >> bm_bit) & cutlass.Int32(1)) != cutlass.Int32(0)
-        else:
-            block_on = True  # Python bool → unconditional
+        # ``block_on`` is a Python ``True`` → ``if block_on:`` below is a
+        # pure-Python if (traces the body unconditionally, dense path is
+        # byte-identical).
+        block_on = True  # Python bool → unconditional
```
egilliam-nv: Can we just eliminate this variable (which is a historical artifact)?

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -2791,3 +2799,640 @@ def sdpa_fwd_wrapper_dsl_sm120(
...
+        every key component is plan-time data (dense B/H/S/D from the frozen
+        graph), so the cache key is Rule-4 clean; the THD wrapper path's
+        packed-total keys are the known issue #604 cleanup.
+        """
+        self._logger.debug("Entering compile (no-op — kernel self-caches)")
+        self._ensure_support_checked()
+        self._compiled_kernel = True
+        self._logger.debug("compile completed")
+
+    def scratch_workspace_bytes(self) -> int:
```
egilliam-nv, line 3151: **Per-execute allocations behind `scratch_workspace_bytes() == 0`** (the issue #514 residue — no regression, everything here is grandfathered, but the new adapter shape makes carving these trivial):

- dense GQA K/V head expansion (`repeat_interleave`) — the largest, two `B*S*H_q*D` device allocations per execute;
- host-side V padding for off-flavor head dims;
- the dense_flex O scratch in `_to_bshd_writable`;
- the LSE fallback when the caller's stats buffer is non-contiguous (kernel-side alloc + copy-back);
- and on **every stats-less execute**, a full `B*H*SQ` fp32 LSE the kernel allocates because execute hardwires `return_lse=True` — the row now declares `lse_optional`, but unlike the SM100/SM120 kernels the SM80 kernels don't compile the LSE store out, so `lse_optional` here means "allocated and thrown away" rather than "no buffer".

I have the workspace-carving half of #514 in flight on a branch (backward done, forward previously done against the old `api.py` layout) — I'll rebase it onto this PR's architecture once it merges and route all of the above through `scratch_workspace_bytes()`/`lower_dsl_prefill`'s existing carver plumbing, so no action needed here unless you'd rather fold it in.

python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py
```diff
@@ -2241,15 +2034,28 @@ def forward(
...
-    LSE = torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
-    # SCORE_MAX / SCORE_SUM [B,H,SQ] (or 1-elem dummies when not requested).
-    if return_score_stats:
-        SCORE_MAX = torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
-        SCORE_SUM = torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
+    # Caller-bound out_* buffers skip the allocation.  Rows a feature path
+    # never touches (seq_len_q trim, zero-length padded batches) rely on
+    # zero-init, so zero a bound buffer only when such a path is active —
+    # the plain / causal / SWA dense paths write every valid element.
+    _needs_zero_init = (seq_len_q is not None) or (seq_kv_lens is not None)
```
egilliam-nv, line 2027: Nit on the comment: per the store paths, none of these features actually *rely* on zero-init — the dense epilogue stores every in-bounds row unconditionally (the empty-loop clamp routes zero-KV-iteration tiles through the store with `row_sum=0`), the LSE/score stores predicate only on `sq_store_bound` (physical `SQ` for dense — trimmed rows are written explicitly with O=0/LSE=-inf), and THD (the one skipped-row case) can't receive bound outputs. So this gate is defensively redundant rather than load-bearing — which is the safe side of wrong and fine to keep, but worth rewording as *defensive* so a later cleanup doesn't 'optimize' a real dependency into existence (I chased SWA/BR empty-window rows and zero-length `seq_kv_lens` batches specifically; all stored).

egilliam-nv (commented):
Reviewed with a focus on the direct output-binding correctness (dirty caller buffers vs. the old fresh-`torch.zeros` guarantee) — it holds up everywhere I could attack it: empty-window SWA/causal-BR rows, uneven tails, padded-Q trim, zero-length KV batches, block-masked tiles, and the `(B,H,S,1)` stats squeeze (always a valid view; non-contiguous #304-randomized stats correctly fall back to alloc + strided `copy_`). The `ExternalStream(0/1/2)` guard migration preserves the original fix's semantics and usefully extends it to the THD paths, and the feature-drop commits keep the capability gates while removing code cleanly — I found no dangling references.

Two inline comments: the per-execute-allocation inventory still behind `scratch_workspace_bytes() == 0` (I'll pick those up when I rebase my #514 carving branch onto this), and a doc nit on `_needs_zero_init`. One PR-body suggestion: worth stating the expected mhas FROST-routing drop (ALiBi / block_mask / score-stat graphs now serve native) so the next person diffing routing stats knows it's intentional.

python/cudnn/sdpa/__init__.py
vedaanta: yeah, lemme remove all.

vedaanta:
## Internal-CI mirror failures root-caused: test under-declaration, fixed in 2afc937e0

The `frost:rel:sdpa:sm80` failures on the internal mirror of this PR (job 405391124: `test_{fwd,bwd}_engine_end_to_end`, 97.5% / 16% mismatch) are **not an engine regression** — the SM80 engine's numerics are exact. Root cause, reproduced and bisected on an A100:

1. The integration test's `_build_fwd_graph` never calls `set_dim/set_stride` on O, so backend layout inference declares O **BHSD-contiguous** — while the test reads its buffer through BSHD-physical torch strides.
2. Under the execute contract, a variant-pack entry is raw storage laid out per the IR descriptor (`lower_dsl_prefill`'s `_ir_view`; the C++ backend only ever sees raw pointers). This PR's SM80 adapter honors that contract — the old standalone-`api.py` lowering it replaces honored the caller's torch strides instead, which masked the test's under-declaration.
3. Fingerprint: reading the "wrong" output buffer under BHSD-contiguous strides matches the torch reference to fp16 rounding (max 9.8e-4), and LSE is exact (1.4e-6) — the kernel computed everything correctly and wrote exactly the declared layout.
4. Control: pinning the **native backend** plan (`eng8_…`) on the same undeclared graph scrambles identically (97.8%, same 4.53 max diff) — the test as written fails against the closed backend too.
5. `test_mhas_v2` in the same CI job (declared strides, 322 graphs FROST-routed to `sdpa_fwd_prefill_sm80`) was already green.

Commit 2afc937e0 brings the test file into this PR with the one-line fix (declare O's dim/stride, mirroring the SM100 sibling test); the whole file passes **8/8** on A100-PCIE-40GB against this PR's engine.

python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py
vedaanta: Done in 7c009d282 — `block_on` and its if-guards are gone entirely, along with the QK sites' orphaned `else` branches (the fully-masked-block dead path), which were the last block_mask remnant in the mainloop. Re-validated on A100: fe_api sm80 fwd+bwd (all levels) + the SM80 integration suite, 113/113 passed.

python/cudnn/sdpa/bwd/engines.py
vedaanta: Done in 7c009d282 — dropped the annotation from both the bwd and fwd `_sm80_spec` rows so the capability tables stay comment-free like every other engine. The user-facing unsupported-features note stays only in the public wrapper/adapter docstrings.

vedaanta:
@cudnn-ci-bot run frost

python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py
```diff
@@ -2241,12 +2248,37 @@ def forward(
...
     if return_score_stats:
-        SCORE_MAX = torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
-        SCORE_SUM = torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
+        SCORE_MAX = out_score_max if out_score_max is not None else torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
+        SCORE_SUM = out_score_sum if out_score_sum is not None else torch.zeros(B, H, SQ, dtype=torch.float32, device=Q.device)
+        if _needs_zero_init:
+            if out_score_max is not None:
+                SCORE_MAX.zero_()
+            if out_score_sum is not None:
+                SCORE_SUM.zero_()
```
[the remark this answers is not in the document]
vedaanta: Great catch — and the real bug was one layer deeper than zero-init. The V tile loads passed the shared `valid_cols` (= runtime **Q/K** head dim) to `load_tile_2d`, zero-filling V columns `[D, d_v)` for any `d_qk < d_v` envelope graph — so O's tail was computed as P·0 *in registers* (the O-store's matching column trim hid the same range). Verified on A100: (96,128) returned zeros in cols [96,128) where the reference has absmax 3.06 — i.e. this was a pre-existing numerics bug even with the old zeros-allocated output, not just uninitialized memory. Fixed in f5115a93c: V loads use `valid_cols=None` (V is always exactly the compile-time d_v, asserted), and the O store now trims rows only. (96,128) now matches torch full-width; 113/113 on A100. `_needs_zero_init` intentionally does **not** grow an `is_even_k` term — with the load/store fixes the full row×d_v range is always written.

python/cudnn/sdpa/fwd/kernels/prefill_f16_sm80.py
vedaanta: Reworded in f5115a93c as defensive-only, quoting your store-path chase (unconditional in-bounds row stores, zero-KV-iteration clamp, explicit trimmed-row writes, THD never bound). Same commit also removes the epilogue's `not is_even_k` **column** trim — it was predicated on the Q/K head dim while O is d_v wide, which turned out to be a real pre-existing bug for d_qk < d_v envelope graphs (see the thread above): the columns weren't skipped-but-zeroed, they were computed as P·0 because the V loads used the same shared valid_cols. Both fixed; rows remain the only store predicate.

python/cudnn/sdpa/fwd/engines.py
```diff
@@ -925,6 +878,18 @@ def _execute(variant_pack, workspace=None, stream=None):
...
+            if feature_ops.bias is not None and "bias_tensor" in _extra_exec_keys:
+                execute_kwargs["bias_tensor"] = feature_ops.bias
+            if feature_ops.alibi and "alibi" in _extra_exec_keys:
+                execute_kwargs["alibi"] = True
+            if feature_ops.block_mask is not None and "block_mask" in _extra_exec_keys:
+                execute_kwargs["block_mask"] = feature_ops.block_mask
+            if binding.score_max is not None and "score_max_tensor" in _extra_exec_keys:
+                execute_kwargs["score_max_tensor"] = resolved.get(id(binding.score_max))
+            if binding.score_sum_exp is not None and "score_sum_tensor" in _extra_exec_keys:
+                execute_kwargs["score_sum_tensor"] = resolved.get(id(binding.score_sum_exp))
```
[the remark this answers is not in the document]
vedaanta: Stale — the flagged score-stat forwarding was removed entirely in e9ebf02d1: the SM80 rows no longer serve score_max/score_sum_exp (such graphs decline at the capability row and the backend serves them), and lower_dsl_prefill's extras now forward only `bias_tensor`. A graph declaring a score-stat output can no longer reach a FROST engine.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -2791,3 +2799,691 @@ def sdpa_fwd_wrapper_dsl_sm120(
...
+            )
+
+        b, h_qo, s_qo, d_qk = self.q_desc.shape
+        _, h_kv, s_kv, _ = self.k_desc.shape
+        _, _, _, d_v = self.v_desc.shape
+
+        self._check_tensor_shape(self.q_desc, (b, h_qo, s_qo, d_qk), name="Q")
+        self._check_tensor_shape(self.k_desc, (b, h_kv, s_kv, d_qk), name="K")
+        self._check_tensor_shape(self.v_desc, (b, h_kv, s_kv, d_v), name="V")
+        self._check_tensor_shape(self.o_desc, (b, h_qo, s_qo, d_v), name="O")
```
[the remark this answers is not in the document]
vedaanta: Done in e8f2036ce — `SdpaFwdDslSm80.check_support` now applies the same `dense_layout_ok` gate as the SM100/SM120 adapters (identical message), so broadcast/overlapping/non-innermost-D layouts decline with a typed error on the direct and wrapper paths too.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -2791,3 +2799,691 @@ def sdpa_fwd_wrapper_dsl_sm120(
...
+                d_v=self.flavor_d_v,
+                mask=self.mask_token,
+                swa_window=int(self.swa_window_runtime),
+                right_bound=int(self.right_bound_runtime),
+                causal_bottom_right=self.causal_bottom_right,
+                seq_kv_lens=seq_kv_lens.reshape(-1) if seq_kv_lens is not None else None,
+                seq_len_q=seq_q_lens.reshape(-1) if seq_q_lens is not None else None,
+                bias=bias_tensor,
+                alibi_slopes=alibi_slopes,
+                sinks=sinks.reshape(-1) if sinks is not None else None,
```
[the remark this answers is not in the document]
vedaanta: Done in e8f2036ce — seq lens bind via `_checked_seq_lens` and sinks via `_checked_sinks_1d`, same as the siblings; an int64/wrong-shape caller tensor now fails fast instead of triggering the kernel-host cast (which becomes a no-op view for contract-conforming int32 input).

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -2791,3 +2799,691 @@ def sdpa_fwd_wrapper_dsl_sm120(
...
+    wl, wr = window_size
+    if is_causal and wl >= 0:
+        mask_token, swa = "causal_swa", wl
+    elif is_causal:
+        mask_token, swa = "causal", 0
+    elif wl >= 0:
+        mask_token, swa = "swa", wl
+    else:
+        mask_token, swa = "none", 0
+    right_bound = wr if (is_causal and wr is not None and wr > 0) else 0
```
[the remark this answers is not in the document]
vedaanta: Done in e8f2036ce — the wrapper's THD branch now raises the same anchored-BR ValueError the dense check_support raises, before reaching the kernel assert.

python/cudnn/sdpa/fwd/engines.py
```diff
@@ -585,79 +588,19 @@ def _sm80_spec() -> EngineSpec:
             stats=True,
             padded_stats=True,
             decode=False,
+            # The kernels implement the dense padded-Q trim natively
+            # (per-batch ``seq_len_q`` forward kwarg): rows >= seq_len_q[b]
+            # rely on the zero-initialized output (O := 0 / LSE := 0).
+            dense_seq_q_trim=True,
```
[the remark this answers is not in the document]
vedaanta: Fixed in e8f2036ce — the comment now says what the kernel does: padded rows are written explicitly (O := 0, LSE := -inf); nothing relies on zero-init.

python/cudnn/sdpa/bwd/api.py
```diff
@@ -1123,18 +513,19 @@ def sdpa_bwd_wrapper_sm80(
 
     Returns ``TupleDict(dq_tensor=..., dk_tensor=..., dv_tensor=...
     [, dbias_tensor=...])`` (BHSD grads; dBias head-major [., H, SQ, SKV]).
+    ALiBi and block_mask are not supported (use the graph API, which routes
+    them to the cuDNN backend); bias/dBias remain fully served.
```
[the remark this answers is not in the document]
vedaanta: Done in e8f2036ce — the Returns doc now lists both optional outputs and the stable order (dq, dk, dv, dbias, dsink).

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -2791,3 +2799,700 @@ def sdpa_fwd_wrapper_dsl_sm120(
...
+            raise RuntimeError("SdpaFwdDslSm80 is not compiled")
+
+        # Graph Stats / score declarations arrive as (B, H, S, 1); the kernels
+        # write [B, H, SQ] — rebind the squeezed view (zero-cost, same storage).
+        if lse_tensor is not None and lse_tensor.ndim == 4:
+            lse_tensor = lse_tensor.squeeze(-1)
+        if score_max_tensor is not None and score_max_tensor.ndim == 4:
+            score_max_tensor = score_max_tensor.squeeze(-1)
+        if score_sum_tensor is not None and score_sum_tensor.ndim == 4:
+            score_sum_tensor = score_sum_tensor.squeeze(-1)
```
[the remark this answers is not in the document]
vedaanta: Not applicable: squeeze(-1) is a pure metadata view — it launches no kernel and touches no allocator, so there is no work to order on the launch stream (Hard Rule 5 covers enqueued torch ops and allocator calls). Every stream-ordered operation in this execute path already runs inside _torch_stream_context. (The score squeezes no longer exist — score stats were dropped from the SM80 rows in e9ebf02d1.)

python/cudnn/sdpa/__init__.py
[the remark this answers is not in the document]
vedaanta: Resolved the other way in e8f2036ce, per the target design: cudnn.sdpa now exports nothing and is reserved for the coming arch-agnostic sdpa_fwd_wrapper / sdpa_bwd_wrapper entry points (one wrapper that resolves the adapter from the device, like the other FE OSS families). The per-arch adapters/wrappers — the pinning/benchmark tier — are imported from cudnn.sdpa.fwd / cudnn.sdpa.bwd directly; tests and the sm120 doc example (which referenced a name this level never actually exported) are updated. The module docstring documents the layering.

python/cudnn/sdpa/fwd/kernels/prefill_d256_f16_sm80.py
vedaanta: Yes — this one stays. Despite the name collision with the deleted SM100 stacks, prefill_d256_f16_sm80.py is the A100 qwen-flavor kernel (the (256,256) envelope with the symmetric K+V prefetch schedule, which regresses at d<=192 — hence the separate file). It is the only way the SM80 row serves d in (192, 256], so it is load-bearing for the sdpa_fwd_prefill_sm80 engine, unrelated to the removed standalone SM100 d256 forward/backward.

test/python/test_cudnn_sdpa_op.py
vedaanta: Partially agree — the knowledge belongs below the test, but node-level validation is the one place it must NOT go: scaled_dot_product_attention.h pre-validation runs before engine ranking, so a d=256-needs-9.23 gate there would also veto the FROST sdpa_fwd_prefill_sm100_d256 engine, which serves d=256 forward on any backend version (that OSS-fills-the-gap coverage is exactly why the deleted custom-op fallback existed for <9.23). The honest per-engine story already works: pre-9.23 backends simply offer no d=256 plans, the walk declines through to FROST (or reports no plans), and the test skip is only about test intent. If we want a friendlier early error for backend-only users, the right hook is the backend-scoped support surface (the AttentionImplementation selection path in sdpa_support_surface.h, where the decline applies to backend plans without shadowing python engines) — happy to do that as a follow-up PR; head dim is available there via the node's tensor dims at that call site.

python/cudnn/sdpa/fwd/api_dsl.py
vedaanta: Agreed on all five, and deliberately left as-is here: everything on that list is pre-existing behavior carried through the port (the adapter shape is what makes carving them trivial now, which was part of the point). Since you have the workspace-carving half of #514 in flight, this PR stays out of its way. One note for that work: the stats-less LSE waste wants a different fix than carving — the real cleanup is compiling the LSE store out like the SM100/SM120 kernels do, which lands with the SM80 TemplateParams conversion (tracked follow-up); carving the buffer would keep paying the bandwidth, just from a different allocation. The GQA repeat_interleave is also on its own track: it disappears entirely once the kernels' native dense-GQA path is qualified on A100 (see graph_analyzer.expand_gqa_heads), rather than being carved.

vedaanta:
@cudnn-ci-bot run frost

test/python/test_cudnn_sdpa_op.py
egilliam-nv: I'm okay leaving this as is without any change in sdpa_support_surface.h. Resolved!

egilliam-nv (approved):

PR #690, opened 2026-08-21
sdpa: require FP32 Stats and default an unset Stats dtype to FP32

vedaanta:
## Problem

The SDPA kernels always compute and store Stats (logsumexp) as FP32, 4 bytes per row. A graph that declared Stats with a narrower dtype — explicitly, or implicitly by leaving it unset with a non-FP32 `io_data_type` (`fill_from_context` assigned the io dtype) — built and executed fine, and the kernel then wrote FP32 rows past the end of the buffer the caller sized from the declared dtype. Depending on what the stray writes hit, this surfaced as silent corruption of adjacent allocations, illegal memory accesses, or `CUDNN_STATUS_EXECUTION_FAILED_CUDA_DRIVER` launch failures.

Root-caused from a customer workload: an fp16 Stats buffer (half the required bytes) was overwritten by batch ≥ 1 stats rows, corrupting whichever allocation happened to be adjacent — verified with compute-sanitizer (invalid f32 global writes starting exactly at byte offset `batch_idx * s_q * 4`, one byte past the fp16-sized allocation).

## Fix

- `Graph::sdpa_internal` sets the Stats output dtype to FLOAT at creation, so no later fill can assign the io dtype.
- `SDPANodeBase::infer_properties_node` defaults a still-unset Stats dtype to FLOAT (covers direct node-construction paths).
- `SDPANodeBase::post_validate_node` rejects an explicitly non-FP32 Stats with `GRAPH_NOT_SUPPORTED` ("The Stats output of sdpa must be an FP32 tensor.") instead of corrupting memory at execute.
- The python IR layer (`nodes.py`) applies the same FP32 default for SDPA / SDPA_FP8 / SDPA_MXFP8 Stats before its generic io-dtype fill.

## Testing

- New `test/python/test_sdpa_stats_fp32_required.py`: explicit fp16/bf16 Stats rejected at `validate()` (per io dtype), unset Stats inferred FP32, FP32 Stats accepted — 7/7 pass on B200 + cuDNN 9.26.0.33.
- Equivalent C++ checks exercised via a local `graph.validate()` harness: FLOAT OK, HALF rejected, unset inferred FLOAT.
- Regression: `test_sdpa_fp32_rejected.py`, `test_sdpa_with_caching.py`, `test_sdpa_custom_features.py` — 23/23 pass (including native-engine paths).
- The customer repro's full shape suite passes once its wrapper binds an FP32 stats buffer; with this change the previous misuse fails loudly at graph build instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * SDPA statistics outputs now consistently use FP32, regardless of the graph’s input or output data types.
  * Invalid non-FP32 statistics output configurations are now rejected during validation.
  * Unspecified statistics output types are automatically inferred as FP32.

* **Tests**
  * Added coverage for FP16, BF16, explicitly defined FP32, and invalid statistics output configurations.

### DIFF
--- a/include/cudnn_frontend/graph_interface.h
+++ b/include/cudnn_frontend/graph_interface.h
@@ -381,6 +381,9 @@ class Graph : public ICudnn, public INode {
         if (attributes.generate_stats == true) {
             sdpa_outputs.Stats = attributes.outputs[SDPA_attributes::output_names::Stats] =
                 output_tensor(attributes.name + "::Stats");
+            // Stats is always computed and stored in FP32; setting it at creation keeps
+            // fill_from_context from assigning the (possibly narrower) io data type.
+            sdpa_outputs.Stats->set_data_type(DataType_t::FLOAT);
         }
 
         // Dropout mask dump (created conditionally based on dropout parameters)
--- a/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
+++ b/include/cudnn_frontend/node/scaled_dot_product_flash_attention.h
@@ -446,6 +446,13 @@ class SDPANodeBase : public NodeCRTP<DerivedT> {
             auto stats     = attributes.outputs.at(output_names::Stats);
             auto stats_dim = stats->get_dim();
 
+            // Stats is always computed and stored in FP32, regardless of the io data type.
+            // Default an unset data type here instead of fill_from_context, which would
+            // wrongly assign the io data type.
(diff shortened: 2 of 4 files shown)


vedaanta:
@cudnn-ci-bot run frost

YangXu1990uiuc (approved):

PR #696, opened 2026-08-21
Fix three Rubin (sm_107) compile/launch failures in the CuTeDSL kernels

Anerudhan:
## Problem

All 26 fp8-path `test_grouped_gemm_quant` tests fail on Rubin (sm_107) with a deterministic pre-flight launch rejection:

```
error[CUDA_LAUNCH_INVALID_CONFIG]: CUDA launch failed: cudaErrorInvalidValue (1)
kernel '...moe_blockscaled_grouped_gemm_quant_rubin...' launch shared memory exceeds
current GPU arch sm_107a allowed. Allocated: 335872 bytes. Max: 334848 bytes.
```

## Root cause

`_compute_stages` picks the A/B pipeline stage count from a byte-sum model (`1024 mbar + epi_bytes + sinfo`) that ignores the 1024-B alignment padding of the `SharedStorage` buffers. For the fp8 256x256 quant configs the model chooses **9 A/B stages with only 480 B of modeled slack**, but the real struct needs ~1.5 KiB of padding (sSFA 4608→5120 B, plus the trailing padding that rounds the struct itself up to 1024 B). The launch then requests 335872 B of dynamic SMEM — exactly 1 KiB over sm_107's 334848 B per-block cap — and the driver rejects it.

fp4 configs fit only because their per-stage sizes happen to leave enough slack, and the identical math on sm100 lands *exactly at* the 232448 B cap, which is why only Rubin fp8 fails.

## Fix

`_compute_stages` in the three Rubin kernels that own a private copy of this logic (grouped GEMM **quant**, **dglu**, **glu_hadamard_quant**) now mirrors its kernel's `SharedStorage` layout exactly — per-buffer 1024-B aligned extents, mbar/scheduler head, tail buffers plus trailing struct padding — and sheds A/B stages until the struct genuinely fits, re-running the leftover-based D-stage expansion where one exists. **Configs that already fit keep identical stage counts** (the clamp only engages on overflow).

The shared `grouped/moe_kernel_helpers.py::compute_stages` (used by the glu Rubin kernel and several sm100 kernels) carries the same modeling gap but no observed overflow; it is intentionally left unchanged here to avoid touching the sm100 kernels' sizing, and can be hardened the same way in a follow-up.

## Verification

- **quant fp8 on sm107 hardware**: all previously failing tests pass (internal CI run: 389 tests, 0 failures, 0 errors). fp8 256x256 lands on 8 A/B + 6 D stages = 333824 B ≤ 334848 B.
- **quant fp4**: stage counts and SMEM sizes byte-identical before/after.
- **dglu / glu_hadamard_quant**: a 152-config sweep of `_compute_stages` (dtypes x tilers x CTA modes x feature flags) produces identical stage counts before/after at the real sm_107 capacity; a capacity scan confirms the clamp engages and sheds exactly one A/B stage at overflow boundaries.

Internal tracking: NVBugs 6645929.

## Second commit: DSA arch-target selection on Rubin (NVBugs 6645933)

A second Rubin compile failure with the same flavor of root cause (target selection, not codegen): the DSA arch map compiled compute-capability-10.7 kernels for the **`sm_100f` family target**, but the MXFP8 indexer kernels emit block-scaled MMAs (`.scale_vec::1X`) that ptxas only accepts on arch-specific targets, so every DSA mxfp8 test on Rubin failed at compile with `ptxas error: Feature '.scale_vec::1X' not supported on .target 'sm_100f'`. The family entry predates CuTeDSL wheels that know the Rubin arch; current wheels carry `sm_107a`, so `_ARCH_MAP` now maps `(10, 7)` to it.

Verified with `nvidia-cutlass-dsl 4.8.0a0`: the `sm_100f` mapping reproduces the exact ptxas failure; with `sm_107a`, all 39 DSA test parametrizations (mxfp8 **and** bf16) compile cleanly.

Internal tracking: NVBugs 6645933.

## Third commit: bias GMEM pointer alignment in the Rubin GLU kernel (NVBugs 6645919)

Every `with_bias` grouped GEMM GLU test on Rubin (~170 parametrizations) failed at `cute.compile` with an IR-verification ICE: `'cute.copy' op src ptr alignment (16 bits) does not meet requirement (128 bits) of atom '!cute_nvgpu.atom.simt_async_copy<bf16, cache = always, 128 b>'`. The kernel's bias GMEM→SMEM `cp.async` path partitions the `gBias` tile directly into a 128-bit copy atom; with dynamic MNKL the tile's provable alignment is only the bf16 element width. The Blackwell kernel (`moe_blockscaled_grouped_gemm_glu_bias.py`) already rewraps the tile with `assumed_align=16` before `partition_S` — this commit applies the identical rewrap to the Rubin kernel (the runtime allocation is genuinely 16-byte aligned, same guarantee the Blackwell path relies on).

Verified: all **172** `with_bias` GLU parametrizations compile cleanly for sm_107a with the change; before it, every sampled config reproduced the ICE. The glu_hadamard_quant kernel has the same textual pattern but its bias compile is clean (its API carries the alignment through), so it is deliberately left untouched.

Internal tracking: NVBugs 6645919.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shared-memory allocation calculations for grouped GEMM operations.
  * Accounts for buffer alignment, staging requirements, synchronization storage, and reserved capacity.
  * Dynamically adjusts processing stages to stay within available accelerator memory, improving reliability and occupancy.
  * Corrected architecture targeting for Rubin-based workloads.
  * Improved bias data handling to support required memory alignment and more reliable execution.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/utils/compiler.py
+++ b/python/cudnn/deepseek_sparse_attention/utils/compiler.py
@@ -31,7 +31,9 @@ _ARCH_MAP = {
     (9, 0): "sm_90a",  # Hopper H100
     (10, 0): "sm_100a",  # Blackwell B200
     (10, 3): "sm_103a",  # Blackwell Ultra B300
-    (10, 7): "sm_100f",
+    # Rubin needs the arch-specific target: the MXFP8 kernels emit block-scaled
+    # MMAs (.scale_vec::1X) that ptxas rejects on the sm_100f family target.
+    (10, 7): "sm_107a",
 }
 
 
--- a/python/cudnn/gemm/cutedsl/grouped/dglu/moe_blockscaled_grouped_gemm_dglu_rubin.py
+++ b/python/cudnn/gemm/cutedsl/grouped/dglu/moe_blockscaled_grouped_gemm_dglu_rubin.py
@@ -3908,12 +3908,11 @@ class BlockScaledMoEGroupedGemmDgluKernel:
             1,
         )
 
-        ab_bytes_per_stage = (
-            cute.size_in_bytes(a_dtype, a_smem_layout_stage_one)
-            + cute.size_in_bytes(b_dtype, b_smem_layout_staged_one)
-            + cute.size_in_bytes(sf_dtype, sfa_smem_layout_staged_one)
-            + cute.size_in_bytes(sf_dtype, sfb_smem_layout_staged_one)
-        )
+        a_bytes_per_stage = cute.size_in_bytes(a_dtype, a_smem_layout_stage_one)
+        b_bytes_per_stage = cute.size_in_bytes(b_dtype, b_smem_layout_staged_one)
+        sfa_bytes_per_stage = cute.size_in_bytes(sf_dtype, sfa_smem_layout_staged_one)
+        sfb_bytes_per_stage = cute.size_in_bytes(sf_dtype, sfb_smem_layout_staged_one)
+        ab_bytes_per_stage = a_bytes_per_stage + b_bytes_per_stage + sfa_bytes_per_stage + sfb_bytes_per_stage
         # Mbar bytes
         mbar_helpers_bytes = 1024
         # Sinfo bytes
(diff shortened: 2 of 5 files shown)


Anerudhan:
@cudnn-ci-bot run oss

Issue #624, opened 2026-08-17
frost(sdpa): THD zero-host-read execute loads uninitialized KV capacity rows — NaN poisons P@V on the f16/fp8 SM100/SM120 rows


Aneureka:
## Summary

The zero-host-read THD execute (#606 for SM100, #608 for SM120) binds K/V views at their
buffers' **capacity** — the packed token totals are device values, so the host can no longer
size the views to the live totals. The K/V TMA descriptors therefore span rows **between the
valid KV length and the storage end**: uninitialized memory the caller never wrote.

Those rows are neutralized on the S side (the shared `apply_mask_chunk` / the SM120 mask
overwrite masked columns with `-inf` via a **select**, which is NaN-safe), but nothing
protects the **V** side: the masked columns' P is `exp(-inf) = 0`, and in `P @ V` the IEEE
rule `0 * NaN = NaN` lets a NaN bit pattern in a garbage V row poison the whole accumulator —
every output row of the tile becomes NaN.

Before the zero-host-read rework this could not happen: the views (and descriptor extents)
were sized to the live totals read on the host, so rows past the packed total were outside
the descriptor and TMA **hardware-zero-filled** them.

The failure is **nondeterministic**: it needs NaN bit patterns in recycled device memory
(fresh CUDA allocations are usually zero pages, which is why the THD suites — whose harnesses
zero-fill their pads — stay green). f16 is especially exposed: ~3% of random bytes decode to
an f16 NaN, so a single 128-element garbage row almost surely contains one. In the wild it
surfaces as nondeterministic whole-sequence NaN on
`test_mhas_v2.py::test_sdpa_fp8_fwd_ragged_L0` (the mhas harness binds rounded-up
allocations).

## Affected

| Engine | Serves THD | Exposed | Evidence |
|---|---|---|---|
| `sdpa_fwd_prefill_sm120` (f16/bf16) | yes | **yes — reproduced** (RTX 5080 & RTX PRO 6000 SE) | probe below: NaN pad → `nan_frac=1.000` |
| `sdpa_fwd_prefill_sm100_d128` (f16/bf16) | yes | **yes — reproduced** (SM100) | same probe, same result |
| `sdpa_fwd_prefill_sm120_fp8` | yes | **yes — same mechanism** | fp8 variant of the probe reproduces it; also the mhas ragged NaN above |
[...]

Issue #538, opened 2026-08-10
frost(sdpa): cu_seq_len (cu_ragged) form assumes packed strides


vedaanta:
Follow-up to #516/#526. The `cu_seq_len_q`/`cu_seq_len_kv` prefix-sum form (cuDNN 9.24+) derives element offsets internally assuming the packed token stride `h*d`. A tensor declaring a wider token stride (kv-interleaved views etc.) would be mis-addressed the same way the THD ragged form was before #526.

Needed: declared-stride handling (scale by the tensor's actual `stride[2]`) or an explicit decline, plus fuzz coverage — #516's `with_ragged_token_gap` knob is deliberately not applied to the `cu_ragged` sweep draws for this reason.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #537, opened 2026-08-10
frost(sdpa): FP8/MXFP8 THD declared-stride support


vedaanta:
Follow-up to #526 (native THD stride support in the f16 fwd kernels).

- The SM120 per-tensor FP8 THD path (#509) keeps the packed contract: `_thd_check_strides_packed` declines non-packed declarations (AGENTS Hard Rule 2 — decline, never adapt). Extending native declared-stride support there needs the FP8 kernel's `compile()` to take per-tensor strides like the f16 kernels, with the 16-byte TMA global-stride rule evaluated at 1 byte/elem (strides x16 elements).
- SM100 FP8/MXFP8 have no THD path wired at all yet; when added, they must adopt the declared-stride serve-or-decline contract from day one.
- The fp8/mxfp8 test harness (`test/python/sdpa/mxfp8.py` etc.) allocates assuming packed strides, which is why #516's `with_ragged_token_gap` fuzz knob is enabled only in the fp16 ragged sweeps — the harness needs the same configured-stride allocation + offset generalization `fp16.py` got.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gavinkvx:
Hi team,

I'm happy to take the SM120 per-tensor FP8 portion of this issue.

My plan is to follow #526's serve-or-decline design: extend the FP8 THD compile specialization to carry declared Q/K/V/O strides, bind the declared-stride views/descriptors natively, and generalize the FP8 THD tests to cover gapped / interleaved token strides. I can validate the path directly on my RTX 5090 / SM120.

I noticed that #648 has since wired packed THD support for the SM100 FP8/MXFP8 paths as well, so unless you'd prefer a single larger PR, I'd keep those as a separate follow-up.

Please let me know, thank you! @vedaanta

vedaanta:
Thanks @gavinkvx. Please take up fp8 per-tensor.

Please run fp8 tests from [test_mhas_v2.py](https://github.com/NVIDIA/cudnn-frontend/blob/develop/test/python/test_mhas_v2.py) for verification.

PR #706, opened 2026-08-23
frost(sdpa): derive THD token capacity from the view's element span (fixes #613)

### LINKED ISSUE #613: frost(sdpa): THD zero-host-read extents break non-packed views; unit decode is extent-sensitive (regression from #606, also in #608)
## Summary

#606's zero-host-read THD execute derives the packed token extents from a host-side buffer heuristic (`t_q = min(q_buf.numel() // q_ts, o_buf.numel() // o_ts)`, same for KV). Two independent problems follow, one breaking non-packed **views**, one showing the kernel is **extent-sensitive** in a way no host heuristic can satisfy. Found while restacking the torch-ops work (#517/#554) on #608; SM100 (merged) and #608's SM120 path (`buf.numel() // ts` as well) both carry it.

## Bug 1 — `numel() // token_stride` under-claims non-packed views

A K/V slice of a kv-interleaved `[T, 2, H, D]` buffer (the `torch.nn.attention.varlen` fused-projection layout, #526's headline case) holds `T` tokens but only `T*H*D` of the record's elements, so the derived extent halves. Same for any declared-stride gap view. Result: TMA extents cut off half the tokens — silently wrong O.

Repro (fails on develop, FROST route; backend route passes):

```
cd test/python && CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1 pytest test_mhas_v2.py::test_repro --repro \
"{'data_type': 'torch.float16', 'diag_align': 'cudnn.diagonal_alignment.TOP_LEFT', 'batches': 2, 'h_q': 8, 'h_k': 8, 'h_v': 8, 'd_qk': 128, 'd_v': 128, 's_q': 256, 's_kv': 256, 'seq_len_q': [192, 256], 'seq_len_kv': [192, 256], 'is_ragged': True, 'is_padding': True, 'is_infer': True, 'with_ragged_token_gap': True, 'rng_geom_seed': 1, 'rng_data_seed': 2}"
```

→ 41.1% O mismatches served by `frost:sdpa_fwd_prefill_sm100_d128`. (The always-on stride fuzzing from #516 catches this — the ragged sweeps are partially red on develop depending on the seed's gap draws and buffer binding style.)

A storage-derived exact-addressable capacity (`(storage_elems_from_offset - row_footprint) // ts + 1`) fixes every view case (verified: repro 41% → 0, kv-interleave probes bit-exact) — **but breaks base-buffer bindings**, because of bug 2.

## Bug 2 — the kernel's unit decode is extent-sensitive; "over-claim is safe" does not hold

Forcing `t_q` for a failing sweep config (`seq_len_q=[17,174,665,911]`, real total 1767, harness buffer capacity 1792, CGA_TILE_M=256):

| forced t_q | result |
|---|---|
| 1767 (exact) | pass |
| 1792 (= 7×256 boundary) | pass |
| 1793 | **fail, ~50% O mismatches** |
| 1794 | fail |
| 1856 (64-aligned!) | fail |

Empirical law: corruption whenever `ceil(t_q / CGA_TILE_M) > ceil(real_total / CGA_TILE_M)` — the extent feeds the unit/tile decode (or something equivalent), so an over-claimed extent that crosses a tile boundary shifts real tiles onto wrong (batch, row) coordinates. The in-tree suites stay green only because the harness allocates `packed_token_capacity` (64-rounded) buffers, and a 64-rounding can never cross a 256 boundary.

## Why no host-side formula can fix this

The adapter must choose `t_q ∈ [real_total, real_total rounded up to the SAME tile count]` — but the real total is device-only by #552's own design. `numel//ts` breaks views; exact-addressable capacity breaks any binding with slack past a tile boundary. The fix has to be kernel-side: make the unit decode (and anything else consuming the token extent) read the device metadata (`cu[n_batch]`) instead of the descriptor extent, so the extent can be a plain safe upper bound. Alternatively the envelope/decode contract needs to tolerate extra dead tile rows.

## Affected

- SM100 (merged, #606): `_execute_thd` capacity sites in `api_dsl.py`.
- SM120 (#608, open): same heuristic (`return buf.numel() // ts`, and the token-major LSE cap).
- Non-packed THD views through any consumer (the torch-ops path in #517/#554 binds views and is currently xfail'd on the kv-interleaved cases pending this fix).

Happy to pair on this — I have the probes, the forced-t_q knob diff, and the storage-capacity patch staged.

vedaanta:
## Problem (issue #613)

The zero-host-read THD execute (#606/#608) derives the packed token extents host-side as `numel() // token_stride`. That is wrong on both edges for the buffers real integrations bind:

1. **Non-packed views halve.** A K/V slice of a kv-interleaved `[T, 2, H, D]` record — the layout `torch.nn.attention.varlen` users produce by slicing a fused KV projection — holds `T` tokens but only `T*H*D` of the record's elements, so the derived extent **halves** and the TMA descriptors cut off half the tokens: silently wrong O (41% mismatches on the #613 repro). Through the PyTorch python-API integration this now breaks **40 upstream `test_varlen_attention` cases**, because the engines recently started claiming tiny-window THD configs they previously declined.
2. **Storage-derived extents poison through allocator slack.** The obvious repair (derive from the untyped storage) over-claims into allocator slack — and that is *not* benign, which is the subtle part: rows between the real packed total and the extent are masked but still **multiplied** (`P(=0) × V`), so they must be *finite*; TMA zero-fill only covers rows **at or beyond** the extent. One slack row carrying NaN bit patterns poisons whole sequences through `0 × NaN` (reproduced: batch-wide NaNs in the ragged sweeps).

## Fix

Capacity = the largest `T` whose final token's **row** still fits in the buffer's own **element span** (`1 + Σ (size−1)·stride`). The span is exact on both edges:

- flat capacity buffers → exactly their token capacity (span = numel, no slack);
- interleaved / gapped views → exactly `T` (the last token needs only its own row footprint, not a full record span).

Every row below the capacity lies in caller-provided finite elements; every row at or beyond it TMA-clips to zeros. One shared `_thd_capacity` helper serves the SM100 f16 sites and the SM120/FP8 `_cap` sites (packed contract included).

## Verification (SM100, isolated worktree + venv, no shared JIT cache)

| Check | develop | this PR |
|---|---|---|
| #613 kv-interleave repro (`test_repro`, seeded) | 41.1% O mismatches, frost-served | pass |
| new deterministic regression test (fused-record K/V views vs packed binding, `torch.equal`) | **fail** | **pass** |
| `test_sdpa_random_fwd_ragged_L0` 5-seed slice (84 tests) | 84/84 | 84/84 — no regressions |
| fp8 THD ragged slice | green | green |
| upstream PyTorch `test_varlen_attention` (with the torch-ops stack on top) | 100 pass / **69 fail** | **140 pass / 29 fail** — the long-standing impl-identity baseline |

The regression test lives next to the #606 suite (`test_dsl_sm100_thd_interleaved_kv_views`) and encodes both failure modes in its docstring.

Fixes #613.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved THD attention handling for non-contiguous tensor views and buffers with unused capacity.
  * Corrected token extent calculations for Q/O and K/V data, improving result reliability.

* **Tests**
  * Added coverage for interleaved, strided K/V views and ragged offsets.
  * Verified that strided inputs produce the same results as equivalent contiguous inputs.

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -463,6 +463,38 @@ class SdpaFwdDsl(APIBase):
                 f"{desc.name}: non-packed THD strides {tuple(desc.stride)} are not supported by the FP8 path yet",
             )
 
+    def _thd_capacity(self, buf: torch.Tensor, desc: TensorDesc, packed: bool = False) -> int:
+        """Token CAPACITY of a THD buffer under the strides the view will
+        bind: the largest T whose final token's ROW still fits inside the
+        buffer's own element SPAN (``1 + sum((size_i - 1) * stride_i)``).
+
+        Why the span, and not numel or the untyped storage (issue #613):
+
+        - ``numel() // token_stride`` halves non-packed VIEWS — a K/V slice
+          of a kv-interleaved ``[T, 2, H, D]`` record holds T tokens but only
+          ``T*H*D`` of the record's elements — silently truncating the TMA
+          extent (half the tokens never load).
+        - The untyped storage over-claims into ALLOCATOR SLACK. That is not
+          benign: rows between the real packed total and the extent are
+          masked but still multiplied (``P == 0`` times V), so they must be
+          FINITE — TMA zero-fill only covers rows at or beyond the extent.
+          A slack row carrying NaN bit patterns poisons whole sequences
+          through ``0 * NaN``.
+
+        The span is exact on both edges: every row below the returned
+        capacity lies fully inside caller-provided (finite) elements, and
+        every row at or beyond it is TMA-clipped to zeros."""
+        h, d = desc.shape[1], desc.shape[3]
+        if packed:
+            ts, hs, es = h * d, d, 1
+        else:
+            (ts, hs, es), _ = self._thd_declared(desc)
+        if buf.numel() == 0:
+            return 0
+        span = 1 + sum((size - 1) * stride for size, stride in zip(buf.shape, buf.stride()))
+        row = (h - 1) * hs + (d - 1) * es + 1
+        return 0 if span < row else (span - row) // ts + 1
+
     def _thd_view(self, buf: torch.Tensor, desc: TensorDesc, tokens: int) -> torch.Tensor:
         """The declared-stride ``(1, T, H, D)`` view over a THD buffer's storage.
 
@@ -1504,9 +1536,7 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
         kv_lens_dev = self._checked_cu_seq_lens(seq_kv_lens, "cu_seq_len_kv") if self.cu_seq_kv_lens else self._checked_seq_lens(seq_kv_lens, "seq_kv_lens")
         lens_form = (1 if self.cu_seq_q_lens else 0) | (2 if self.cu_seq_kv_lens else 0)
 
-        (q_ts, _, _), _ = self._thd_declared(self.q_desc)
-        (o_ts, _, _), _ = self._thd_declared(self.o_desc)
-        t_q = min(q_buf.numel() // q_ts, o_buf.numel() // o_ts)
+        t_q = min(self._thd_capacity(q_buf, self.q_desc), self._thd_capacity(o_buf, self.o_desc))
         if lse_tokens_cap is not None:
             t_q = min(t_q, lse_tokens_cap)
         if t_q == 0:
@@ -1527,9 +1557,7 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
 
         Q = self._thd_view(q_buf, self.q_desc, t_q)
         O = self._thd_view(o_buf, self.o_desc, t_q)
-        (k_ts, _, _), _ = self._thd_declared(self.k_desc)
-        (v_ts, _, _), _ = self._thd_declared(self.v_desc)
-        t_kv = min(k_buf.numel() // k_ts, v_buf.numel() // v_ts)
+        t_kv = min(self._thd_capacity(k_buf, self.k_desc), self._thd_capacity(v_buf, self.v_desc))
         if t_kv == 0:
             # No KV storage at all:
             # every query row is dead — served by the KERNEL's own dead-row
@@ -3002,8 +3030,7 @@ class SdpaFwdDslSm120(SdpaFwdDsl):
         def _cap(buf, desc, heads, d):
             # Token CAPACITY under the strides the view will bind: declared
             # (f16, TMA-expressible by check_support) or packed (FP8).
-            ts = self._thd_declared(desc)[0][0] if declared_views else heads * d
-            return buf.numel() // ts
+            return self._thd_capacity(buf, desc, packed=not declared_views)
 
         # Q/O (and a token-major LSE) bind ONE dynamic token symbol; K/V the
         # other — shared floors.
--- a/test/python/sdpa/frost/test_sdpa_fwd_dsl_sm100.py
+++ b/test/python/sdpa/frost/test_sdpa_fwd_dsl_sm100.py
@@ -1393,6 +1393,74 @@ def test_dsl_sm100_thd_execute_cuda_graph_capture():
     _check([64, 33])
 
 
+@pytest.mark.L0
+@torch_fork_set_rng(seed=0)
+def test_dsl_sm100_thd_interleaved_kv_views():
+    """THD K/V bound as strided VIEWS of one fused [T, 2, H, D] record — the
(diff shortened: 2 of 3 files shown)


vedaanta:
Added **bd7a8f2** — `test(sdpa): actually fuzz ragged token gaps in the randomized sweeps` — the answer to "why didn't the fuzzer catch this?".

**The coverage seam.** The seeded token-gap draw (#516) lives in `ExecConfig.fill_derived_fields` and only fills strides left `None` — but `RandomizationContext`, which drives every `test_sdpa_random_*_ragged` sweep, explicitly assigned packed `bshd` strides in its ragged branch. So the randomized ragged fleet has **never bound a non-packed THD stride**, and for packed buffers `numel() // token_stride` is exact — no failure surface. Census over the fwd ragged L0 slice (84 configs): **0/84** drew a gap, although each config's own `rng_geom_seed` hand-draws nonzero gaps (e.g. `[2,1,1,2]`); the pre-fix adapter passes all 84, frost-served. That is exactly how this bug shipped and had to be found through an external integration.

**The commit** makes the ragged branch leave Q/K/V/O strides `None` and ends `__call__` with `fill_derived_fields()` — one source of truth for the gap draw and its auto-packed fallbacks (cu / offset-multiplier forms #538, 1-byte dtypes #537). Head-major stats stride and the dense branch are untouched.

**Red-team evidence** (SM100, cuDNN 9.26.0.33, `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`):

| suite (wired fleet) | pre-fix adapter | with this PR's fix |
|---|---|---|
| fwd ragged L0 slice (84, all frost-served) | **80/84 FAIL** | **84/84 pass** |

i.e. this wiring alone would have caught #613 the day the capacity heuristic merged. Post-fix matrix, all green:

- fwd ragged L0 slice: 84/84 (census: 84/84 now gapped, all 84 vulnerable under the old formula)
- bwd ragged L0 slice: 158/158 — identical to the unwired control on the same lib (backend serves every gapped gradient combination; bwd FROST engines decline THD, so routing is unchanged)
- `ragged_unified_L1`: 24/24 · `offset_multiplier_unified_L1`: 24/24 (cu / mult forms stay packed via the existing fallbacks, 20/20 each in an offline census)
- `test_ragged_token_gap_stable_under_stride_overrides`: pass

Note: the bwd baseline requires a backend with the dSink+ragged-stats fixes (9.26 here); on a 9.24 lib the bwd ragged sweep fails for that unrelated pre-existing backend reason with or without this commit.

Anerudhan:
Run CI.
Moving to 1.29

vedaanta:
Follow-up filed as #718. While reasoning about how much of the capacity contract this PR actually buys, I confirmed the other half is still open: `#706` guarantees every row below capacity is **caller-owned**, but not that it is **initialized**. The f16 THD paths have no equivalent of the FP8 packed-total clamp, so an over-allocated K/V buffer with an unwritten tail still poisons O — measured 49.6% NaN at `seq_lens=[200,150,47]`, `CAP=640`, and exactly 0 when the extent equals the packed total. Repro and the port plan (the FP8 `build_thd_meta_o_kv_descs_kernel` already does all of it) are in #718.

Nothing here changes this PR — the span fix is still needed and still correct, since the clamp only helps tensors that get one. Flagging the linkage so the two are reviewed together.

vedaanta:
Correction to my note above: the follow-up I filed as #718 is a duplicate of #624 (same defect, already tracked, with harness hardening in flight as #646). #718 is closed and the analysis is consolidated onto #624. The linkage to this PR is unchanged — #706 guarantees rows below capacity are caller-*owned*; #624 covers their being *initialized*.

vedaanta:
Superseded by #740 — closing.

#740 was stacked on this branch and merged as a **squash** (`3631ecb44`), which folded in all four commits, including both of this PR's. Verified on `develop` after the merge:

- **span-derived capacity** — `_thd_capacity` present with the exact formula (`span = 1 + sum((size - 1) * stride ...)`), wired at all three extent sites: SM100 `t_q` (api_dsl.py:1612), SM100 `t_kv` (:1634), and the SM120/FP8 `_cap` (:3126).
- **the #613 regression test** — `test_dsl_sm100_thd_interleaved_kv_views` present.
- **the fuzzer wiring** — `random_config.py`'s ragged branch now leaves Q/K/V/O strides `None` and `__call__` ends with `fill_derived_fields()`; the old `get_strides_from_layout` assignments that bypassed the seeded token-gap draw are gone. This is the change that makes `test_mhas_v2`'s randomized ragged sweeps actually fuzz non-packed THD strides — the coverage seam that let #613 ship. It is on `develop`.

So nothing here is lost, and #613 stays fixed by the merged code rather than by this PR.

PR #710, opened 2026-08-24
CSA compressor: extend the validated ratio envelope to ratio=2

zkyue:
## What

Widen `check_support` in `python/cudnn/csa/compressor/api.py` to accept `ratio=2` in the
generic-kernel family: the validated envelope becomes `ratio in {2, 4, 128}` (with
`coff in {1, 2}` at every ratio, as before).

**No kernel code changes.** The generic kernels in `compressor_sm100.py` are
JIT-specialized per `(ratio, head_dim, coff)` and are ratio-generic; the `{4, 128}` gate
encoded validation scope, not a kernel limitation. The dedicated ratio=128 family
(`compressor_sm100_r128.py`) and the ratio-based routing are untouched.

## Why

`ratio=2` (with `coff=2`) is the compressor configuration used in production training of
the model family this operation serves (the kernels were ported from Megatron-LM in
#427). With the gate at `{4, 128}`, integrating the module at ratio=2 requires bypassing
the validation layer and calling `compressor_sm100.run_fwd`/`run_bwd` directly; widening
the envelope lets that configuration go through the supported, validated API.

Why ratio=2 needs no kernel work (verified by reading the kernels and by the test
additions below):

- the window length `win = coff * ratio` is a compile-time constant; register pressure
  shrinks with ratio (the family is documented register-bound only beyond `ratio ~ 32`,
  and ratio=4 is already validated);
- APE indexing (`k % ratio`), the overlap-window first-block invalidation, the backward
  zero-write ownership classes (`seqlen % ratio` tails, `seqlen < ratio` segments, the
  last block's first-half columns at `coff=2`, token-capacity padding) and the dAPE
  atomic targets are all expressed in `ratio` with no lower bound;
- the launch schedules do not depend on ratio (`_fwd_schedule` is a function of
  `head_dim` only; the backward uses fixed `rows_per_cta`/`threads`);
- the int32-offset bounds in `check_support` are checked independently on
  `total_tokens * coff * head_dim`, `total_comp * head_dim` and
  `ratio * coff * head_dim`, so the (relatively larger) `total_comp` at ratio=2 is
  already covered.

## Numerics at ratio=2

B200 (CC 10.0), torch 2.13.0 / CUDA 13.3 / nvidia-cutlass-dsl 4.5.1, seeded inputs from
this suite's generator; fused wrappers vs the suite's fp32-intermediate eager reference
(the numerics-contract comparator):

| config | out | dKV | dScore | fwd n_diff | dAPE max_abs (rel to ref max) |
|---|---|---|---|---|---|
| r2 c2 d128 1x8192 | bitwise | bitwise | bitwise | 0/524288 | 2.1e-05 (6.4e-07) |
| r2 c2 d128 3x8192 | bitwise | bitwise | bitwise | 0/1572864 | 8.0e-05 (1.1e-06) |
| r2 c2 d512 1x8192 | bitwise | bitwise | bitwise | 0/2097152 | 3.2e-05 (7.5e-07) |
| r2 c2 d512 3x8192 | bitwise | bitwise | bitwise | 0/6291456 | 9.2e-05 (1.0e-06) |
| r2 c2 d128 ragged [8191,8192,4093] | bitwise | bitwise | bitwise | 0/1310336 | 4.2e-05 (6.1e-07) |
| r2 c1 d128 1x8192 | bitwise | bitwise | bitwise | 0/524288 | 3.1e-05 (8.0e-07) |
| r4 c2 d128 1x8192 (control) | 8 elems differ | bitwise | bitwise | 8/262144 | 1.1e-05 (4.9e-07) |
| r4 c2 d512 3x8192 (control) | 109 elems differ | bitwise | bitwise | 109/3145728 | 2.9e-05 (7.5e-07) |

`dKV`/`dScore` carry the generic family's bitwise contract at ratio=2, exactly as at
ratio=4. On every tested ratio=2 config the forward is additionally fully bitwise (the
contract only requires "within one bf16 rounding step on a tiny fraction of elements";
the ratio=4 controls show that documented behavior). `dAPE` differs at fp32
atomic-order level only (~1e-6 relative), per the documented contract (`dAPE` is not
run-to-run deterministic in either implementation; the backward still refuses strict
`torch.use_deterministic_algorithms(True)`).

## Performance at ratio=2

nsys pure-kernel times (sum of kernel durations inside a `cudaProfilerApi`-gated
50-iteration unsynced loop / 50; JIT + fast-launch warmup before the window), ratio=2
`coff=2`, single B200; eager = the fp32-intermediate reference region (fwd + autograd
bwd, ~93 kernel launches per step):

| config | fused fwd | fused bwd | fused fwd+bwd | eager fwd+bwd | speedup |
|---|---|---|---|---|---|
| d128, 1x8192 | 4.7 us | 11.4 us | 16.2 us | 331.6 us | **20.5x** |
| d128, 3x8192 | 10.4 us | 25.6 us | 36.1 us | 650.5 us | **18.0x** |
| d512, 1x8192 | 13.5 us | 27.7 us | 41.2 us | 785.5 us | **19.1x** |
| d512, 3x8192 | 39.6 us | 69.2 us | 108.8 us | 2104.6 us | **19.3x** |

(The fused ratio=2 kernels run at the same throughput class as the validated ratio=4
configs — same kernel bodies, smaller unrolled window.)

## Tests

`test/python/fe_api/csa/test_CSA_compressor.py` adds ratio=2 cases following the exact
ratio=4 patterns, both window forms (`coff` 1 and 2):

- numerics vs the fp32-intermediate eager reference / verbatim upstream eager / fp64
  oracle (12 new `_SHAPES` cases: single/ragged/short-segment/empty-segment packs, d
  128/512 and the odd `d=65` scalar layout), with the bitwise `dKV`/`dScore` assertion
  now applied at `ratio in {2, 4}`;
- static-capacity padding rows (6 new cases, including a padding-only pack — every segment shorter than ratio, positive static capacity);
- NaN-canary zero-write ownership over every never-consumed slot class (18 new cases;
  the tiny-segment packs use 1-token segments so the `seqlen < ratio` class is still
  exercised at ratio=2);
- run-to-run replay determinism (ratio now parametrized over {2, 4});
- `check_support` acceptance at ratio=2 and rejection cases (`coff` 0/3 at ratio=2, a
  `total_tokens < ratio` case at ratio=2, and the updated `{2, 4, 128}` rejection
  message).

Module suite (`-m "L0 or L1"`, single B200): **145 passed, 1 skipped**
(`test_multi_device_launch`, single-GPU runner). On the unpatched gate the new ratio=2
cases fail with the expected envelope rejection
(`ValueError: CSA compressor is validated for ratio in {4, 128} only, got ratio=2, ...`),
so the tests pin the widening in both directions.

Docs: `docs/fe-oss-apis/csa.md` support surface and numerics-contract sections updated
(`ratio in {2, 4}` for the generic family; ratio=128 family unchanged).

## Summary by CodeRabbit

- **New Features**
  - Added support for CSA compression with ratio 2, alongside existing ratios 4 and 128.
  - Added support for coefficient values 1 and 2 with ratios 2 and 4.

- **Bug Fixes**
  - Improved validation for unsupported ratio and coefficient combinations.
  - Ensured consistent bitwise results for key outputs across supported generic compression ratios.

- **Documentation**
  - Updated API and numerical behavior documentation to describe ratio 2 support and its guarantees.

### DIFF
--- a/docs/fe-oss-apis/csa.md
+++ b/docs/fe-oss-apis/csa.md
@@ -48,12 +48,13 @@ gradients on such padding rows.
 All arithmetic is fp32 with one final bf16 rounding; `mul.rn.f32` / `fma.rn.f32` are
 pinned in PTX so results do not depend on compiler FMA contraction. The numerics
 contract is **per ratio family** (the two families intentionally differ — do not
-assume the ratio=4 guarantees at ratio=128):
+assume the ratio {2, 4} guarantees at ratio=128):
 
-- **`ratio == 4`** (production, unchanged): against an fp32-intermediate eager
-  reference (same op order, fp32 throughout), `dKV`/`dScore` are **bit-identical**
-  and the forward matches within one bf16 rounding step on a tiny fraction of
-  elements.
+- **`ratio in {2, 4}`** (generic kernels; the bitwise contract — ratio=4 as originally
+  validated, ratio=2 added on the same terms): against an
+  fp32-intermediate eager reference (same op order, fp32 throughout), `dKV`/`dScore`
+  are **bit-identical** and the forward matches within one bf16 rounding step on a
+  tiny fraction of elements.
 - **`ratio == 128`** (deterministic tolerance contract): the kernels are
   **deterministic and faithful to the fp32-intermediate eager reference** (the
   same eager region computed with fp32 intermediates and one final bf16 rounding —
@@ -132,8 +133,9 @@ that need a fully deterministic backward must use an eager implementation.
 
 - Compute-capability major **>= 10** (SM100 and newer; the kernels use no
   architecture-specific features beyond the SM100 baseline)
-- `ratio == 4`, `coff in {1, 2}` (`coff == 2` is the production CSA/HCA configuration,
-  `coff == 1` the own-block window form) — served by the generic kernels, which are
+- `ratio in {2, 4}`, `coff in {1, 2}` (`coff == 2` is the production CSA/HCA configuration,
+  `coff == 1` the own-block window form; `ratio == 2` is the configuration used in
+  production training) — served by the generic kernels, which are
   generic over `(ratio, head_dim, coff in {1, 2})` but keep the whole pooling window in
   registers (register-bound beyond `ratio = 32`)
 - `ratio == 128`, `coff in {1, 2}`, `head_dim in {128, 512}` — served by dedicated
@@ -338,7 +340,7 @@ never-consumed slots and fp32-atomic `dAPE`, exactly as at ratio=4. The backward
 defaults to the same fast exp outside the d=128 small-pack (vec=1) buckets.
 
 **The ratio=128 numerics contract is the deterministic tolerance contract described in
-[Numerics](#numerics), NOT the ratio=4 bitwise-`dKV`/`dScore` contract.** The
+[Numerics](#numerics), NOT the generic-family (ratio {2, 4}) bitwise-`dKV`/`dScore` contract.** The
 reduction orders (forward chunk merge, backward `den`/`S` partial merge), the
 backward's hoisted reciprocal, and the fast-exp buckets all differ from the eager op
 order by design, each adopted on a measured same-GPU win and gated on tolerance +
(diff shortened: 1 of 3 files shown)


Anerudhan:
@cudnn-ci-bot run oss

Anerudhan (approved):

Anerudhan (approved):

PR #720, opened 2026-08-24
frost(sdpa): run the KV split the heuristic chose, on the true cluste…

yanzhuo607:
…r extent

choose_split_kv computed a split and then nothing used it. Two defects, both on the delivery path rather than in the cost model:

- _split_points returned [no_split, chosen], so the chosen split landed at plan[1]. build_plans() stops at the first entry that builds and execute() runs _plan_index, so a plain build ALWAYS took the unsplit plan; the split was reachable only through select_plan or an ALL-policy autotune. Return [chosen, no_split] instead — the split leads, and no-split stays reachable behind it.

- The model was fed rows_per_tile = tile_m * cga, but an SM100 d128 cluster covers TILES_Q * TILE_M * CTA_MMA Q rows on its CTA pair — twice that. The doubled tile count reads a half-empty machine as full, so the chooser under-splits or declines to split at all. Use _pack_gqa_tile_q, the helper that already answers "Q rows one grid tile covers", and the same extent every test in test_split_kv_heuristic.py already assumed.

Flipping the lead moved the split into the base knob set, which exposed a third: the "a split set rides the plain scheduler" coupling lived only in the splits[1:] runner-up loop, so a LEADING split inherited the derived LPT_L2 policy on causal graphs — unbuildable on SM120, which raises on split_kv > 1 under an LPT remap. The coupling now binds whichever leg leads, and scheduler runners ride an unsplit leg.

The chooser itself is unchanged, so a grid that already fills the machine still does not split.

## Before submitting

## Summary by CodeRabbit

- **Performance**
  - Improved SDPA forward-pass configuration selection using device-aware, KV-size-bounded split-KV candidates.
  - More accurately accounts for combine-work costs and launch geometry when choosing execution plans.
  - Split-KV execution is prioritized when beneficial, while no-split remains available as a fallback.
  - Improved scheduler selection for causal and mask-free attention workloads.

- **Bug Fixes**
  - Corrected split-KV planning for grouped-query attention, decode workloads, and large sequence dimensions.
  - Improved consistency between selected execution plans and scheduler behavior.
  - Improved handling of devices with different split-KV capabilities.

### DIFF
--- a/python/cudnn/sdpa/fwd/engines.py
+++ b/python/cudnn/sdpa/fwd/engines.py
@@ -242,10 +242,14 @@ class Capabilities:
     tile_ns: frozenset[int] = frozenset()
     cgas: frozenset[int] = frozenset()
     pack_gqas: frozenset[bool] = frozenset({False})
-    # Split-KV domain. {1} = the axis exists but only "off" is served; rows
-    # whose kernels wire the split path AND whose adapter launches the combine
-    # widen this (the SM100 f16 rows today).
-    split_kvs: frozenset[int] = frozenset({1})
+    # Does this row's lowering wire the KV-split path (kernel SplitHelpers +
+    # adapter carving the partial slabs + launching the combine)? A GATE, not a
+    # domain: WHICH splits are worth trying is a device-derived search space
+    # (heuristics.split_kv_candidates), not a per-row constant. Fail-closed —
+    # a row that accepts split_kv > 1 without the plumbing leaves untouched
+    # partial slots at lse_partial = 0, which corrupt the combine's
+    # log-sum-exp rather than raising.
+    split_kv_supported: bool = False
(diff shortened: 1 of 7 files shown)


yanzhuo607:
@cudnn-ci-bot run frost

vedaanta (approved):

yanzhuo607:
@cudnn-ci-bot run frost

vedaanta:
The sm120 CI failure here (job 412040217, \`frost:rel:sdpa:sm120\`) is a missing \`combine_rows\` kwarg in \`test_sdpa_fwd_split_kv_sm120.py\`'s \`_expected_split\` helper — \`choose_split_kv\` gained that required keyword-only arg on this branch, and the sm100 test sibling already passes it but sm120's didn't.

I rebased this branch onto latest develop (clean, no conflicts) and fixed the sm120 test to match: yanzhuo607/cudnn-frontend#1

PR #691, opened 2026-08-21
Fix SM100 D192 E5M2 attention sink accuracy

adshen:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Fix the SM100 per-tensor FP8 `D_QK=192, D_V=128` forward kernel for E5M2 Q/K/V with an FP32 attention sink.

The change:

- removes the E5M2 sink routing restriction;
- initializes the E5M2 online-softmax state with the sink as a virtual score;
- uses true negative infinity for masked E5M2 sink scores;
- keeps the first real KV tile on a numerically stable max trajectory for BF16/FP16 output;
- consumes the combined sink/KV max and sum directly in the correction epilogue;
- converts the existing dtype rescale threshold to the kernel's log2 score units inside the kernel; and
- releases next-tile TMEM ownership only after the current output accumulator loads complete.

The review follow-up removes all `CUDNN_RESCALE_THRESHOLD` environment parsing, template-parameter plumbing, and cache-key specialization. E4M3 sink, no-sink, MXFP8, and other head-dimension routes retain their existing compile-time paths.

## Why

The FP8 reference initializes online softmax with `max=sink` and `sum=1`. The previous D192 kernel processed and quantized QK probabilities first, then merged the sink into the denominator in the final correction epilogue. Although equivalent in exact arithmetic, the two evaluation orders can choose different lazy-rescale maxima and therefore produce different E5M2 probability rounding near rescale cutoffs.

The sparse CI blocker had two mismatched output elements out of 17,372,160, with a maximum absolute error of `0.1484375` against an allowed absolute tolerance of `0.08`. Initializing the kernel state with the sink, preserving masked scores as true negative infinity, and using the correct threshold units reduces the same case's maximum absolute error to `0.03125` with zero mismatches.

[...]

PR #743, opened 2026-08-25
test_mhas_v2: naturally produce deeply negative attention scores in every test

vedaanta:
## Why

Sparse-small-int test data concentrates attention scores near 0, so the suite never exercised the regime where an entire score row is deeply negative. Real models hit it routinely (Qwen2.5 attention logits reach +-2500), and it is exactly the regime that let the SM100 decode row-max/OOB-masking bug ship, silently corrupting generation (pytorch/pytorch#193893): with all valid scores negative, an unmasked OOB score of 0 poisons the row max and every valid probability underflows to exactly 0 in fp32.

## What

New `inject_negative_score_rows` helper in `test/python/sdpa/helpers.py`, wired into all three dtype generation paths (fp16/bf16 dense + ragged, fp8, mxfp8; perf mode excluded):

- draw a random +-1 direction `u` of length d,
- shift all of `k` by `+2u`,
- overwrite 1/16 of q rows (min 1) with `-m*u`, `m` a power of two sized per head-dim and attn_scale so selected rows' scores land at -200..-490 (past the fp32 `exp2f` underflow cliff, ~-87),
- every value stays exactly representable in fp8 e4m3/e5m2, fp16, bf16 and fp32, preserving the suite's exact-arithmetic philosophy (`|k| <= 4`, `|q| = m` a power of two). Verified for d in {32..256} and both attn_scale conventions.

Summary of current fix:
#744 : fp64 model-fit shows the SM100 bwd kernel matching its exact reference to ~0.04% — the "fp16-rounded score recompute" hypothesis was wrong. Two causes: the reference used its own rounded o_ref for D instead of the quantized fwd O the kernel consumes; and under GQA the injected rows make the cross-head dK/dV reduce ill-conditioned. Fixes: feed the kernel's O; confine injections to the first q-head per group.

#745: two reference bugs. The rescale threshold was compared in the natural-log domain while the kernel's softmax runs in log2 (off by ln2) — 13 mxfp8 fwd failures. And the bwd reference renormalized with its own softmax instead of rebuilding P from the fwd stats the kernel is fed — 3 fp8 bwd failures. Plus dtype-aware bwd atol (fp8 0.04→0.08), previously tighter than the forward's for no reason.

Frost fp8 MLA kernel race: fp8 MLA (d_qk=192, d_v=128) fwd tests fail intermittently — test_sdpa_fp8_fwd_L0[test233] (s=3979 causal, e4m3→fp16) corrupts a few hundred output elements ~3/4 runs. Root cause: the correction warp arrived mb_stats_read[qs] right after the stats read, releasing the next tile early — its first BMM2 overwrites the O accumulator TMEM while the epilogue is still reading it. Fix: arrive after the epilogue's last O TMEM load, generalizing #691's E5M2+sink fix to all paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attention output synchronization across FP8, BF16, and FP16 configurations.
  * Corrected softmax rescaling and backward probability calculations when forward statistics are available.

* **Tests**
  * Expanded scaled dot-product attention coverage for FP16, FP8, and MXFP8 formats.
  * Added strongly negative attention-score cases, including ragged and uniform inputs.
  * Improved backward validation, sink-token handling, and format-specific comparison tolerances.


vedaanta:
The surfaced backend follow-ups now have tracking issues: #744 (fp16/bf16 bwd dK/dQ ~7% loss on deeply-negative rows, fp16-rounded score recompute signature) and #745 (fp8 fwd O mismatch + mxfp8 kernel/emulator exact-match disagreements in the negative regime).

vedaanta:
@cudnn-ci-bot run frost,backend

Adnios:
@cudnn-ci-bot run frost,backend

Adnios:
@cudnn-ci-bot run frost,backend

Adnios:
@cudnn-ci-bot run frost,backend

adshen (approved):

PR #757, opened 2026-08-26
frost(sdpa): guard empty d192 quantized O TMEM loads

### LINKED ISSUE #702: frost(sdpa): latent dead-row uninitialized-TMEM NaN hazard in the d192 fp8/mxfp8 and f16 SM100 forward epilogues
### Summary

The SM100 TMEM-epilogue forward kernels zero dead rows (Q rows with no visible KV column: zero-length KV sequence, or fully padded/masked row) by multiplying the O accumulator read from TMEM by `inv_sum := 0`:

```python
row_dead = total_sum <= 0.0
inv_sum  = select(row_dead, 0.0, inv_sum)
...
o_scaled = o_chunk * inv_sum   # o_chunk = tcgen05_ld from the O TMEM region
```

For a dead row the mainloop is empty and **BMM2 never writes the O TMEM region**, so `o_chunk` is whatever bits were already there. TMEM is not zero-initialized at launch: on the *first* tile a tensor-core slot processes, those bits can decode as NaN/inf, and `NaN * 0 = NaN` leaks straight into O (and, on fp8 kernels, into the `amax_o` reduction). On any later tile TMEM holds the previous tile's finite fp32 results, so `finite * 0 = 0` and the guard works — which is why the bug is order-dependent and almost never fires.

### Where this was already fixed

PR #648 hit this in CI (mhas fp8 ragged, `seq_len_q=[126, 0, 60]` / `seq_len_kv=[0, 83, 77]`: the zero-KV sequence in **batch 0** landed on first tiles and produced deterministic NaN O, while `seq_len_kv=[174, 0]` passed because batch 1's tiles reused warmed TMEM). The fix shipped there for the three d128 kernels — `prefill_d128_fp8_sm100.py`, `prefill_d128_fp8_sm107.py`, `prefill_d128_mxfp8_sm100.py`:

- hoist `row_dead = total_sum <= 0` above the sink branch (with a sink the denominator is finite but the O numerator is an empty sum, so O must still be 0), and
- `arith.select` each stored O element to literal `0.0` on `row_dead` before the fp8 pack / half cast and before the `amax_o` accumulation, instead of relying on multiply-by-zero.

### Where it is still latent

- `prefill_d192_d128_fp8_sm100.py` and `prefill_d192_d128_mxfp8_sm100.py` — same epilogue pattern (`row_dead` guard exists but only via `inv_sum := 0`; `o_scaled = o_chunk * inv_sum` is unsanitized). Dense-only kernels, reachable via `use_padding_mask` with `seq_len_kv[b] == 0` when that sequence's tiles are the first on their SMs.
- All f16 SM100 forward kernels (`prefill_d128_f16_sm100.py`, `prefill_d192_d128_f16_sm100.py`, `prefill_d256_f16_sm100.py`, `prefill_d512_f16_sm100.py`) — the pattern originates there. Additionally the f16 test harness does not NaN-poison its buffers, so current coverage cannot catch a leak.
- `prefill_fp8_sm120.py` is **not** affected: register accumulators (zero-initialized) and an explicit `O := 0` dead-row branch.

### Suggested fix

Port the #648 dead-row sanitize (per-element `arith.select` to 0 on `row_dead`, hoisted above the sink branch) into the epilogues listed above, and add a zero-length-KV-in-batch-0 regression test per suite (the batch-0 placement is what makes the failure deterministic instead of flaky).

Repro sketch: fp8/mxfp8 d128 THD with `seq_len_kv[0] == 0` reproduced deterministically before the #648 fix; the d192/f16 analogue is the dense padded equivalent.

Aneureka:
## Before submitting

## Affected area

- FE OSS kernels or CuTeDSL

## Summary

- Sanitize dead rows in the SM100 d192/d128 per-tensor FP8 and MXFP8 forward epilogues by selecting literal zero before output conversion and `amax_o` accumulation.
- Add deterministic NaN-primed regressions for a leading zero-length KV batch across FP8 and half output types, with and without attention sinks.
- Also fix a style issue introduced by 781e3383b6775b7130dc31cda5228444825dff4e.

## Why

When a row has no visible KV column, BMM2 may leave its output accumulator in TMEM unwritten. Multiplying the loaded value by a zero normalization factor is insufficient because an unwritten value can decode as NaN and `NaN * 0` remains NaN. That can contaminate both O and `amax_o`.

The epilogues now apply a per-element `arith.select` to replace every dead-row value with literal zero before amax reduction, FP8 packing, or half conversion. This also covers dead rows inside otherwise live tiles and preserves the existing TMEM load pipeline and barrier protocol.

## Related issues

Fixes #702.

No public API or ABI change. Dead rows now produce zero O, retain the expected LSE and sink semantics, and do not affect output amax.

## Testing

On SM100 with cuDNN backend 9.26, all five targeted regression cases passed:

```bash
pytest -q \
  sdpa/frost/test_sdpa_fwd_fp8_sm100.py::test_fp8_d192_d128_leading_zero_length_kv \
  sdpa/frost/test_sdpa_fwd_mxfp8_sm100.py::test_mxfp8_d192_d128_leading_zero_length_kv
```

Also passed:

- `pre-commit run` on the four changed files
- `python -m py_compile` on the two changed kernels and two changed test modules
- `git diff --check`

The full test suite was not run; validation was targeted to the affected SM100 kernels.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected scaled dot-product attention handling for rows with empty or invalid key/value ranges.
  * Such rows now reliably produce zero outputs, negative-infinity log-sum-exp values, and valid amax results.
  * Improved behavior across FP8, MXFP8, FP16, BF16, and attention-sink configurations.

* **Tests**
  * Added coverage for leading zero-length key/value sequences, padding-mask scenarios, and declared packed-token totals.
  * Added validation for zero outputs and finite quantization metrics.

### DIFF
--- a/python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_fp8_sm100.py
+++ b/python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_fp8_sm100.py
@@ -1955,6 +1955,12 @@ def _correction_warp_group(
             lse_val = cutlass.Float32(0.0)  # pre-declare; computed in both branches
             LN2 = cutlass.Float32(0.6931471805599453)
             total_max_nat = total_max_scaled * LN2
+            # E5M2+sinks seeds total_sum with the sink contribution, so its
+            # empty-numerator predicate must come from the KV bounds instead.
+            if cutlass.const_expr(_E5_SINK):
+                row_dead = bounds.right <= bounds.left
+            else:
+                row_dead = total_sum <= cutlass.Float32(0.0)
             # Row identity (PackGQA): this lane's row decodes to (token,
             # head-in-group) — q_row_global is the TOKEN index and
             # row_head_idx the row's true query head (lane-varying under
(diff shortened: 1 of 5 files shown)


Aneureka:
@cudnn-ci-bot run frost

adshen (approved):

PR #760, opened 2026-08-26
perf(dsa): specialize SM100 sparse-attention backward for H32/D576

icavan:
## Before submitting

- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (maintainer action required for external contributors).
- [x] I set the **Milestone** and **Projects** fields in the sidebar (maintainers can set these for external contributions).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

This PR adds a dedicated SM100 DeepSeek Sparse Attention backward kernel for the exact `H=32, Dqk=576, Dv=512` shape and routes that shape to an H32/M64 backend.

The specialization:

- subclasses the H16 specialization and reuses its established kernel skeleton—loaders, warp-role dispatch, pipeline/barrier construction, reducer helpers, and dQ epilogue—while rebuilding the H32-dependent MMA tilers, SMEM views, TMEM offsets, and steady-state schedule;
- uses a native 32-head tensor-core tile instead of executing the generic 64-head tile with half of the head extent inactive;
- swaps the score GEMM operands relative to the generic path, computing the equivalent transposed form `Sᵀ[KV,H] = K[KV,D] · Qᵀ[D,H]`; this maps sparse KV rows to UMMA M64 and the 32 heads to the supported N32 extent;
- keeps a single CTA per query token and processes sparse KV rows in M64 tiles;
- uses compact 4 KiB `P` and 4 KiB `dS` shared-memory staging buffers;
- assigns disjoint TMEM columns to the 64-wide dKV tail, while dKV main fragments use a two-generation alias ring;
- releases the single-stage K buffer immediately after the final K-dependent dQ MMA, allowing the next sparse KV gather to overlap the remaining dKV MMA/TMEM/global-reduction work;
- detaches dKV2/dKV3 from TMEM before publishing the reuse barrier, so their global FP32 atomics can overlap the next producer generation;
- retains the 16-warp, 8-lane-subwarp sparse KV gather path, with an H32-specific register budget.

The generic SM100 implementation remains unchanged for all other head counts and dimensions.

## High-level architecture

[SM100 H32/D576 DSA backward kernel architecture](https://github.com/icavan/random-notes/blob/main/pictures/sm100-dsa-bwd-h32-d576-single-q-kernel-architecture.svg)

The H32 backend is intentionally implemented as a shape-specialized subclass of the H16 kernel rather than as a second independent kernel. H16 already defines the proven control-plane architecture for the same `Dqk=576, Dv=512` backward dataflow: tensor and workspace preparation, static and indexed KV loaders, warp-role dispatch, pipeline and named-barrier construction, reducer utilities, synchronization/fence contracts, and the dQ epilogue. H32 reuses that structure and replaces only the shape-sensitive data plane: the H32/M64-N32 MMA tilers and A/B-swapped score orientation, SMEM views and lifetime reuse, TMEM column placement, register budgets, and the per-tile producer/consumer schedule. Starting from a separate full copy would duplicate most of the kernel and make correctness fixes to barriers, proxy fences, top-k boundaries, and epilogues easy to apply to one variant but miss in the other. The subclass boundary keeps shared synchronization machinery centralized while isolating H32 code-generation decisions, so the specialization remains reviewable without injecting H32 conditionals into the tuned H16 hot path.

## Why

The generic SM100 DSA backward kernel uses an M64 × N64 score tile. For `H=32`, half of the N extent is inactive, but the CTA still pays much of the N64 instruction, storage, and synchronization cost.

The H32 specialization changes the physical execution shape without changing the backward equations or accumulation precision. It preserves the descending sparse-tile traversal used by the generic implementation for dQ accumulation, keeps dKV accumulation in the existing FP32 global workspace, and converts to the requested output dtype only in the existing epilogue.

The TMEM allocation is kept within 416 of the 512 available columns:

| TMEM columns | Purpose |
|---|---|
| `0:32` | score `S` |
| `32:64` | `dP` |
| `64:192` | dKV0/dKV1; reused by dKV2/dKV3 after T2R |
| `192:352` | five persistent dQ fragments, including the 64-wide D576 tail |
| `352:416` | disjoint dKV tail |

This layout avoids extending the lifetime of `S`/`dP`, and the early K release exposes the next random KV gather while the current tile finishes K-independent work.

## Where the speedup comes from, and why M128 was not selected

The first-order gain comes from specializing the tensor-core N extent for 32 heads. Instead of the generic orientation `S[H,KV] = Q[H,D] · Kᵀ[D,KV]`, the H32 kernel performs an A/B swap and computes the mathematically equivalent transposed result `Sᵀ[KV,H] = K[KV,D] · Qᵀ[D,H]`. This places the 64 sparse KV rows on UMMA M64 and the 32 heads on N32; the upstream N64 path otherwise leaves half of the head lanes inactive while still paying most of its instruction, storage, and synchronization overhead. The remaining gain comes from a more balanced steady-state pipeline rather than from changing the equations: K is released immediately after the last K-dependent dQ MMA, the next random KV gather starts while dKV tail MMA/T2R/global atomics are still draining, and dKV2/dKV3 are detached from their aliased TMEM columns before those columns are reused. The 16-warp subwarp gather supplies enough independent sparse loads to make this overlap useful.

An M128 KV-tile prototype was also evaluated because it halves the number of loop iterations and can amortize per-tile control overhead. It did not outperform M64. Doubling the live K/V tile grows dynamic SMEM from about 149 KiB to roughly 221–222 KiB, leaving almost no staging flexibility, while each producer and reducer burst becomes twice as large. More importantly, the next random gather cannot start until the final dQ MMA releases the entire 128-row K tile, so M128 provides half as many early-release/overlap opportunities as M64. Total sparse KV traffic, dKV arithmetic, and global FP32 atomic traffic are not reduced; the larger T2R/atomic burst and higher register/TMEM lifetime pressure therefore outweigh the saved loop overhead. M64 is retained because it gives the better producer/consumer balance and measured latency.

## Related issues

None.

- No public API or signature changes.
- No additional global workspace or saved forward tensor.
- Dispatch changes only for SM100 `num_heads == 32 && head_dim == 576`.
- BF16 and FP16 input/output paths are covered; dKV accumulation remains FP32.
- dQ and dSink remain bitwise identical to the upstream generic kernel in a same-input comparison. BF16 dKV may differ by one output ULP because the specialized schedule changes the arrival order of global FP32 reductions.

## Performance

Environment:

- GPU: NVIDIA GB200 (SM100)
- CUDA: 13.0
- cuDNN Frontend: 1.28.0
- cuDNN backend: 9.15.1 (`91501`)
- CuTe DSL: 4.5.2
- dtype: BF16
- shape: `H=32, Dqk=576, Dv=512`
- attention sink enabled; `topk_length` provided
- 20 warmup iterations and 100 measured iterations
- table values are the mean of a `baseline → candidate → candidate → baseline` run order

Baseline: `develop@a6c89246` using `generic_m64`  
Candidate: `2c69a251` using `h32_m64`

| Sq = Skv | top-k | Upstream (ms) | H32 kernel (ms) | Speedup | Latency reduction |
|---:|---:|---:|---:|---:|---:|
| 4096 | 512 | 1.9741 | 1.1624 | 1.6983× | 41.12% |
| 4096 | 2048 | 7.1245 | 3.8841 | 1.8342× | 45.48% |
| 8192 | 512 | 3.8805 | 2.2613 | 1.7160× | 41.73% |
| 8192 | 2048 | 14.0669 | 7.7088 | 1.8248× | 45.20% |

Benchmark command:

```bash
PYTHONPATH=$CHECKOUT/python python benchmark/dsa/benchmark_dsa_sparse_attention_backward.py \
  --seqlens 4096,8192 \
  --topks 512,2048 \
  --nheads 32 \
  --head-dim 576 \
  --warmup 20 \
  --repeat 100
```

## Numerical comparison

For identical seeded inputs with `Sq=Skv=top-k=512`, `H=32`, `Dqk=576`, `Dv=512`:

| Output | Upstream vs candidate |
|---|---|
| dQ | bitwise identical |
| dSink | bitwise identical |
| dKV | max abs `4.8828125e-4`; RMSE `1.1126e-6`; relative L2 `4.6817e-5` |

The dKV difference is one BF16 output ULP and is consistent with a different cross-CTA FP32 atomic arrival order; accumulation and workspace precision are unchanged.

## Testing

Focused SM100 regression suite:

```bash
python -m pytest -q -s \
  test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm100_auto_dispatch \
  'test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[False-512-576-512-32-acc_dtype0-dtype0]' \
  'test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[True-512-576-512-32-acc_dtype0-dtype0]' \
  'test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm100_zero_topk_length[d576-h32-m64-boundaries]' \
  'test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm100_partial_max_topk[h32-m64]' \
  test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_fp16_sm100_numerics
```

Result: `11 passed`.

Coverage includes BF16 with and without `topk_length`, FP16 numerics, empty and partial top-k rows around the M64 boundary, a partial maximum top-k (`2047`), and auto-dispatch.

Formatting:

```text
black: passed
black-jupyter: passed
```

Compute Sanitizer:

| Tool | Exercised case | Result |
|---|---|---|
| memcheck | H32/D576 wrapper, top-k 512 | `0 errors` |
| racecheck | H32/D576 M64 top-k boundary regression | `0 errors, 0 warnings` |
| initcheck | H32/D576 M64 top-k boundary regression | `0 errors` |

## Summary by CodeRabbit

* **New Features**
  * Added optimized SM100 support for 32-head, 576-dimensional sparse attention backward operations.
  * Added specialized processing for H32/M64 tiling, including sparse-tile computation, gradient accumulation, and top-k reduction.

* **Bug Fixes**
  * Improved handling of zero and partial top-k configurations across supported kernel variants.

* **Tests**
  * Expanded numerical and boundary coverage for H16/M128 and H32/M64 configurations, including FP16 scenarios.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm100.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm100.py
@@ -24,6 +24,8 @@ def _select_sm100_backend(num_heads: int, head_dim: int) -> Tuple[str, int]:
     """Return the tuned SM100 kernel variant and its sparse-row tile size."""
     if num_heads == 16 and head_dim == 576:
         return "h16_m128", 128
+    if num_heads == 32 and head_dim == 576:
+        return "h32_m64", 64
     return "generic_m64", 64
 
 
@@ -197,6 +199,16 @@ def flash_attn_bwd_sm100(
             from .dsa_bwd_sm100_h16 import FlashAttentionDSABackwardSm100H16
 
             kernel_obj = FlashAttentionDSABackwardSm100H16(
+                element_dtype=dtype,
+                head_dim=head_dim,
+                head_dim_v=head_dim_v,
+                block_tile=block_tile,
+                max_topk=max_topk,
+            )
+        elif backend == "h32_m64":
+            from .dsa_bwd_sm100_h32 import FlashAttentionDSABackwardSm100H32
+
+            kernel_obj = FlashAttentionDSABackwardSm100H32(
                 element_dtype=dtype,
                 head_dim=head_dim,
                 head_dim_v=head_dim_v,
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h16.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h16.py
@@ -1101,6 +1101,7 @@ class FlashAttentionDSABackwardSm100H16:
         sSum_OdO: cute.Tensor,
         pipelines,
     ):
+        """Load the CTA's Q/dO tiles and per-head LSE/O·dO scalars into SMEM."""
         tidx, _, _ = cute.arch.thread_idx()
         token_idx, head_block_idx, batch_idx = cute.arch.block_idx()
         local_tidx = tidx % self.threads_per_warp
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h16.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h16.py
@@ -1141,13 +1141,13 @@ class FlashAttentionDSABackwardSm100H16:
         load_mma_QdO_producer_state.advance()
 
         async_copy_atom = cute.make_copy_atom(cpasync.CopyG2SOp(cache_mode=cpasync.LoadCacheMode.ALWAYS), self.acc_dtype, num_bits_per_copy=64)
-        lse_copy_threads = 8
+        lse_copy_threads = self.h_tile // 2
         thr_layout = cute.make_layout((lse_copy_threads), stride=(1))
         val_layout = cute.make_layout((2), stride=(1))
         async_tiled_copy = cute.make_tiled_copy_tv(async_copy_atom, thr_layout, val_layout)
         thr_async_copy = async_tiled_copy.get_slice(local_tidx % lse_copy_threads)
 
-        softmax_rows = 16
+        softmax_rows = self.h_tile
         gLSE = cute.flat_divide(mLSE, (softmax_rows,))
         gSum_OdO = cute.flat_divide(mSum_OdO, (softmax_rows,))
 
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h32.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h32.py
@@ -28,6 +28,7 @@ class FlashAttentionDSABackwardSm100H32(FlashAttentionDSABackwardSm100H16):
         block_tile: int,
         max_topk: int = 0,
     ):
+        """Configure H32/D576 M64 tiling, TMEM offsets, and register budgets."""
         # Initialize the shared H16 machinery with its native K128 shape;
         # every shape-dependent tensor below is then rebuilt for this class.
         super().__init__(
@@ -713,6 +714,7 @@ class FlashAttentionDSABackwardSm100H32(FlashAttentionDSABackwardSm100H16):
         rTopkIdx: cute.Tensor,
         sub_tile_idx: int,
     ):
+        """Atomically accumulate one M128xN64 register fragment into global dKV."""
         tidx, _, _ = cute.arch.thread_idx()
         _, _, batch_idx = cute.arch.block_idx()
         tidx_in_wg = tidx - self.reduce_warp_id[0] * self.threads_per_warp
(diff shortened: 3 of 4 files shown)


icavan:
@Anerudhan Hi , could you please take a look at this PR when you have a chance? Thanks!

Anerudhan:
@cudnn-ci-bot run oss

Anerudhan (approved):

jiayus-nvidia (approved):

PR #767, opened 2026-08-26
767: Ship the SDPA repro tool as `python -m cudnn.repro`

hwanseoc:
Moves `tools/cudnn_repro` into the `cudnn` package so the SDPA reproducer is
available as `python -m cudnn.repro` (and the `cudnn-repro` console script)
without a second `pip install`.

## Why

The tool is pure stdlib with zero dependencies, yet shipped as its own
distribution requiring `pip install -e tools/cudnn_repro`. It now sits next to
`collect_env.py`, the other `python -m cudnn.<tool>` utility. The root
`packages.find` glob (`cudnn*`) picks the subpackage up, so no packaging change
was needed beyond deleting the old `pyproject.toml` and moving the console
script entry to the root `[project.scripts]`.

## What changed

- `tools/cudnn_repro/cudnn_repro/` → `python/cudnn/repro/` (pure move, all
  imports were already relative)
- `tools/cudnn_repro/tests/` → `test/python/repro/`
- New optional `output` positional:

  ```bash
  python -m cudnn.repro sdpa.log            # stdout, as before
  python -m cudnn.repro sdpa.log repro.sh   # write to a file instead
  python -m cudnn.repro --all sdpa.log out  # all commands, blank-line separated
  ```

- Tests now run in CI for the first time — under `tools/` they were not wired
  into any pipeline. Marked `L0` for the pure-Python unit tests, `L1` for the
  closed-loop ones that spawn real GPU SDPA runs and replay them.

## Caveat

`python -m cudnn.repro` imports the parent `cudnn` package, which loads
`_compiled_module`. Triaging a log now needs the wheel installed; before, the
tool ran on any bare Python.

## Testing

- `pytest test/python/repro/` — 45 passed
- `pytest test/python --collect-only` — 53978 collected, no regressions
- All three CLI modes (stdout / output file / `--all` + output file) and the
  `cudnn-repro` console script verified by hand
- L1 closed-loop: `test_sdpa_random_fwd_L0[test1]` passes.
  `test_sdpa_random_bwd_L0[test1]` fails on a payload mismatch — **pre-existing**,
  reproduced on unmodified `develop` via a worktree, not introduced here.

@coderabbitai ignore


YangXu1990uiuc (approved):

Reviewed `1be2541a9d16` against base `64f6b391144c`.

The package move and lazy console entrypoint make sense. I verified the built wheel, both installed CLI entrypoints, and a real forward log/replay round trip. One small correction for the documented CPU-only test path below; no P0 found. No test pipeline is currently attached to this head, so my following standalone `@cudnn-ci-bot run python_tests` comment will request it. For subsequent updates, that exact command should be a comment by itself; a maintainer can help if access is restricted.

**[P2] Keep the documented CPU-only test command outside the parent GPU conftest** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/1be2541a9d1625884da9a464144a52fc20cd9c96/python/cudnn/repro/README.md#L79))

Moving these tests under `test/python` makes this command load the parent conftest, whose `pytest_configure` unconditionally asserts CUDA availability. With GPUs hidden, the advertised CPU-only command exits with an INTERNALERROR before any test runs. Please document `pytest --confcutdir=test/python/repro test/python/repro -m L0` for the unit-only path, or isolate the CPU tests from that parent fixture. The amended command passed all 45 unit tests without a visible GPU.

Evidence: Built and installed the reviewed wheel. With CUDA_VISIBLE_DEVICES empty: documented command exits 3 at the parent CUDA assertion; adding --confcutdir=test/python/repro gives 45 passed, 30 deselected. Attribution: Introduced by relocating the tests beneath the existing GPU-only parent conftest; the unconditional assertion itself predates this PR.

Validation: Built a wheel from 1be2541a9d16 with its matching compiled extension and installed it into an isolated directory; confirmed both cudnn and cudnn.repro import from that installation. Installed module and console entrypoints agree for last/all entries, stdin, and output-file modes, with clean stderr. All 45 L0 tests pass with GPUs hidden and the parent conftest excluded. SM100, Python 3.12, PyTorch 2.13/cu130 and cuDNN 9.25.1: the explicit L1 forward test1 closed-loop case passes, generating a real log, replaying its command, and comparing normalized graph JSON. A throwaway merge onto develop 64f6b391144c completed without conflicts; package discovery and the console entrypoint survive integration.

Limitations: Focused forward replay only; this does not certify the full backward/FP8/MXFP8 replay matrix. The author reports a pre-existing backward payload mismatch, which I did not independently revalidate in this packaging review. No performance comparison needed: the change relocates an offline CLI and adds file output; it does not change the execution hot path. CI requested after this review is pending evidence.

YangXu1990uiuc:

PR #779, opened 2026-08-27
benchmark/e2e: bucket causal_conv1d and the rest of the GDN kernels

YangXu1990uiuc:
## What

`benchmark/e2e/_perfshare.py`'s `categorize()` matched `chunk_gated` and `wy_fast` against
kernels that are actually named `chunk_fwd_kernel_*`, `chunk_bwd_kernel_*`,
`prepare_wy_repr_*` and `recompute_w_u_*`, and had **no entry for the short convolution at
all**. Nine kernels from a Gated DeltaNet training step land in `other` on `develop`:

```
other  causal_conv1d_fwd_kernel        other  chunk_fwd_kernel_o
other  causal_conv1d_bwd_kernel        other  chunk_fwd_kernel_h
other  causal_conv1d_update_kernel     other  chunk_bwd_kernel_dqkwg
other  prepare_wy_repr_fwd_kernel      other  recompute_w_u_fwd_kernel
other  prepare_wy_repr_bwd_kernel
```

## Why it matters

The share table still sums to 100%, so the under-count is invisible in the output. On a
12-layer GDN proxy the leak was 9.2% of kernel time: `linear_attn` read **15.1%** when the
GDN kernels actually cost **24.0%**, and the short convolution had no row at all. This
benchmark's output is used to rank components for investment, so a component reading low —
or not appearing — is the failure mode that matters most here.

## Changes

- add a `short_conv` category (`causal_conv1d`, `conv1d`) and print it in the table
- complete the `linear_attn` name list: `chunk_fwd_kernel`, `chunk_bwd_kernel`,
  `prepare_wy_repr`, `recompute_w_u`
- make `other` name its contents instead of reporting only a total, so the next upstream
  rename is visible rather than silent
- `benchmark/e2e/tests/test_perfshare_categories.py` pins `categorize()` to kernel names
  taken from a profiler trace (FLA Triton GDN, cuDNN fused GDN, both SDPA arms, GEMM,
  SwiGLU, norm). It fails on the pre-fix table with all nine names above.

No change to timing, model construction, or any measured value — only which bucket a
kernel is counted in.

## Test

```
$ python -m pytest benchmark/e2e/tests/test_perfshare_categories.py -q
3 passed

$ git stash && python -m pytest benchmark/e2e/tests/test_perfshare_categories.py -q
2 failed, 1 passed
FAILED ...::test_observed_kernels_are_categorized
FAILED ...::test_nothing_observed_falls_into_other
```

[...]

PR #782, opened 2026-08-27
benchmark/e2e: per-block perf share for every published Qwen3.5/3.6/3.8 configuration

YangXu1990uiuc:
## What

`benchmark/e2e/_blocks.py` plus a `Qwen3.5/` leaf that times one block of each distinct type
on a random residual stream and weights it by the model's real layer counts. Adds the eight
published Qwen3.5 architectures; Qwen3.6-27B and Qwen3.6-35B-A3B are field-for-field identical
to their Qwen3.5 counterparts and are aliases.

```
python benchmark/e2e/Qwen3.5/run_blocks.py --model Qwen3.5-27B --seq 16384
```

This is an addition, not a replacement. `Qwen3.8/run_model.py` stays: the `--accelerate_mlp` /
`--accelerate_attn` / `--full_attn_backend` treatments and the factorial leaves are whole-model
measurements and remain the right tool for those.

## Why

Three approximations disappear, and one class of configuration becomes measurable at all.

**Memory.** Qwen3.5-397B-A17B carries 12 GB of experts per MoE layer; four layers is 48 GB
before activations. One block at a time makes every published configuration reachable on a
single GPU.

**Attention shape.** FLA's attention derives `head_dim = hidden_size // num_heads`. Six of the
eight published architectures have `num_attention_heads * 256 > hidden_size`, so an assembled
proxy cannot express the published pair — the current 27B preset documents using 20 heads where
the model has 24. A block sets `head_dim` directly. There is a test pinning this.

**The LM head.** Measured once at the real 248320 vocabulary instead of scaled down to preserve
its FLOP ratio, which is how it becomes visible that it is **24% of a composed Qwen3.5-0.8B step
and 3.6% of a Qwen3.5-27B one**.

Forward and forward+backward are timed separately, so a recipe with several forward passes per
backward — GRPO runs three — is a weight (`--forwards-per-backward`) rather than a different
model.

## Measured

seq 16384, bs 1, B200, bf16, stock FLA Triton GDN + torch SDPA:

| model | linear_attn | full_attn | head | composed step |
|---|---|---|---|---|
| Qwen3.5-0.8B | 41.9% | 33.6% | 24.5% | 405 ms |
| Qwen3.5-2B | 48.5% | 27.2% | 24.3% | 562 ms |
| Qwen3.5-4B | 61.6% | 27.7% | 10.6% | 1422 ms |
| Qwen3.5-9B | 66.4% | 23.2% | 10.3% | 2011 ms |
| Qwen3.5-27B | 72.3% | 24.1% | 3.6% | 6933 ms |

Cross-checked against an assembled model at the 27B layer dimensions: 3 GDN + 1 full-attn + head
measured 334.52 ms, the weighted sum of the same blocks 332.46 ms, **-0.6%**.

## A note on how these are measured

Do not reconstruct a block by subtracting whole-model step times. An earlier revision of this
harness did, and on Qwen3.5-0.8B — a ~9 ms block behind a ~98 ms LM head — `t(2 layers) -
t(1 layer)` returned 0.45 ms. Two nearly equal numbers, and the answer was noise: linear
attention read 2.9% of the step instead of 42%, and full attention looked like the largest
component when it is not. Direct measurement of the same block reproduces to within 1% across
runs. Every block here is reachable on its own, so it is reached directly.

## Not in this PR

The MoE block. The runner refuses the three MoE configurations with a reason rather than
reporting a number: a synthetic uniform router produces grouped-GEMM shapes that do not occur
in practice, and routing imbalance is what sets MoE throughput. `transformers`'
`Qwen3_5MoeSparseMoeBlock` gives the right shapes and routing but runs a per-expert loop with
`one_hot` + `index_add_`, so it would reproduce neither the permute/unpermute nor the grouped
GEMM that the bucket exists to capture.

## Test

```
$ python -m pytest benchmark/e2e/tests/test_blocks.py -q
7 passed
```

The tests pin the published dimensions and the fields derived from them, because a wrong
dimension does not fail — it produces a plausible share table for a model nobody ships.

## Checklist

- [x] I have read the [contributing guidelines](../CONTRIBUTING.md)
- [x] The change is covered by tests
- [x] `black --line-length 160 --fast` (pinned 26.3.1) reports no change; pre-commit passes
- [x] License headers present on the new files (Apache-2.0)

---
<sub>note to self: claude::473deaf5-2130-4c34-bbf3-9c8099297886 — "causal conv1d for linear attention: perf vs attention-gym" · cwd /home/scratch.yanxu_libs/cudnn_frontend · workspace /home/scratch.yanxu_gpu/wt-blocks</sub>

## Summary by CodeRabbit

- **New Features**
  - Added CUDA benchmarking for individual model components, including forward and backward timing.
  - Added weighted performance reports showing component timings, contributions, and optional total-time comparisons.
  - Added configuration support for published Qwen3.5, Qwen3.6, Qwen3.8, and Qwen3.8-Flash-Next model variants and aliases.
  - Added profiling categories for short-convolution and additional linear-attention kernels.

- **Bug Fixes**
  - Improved handling and reporting of uncategorized profiling kernels.

- **Tests**
  - Added coverage for model configurations, aliases, timing composition, supported models, and profiling categorization.


YangXu1990uiuc:
Updated: folded in #779, and added the Qwen3.8 configurations, which I had missed — the directory is named after that release.

**Coverage is now 10 architectures and 3 aliases.**

| | conv width | note |
|---|---|---|
| Qwen3.5-0.8B / -2B | 6144 | |
| Qwen3.5-4B / -9B | 8192 | |
| Qwen3.5-27B | 10240 | **= Qwen3.6-27B = Qwen3.8-27B**, one architecture across three releases |
| Qwen3.5-35B-A3B | 8192 | = Qwen3.6-35B-A3B |
| Qwen3.5-122B-A10B / -397B-A17B | 12288 | |
| **Qwen3.8-2.4T-A95B** | **20480** | new: hidden 8192, 92 layers, 128 linear value heads, 64q/4kv |
| **Qwen3.8-Flash-Next** | 10240 | `Qwen4Exp`, **not** in `MODELS` — see below |

#779 is folded in as the middle commit rather than kept as a separate PR; the categoriser is
still live because `Qwen3.8/run_model.py` stays for the `--accelerate_*` treatments and the
factorial leaves, which the block harness does not replace.

### Qwen3.8-Flash-Next is deliberately excluded

It shares the Gated DeltaNet, short convolution and MoE shapes, so it would have slotted into
`MODELS` without complaint — and produced a plausible table that silently omits three
components its decoder layer carries:

- **`Qwen4ExpTextQSAIndexer`** — block-sparse attention selection on the 12 full-attention
  layers: 4 q heads / 1 kv head at head_dim 128, token budget 2048, compress ratio 4, so block
  top-k 512. The same family as the DSA and NSA indexers cuDNN already ships.
- **`Qwen4ExpTextPLELayer`** — n-gram embedding plus a *dilated* depthwise convolution
  (kernel 4, dilation = ngram_size = 3). One site: `ple_layer_ids` is `[2]`, layer 2 of 48.
  `transformers` declines to route it through the short-conv kernels because of the dilation,
  and no dilation parameter exists in cuDNN's causal-conv1d API, in FLA's, or in
  attention-gym's.
- **`Qwen4ExpTextGatedResidual`** — hyper-connections in place of the plain residual adds,
  twice per layer.

The runner refuses it and names all three; the parameters are recorded in `QWEN4EXP` so a
future runner does not have to rediscover them. It needs its own leaf, which is follow-up work
alongside the MoE block.

```
$ python benchmark/e2e/Qwen3.8/run_blocks.py --model Qwen3.8-Flash-Next
Qwen3.8-Flash-Next is a Qwen4Exp model. Its Gated DeltaNet, short convolution and MoE block
match the shapes here, but its decoder layer also carries qsa_indexer, ple, gated_residual,
and composing it from the Qwen3.5 blocks alone would omit all three while still printing a
plausible table. It needs its own runner; see QWEN4EXP in _qwen_configs.py for the parameters.

$ python -m pytest benchmark/e2e/tests/ -q
25 passed
```

YangXu1990uiuc (commented):

Reviewed `4542d15e8313` against base `2a0343f079b0`.

Direct block timing and explicit exclusions are a useful direction, but the current component shares are not yet faithful to the advertised model/recipe. I independently confirmed four existing CodeRabbit findings below; please address them before relying on or refreshing the performance tables. Constructor/operation parity tests would add more value here than further config-table assertions.

- **[P1] Still open:** [Construct the declared GDN value-head geometry](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559565).

- **[P1] Still open:** [Measure the complete declared full-attention block](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559553).

- **[P2] Still open:** [Apply the forward-count weighting to the head too](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559572).

- **[P2] Still open:** [Release the setup graph before timing the block](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559544).

Validation: Reviewed all six changed files and the current upstream FLA/Qwen contracts; the PR merges cleanly with develop, and the three core new benchmark files are identical in that prospective merge. Exact-head benchmark suite:25 tests passed, with25 subtests. CPU/meta probes reproduced the geometry mismatches, head-weighting omission and retained setup graph. SM100 smoke:BF16 Qwen0.8B full-attention block and the real-vocabulary head both produced finite forward outputs and parameter gradients at B=1,S=64. This establishes execution only. Exact-head Style check passed. The standard test CI targets do not collect benchmark/e2e/tests, so I ran its targeted suite locally.

Limitations: I did not run the full FLA GDN GPU path:the constructor probe executes its pinned upstream initializer with metadata-only conv/norm placeholders. The new runner has no baseline-equivalent block benchmark, and its model-fidelity findings make a reported model-performance comparison premature. The GPU smoke is not validation of the published16K/B200 timing table.

vedaanta (approved):

Reviewed `4542d15e8313` against base `9b9f32781d0c`.

Thank you for your contribution! The four existing benchmark findings remain reproducible. The configuration tests pass, but they do not establish that the instantiated blocks match the advertised models or recipe. Please correct these issues before relying on the component-share tables. Under the configured risk-v2 policy these remain nonblocking: the two P1 findings are confined to an optional benchmark estimator, with no demonstrated cuDNN runtime corruption, unsafe memory access or core-path regression. Approval does not validate the reported performance numbers.

- **[P1] Still open:** [Construct the declared GDN value-head geometry](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559565).

- **[P1] Still open:** [Measure the complete declared full-attention block](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559553).

- **[P2] Still open:** [Apply the forward-count weighting to the head too](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559572).

- **[P2] Still open:** [Release the setup graph before timing the block](https://github.com/NVIDIA/cudnn-frontend/pull/782#discussion_r3876559544).

Validation: Reviewed all six changed files. Isolated integration onto develop 9b9f32781d0c merges cleanly; the new runner, timing helper and configuration table are unchanged by the merge. Both new test files pass at exact head and at integration: 11 tests plus 18 subtests in each run. Fresh constructor-argument/meta checks reproduce the 27B value/conv mismatch 18432/22528 versus 6144/10240, and Q projection 6144 versus 12288. Checked the primary contracts in [FLA's pinned initializer](https://github.com/fla-org/flash-linear-attention/blob/516143e31fce09925e6c39ac37148444bad176c4/fla/layers/gated_deltanet.py#L109), [Transformers' pinned Qwen attention](https://github.com/huggingface/transformers/blob/6cd7ae39bdf89ece8cc11761293ed28a4b70b838/src/transformers/models/qwen3_5/modeling_qwen3_5.py#L677), and [Qwen's published 27B config](https://huggingface.co/Qwen/Qwen3.5-27B/blob/main/config.json). Fresh runner-weighting probe confirms head time is unchanged between one and three forwards per backward. Fresh saved-tensor lifetime probe confirms the unused sizing-forward graph stays live across both timing phases. Current-head Style passes. Merge-requirements annotations report missing Milestone and Project metadata. No pending fork workflow needs approval. Standard test CI does not collect these benchmark tests; broad CI is not useful while the identified benchmark-fidelity fixes remain outstanding.

Limitations: No new GPU/model-throughput claim: constructor kwargs were captured and evaluated using verified upstream FLA equations, not a full GDN forward. Full-attention modules were instantiated on the meta device. The reported 16K/B200 share table has not been reproduced or qualified. Reused the earlier same-head review's limited GPU execution evidence only as context; it does not resolve these fidelity findings.

PR #487, opened 2026-08-05
Add HSTU attention CuTe DSL kernels for Blackwell

jiayus-nvidia:
## Summary

- Add packed variable-length HSTU attention forward and backward CuTe DSL kernels for Blackwell SM100/SM103 GPUs.
- Support FP16/BF16 head dimensions 64, 128, and 256; full, causal, local, and arbitrary masks; paged-KV forward; and strided or preallocated gradient outputs.
- Add the experimental FE API, lazy top-level exports, documentation, and L0/L1 pytest coverage.
- Apply the repository's existing licensing pattern: NVIDIA-owned implementation, API, and test files use Apache-2.0, while seven retained low-level derived helpers use MIT and preserve their original author notices.
- Make explicit-stream execution allocation- and lifetime-safe, reject cross-device PyTorch streams, and validate output overlap with all read-only metadata.
- Preserve the original HSTU development history while adapting the port to the current `develop` package layout and conventions.

## Validation

- `pre-commit run --files <changed files>`
- `python -m compileall -q python/cudnn/hstu_attention`
- `cd test/python && PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=../../python python -m pytest -p no:cacheprovider -q -m "L0 or L1" fe_api/hstu_attention/`
  - 185 passed on NVIDIA B200 (SM100), CUDA 13.2, cuDNN 9.20, PyTorch 2.11, CUTLASS DSL 4.6.1
- License audit: 19 Apache-2.0 files and 7 MIT files in the HSTU package and tests, with no file-level BSD source in the HSTU tree.

Closes #369

## Summary by CodeRabbit

* **New Features**
  * Added experimental HSTU attention support for Blackwell SM100/SM103 GPUs.
  * Added packed variable-length forward and backward APIs with causal, local, arbitrary, and paged-KV masking.
  * Added SiLU score transformation without softmax, automatic block-sparse metadata, and head-dimension-256 backward support.
  * Exposed HSTU operations through the Python API.

* **Documentation**
  * Added comprehensive API documentation, usage details, limitations, and support information.

* **Tests**
  * Added extensive correctness and integration coverage.

* **Chores**
  * Updated licensing and third-party attribution records.

PR #784, opened 2026-08-28
Fix HSTU TMA store synchronization

jiayus-nvidia:
## Summary

- Publish HSTU epilogue shared-memory writes to the async proxy before issuing the SMEM-to-GMEM TMA store.
- Preserve bitwise-deterministic forward results under Blackwell GPU context contention.
- Keep the exact determinism assertion and the TMA output path enabled.

## Root cause

The epilogue wrote `sO` through the generic shared-memory proxy and then synchronized its warps with a named barrier. The barrier orders the participating threads, but the TMA engine reads through the async proxy and requires an explicit `fence.proxy.async.shared::cta`. Without that fence, CUTLASS DSL 4.5.x could occasionally store partially visible `sO` data.

This was exposed by the four-worker CI job as 120 mismatched BF16 elements between two identical same-stream launches, with a greatest absolute difference of `0.0019073486328125`.

## Validation

- NVIDIA B200, four concurrent processes, 1000 invocations per process of `test_current_stream_and_compile_cache`:
  - CUTLASS DSL 4.5.0: pass, 0 mismatches
  - CUTLASS DSL 4.5.1: pass, 0 mismatches
  - CUTLASS DSL 4.6.1: pass, 0 mismatches
- `python3 -m black --check --line-length 160 python/cudnn/hstu_attention/_kernels/hstu_fwd.py`
- `python3 -m py_compile python/cudnn/hstu_attention/_kernels/hstu_fwd.py`
- `git diff --check`

Follow-up to #487.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization during accelerated attention output operations for more reliable results.

### DIFF
--- a/python/cudnn/hstu_attention/_kernels/hstu_fwd.py
+++ b/python/cudnn/hstu_attention/_kernels/hstu_fwd.py
@@ -3121,6 +3121,9 @@ class HSTUAttentionForwardSm100:
                 cute.arch.fence_view_async_tmem_load()
                 pipeline_o.consumer_release(o_consumer_state)
 
+        if const_expr(self.use_tma_O):
+            # Publish regular SMEM writes before the async proxy reads sO for the TMA store.
+            cute.arch.fence_proxy("async.shared", space="cta")
         cute.arch.barrier(barrier_id=EPILOGUE_BARRIER_BASE + stage, number_of_threads=cute.arch.WARP_SIZE * len(self.silu1_warp_ids))
 
         logical_stage_start = m_block * self.kBlockM - offset_dynamic


jiayus-nvidia:
I do not have repository triage or Projects permission to set the required PR metadata. Please add labels cat-bug, mod-cutedsl, and orig-nv-eng; assign milestone Frontend 1.29.0; and add this PR to the appropriate Project board, then re-run merge-requirements.

Anerudhan:
@cudnn-ci-bot run oss

vedaanta (approved):

Issue #676, opened 2026-08-20
[Bug]: DSA sparse_attention_backward (SM90/SM100): NaN gradients on invalid/empty/all-sink top-k rows, GPU deadlock on empty rows, wrong d_sink, and accuracy loss under large logits


wangfanChris:
### Before submitting

- [x] #677

### Affected area

None

### Description

The DeepSeek Sparse Attention backward kernel
(`python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py`,
exposed through `cudnn.DSA.sparse_attention_backward_wrapper`) is numerically
unsafe for several categories of input that are valid and common in practice.
The failures are deterministic and reproduce on a single GPU. Any training run
that hits a padded/empty/saturating top-k row gets NaN gradients (or a hung GPU),
which silently corrupts or stalls the whole step.

Five distinct, independently-reproducible symptoms:

1. **NaN gradients for invalid / padded top-k rows.** When a `topk_idxs` row
   contains entries outside `[0, s_kv)` — the `-1` padding used when the active
   key count is below the allocated `topk` width, out-of-range positive indices,
   or entries past `topk_length` — those lanes are not masked out of the softmax.
   The kernel dereferences the raw index and feeds a garbage score into
   `exp2(score − lse)`, which overflows to `inf` and then `inf·dP = NaN` in `dq`.

2. **GPU deadlock on fully-empty rows.** A query row with `topk_length = 0` and
   the sink disabled (`attn_sink = -inf`) never returns: `torch.cuda.synchronize()`
   hangs indefinitely (>150 s, 100% GPU utilization). The empty-row CTA does not
   reach the barrier that the rest of the launch waits on.

3. **All-NaN `dq` when the sink owns all probability mass.** With
   `attn_sink → +inf` the KV probability is 0 for every key. The log-sum-exp with
   sink computes `+inf − +inf = NaN` (and the delta path computes `0·inf`), so
   every head of `dq` becomes NaN.

4. **Wrong `d_sink`.** The softmax delta `Σ p·dP` is read from the cached BF16
   forward output as `O·dO` instead of being recomputed from `Q·K·dO`. After BF16
   rounding of `O`, the delta that feeds `d_sink` is off by a term that only
   cancels in full precision, so `d_sink` is wrong for **all** heads (observed max
   abs error 9.77e-4 against an FP64 reference; tolerance 2e-5).

[...]

PR #785, opened 2026-08-28
DSA backward SM90: fix three top-k and attention-sink boundary failures

SuperGoodGame:
## Before submitting

- [ ] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*`. *(External contributors cannot set labels — requested in a comment below: `cat-bug`, `mod-cutedsl`, `orig-external`, matching #439.)*
- [ ] I set the **Milestone** and **Projects** fields in the sidebar (required to merge; maintainers can set these for external contributions).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Three independent failures in `python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py`, each reachable from documented inputs, plus regression coverage for all of them.

**1. A nonpositive `topk_length` corrupts memory or hangs.** `n_block_max` is 0 and `n_block` is `-1`, but WG0 still runs its unconditional first n_block, so the KV gather indexes `topk_idxs[-64 + row]` out of bounds and dereferences the result as a KV row. WG1 meanwhile runs zero mainloop iterations, leaving WG0 to wait alone on `G4_half_ready` and `sdS_consumed`, both 256-thread named barriers. Whichever lands first decides whether the symptom is `cudaErrorIllegalAddress` or a hang. WG1's `acc_dQ_2/3` are also never zero-initialised in this case (`zero_init=first_iter` runs only on the first mainloop iteration), so the epilogue TMA-stores stale registers into the caller's `dq`.

**2. Padded top-k columns are never masked.** The gather zero-fills their KV row in SMEM, so their score is exactly `0` rather than `-inf` and their probability comes out as `exp2(-LSE)`. For a sufficiently negative LSE that overflows to `+inf`, and GEMM4 multiplies it by the zeroed KV row, so `inf * 0 = NaN` propagates across the entire dQ tile. Both padding layouts are affected: a compact `topk_length` whose tail does not fill the 64-row tile, and the non-compact layout where `-1` marks padding.

**3. A saturating `attn_sink` NaNs every gradient.** Folding the sink into the LSE uses a max-shifted logaddexp; once `fmax(lse_log2, sink_log2)` is infinite the shift evaluates `inf - inf`. `attn_sink` does not have to be infinite — the `log2(e)` rescale saturates for any finite `|sink| > 3.4e38 / log2(e) ≈ 2.36e38`. This hits both the `d_sink` weight and the LSE handed to the main kernel, so all of dQ, dKV and d_sink become NaN.

## Why

**1 — a symmetric guard rather than an early exit.** `topk_length` is CTA-uniform: both warpgroups read `mTopkLength[batch_idx, seq_idx]` for the same scheduler tile, so guarding both sides on `topK > 0` leaves every cross-warpgroup barrier unarrived on both sides instead of one side waiting alone. The mainloop is structurally untouched, and the existing TMA epilogue is reused to write the zero tile, so no raw `mdQ` has to be plumbed into the kernel and the zero store inherits the same addressing, row predication and d=576 tail handling as the normal path. Zeroing the four dQ accumulators in the guarded branch also removes the stale-register store described above — that is the same defect's other half, not a separate change.

A second-order hazard had to be handled. Guarding only the barriers leaves WG0's Q/dO TMA in flight, and it lands in `sQ`, which WG1's epilogue writes. On the mainloop path the `sP_ready`/`sdS_ready` handshake transitively orders that load ahead of both epilogues; with the mainloop skipped there is no such ordering, and the in-flight load overwrote the zeros WG1 had already stored to `sQ[256:]` — `dq[256:576]` came back holding `q`, nondeterministically. The guard therefore covers WG0's whole prologue; an empty row needs no Q, dO, LSE or dP_sum.

**2 — mask in the softmax, where the padding is already known.** A column's KV coordinate comes from a loop-invariant identity tensor built once per CTA, mirroring the existing pattern in `scatter_dkv_atomic`. The compact-tail test is emitted only for the peeled first n_block (`is_first` is a trace-time constant), so it costs nothing in the steady-state loop. The negative-index test sits under `const_expr(not self.have_topk_length)` and is not emitted at all when `topk_length` is supplied, matching the contract documented at `dsa_bwd_sm100.py:321` ("None means non-compact (use full topk, -1 entries in topk_idxs)").

**3 — keep the value finite at the two sites that saturate.** `p_sink` becomes a sigmoid, algebraically identical to `exp2(sink - logaddexp2(lse, sink))` but finite when either term saturates. The LSE-with-sink logaddexp now shifts by its maximum only while that maximum is finite; an infinite maximum is already the answer.

## Related issues

Related to #676. This addresses the empty-row hang, the padded/invalid top-k NaN and the saturating-sink NaN. The remaining two items in that issue are analysed under "Not addressed" below and deliberately left unpatched.

- No public API, signature or kernel-parameter change.
- `dq` and `dkv` are bit-identical to `develop` on ordinary inputs, verified by dumping both builds on the same inputs and comparing.
- `d_sink` moves by at most **2.5e-6 relative** on ordinary inputs — the sigmoid rewrite of an algebraically identical expression, well inside the 5e-2 tolerance the existing tests use.
- Performance (H200 / SM90, whole `sparse_attention_backward_wrapper` wall clock, same-GPU interleaved A/B, best of 5 reps × 30 iterations, d=576 h=64):

  | case | `develop` | this PR | delta |
  |---|---|---|---|
  | compact, topk=1024 (16 n_blocks) | 2.1681 ms | 2.1667 ms | **−0.07% (noise)** |
  | compact, topk=64 (1 n_block, worst case for the tail mask) | 2.1660 ms | 2.1736 ms | **+0.35%** |
  | non-compact, topk=1024 (no `topk_length`) | 2.3829 ms | 2.4037 ms | **+0.87%** |

  The tail-mask cost amortises as `1/n_block_max`, hence nothing measurable at topk=1024 and +0.35% at topk=64. The +0.87% is the per-column `topk_idxs` read and applies only to the non-compact layout; that code is not emitted when `topk_length` is supplied. Hoisting the read out of the row loop was implemented and measured no better (2.4058 ms vs 2.4037 ms), so the simpler form was kept. Happy to split the negative-index half into its own PR if that trade is not wanted here.

## Testing

```
cd test/python
pytest fe_api/dsa/ -q          # 71 passed, 61 skipped, 14 deselected   (H200, SM90)

pre-commit run --files \
  python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py \
  test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py            # clean
```

Eleven test cases newly execute on SM90 — four from widening the existing zero-top-k test, seven new. **Ten of the eleven fail on `develop`**; the eleventh (`saturating_attn_sink[neg-inf]`) passes there and is a companion guard for the negative side, which `fmax` never sends through the NaN path.

- `test_DSA_sparse_attention_backward_zero_topk_length` — the SM100 test from #439, generalised rather than duplicated: renamed off `sm100_`, gate widened from `major*10+minor < 100` to `major < 9` (matching `api.py`'s own "requires SM90+"), and one `d576-mixed` parameter added so empty and non-empty rows land in adjacent CTAs, which is what catches a warpgroup-asymmetric skip. On `develop` all four parameters die with `cudaErrorIllegalAddress`.
- `..._sm90_padded_topk_columns_contribute_zero` — 4 parameters over {d512-h64, d576-h32} × {compact, non-compact}, asserting finiteness and agreement with the reference. Self-validating: it asserts `lse.max() < -80`, so it cannot silently stop exercising the overflow. On `develop` all four fail on `assert torch.isfinite(dq).all()`.
- `..._sm90_saturating_attn_sink` — 3 parameters {2.4e38, +inf, -inf}. For a positive saturating sink it also pins the limit: the sink takes the whole denominator, so `dq == 0`, `dkv == 0` and `d_sink == -sum(dP_sum)`.

Beyond the suite:

- 800 launches of the empty-row path in one process (400 all-empty + 400 mixed, d=576), 0 failures. The `sQ` overwrite described above was nondeterministic, so process-level repeats alone were not sufficient evidence.
- The saturating-sink limits were checked numerically: with `sink = +inf` the kernel returns `dq == 0`, `dkv == 0` and `d_sink == -sum(dP_sum)` to 1.2e-7.

Not tested: SM100 / SM100-h16, no such GPU available here — see below.

## Alternatives considered

- **#439's shape (early exit plus a hand-written zero store) was not ported.** SM90 splits dQ by column across the two warpgroups and writes it with a TMA epilogue, so exiting before that epilogue would skip the dQ store entirely and leave the caller's buffer untouched — exactly what #439's own NaN-filled-buffer assertion is designed to catch. It would also have required plumbing the raw gmem `mdQ` into the kernel, since the parameters named `mdQ`/`mdQ_64` are the TMA tensors. Reusing the epilogue avoids both. (For the record, SM90's tile loop is not persistent — `SingleTileScheduler.advance_to_next_work()` only clears `_is_first_block` — so an early exit would not have dropped tiles; the dQ store is the reason, not tile loss.)
- **SM100's pre-negated `scaled_lse`** (`dsa_bwd_sm100.py:713`) does not help here. It is a representation choice that pairs with `fma_packed_f32x2`, and the negation happens two lines *after* the NaN is produced, so it fixes neither failure; on SM90 `a*b - c` already maps to a single FFMA with a negated addend.
- **No guard was added for a negative `topk_idxs` entry inside `[0, topK)` when `topk_length` is supplied.** Per `dsa_bwd_sm100.py:321` that layout is compact and every entry in range is valid, so the check would sit on the hot path guarding against out-of-contract input. Glad to add it if the contract is meant to be looser.

## Not addressed (context for #676)

- **`d_sink` precision — not a formula bug; the precision is lost upstream, in the forward.** Against an fp64 reference the kernel reproduces the fp64 result *recomputed from the BF16-stored `out`* to **7.06e-9**, and that BF16 quantisation of `out` is the whole of the 5.33e-5 gap against an fp64-exact `out`. So the kernel implements the standard FlashAttention `delta = O · dO` identity exactly, and the reduction at `:317` is already FP32 — nothing in the backward can recover precision the forward discarded before `out` was handed over.

  There is real accuracy on the table, though. Feeding two different `delta` values into an otherwise-fp64 backward (d=576, h=32, topk=128; max relative error vs fp64):

  | `delta` source | delta | dq | dK | d_sink |
  |---|---|---|---|---|
  | `BF16_O · dO` — what the kernel does | 1.4e-3 | 3.1e-4 | 1.4e-4 | 1.1e-3 |
  | `sum_j P_j · dP_j` accumulated in FP32 | 5.6e-7 | 1.4e-7 | 3.0e-8 | 1.6e-7 |

  The second form needs every `P_j·dP_j` before any `dS_j` can be formed, so it costs an extra GEMM1+GEMM2 pass over the top-k (~40% of the mainloop MACs by K/N extent, plus a second KV gather); caching P/dP instead would be tens of GB. That is why no FlashAttention implementation does it, and it is not something to slip into a bug-fix PR. The cheap exact fix is forward-side: FlashMLA already holds `O` in FP32 accumulators and could emit `delta` (s_q × h floats) directly, after which the backward just reads it. Happy to open a separate issue for that if it is useful.

- **Online-softmax / running-maximum rescale — the premise does not hold for this kernel, and the clamp suggested in the issue is unreachable after fix 2.** This kernel has neither an online softmax nor a running maximum; the only `fmax` uses are the two sink logaddexp sites fixed here. Once padded columns are masked, `S*scale - lse <= 0` holds by construction for every live column (the sink only raises the denominator), so the overflow is no longer reachable for self-consistent inputs, and a clamp would cost an `fmin` on the hottest instruction in the kernel.

- **SM100 / SM100-h16 appear to share fixes 2 and 3.** The logaddexp expression is byte-identical at `dsa_bwd_sm100.py:710-712` and `dsa_bwd_sm100_h16.py:678-680`, and I see no column mask before the `exp2` at `dsa_bwd_sm100.py:1908-1909`. I have no SM100 GPU, so I left them alone rather than ship untested changes. Glad to port both if someone can run the tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sparse-attention backward stability for infinite or saturated attention values.
  * Correctly handles empty top-k rows, padded entries, and tail blocks without producing invalid gradients.
  * Prevented non-finite results when masked scores interact with infinite attention sinks.
  * Ensured valid zero gradients for queries with no selected top-k entries.

* **Documentation**
  * Clarified sparse top-k padding semantics, valid index ranges, and length requirements.

* **Tests**
  * Updated regression coverage for empty selections, padding, overflow, and infinite attention sinks.

### DIFF
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm90.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/_interface_sm90.py
@@ -63,8 +63,12 @@ def flash_attn_bwd_sm90(
         dq: pre-allocated (total_S_q, nheads, headdim), optional
         dkv: pre-allocated (total_S_kv, headdim), optional
         d_sink: pre-allocated (nheads,), optional
-        topk_idxs: (total_S_q, topk_max) int32, global indices
-        topk_length: (total_S_q,) int32, per-query valid count, optional
+        topk_idxs: (total_S_q, topk_max) int32, global KV indices.
+            Compact (`topk_length` given): prefix `[0, topk_length)` is valid.
+            Non-compact (`topk_length` is None): a negative entry is padding.
+            Positive indices must be in `[0, S_kv)`; oversized `topk_length`
+            is out of contract.
+        topk_length: (total_S_q,) int32, per-query valid prefix length, optional
         need_d_sink: return and compute d_sink when True
 
     Returns:
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py
@@ -335,15 +335,12 @@ class _FlashAttentionDSABackwardPreprocessSm90:
                         LOG2_E = math.log2(math.e)
                         lse_log2 = lse_row * LOG2_E
                         sink_log2 = mAttnSink[head_idx] * LOG2_E
-                        # Sigmoid form of the two-way softmax weight; stays finite
-                        # when either operand saturates.
-                        p_sink = Float32(1.0) / (Float32(1.0) + cute.math.exp2(lse_log2 - sink_log2))
-                        # inf == inf is true, so both-saturating operands split
-                        # evenly instead of leaving p_sink as NaN from inf - inf.
-                        if lse_log2 == sink_log2:
-                            p_sink = Float32(0.5)
-                        if lse_row == Float32.inf:
-                            p_sink = Float32(0.0)
+                        # KV LSE +inf keeps the develop convention p_sink = 0.
+                        p_sink = Float32(0.0)
+                        if lse_row != Float32.inf:
+                            p_sink = Float32(1.0) / (Float32(1.0) + cute.math.exp2(lse_log2 - sink_log2))
+                            if lse_log2 == sink_log2:
+                                p_sink = Float32(0.5)
                         if row_valid:
                             atomic_add_fp32(
                                 -p_sink * dP_sum[m],
@@ -390,21 +387,14 @@ class _FlashAttentionDSABackwardPreprocessSm90:
                     if cutlass.const_expr(mAttnSink is not None):
                         sink_log2 = mAttnSink[head_idx] * LOG2_E
                         lse_max_log2 = cute.arch.fmax(lse_log2, sink_log2)
-                        # Max-shift only while the maximum is finite; an
-                        # infinite max is already the logaddexp result.
-                        # Seed before the staged ifs (CuTeDSL SSA).
-                        lse_with_sink_log2 = lse_max_log2
-                        if lse_max_log2 == Float32.inf:
-                            lse_with_sink_log2 = Float32.inf
+                        # Infinite max is already the logaddexp; shifting it is inf-inf.
+                        if cute.math.isfinite(lse_max_log2):
+                            sum_exp2 = Float32(
+                                cute.math.exp2(lse_log2 - lse_max_log2) + cute.math.exp2(sink_log2 - lse_max_log2)
+                            )
+                            lse_log2 = lse_max_log2 + cute.math.log2(sum_exp2)
                         else:
-                            if lse_max_log2 == -Float32.inf:
-                                lse_with_sink_log2 = -Float32.inf
-                            else:
-                                sum_exp2 = Float32(cute.math.exp2(lse_log2 - lse_max_log2) + cute.math.exp2(sink_log2 - lse_max_log2))
-                                lse_with_sink_log2 = lse_max_log2 + cute.math.log2(sum_exp2)
-                        if lse == Float32.inf:
-                            lse_with_sink_log2 = Float32.inf
-                        lse_log2 = lse_with_sink_log2
+                            lse_log2 = lse_max_log2
                     gLSElog2[tidx] = lse_log2
 
 
@@ -1164,8 +1154,6 @@ class FlashAttentionDSABackwardSm90:
                 topK = mTopkLength[batch_idx, seq_idx]
             else:
                 topK = self.max_topk
-            n_block_max = (topK + self.tile_n - 1) // self.tile_n
-            topk_tail_rows = topK - (n_block_max - 1) * self.tile_n
 
             mKV_cur = mKV[None, None, head_idx_kv, batch_idx]
             mTopkIdxs_cur = mTopkIdxs[batch_idx, seq_idx, None]
@@ -1178,12 +1166,12 @@ class FlashAttentionDSABackwardSm90:
             load_Q, _, _ = tma_get_copy_fn(tma_atom_Q, 0, cute.make_layout(1), gQ, sQ, single_stage=True)
             load_dO, _, _ = tma_get_copy_fn(tma_atom_dO, 0, cute.make_layout(1), gdO, sdO, single_stage=True)
 
-            # Empty top-k: skip everything WG1 also skips. The 256-thread
-            # barriers (G4_half_ready, sdS_consumed) would otherwise wait on
-            # WG0 alone. The Q/dO TMA lands in sQ, which WG1's epilogue
-            # writes; without the mainloop handshake it races that store.
-            # The epilogue still runs so caller-owned dQ is written.
+            # Skip the WG0 prologue together with WG1 for empty rows.
+            # This avoids asymmetric 256-thread barriers and prevents the Q TMA
+            # from racing with WG1's dQ epilogue in sQ.
             if topK > 0:
+                n_block_max = (topK + self.tile_n - 1) // self.tile_n
+                topk_tail_rows = topK - (n_block_max - 1) * self.tile_n
                 if warp_idx_in_wg == 0:
                     with cute.arch.elect_one():
                         cute.arch.mbarrier_arrive_and_expect_tx(mbar_QdO_ptr, self.tma_copy_bytes["Q"] + self.tma_copy_bytes["dO"])
@@ -1232,7 +1220,8 @@ class FlashAttentionDSABackwardSm90:
 
                 n_block = n_block_max - 1
 
-                # first n_block
+                # Peel the tail block so its validity mask does not enter the
+                # steady-state loop.
                 self._wg0_one_n_block(
                     n_block,
                     wg_tidx,
@@ -1256,12 +1245,10 @@ class FlashAttentionDSABackwardSm90:
                     softmax_scale_log2,
                     softmax_scale,
                     dQ_accumulate=False,
-                    is_first=True,
-                    num_valid_rows=topk_tail_rows,
+                    valid_rows=topk_tail_rows,
                 )
                 n_block -= 1
 
-                # remaining n_blocks
                 while n_block >= 0:
                     self._wg0_one_n_block(
                         n_block,
@@ -1286,8 +1273,7 @@ class FlashAttentionDSABackwardSm90:
                         softmax_scale_log2,
                         softmax_scale,
                         dQ_accumulate=True,
-                        is_first=False,
-                        num_valid_rows=self.tile_n,
+                        valid_rows=self.tile_n,
                     )
                     n_block -= 1
 
@@ -1449,11 +1435,10 @@ class FlashAttentionDSABackwardSm90:
         softmax_scale_log2: Float32,
         softmax_scale: Float32,
         dQ_accumulate: Boolean = False,
-        is_first: Boolean = True,
-        num_valid_rows: Int32 = 64,
+        valid_rows: Int32 = 64,
     ):
         """WG0 one n_block: load KV(cp.async) + GEMM1/2 + softmax/dsoftmax + R2S + GEMM4_WG0(x2)"""
-        if not is_first:
+        if dQ_accumulate:
             cute.arch.barrier(
                 barrier_id=int(NamedBarrierBwd.KV_empty),
                 number_of_threads=self.num_mma_threads,  # 256 = WG0(128) + WG1(128)
@@ -1465,7 +1450,7 @@ class FlashAttentionDSABackwardSm90:
         for r in cutlass.range_constexpr(ROWS_PER_GROUP):
             row = r * NUM_GROUPS + group_idx
             global_topk_row = n_block * self.tile_n + row
-            if row < num_valid_rows or not is_first:
+            if row < valid_rows or dQ_accumulate:
                 if const_expr(self.have_topk_length):
                     self._copy_row(
                         mKV_cur,
@@ -1493,7 +1478,7 @@ class FlashAttentionDSABackwardSm90:
                     else:
                         self._zero_row(sKV, row, idx_in_group)
             else:
-                # clear for OOB rows (only in first n_block)
+                # Tail padding of the peeled n_block.
                 self._zero_row(sKV, row, idx_in_group)
 
         cute.arch.cp_async_commit_group()
@@ -1516,7 +1501,7 @@ class FlashAttentionDSABackwardSm90:
         #
         # Padded columns are zero-filled in sKV, so their score is 0 rather
         # than -inf. Force those lanes to probability zero.
-        # Compact: past the top-k tail (peeled first n_block only).
+        # Compact: past valid_rows on the peeled tail n_block.
         # Non-compact: a negative top-k index. Positive OOB indices and
         # compact entries in [0, topK) are trusted as valid KV rows.
         acc_S_mn = make_acc_tensor_mn_view(acc_S, transpose=self.SdP_swapAB)
@@ -1525,8 +1510,8 @@ class FlashAttentionDSABackwardSm90:
             for c in cutlass.range(cute.size(acc_S_mn, mode=[1]), unroll_full=True):
                 p = cute.math.exp2(acc_S_mn[r, c] * softmax_scale_log2 - tLSErLSE[r], fastmath=True)
                 col = tScS_mn[r, c][COL]
-                if is_first:
-                    p = Float32(0.0) if col >= num_valid_rows else p
+                if not dQ_accumulate:
+                    p = Float32(0.0) if col >= valid_rows else p
                 if const_expr(not self.have_topk_length):
                     # Peeled tile still spans tile_n columns; clamp so a
                     # partial non-compact row is not read past its end.
--- a/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py
+++ b/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm90.py
@@ -1245,7 +1245,7 @@ class FlashAttentionDSABackwardSm90:
                     softmax_scale_log2,
                     softmax_scale,
                     dQ_accumulate=False,
-                    valid_rows=topk_tail_rows,
+                    num_valid_rows=topk_tail_rows,
                 )
                 n_block -= 1
 
@@ -1273,7 +1273,7 @@ class FlashAttentionDSABackwardSm90:
                         softmax_scale_log2,
                         softmax_scale,
                         dQ_accumulate=True,
-                        valid_rows=self.tile_n,
+                        num_valid_rows=self.tile_n,
                     )
                     n_block -= 1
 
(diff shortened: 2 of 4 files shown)


SuperGoodGame:
Labels for this PR (I do not have permission to set them): `cat-bug`, `mod-cutedsl`, `orig-external` — the same set as #439, which is the closest precedent (external contributor, DSA backward, zero-length top-k). Milestone and Projects would also need to be set on your side; the `merge-requirements` check is currently red only for those two.

Two things I would specifically like a maintainer opinion on:

1. **The +0.87% on the non-compact path** (no `topk_length`) buys the negative-index column mask. If that trade is not wanted here I am happy to split it into its own PR and land the rest, which is free on the compact path.
2. **SM100 / SM100-h16 appear to share fixes 2 and 3** (details in the PR body). I have no SM100 GPU so I left them untouched rather than ship untested changes — glad to port them if someone can run the tests.

SuperGoodGame:
@codex review

Focus areas, in rough order of risk:

1. **Warpgroup symmetry.** `topK` is read independently by `mma_wg0` and `mma_wg1` from `mTopkLength[batch_idx, seq_idx]` for the same scheduler tile. The whole fix rests on both sides taking the same branch. Is there any path where the two disagree, or where a named barrier (`sP_ready`, `sdS_ready`, `sP_consumed`, `sdS_consumed`, `G4_half_ready`, `KV_empty`) ends up with an arrival count that differs from `develop` for a non-empty tile?

2. **SMEM lifetime on the guarded path.** The guard skips WG0's Q/dO TMA precisely because the in-flight load lands in `sQ`, which WG1's epilogue writes. Is any other shared buffer (`sKV`, `sdO`, `sP`, `sdS`, `sLSE`, `sdPsum`) written by one warpgroup and read by the other without an ordering point once the mainloop is skipped?

3. **Column masking correctness.** `tScS_mn[r, c][COL]` is assumed to give the KV (N) coordinate of the accumulator element under `make_acc_tensor_mn_view(..., transpose=SdP_swapAB)`, mirroring `scatter_dkv_atomic`. Is `COL = 1 if not SdP_swapAB else 0` right for both orientations? Note `SdP_swapAB` is hard-coded `False` at `_interface_sm90.py:94`, so only one orientation is exercised by the tests.

4. **The `p_sink` rewrite.** `1 / (1 + exp2(lse_log2 - sink_log2))` with an `lse_log2 == sink_log2 -> 0.5` special case, replacing `exp2(sink_log2 - logaddexp2(lse, sink))`. Any finite input where these disagree beyond rounding, or any remaining NaN/Inf combination?

5. **`mbar_QdO_phase`.** It is now assigned only inside the `topK > 0` branch. The tile loop traces once (`SingleTileScheduler` is not persistent), so this should be moot — but please confirm the value carried out of the conditional is well-defined.

SuperGoodGame:
Good catch — this is real, and fixed in 3a1d8da4.

The peeled tile spans all 64 columns even when it is the partial one, so with `max_topk = 100` the load reaches index 127 against an extent of 100 (28 entries past the row); at `max_topk = 70` it is 58 past. Confirmed by walking the index arithmetic on this host — no `compute-sanitizer` available here, so I could not get a fault report.

Worth recording that the *results* were unaffected: lanes at or past the tail are already zeroed by the tail mask immediately above, and this guard only ever zeroes, so whatever the stray entry held could not change `p`. It is a memory-safety defect only, which matches the "Stability & Availability" classification rather than correctness.

Fixed by clamping the index rather than gating the load behind the tail test, which keeps the inner loop free of nested dynamic control flow:

```python
idx = n_block * self.tile_n + col
idx = idx if idx < self.max_topk else Int32(self.max_topk - 1)
if mTopkIdxs_cur[idx] < 0:
    p = Float32(0.0)
```

`self.max_topk` is `topk_idxs.shape[-1]`, which is exactly the extent of `mTopkIdxs_cur`, so the index is in bounds by construction. The clamp is emitted only in the non-compact path, so the compact path is untouched.

The regression test now sizes the non-compact case's `topk_idxs` to `n_valid + 18 = 118`, i.e. a deliberate non-multiple of 64 with a 54-column tail, so the peeled tile addresses past the row unless the clamp holds. `pytest fe_api/dsa/ -q` is still 71 passed / 61 skipped on SM90, and `pre-commit` is clean.

SuperGoodGame:
On the second finding — *"Synchronize both warpgroups after an empty tile"* — the reasoning is sound as a conditional, but the premise it rests on ("the next tile's Q TMA") does not hold for this kernel, and the shape it describes is not something this PR introduces. I would rather not add the barrier; happy to be overruled.

**There is no next tile.** `SingleTileScheduler` maps one CTA to exactly one tile, two independent ways:

- `get_grid_shape()` returns `(num_block, num_head * num_splits, num_batch)` — the entire tile space — and `get_current_work()` returns `tile_idx` straight from `cute.arch.block_idx()`. The mapping is a bijection, so no CTA has a second tile to advance to.
- `advance_to_next_work()` only does `self._is_first_block = False`, a plain Python attribute assignment, and `is_valid_tile` is that same Python bool. `while work_tile.is_valid_tile:` is therefore a trace-time loop that emits exactly one body. (Were it a real `scf.while`, a Python attribute mutation could not act as a loop-carried value and it would never terminate.)

**The normal path has the identical shape.** With `topK > 0`, WG0's last cross-warpgroup barrier for a tile is `sdS_consumed` *before* `epilogue_dQ`, and the first thing it would do on a hypothetical next tile is issue the Q TMA into `sQ` — with no 256-thread barrier between that and WG1's `epilogue_dQ_wg1` TMA store. So if the scheduler ever became persistent, the hazard would exist on the ordinary path too, not just the empty one. It is a latent invariant of the current design rather than a regression here.

**One barrier would not actually make the kernel persistence-safe**, which is my main reason for leaving it out — it would read as assurance that is not there. Two of the named barriers already carry state across tiles today: `KV_empty` and `sP_consumed` each get `K` arrivals from WG1 but only `K - 1` waits from WG0 per tile (WG0's arrive-and-wait at the top of the next n_block doubles as its arrival, and `sP_consumed` is skipped on the first n_block via `dQ_accumulate`). That leaves 128 unconsumed arrivals per tile. Harmless with one tile per CTA; under a persistent scheduler it would drift and release barriers early, independently of `sQ`.

For what it is worth, the `sQ` ordering *is* the thing this PR had to get right within a tile — guarding only the barriers left WG0's in-flight Q TMA overwriting the zeros WG1 had already stored to `sQ[256:]`, which is why the guard covers WG0's whole prologue rather than just the barrier sites. That failure was reproducible (`dq[256:576]` came back holding `q`, nondeterministically) and is covered by the 800-launch stress described in the PR body.

If a maintainer would still like the invariant pinned down, I would suggest a comment at the scheduler construction site stating that the epilogues rely on one tile per CTA, rather than a barrier — and if the intent is to actually allow a persistent scheduler later, that is a larger change than this PR (the two barrier ledgers above would need fixing as well). Glad to do either.

SuperGoodGame:
`_wg0_one_n_block` dropped `is_first`. Develop already passed `dQ_accumulate=False` on the peeled first n_block and `True` on the rest, which was always the opposite of `is_first`, so the compact tail mask now uses `not dQ_accumulate`. `num_valid_rows` is unchanged. The only new parameter is `tScS_mn` (KV-column identity of each S accumulator lane).

jiayus-nvidia:
@cudnn-ci-bot run oss

jiayus-nvidia (approved):

SuperGoodGame:
@jiayus-nvidia  @Anerudhan tks

vedaanta (changes requested):
I think these errors on sm90 are from this PR:
```
=========================== short test summary info ============================
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm90_saturating_attn_sink[pos-inf] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_qh32_uses_per_query_topk_without_padding - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[False-512-512-512-16-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[False-512-512-512-64-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[False-512-576-512-16-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[False-512-576-512-32-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[True-512-512-512-16-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[True-512-512-512-64-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[True-512-576-512-16-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_wrapper[True-512-576-512-32-acc_dtype0-dtype0] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_zero_topk_length[d512-mixed] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_zero_topk_length[d576-mixed] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_zero_topk_length[d576-h16-m128-boundaries] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm90_padded_topk_columns_contribute_zero[compact] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm90_padded_topk_columns_contribute_zero[non-compact] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
FAILED test/python/fe_api/dsa/test_DSA_sparse_attention_backward.py::test_DSA_sparse_attention_backward_sm90_saturating_attn_sink[finite-but-rescale-overflows] - AttributeError: module 'cutlass.cute.math' has no attribute 'isfinite'
===== 16 failed, 113 passed, 4477 skipped, 57 warnings in 72.41s (0:01:12) =====
```

SuperGoodGame:
@vedaanta thanks — those SM90 failures are from `cute.math.isfinite` in the sink LSE fold. That symbol is a newer CuTeDSL API; this repo's `cutedsl` extra is `nvidia-cutlass-dsl>=4.5.0` (not pinned) and the version CI installs does not have it, so compile raises `AttributeError` before any of the tests run.

Replaced it with the same `±Float32.inf` compares this kernel already uses on develop (`46a6d66c`). Behavior is unchanged: an infinite max is already the logaddexp, and shifting it is `inf-inf`.

vedaanta:
@cudnn-ci-bot run oss

vedaanta (approved):

vedaanta:
@SuperGoodGame please feel free to merge after resolveing conflicts

SuperGoodGame:
@vedaanta Conflicts with latest `develop` are resolved — rebased onto f12d0527, keeping your new SM100 invalid-row tests and the `d512-h128-two-cta` param alongside the SM90 cases (that param skips via `_require_exact_sm100()` on SM90; the test itself now gates SM90+).

All checks are green, and the full `fe_api/dsa/` suite passes locally on an SM90 host (70 passed / 76 skipped, SM100 cases skip as designed). GitHub doesn't offer me the merge button as an external contributor — could you merge when you get a chance? Thanks!

PR #791, opened 2026-08-28
791: remove non-functional CUDA IMA guard

hwanseoc:
Remove the `torch.cuda.synchronize` monkeypatch guard from `test/python/conftest.py` (and its landmine note in `test/AGENTS.md`).

The guard hard-exits the entire pytest worker (`os._exit`) on any async CUDA error. Groundwork for multi-GPU xdist testing: with one worker pinned per GPU, a hard exit kills that GPU's whole test shard instead of failing one test. Async CUDA errors now surface as normal test failures.

@coderabbitai ignore


hwanseoc:

PR #794, opened 2026-08-28
remove torch dependency from conftest

hwanseoc:
The shared `test/python/conftest.py` hard-imported torch for the device-availability gate, the pytest-xdist memory gate, the crash-isolation context probe, and the session cudnn handle's stream — even though torch is an opt-in dependency group of the package ("Framework-neutral core") and the `*_jax.py` / driver-level tests run without it.

This is also a prerequisite for multi-GPU test routing (#791 was the first part): nothing in conftest before the future per-worker `CUDA_VISIBLE_DEVICES` pin point may initialize CUDA. With this PR, nothing at import time touches CUDA at all; the first cudart call happens in `pytest_configure`, after the pin point.

## Changes

- **`test/python/conftest.py`**: drop the module-level `import torch`; use cudart via **cuda-python** instead.
  - Device gate → `cudaGetDeviceCount`; banner → `cudaGetDeviceProperties`.
  - Mem-gate → `cudaMemGetInfo`; torch-allocator work (`empty_cache`, `OutOfMemoryError` detection) is done lazily via `sys.modules.get("torch")` — unchanged when tests use torch, no-op otherwise.
  - Crash-guard probe (`_dead_cuda_context`, from #639): `torch.cuda.is_initialized()` → `cuCtxGetCurrent()` (non-null only if this thread already holds a context, so the probe never creates one), `torch.cuda.synchronize()` → `cudaDeviceSynchronize()` with the error string from `cudaGetErrorString`. Bonus: now also covers tests that touch CUDA without torch (cutedsl / cuda-python direct).
  - Handle fixture stream → `cudaStreamCreateWithFlags(cudaStreamNonBlocking)` (matching `torch.cuda.Stream()` semantics), destroyed at teardown.
- **`pyproject.toml` / `requirements.txt`**: add `cuda-python` to the dev dependency group (previously only pulled in via the `cutedsl` extra).

Per review feedback, the cudart calls live in the test harness via cuda-python rather than new bindings in the compiled module — nothing new to maintain in `pycudnn`.

## Testing

On sm100 (cuDNN 9.27, CUDA 13, cuda-python 13.3.1), rebased on develop @ `3552ee771`:
- torch venv: `test_mhas_v2.py::test_sdpa_random_fwd_L0` full sweep under `-n 2` (245 passed) — exercises the mem-gate path; default (crash-isolated, `-n 1` injected) run passes.
- torch-free venv (no torch installed): conftest loads, banner shows `PyTorch: not installed`, `test_device_info.py` + `test_import_boundaries.py` pass default and under `-n 2`.
- crash-guard probe checked directly: returns `None` before any CUDA use and leaves the driver uninitialized; `None` after healthy torch work (sees torch's context via the driver); returns `device-side assert triggered` after a torch out-of-bounds index poisons the context.
- `black`: changed lines clean; remaining complaints on conftest are pre-existing at HEAD.

@coderabbitai ignore

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,6 +90,7 @@ dev = [
     "pybind11[global]>=2.13,<3",
     "pytest",
     "pytest-xdist",
+    "cuda-python",
     "looseversion",
     "black==26.3.1",
     "clang-format==21.1.6",
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ numpy<2.0.0
 pybind11[global]>=2.13,<3
 pytest
 pytest-xdist
+cuda-python
 looseversion
 black==26.3.1
 clang-format==21.1.6
--- a/test/python/conftest.py
+++ b/test/python/conftest.py
@@ -31,7 +31,20 @@ except (ImportError, OSError):
     pass
 
 import cudnn
-import torch
+
+# cudart via cuda-python instead of torch: torch is optional, and startup must
+# not create a CUDA context so per-worker CUDA_VISIBLE_DEVICES routing works.
+import cuda.bindings.driver as cuda_driver
+import cuda.bindings.runtime as cudart
+
+
+def _cudart_call(fn, *args):
+    err, *rest = fn(*args)
+    if int(err) != 0:
+        raise RuntimeError(
+            f"{fn.__name__} failed: {cudart.cudaGetErrorString(err)[1].decode()}"
+        )
+    return rest[0] if len(rest) == 1 else tuple(rest)
 
 # fmt: off
 
@@ -84,13 +97,15 @@ def _dead_cuda_context():
     # A sticky context error is returned by every later CUDA call, so one
     # synchronize is both the cheapest and the most complete probe; it also
     # surfaces async faults the test itself swallowed.
-    if not torch.cuda.is_initialized():
+    # Probe only if this thread already holds a context; cudaDeviceSynchronize
+    # would otherwise create one in a worker that never touched the GPU.
+    err, ctx = cuda_driver.cuCtxGetCurrent()
+    if int(err) != 0 or int(ctx) == 0:
         return None
-    try:
-        torch.cuda.synchronize()
-    except Exception as e:
-        return str(e).strip().splitlines()[0]
-    return None
+    (err,) = cudart.cudaDeviceSynchronize()
+    if int(err) == 0:
+        return None
+    return cudart.cudaGetErrorString(err)[1].decode()
 
 
 def pytest_runtest_logstart(nodeid, location):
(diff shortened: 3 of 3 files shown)


YangXu1990uiuc (approved):

Reviewed `5bb93bc41bd0` against base `64f6b391144c`.

Moving the shared harness to cuda-python makes sense, and the torch-free execution and real crash-isolation paths work in my focused checks. One nonblocking error-path correction below; no P0 found. Before merge, please rebase the conftest conflict while keeping develop's `_log_to_real_stderr` and this PR's `_torch_empty_cache`. The current Style failure is on the newly added `_cudart_call` formatting (one-line RuntimeError plus a blank line), so it is worth running the formatter on this file. After rebasing, request test CI with a standalone `@cudnn-ci-bot run python_tests` comment; a maintainer can help with access. I did not request a pipeline against the known merge conflict.

**[P2] Do not treat every cuCtxGetCurrent error as a healthy/no-context result** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/5bb93bc41bd0cf9b05b55b9d6d5b40c98f239423/test/python/conftest.py#L104))

`cuCtxGetCurrent` can itself report an earlier asynchronous CUDA failure, but this branch converts every non-success into `None`. In that case logfinish disarms the watchdog and keeps the worker, defeating the crash guard. Please special-case the expected not-yet-initialized state, and return a diagnostic for other driver errors (using the driver error-string API); success with a null context can still return `None`. The [CUDA context API documentation](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CTX.html) explicitly permits errors from earlier asynchronous launches.

Evidence: Injecting CUDA_ERROR_ILLEGAL_ADDRESS at the driver-query boundary makes the reviewed function return None. This is an API-error-path reproduction, not a claim that the ordinary device-side-assert path fails: my actual GPU assert returned a successful context query and was detected correctly by cudaDeviceSynchronize. Attribution: Introduced by the new unconditional early return on any driver-query error.

Validation: Built a matching FE extension for 5bb93bc41bd0 and used a genuinely torch-free Python 3.12 environment with pytest 9.1.1, pytest-xdist 3.8.0 and cuda-python 13.4.1. Device-info/import-boundary suites: 11 passed, 1 optional DSL test skipped, both with default isolation and -n 2. The baseline fails immediately on its hard torch import. Fresh-process checks: importing conftest and probing before CUDA use preserve the uninitialized driver state; a healthy driver-only context is handled without importing torch. SM100, cuDNN 9.25.1, PyTorch 2.13/cu130: two forward SDPA cases pass under -n 2, exercising the memory-gate path. A real device-side assert is correctly detected. Private fault-injection suite: the expected poisoned-worker failure is reported (pytest exit 1), xdist replaces that worker, and both subsequent GPU tests pass in a different PID. This is successful recovery evidence, not an all-green suite. Inspected the current public Style failure and an isolated merge against develop 64f6b391144c; the only integration conflict is the conftest OOM-retry logging/cache-clear block.

Limitations: Focused harness validation; no full Python/GPU matrix or CUDA 12 environment was rerun. No product hot-path performance comparison is warranted for this test-harness/dependency change. The injected driver-query error verifies an API-permitted failure branch; the real GPU fault exercised the successful-query plus failing-synchronize branch. Test CI still needs to run after the owner resolves the merge conflict and formatting.

hwanseoc:
@cudnn-ci-bot run

hwanseoc:
@cudnn-ci-bot run frost,python_tests,oss

YangXu1990uiuc (commented):

Reviewed `74160a41393b` against base `61070749e3f4`.

The rebase preserves the previously validated torch-free harness and retains develop's crash diagnostics. No new approval blocker found. My earlier driver-query error-handling P2 remains open. Existing approval is retained. Please also apply Black's small conftest formatting correction; that is the current Style failure. The owner already started exact-head pipeline 69550198, so I have not launched another run.

- **[P2] Still open:** [Do not treat every cuCtxGetCurrent error as a healthy/no-context result](https://github.com/NVIDIA/cudnn-frontend/pull/794#pullrequestreview-5204881882).

Validation: Seven relevant helpers, including the driver probe, stream fixture, optional torch handling and pytest_configure, are AST-identical to the prior reviewed head. Inspected the rebased logging integration and base additions separately. Exact-head torch-free environment: device-info/import-boundary suites passed with 17 passed and 1 optional-dependency skip in each of default isolation and -n 2. Verified torch is absent and the imported FE source is this checkout; native bindings were reused only after native-source equality checks. Fresh-process conftest import and pre-CUDA probe preserve the uninitialized driver state; a healthy driver-only context works without torch. Repeated the driver-error injection supporting the linked P2. Verified pipeline 69550198 mirrors the full reviewed SHA. Its completed SM107 FROST lane has 3230 passed, 2011 skipped and 12 strict XPASS outcomes from inherited native ragged-decode expectations; no crash-recovery failure was reported in that lane. Other selected jobs remain in progress. Current Style log isolates Black changes to _cudart_call formatting and the separating blank line. No product hot-path performance comparison is needed for this test-harness change.

Limitations: Focused follow-up: the prior actual poisoned-worker restart evidence is reused for unchanged code; no full local GPU suite or fresh CUDA 12 run was repeated. CI is still in progress. The inherited strict-XPASS guard drift is distinct from this PR's remaining driver-query P2.

PR #803, opened 2026-08-29
Fix FROST linear-attention compiled cache device identity

YangXu1990uiuc:
## Before submitting

- [ ] I set the **Milestone** and **Projects** fields in the sidebar (required to merge; maintainers can set these for external contributions). The milestone is set; the current token cannot set organization Projects.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

- Resolve the FROST linear-attention compilation target from the cuDNN handle carried by the plan-build `ExecutionContext`.
- Record that device ordinal and its SM count once on the compiled plan, then pass the same facts to every GDN, GDN2, and KDA prefill, recompute, and backward cache/JIT path.
- Reject an execute-time handle targeting another device before launch.
- Clamp ragged final-chunk GDN backward gate loads to the last valid token before the dynamic select, preventing invalid lanes from forming out-of-range GMEM addresses.

These are correctness fixes. The device-identity change does not alter kernel math, and neither change claims model-level performance.

## Why

Equivalent linear-attention shapes can run on multiple CUDA devices in one process. The compiled callable bakes device-dependent launch facts such as `num_sm`; a shape-only cache could otherwise reuse device 0's callable on another GPU. The build handle is the authoritative device, rather than ambient `torch.cuda.current_device()`, so cache identity, compilation, and execution now describe the same target.

## Related issues

None.

There is no public operator-signature, accepted-shape, dtype, or architecture change.

FROST linear-attention plans are now explicitly device-bound. Reusing a built plan with a handle for another GPU raises `ValueError`; multi-GPU callers build one plan per device. Calls without an explicit handle retain the existing current-device fallback. Same-device behavior is unchanged.

## Testing

- ComputeLab B200, cuDNN 9.26: generic dispatch plus focused handle/device-contract tests — **62 passed**.
- ComputeLab B200 at `a442b7b84`: all six FROST examples passed: GDN, GDN2, and KDA forward/backward; both GDN and KDA backward exercised checkpoint recompute.
- ComputeLab B200: the new explicitly L0-classified public GDN backward regression for a 68-token ragged sequence (four live lanes in the final 64-token chunk) passed at current head `bf82311bb`; the related varlen selection was **4 passed** at the kernel-identical prior head.

```bash
for example in test/python/linear_attention/frost/examples/0{1..6}_*.py; do python "$example"; done
python -m pytest -q test/python/linear_attention/test_la.py::test_gdn_bwd_partial_final_chunk
```

- Host compile/static coverage, changed-file pre-commit, and diff checks passed at `bf82311bb`.

The prior gate code can still pass an ordinary numerical test because the later select masks the invalid-lane value and nearby allocator mappings often hide the physical over-read. The address expression itself was statically out of range; the fix makes every eagerly formed load address valid. No private generated-code string is asserted.

The focused tests verify the observable plan contract: the build handle wins over ambient-device fallback, a mismatched execute handle fails before launch, and handleless building keeps the fallback behavior. They do not pin private cache tuple layouts or leaf-function signatures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA device handling for FROST plan execution, preventing launches on mismatched devices.
  * Ensured operations consistently use the hardware configuration captured during plan initialization.
  * Added support for float16 and bfloat16 gate tensors in GDN2 prefill operations.
  * Improved handling of partial sequence chunks and boundary gate reads.
* **Tests**
  * Added coverage for matching, mismatched, and handleless execution-device scenarios.
  * Added regression coverage for partial-chunk GDN backward processing.

### DIFF
--- a/python/cudnn/linear_attention/frost/engine.py
+++ b/python/cudnn/linear_attention/frost/engine.py
@@ -75,6 +75,13 @@ class FrostLaPlan(CompiledPlan):
         return self.compiled.workspace_bytes()
 
     def execute(self, graph, variant_pack, ctx) -> None:
+        handle = ctx.handle
+        launch_device = handle.device.ordinal if hasattr(handle, "device") else variant_pack.device
+        if launch_device != self.compiled.device:
+            raise ValueError(
+                f"{self.compiled.plan_name}: plan was built for cuda:{self.compiled.device}, "
+                f"but the execution context targets cuda:{launch_device}; build one plan per device"
+            )
         ports = self.ports
         if ports is None:
             ports = self.ports = bind_ports(graph, variant_pack)
--- a/python/cudnn/linear_attention/frost/gdn2_engine.py
+++ b/python/cudnn/linear_attention/frost/gdn2_engine.py
@@ -118,6 +118,7 @@ class CompiledGdn2:
         self.table = None
         self.kcache = None
         self.plan_name = "Gdn2FrostEngine (GDN2)"
+        self.device = current_device()
         scale = node.params.get("scale")
         self.scale = float(scale) if scale is not None else 1.0 / math.sqrt(node.inputs["q"].dim[-1])
         self.use_qk_l2norm = bool(node.params.get("use_qk_l2norm", False))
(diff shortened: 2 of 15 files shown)


YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:
@coderabbitai review

YangXu1990uiuc:

YangXu1990uiuc:
@coderabbitai review

YangXu1990uiuc:

YangXu1990uiuc:

jhjpark:
Closing. Absorbed in #890.

PR #804, opened 2026-08-29
Add MiniMax block-128 causal selection attention

YangXu1990uiuc:
## Before submitting

- [ ] I set the **Milestone** and **Projects** fields in the sidebar (required to merge; maintainers can set these for external contributions). The milestone is `Frontend 1.29.0`; Project assignment remains for a maintainer.

## Affected area

FE OSS kernels or CuTeDSL.

## Summary

- Admit block size 128 and GQA ratio 16 in selection attention.
- Add semantic `is_causal` support for token-level masking inside any selected block.
- Keep fully causal-masked selected tiles from introducing NaNs into online softmax state.
- Validate arbitrary sparse and non-monotonic MiniMax-M3 metadata through the public API.
- Include causal mode and tensor device/dtype contracts in wrapper cache identity.
- Fix the existing block-index rank/shape predicate.

This covers the selected-attention geometry used by MiniMax-M3's 57 sparse layers. It consumes already-produced block indices; the Lightning indexer itself remains outside this change.

## Why

MiniMax-M3 uses block size 128, GQA ratio 16, and causal token-level masking within selected blocks. The existing selection-attention admission and cache contracts did not cover that geometry. Extending the existing selection path keeps the change scoped to consuming already-produced block indices rather than adding a model-specific indexer.

Selected blocks are traversed in reverse metadata order. If an entirely future block is encountered before a valid block, every score in that tile is `-inf`; directly evaluating `exp(-inf - -inf)` produces NaNs and poisons the running probability, row-sum, and output state. The fix preserves the true `row_max == -inf` statistic while using zero only as the safe exponent-centering value, so the empty tile contributes zero and a later valid tile recovers normally.

## Related issues

None.

The existing selection-attention API now admits block size 128 and GQA ratio 16 and accepts semantic `is_causal` behavior for token-level masking inside selected blocks. Causal mode and tensor device/dtype contracts are included in wrapper cache identity, and the block-index rank/shape predicate is corrected.

This change does not add the Lightning indexer. Runtime validation reported here is on NVIDIA B200 SM100 with cuDNN backend 9.26.0. No performance claim is made in this PR.

## Testing

From `test/python` on NVIDIA B200 SM100:

```bash
pytest -q -m "L0 or L4" \
  fe_api/nsa/test_NSA_selection_attention.py::test_selection_attention_wrapper_cache_separates_is_causal \
  fe_api/nsa/test_NSA_selection_attention_minimax_m3_probe.py::test_selection_attention_is_causal_on_existing_geometry \
  fe_api/nsa/test_NSA_selection_attention_minimax_m3_probe.py::test_selection_attention_causal_future_first_block_recovers_without_nan
```

Result: `3 passed`.

```bash
pytest -q -m L4 \
  fe_api/nsa/test_NSA_selection_attention_minimax_m3_probe.py::test_minimax_m3_selection_attention_block128_causal_probe
```

Result: `1 passed`.

The targeted coverage verifies causal semantics independently on the existing block-64/GQA-4 geometry, causal/non-causal wrapper cache separation, and recovery when a fully masked future block is traversed before a valid block. The exact-shape MiniMax probe continues to validate block-128/GQA-16 arbitrary non-monotonic selection metadata.

Earlier B200 coverage of the existing selection-attention regression suite reported `2 passed, 2 skipped`; the skips are unsupported BSHD cases.

## Summary by CodeRabbit

* **New Features**
  * Added optional causal masking for selection attention, including masking within selected blocks.
  * Added support for block sizes up to 128 and GQA group sizes up to 16.
  * Improved handling of fully masked attention tiles to prevent invalid numerical results.

* **Bug Fixes**
  * Strengthened validation for block indices, causal settings, and supported configurations.
  * Ensured causal and non-causal execution plans remain correctly separated.

* **Documentation**
  * Updated selection attention guidance, examples, supported block sizes, and causal-mode parameters.

* **Tests**
  * Added coverage for causal attention, future-block selection, numerical stability, and plan caching.


YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `f0a0aa204429` against base `64f6b391144c`.

Extending the existing selection kernel is a sensible scope, and the future-first all-masked-tile fix works for the normal positive scale. One remaining P2 affects accepted nonpositive scales. Please also promote the two small causal GPU regressions from L4 into the default L0 selection; currently the default run only checks causal cache separation. The prior frost/python_tests request skipped OSS, which owns these fe_api tests. My following standalone `@cudnn-ci-bot run oss` comment will request that target.

**[P2] Apply the causal mask after scaling the logits** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f0a0aa20442904b90e39f0c1b4c443350d8136bc/python/cudnn/native_sparse_attention/selection/NSA_select_attn_fwd_hmma.py#L912))

The new mask writes -inf into acc_QK_mn before line 917 multiplies it by the runtime softmax scale. Neither constructor nor execute restricts that scale to positive values: scale 0 turns masked entries into NaN, and a negative scale turns them into +inf. Both contaminate otherwise valid causal rows. Apply the mask to scaled logits, and cover scale 0 plus one negative scale. In particular, zero Q/K should still produce uniform attention over the visible selected keys at either scale.

Evidence: On SM100 at f0a0aa204429, S=128, Hq=4, Hkv=1, D=128, block=64, one selected current block, Q=K=0, and V[t]=t/128: scale 1 is finite; scales 0 and -1 each produce 64,512 NaN O elements. At scale 0 row 0 has O/L/M=NaN; at scale -1 its M is +inf. The baseline noncausal implementation is finite for all three scales. Attribution: Introduced by the causal masking branch in this PR. The older noncausal path already accepts the same runtime scale values.

Validation: SM100, CuTe DSL 4.8: explicit -m "L0 or L4" selection of both affected test files gave 6 passed and 2 unsupported-BSHD skips, including block-128/GQA-16 nonmonotonic metadata and future-first recovery. Reused the matching extension only after checking C++/header equality. Additional NaN-poisoned output checks passed: an all-future selected block and zero selected blocks produce exact O=0, L=0, M=-inf. L here is the kernel row-sum statistic, not LSE. Noncausal baseline/head ABBA spot check on a 68-SM SM100 device, BF16 D=128 and block=64. S=128/Hq=4/Hkv=1/two blocks: CPU prepared 16.26 -> 15.86 us, wrapper 26.84 -> 28.42 us, GPU 6.788 -> 6.788 us. S=2048/Hq=16/Hkv=4/four blocks: CPU prepared 16.26 -> 16.24 us, wrapper 27.98 -> 28.99 us, GPU 221.012 -> 221.018 us. Correctness was checked against analytic uniform attention before timing. GPU timing used 16 calls per captured graph, with changed V and poisoned outputs verified on replay. Compilation and cold cache work are excluded from CPU times. The extra wrapper cost is consistent with the added device/dtype cache-key fields; prepared execution and GPU performance are unchanged in these cases. Isolated merge onto develop 64f6b391144c is conflict-free; current-head Style passed. The previous pipeline did not run OSS.

Limitations: These measurements cover the listed SM100 workloads, not MiniMax model throughput or all supported architectures. The small causal regressions are marked L4 and are not selected by the default L0 CI invocation. The requested OSS pipeline will initially be pending evidence.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc (commented):

Reviewed `c89a4fae912c` against base `fa7a433ee844`.

The nonpositive-scale issue from the previous review is resolved: scaling before the causal mask preserves masked -inf entries at zero and negative scales. The two small causal regressions are now L0. No approval blocker. I also confirmed the existing SM90 test-gate comment: the same six small cases pass on a real H100 when only that local test restriction is broadened. Fresh OSS CI will be requested after this review because the previous green pipeline tested f0a0aa2.

- **[P2] Still open:** [Retain the existing suggestion to run the small L0 regressions on SM90](https://github.com/NVIDIA/cudnn-frontend/pull/804#discussion_r4022273784).

Validation: SM100 at c89a4fae912c: both affected suites with -m 'L0 or L4' yielded 10 passed and two existing unsupported-BSHD skips, including the block-128/GQA-16 nonmonotonic MiniMax probe. Controlled RED/GREEN using identical current tests/source and restoring only the prior kernel: the four zero/negative-scale cases fail with NaNs and both positive-scale controls pass. The fixed kernel passes all six. Separately, those six small cases pass on actual SM90 with the local test gate broadened. SM100 noncausal pre-fix/head/head/pre-fix performance spot check, BF16 D128/block64: S128/Hq4/Hkv1/two blocks remains about 6.788 us GPU; S2048/Hq16/Hkv4/four blocks remains about 221 us. Prepared CPU calls are about 16 us and wrapper calls about 28-29 us in both arms, with no material regression established. Both paths pass analytic correctness and changed-input, poisoned-output captured replay controls.

Limitations: The performance check isolates this small follow-up, not model throughput. The committed small-case test gate still skips SM90 until the existing comment is addressed. Fresh OSS CI is pending; prior-head green results are not current-head qualification.

YangXu1990uiuc:
I am requesting test CI for `c89a4fae912c` with the following standalone command: `@cudnn-ci-bot run oss`. For future updates, request the appropriate test target before review; a maintainer may need to trigger it.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc:
CI follow-up for `c89a4fae912c` (pipeline 68108845): the pipeline is red, but the inspected failures do not come from the causal-scale fix.

The dev-backend build fails because `CUDNN_ATTR_OPERATION_SDPA_FWD_STATS_LOG2` is not defined in its headers. The reference comes from the newly merged #931 on develop; #804's head does not contain that backend-lowering change. The `CUDNN_VERSION >= 92800` guard still admits the dev headers before the corresponding backend attribute is available. The lowering needs to land with compatible backend headers, or remain disabled until that counterpart is available. This affects the integration base and should be resolved before relying on new CI runs.

The text-scan failure is also in #931's `graph_properties.h` documentation. The inspected SM100 OSS failures are the existing BSA missing-`torch` and DSL-4.6.2 compatibility failures. The targeted correctness and performance results for this PR remain in [the review](https://github.com/NVIDIA/cudnn-frontend/pull/804#pullrequestreview-5218585123); this pipeline does not establish an all-green integrated result.

vedaanta (approved):

Reviewed `c89a4fae912c` against base `9b9f32781d0c`.

No approval-blocking finding in this independent review. The scale-before-mask ordering and safe online-softmax center handle nonpositive scales and future-only tiles correctly in the focused tests. Causal specialization is included in the wrapper cache, and the new keyword-only argument preserves positional callers. The existing SM90 regression-test coverage suggestion remains nonblocking. This approval is a code-review decision, not an all-green CI claim.

- **[P2] Still open:** [Run the small L0 causal regressions on supported SM90 devices](https://github.com/NVIDIA/cudnn-frontend/pull/804#discussion_r4022273784).

Validation: Reviewed all five changed files, the selected-block traversal, sequence-local causal coordinates, online-softmax update, public call signatures, and cache identity. An isolated merge onto develop 9b9f32781d0c completed without conflicts. Fresh integration run on SM100 with CuTe DSL 4.7, PyTorch 2.13.0+cu130 and cuDNN 9.26: both affected suites selected explicitly with -m 'L0 or L4' yielded 10 passed and two expected unsupported-BSHD skips in 11.59 seconds. The import banner confirmed the integration checkout. Native bindings were reused only after verifying this integration changes no native sources. The passing cases include the block-128/GQA-16 nonmonotonic MiniMax reference comparison, six positive/zero/negative-scale causal and future-first regressions, noncausal execution/wrapper references, and causal cache separation. Reused the prior same-head review's controlled RED/GREEN and noncausal ABBA performance evidence: approximately 6.788 us and 221 us GPU for its two workloads, with no material prepared-execution regression established. These timings were not rerun here and do not measure MiniMax model throughput. Current-head Style and Docs checks pass. The latest merge-requirements annotation reports a missing Project board. Pipeline 68108845 remains failed; the existing CI follow-up documents inspected inherited backend-header, text-scan and BSA failures: https://github.com/NVIDIA/cudnn-frontend/pull/804#issuecomment-5692457407. No duplicate CI run was triggered.

Limitations: Fresh GPU validation covers the listed SM100 cases, not every supported architecture or metadata shape. SM90 and performance evidence is reused from the earlier same-head review. The failed OSS pipeline is not full integration qualification; not every failed job was freshly retriaged. Maintainer project metadata and the existing SM90 test gate remain outstanding.

PR #806, opened 2026-08-29
Tune underfilled KDA forward split tables on B200

YangXu1990uiuc:
## Before submitting

- [ ] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [ ] I ran `pre-commit run` and committed any formatting changes. The current Style / pre-commit CI check is green, but the live PR body does not record a local invocation.
- [x] I added GitHub labels: `cat-perf-bug`, `mod-frost`, and `orig-nv-eng`.
- [x] I set the **Milestone** to `Frontend 1.29.0`.
- [ ] The PR still needs a **Project** assignment; the current token does not have Projects scope.

## Affected area

- FE OSS kernels or CuTeDSL

## Summary

- Add a runtime `piece_cap` to the FROST split-table planner. Zero preserves the existing adaptive search; a positive value clamps the result after the normal search and uncut margin.
- Select cap 2 only for checkpoint-free KDA forward on exact SM100 when `B=1`, output heads are 64 or 96, the head grid underfills the GPU, and `ideal_chunks <= 5 * num_sms`.
- Keep KDA backward and all GDN/GDN2 forward and backward routes on the existing adaptive default.
- Replay the cap through `TableRecipe` without adding it to the compilation-cache key. Workspace sizing remains conservative and unchanged.

## Why

For the measured single-sequence B200 KDA-forward geometries, the existing adaptive planner can split an underfilled head grid into more pieces than are beneficial. A narrow runtime cap improves those measured cases while leaving the adaptive planner and all unmeasured routes unchanged. Keeping the value runtime-only avoids a new compiled-kernel specialization and preserves the existing workspace bound.

## Related issues

None.

There is no public API change. `piece_cap` is an internal FROST split-table planning input and is replayed at runtime rather than included in the compilation-cache key.

The behavior change is limited to exact SM100, checkpoint-free KDA forward, `B=1`, 64 or 96 output heads, an underfilled head grid, and `ideal_chunks <= 5 * num_sms`. Other architectures, checkpointed forward, batches above one, other head counts, KDA backward, and every GDN/GDN2 forward or backward route retain the existing adaptive planner. Workspace sizing is unchanged.

The performance results below are direct KDA-forward measurements, not backward, an assembled layer, or a full-model speedup.

## Testing

- `python -m pytest -q -c test/python/pytest.ini --confcutdir=test/python/linear_attention/frost test/python/linear_attention/frost/test_kda_piece_cap_policy.py` — 14 passed, covering policy boundaries, runtime replay, cache identity, workspace sizing, and call-site ownership.
- ComputeLab job `4014222`, NVIDIA B200 SM100 with 148 SMs, cuDNN backend `92600`; Frontend base `606e16f9786ea7a13e0462c8a63edf0d7f72ae85`, PR source `70f1fa0916be5f902bc9d3a9e89408dd50087c08`.

The PR policy and adaptive control were compared in one process using the same source. The control differed only by returning zero from the host selector while its plan was built; selector replacement happened outside correctness, warmup, and timing. Each arm used 3 warmups, 9 balanced CUDA-event samples, and a 256 MiB cache flush before every timed call.

| Model geometry | Adaptive control median | PR median | Paired control / PR | Wins |
|---|---:|---:|---:|---:|
| Kimi-K3, H96, 8,192 tokens | 0.646112 ms | 0.630752 ms | 1.0322x | 9/9 |
| Kimi-K3, H96, 16,384 tokens | 1.165312 ms | 1.150016 ms | 1.0133x | 9/9 |
| GLM-5.3, H64, 16,384 tokens | 0.613472 ms | 0.590816 ms | 1.0417x | 9/9 |
| GLM-5.3, H64, 32,768 tokens | 1.099744 ms | 1.094656 ms | 1.0009x | 6/9 |

The 32K GLM-5.3 case is a same-table control: the policy remains off and both arms select identical 9-piece tables, so it is not a claimed speedup.

All output checks passed. Worst maximum-relative output differences across the four rows were `0.0012563`, `0.0012255`, `0.0005252`, and `0`; final states were bit-exact in every case. These measurements exclude backward, projections, causal convolution, output gate, residual, MLP, assembled layers, and full models.

### DIFF
--- a/python/cudnn/linear_attention/frost/common/split_k.py
+++ b/python/cudnn/linear_attention/frost/common/split_k.py
@@ -179,8 +179,10 @@ def clamped_log2(log_gate: cutlass.Constexpr[bool], gate_val: cutlass.Float32) -
 
 
 @cute.jit
-def piece_choice(overhead_chunks: cutlass.Constexpr[int], num_sms: cutlass.Constexpr[int], batch_num_chunks, n_tiles, ideal_chunks):
-    """Per-tile piece choice.  Returns ``(span, num_blocks)``."""
+def piece_choice(overhead_chunks: cutlass.Constexpr[int], num_sms: cutlass.Constexpr[int], batch_num_chunks, n_tiles, ideal_chunks, piece_cap):
+    """Per-tile piece choice.  A positive runtime ``piece_cap`` clamps the
+    normal search result; zero leaves it adaptive.  Returns ``(span,
+    num_blocks)``."""
     # ---- even-spread cap: no span past ideal_chunks ----------------------------------
     p_hi = batch_num_chunks if batch_num_chunks < cutlass.Int32(MAX_BLOCKS) else cutlass.Int32(MAX_BLOCKS)
     p_hi = p_hi if p_hi > 0 else cutlass.Int32(1)
@@ -207,6 +209,8 @@ def piece_choice(overhead_chunks: cutlass.Constexpr[int], num_sms: cutlass.Const
         uncut_estimate = ((n_tiles + num_sms - cutlass.Int32(1)) // num_sms) * (batch_num_chunks + cutlass.Int32(overhead_chunks))
         if cutlass.Int32(4) * best > cutlass.Int32(3) * uncut_estimate:
             p = cutlass.Int32(1)
+    if (piece_cap > 0) and (p > piece_cap):
+        p = piece_cap
     span = cutlass.Int32(0)
     num_blocks = cutlass.Int32(0)
     if batch_num_chunks > 0:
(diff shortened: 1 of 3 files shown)


YangXu1990uiuc:
Closing after a broader B200 distribution sweep invalidated the default policy. At identical H=80/T shapes, cap=2 changed from a 5-8% win on a Kimi-like gate fixture to a 37-46% regression on a generic valid fixture; generic H=96 regressed 25-29%. All 42 correctness comparisons passed and final state was bit-exact, so this is an input-distribution-dependent performance flip, not a numerical bug. The current host policy sees only shape and cannot distinguish those cases. A safe replacement needs a separately designed device-informed planner; retaining the unused piece_cap plumbing alone is not worth the added JIT/API surface.

PR #813, opened 2026-08-30
fix(frost): drain final MoE scheduler broadcast

YangXu1990uiuc:
## Before submitting

- [ ] I set the **Milestone** and **Projects** fields in the sidebar (required to merge; maintainers can set these for external contributions). The milestone is set; the current token cannot set organization Projects.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

- Retain the final grouped-MoE broadcast-ring stage and its completion parity.
- Keep the leader CTA alive until every peer scheduler warp acknowledges the final invalid record.
- Compile the snapshot and drain out for singleton clusters.
- Apply the DSM-lifetime fix symmetrically to plain and block-scaled SM100 grouped-MoE templates.

This is a correctness/lifetime fix with no public API or model-performance claim.

## Why

Normal ring reuse waits for acknowledgements before a stage is reused. The final invalid scheduler record exits the loop instead, so no later reuse keeps the leader CTA and its distributed shared-memory barrier alive while peers still acknowledge it.

The fix snapshots the final stage before ring wrap. Each acknowledgement flips the saved pre-wrap empty phase, so `pre-wrap bcast_empty_phase ^ 1` is the completed parity waited on by the leader.

## Related issues

None.

None. This changes only internal scheduler lifetime in the existing FROST grouped-MoE kernels. Singleton clusters compile away the additional state and wait.

## Testing

- Current review head `d784e62ae`: removed the renderer helper, private schedule constants, regex/exact generated-source assertions, and the two codegen tests. Changed-file pre-commit, Python compilation, and diff checks passed.
- ComputeLab B200 at current head: one existing clustered plain-MoE E2E and one existing block-scaled NVFP4 grouped-MoE E2E reported **2 passed in 5.35 s**.
- ComputeLab B200 at the production-fix head `daf9ba5ba`: the complete L0 suites for both affected grouped-MoE families reported **229 passed, 20 skipped**. Prior functional smoke covered all six 1x1, 1x2, and 2x1 template/cluster combinations.
- An attempted observable CUDA Graph stress test queued 33 poisoned-output replays for each template, but the unfixed parent also passed both cases. It was therefore removed rather than presented as regression evidence.
- Compute Sanitizer `synccheck` reported zero errors before and after; `racecheck` produced identical, non-discriminating DSM reports in both versions. Sanitizer silence is supporting evidence rather than proof.

No deterministic black-box RED test was found for the DSM lifetime violation. The fix is supported by the broadcast-ring protocol argument and existing fixed-path functional coverage; no timing difference is expected or claimed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Mixture-of-Experts grouped matrix multiplication by ensuring coordinated processing completes before resources are released.
  * Improved completion handling across supported multi-cluster configurations while preserving efficient behavior for single-cluster scenarios.
  * Standardized routed outputs as a flat two-dimensional surface for more consistent downstream processing.

* **Tests**
  * Updated validation coverage for grouped matrix multiplication and output layout behavior.

### DIFF
--- a/python/cudnn/gemm/frost/kernel_templates/sm100_moe_grouped_block_scale_matmul_fwd.py
+++ b/python/cudnn/gemm/frost/kernel_templates/sm100_moe_grouped_block_scale_matmul_fwd.py
@@ -703,6 +703,8 @@ def _kernel(
         # leader slot. A singleton cluster has no remote DSM lifetime to drain.
         if cutlass.const_expr(cluster_size > 1):
             if cta_rank_in_cluster == 0:
+                # Each acknowledgement flips the saved pre-wrap empty phase, so
+                # pre-wrap ``bcast_empty_phase ^ 1`` is the completed parity.
                 while not nvvm.mbarrier_try_wait_parity(
                     sched_bcast_empty_mbar_ptr.subview(last_bcast_stage),
                     last_bcast_empty_done_phase,
--- a/python/cudnn/gemm/frost/kernel_templates/sm100_moe_grouped_block_scale_matmul_fwd.py
+++ b/python/cudnn/gemm/frost/kernel_templates/sm100_moe_grouped_block_scale_matmul_fwd.py
@@ -446,6 +446,8 @@ def _kernel(
         bcast_stage = cutlass.Int32(0)
         bcast_full_phase = cutlass.Int32(0)
         bcast_empty_phase = cutlass.Int32(1)
+        last_bcast_stage = cutlass.Int32(0)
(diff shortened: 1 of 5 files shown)


YangXu1990uiuc:

yanqinz2 (approved):

YangXu1990uiuc:
@coderabbitai review

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:
@coderabbitai review

PR #815, opened 2026-08-31
fix(kernelcache): thread-safety guarantees

tp5uiuc:
## Before submitting

## Affected area

C++ frontend API or graph construction

## Summary

Add `mutable std::mutex` to `KernelCache` so that `build()`, `from_json()`, `to_json()`, and `is_finalized()` are thread-safe. Also add `Graph::serialize()` candidate-index guard and `get_plan_name_at_index()` OOB check.

## Why

Without the mutex, two threads calling `build()` concurrently can both observe `is_finalized() == false` before either sets the flag, causing both to call `initialize()` and both to call `cudnnBackendFinalize()`. The second finalization returns `CUDNN_STATUS_BAD_PARAM`, propagating as an error from `build_operation_graph()`. This affects multi-stream workloads where streams share one `KernelCache`.

The `serialize()`/`get_plan_name_at_index()` issues are independent latent bugs, I thought I would fold those fixes in as well.

**Changes:**

- `include/cudnn_frontend/kernel_cache.h`: add `mutable std::mutex mutex_`; wrap `build`, `from_json`, `to_json`, `is_finalized`, `get_ptr` in `std::lock_guard`; add private `finalized_unlocked()` / `get_ptr()` helpers for use inside the held lock.
- `include/cudnn_frontend/graph_interface.h`: `Graph::serialize()` returns a diagnostic error when `selected_execution_plan_` is `-1` (plan not yet built); `get_plan_name_at_index()` adds OOB range check.
- `test/cpp/kernel_cache_thread_safety.cpp` (new): five Catch2 tests covering concurrent `build_operation_graph()` calls, `to_json`/`from_json` races, and idempotent `build()`.
- `test/cpp/serialize.cpp`: one new test covering `serialize()` candidate guard and `get_plan_name_at_index()` OOB.
- `test/cpp/CMakeLists.txt`: add new source file.

## Related issues

None.

None. All changes are internal to `KernelCache`. The public interface is unchanged.

## Testing

```
cd build && cmake --build . --target tests
LD_LIBRARY_PATH=<cudnn-9.x>/lib ./bin/tests [kernel_cache],[serialize]
```

All 6 new Catch2 tests pass on cuDNN 9.25.0. Existing OSS test suite unaffected.

Unpatched failure rate measured on cuDNN 9.25, N=16 threads, 40 iterations: 11 of 640 `build_operation_graph()` calls returned `CUDNN_STATUS_BAD_PARAM`. A small run can pass unpatched by luck.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved thread safety for concurrent kernel-cache building, initialization, and serialization.
  - Made repeated kernel-cache builds safely idempotent.
  - Strengthened serialization validation for unbuilt or unavailable execution plans.
  - Added clearer errors for invalid execution-plan index requests.

- **Documentation**
  - Clarified documented kernel-cache synchronization behavior.

- **Tests**
  - Added coverage for concurrent kernel-cache operations and serialization edge cases.

### DIFF
--- a/include/cudnn_frontend/backend/kernel_cache.h
+++ b/include/cudnn_frontend/backend/kernel_cache.h
@@ -67,12 +67,17 @@ class KernelCache : public detail::backend_descriptor {
 
     error_t
     to_json(std::string &str_json) const {
-        cudnnBackendDescriptor_t desc = get_ptr_locked();
         str_json.clear();
 #if (CUDNN_VERSION >= 91000)
         RETURN_CUDNN_FRONTEND_ERROR_IF(detail::get_backend_version() < 91000,
                                        error_code_t::CUDNN_BACKEND_API_FAILED,
                                        "CUDNN_ATTR_KERNEL_CACHE_JSON_REPRESENTATION is only available starting 9.10.");
+
+        cudnnBackendDescriptor_t desc = get_ptr_locked();
+        RETURN_CUDNN_FRONTEND_ERROR_IF(
+            desc == nullptr,
+            error_code_t::CUDNN_BACKEND_API_FAILED,
+            "KernelCache::to_json: descriptor not initialized; call build() or from_json() first.");
 
         int64_t serializationSize;
         std::vector<char> serialization_buf;
--- a/include/cudnn_frontend/backend/kernel_cache.h
+++ b/include/cudnn_frontend/backend/kernel_cache.h
@@ -67,25 +67,25 @@ class KernelCache : public detail::backend_descriptor {
 
     error_t
     to_json(std::string &str_json) const {
+        std::lock_guard<std::mutex> lk(mutex_);
         str_json.clear();
 #if (CUDNN_VERSION >= 91000)
         RETURN_CUDNN_FRONTEND_ERROR_IF(detail::get_backend_version() < 91000,
                                        error_code_t::CUDNN_BACKEND_API_FAILED,
                                        "CUDNN_ATTR_KERNEL_CACHE_JSON_REPRESENTATION is only available starting 9.10.");
 
-        cudnnBackendDescriptor_t desc = get_ptr_locked();
         RETURN_CUDNN_FRONTEND_ERROR_IF(
-            desc == nullptr,
+            get_ptr() == nullptr,
             error_code_t::CUDNN_BACKEND_API_FAILED,
             "KernelCache::to_json: descriptor not initialized; call build() or from_json() first.");
 
         int64_t serializationSize;
         std::vector<char> serialization_buf;
         _CUDNN_CHECK_CUDNN_ERROR(detail::get_attribute(
-            desc, CUDNN_ATTR_KERNEL_CACHE_JSON_REPRESENTATION, CUDNN_TYPE_CHAR, 0, &serializationSize, nullptr));
+            get_ptr(), CUDNN_ATTR_KERNEL_CACHE_JSON_REPRESENTATION, CUDNN_TYPE_CHAR, 0, &serializationSize, nullptr));
         serialization_buf.resize(static_cast<size_t>(serializationSize));
 
-        _CUDNN_CHECK_CUDNN_ERROR(detail::get_attribute(desc,
+        _CUDNN_CHECK_CUDNN_ERROR(detail::get_attribute(get_ptr(),
                                                        CUDNN_ATTR_KERNEL_CACHE_JSON_REPRESENTATION,
                                                        CUDNN_TYPE_CHAR,
                                                        serializationSize,
--- a/include/cudnn_frontend/backend/kernel_cache.h
+++ b/include/cudnn_frontend/backend/kernel_cache.h
@@ -9,6 +9,7 @@
 #include <array>
 #include <functional>
 #include <memory>
+#include <mutex>
 #include <sstream>
 #include <utility>
 
@@ -40,8 +41,19 @@ class KernelCache : public detail::backend_descriptor {
     }
 
     bool
-    is_finalized() {
-        return finalized;
+    is_finalized() const {
+        std::lock_guard<std::mutex> lk(mutex_);
+        return finalized_unlocked();
+    }
+
+    // Returns the underlying descriptor pointer by VALUE under the lock.
+    // Use this instead of get_ptr() when the cache may still be under concurrent build().
+    // IMPORTANT: build(), from_json(), to_json() must NOT call this — they already hold mutex_
+    // and std::mutex is non-recursive. Use the inherited get_ptr() inside those methods.
+    cudnnBackendDescriptor_t
+    get_ptr_locked() const {
+        std::lock_guard<std::mutex> lk(mutex_);
+        return get_ptr();
     }
 
     // Used to check kernel cache status (particularly after initialization)
@@ -56,6 +68,7 @@ class KernelCache : public detail::backend_descriptor {
 
     error_t
     to_json(std::string &str_json) const {
+        std::lock_guard<std::mutex> lk(mutex_);
         str_json.clear();
 #if (CUDNN_VERSION >= 91000)
         RETURN_CUDNN_FRONTEND_ERROR_IF(detail::get_backend_version() < 91000,
@@ -86,6 +99,7 @@ class KernelCache : public detail::backend_descriptor {
 
     error_t
     from_json(const std::string &json_cache) {
+        std::lock_guard<std::mutex> lk(mutex_);
 #if (CUDNN_VERSION >= 91000)
         RETURN_CUDNN_FRONTEND_ERROR_IF(detail::get_backend_version() < 91000,
                                        error_code_t::CUDNN_BACKEND_API_FAILED,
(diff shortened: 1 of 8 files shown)


include/cudnn_frontend/graph_interface.h
```diff
@@ -1683,14 +1683,24 @@ class Graph : public ICudnn, public INode {
...
-            j["cudnn_backend_data"] = serialized_plan;
-            j["variant_pack_uids"]  = variant_pack_uids;
+        // OSS-sentinel candidates (-2, -3) are NVRTC kernels that bypass the cuDNN backend; they
+        // have no cudnnBackendExecutionPlan and nothing to write — reject with a specific message.
+        if (plans.candidate == graph::Execution_plan_list::OSS_SDPA_ENGINE_CANDIDATE ||
+            plans.candidate == graph::Execution_plan_list::OSS_RMS_NORM_SILU_ENGINE_CANDIDATE) {
+            return {error_code_t::GRAPH_NOT_SUPPORTED, "OSS engine graphs are not serializable"};
         }
 
+        auto const candidate = plans.candidate;
```
tp5uiuc: Move this up before 1688 and use candidate to compare the `if`.

include/cudnn_frontend/graph_interface.h
```diff
@@ -1683,14 +1683,24 @@ class Graph : public ICudnn, public INode {
...
         }
 
-        auto const candidate = plans.candidate;
-        auto execution_plan  = plans.execution_plans[candidate];
-        if (execution_plan != nullptr) {
-            auto serialized_plan    = execution_plan->getJsonRepresentation();
-            j["cudnn_backend_data"] = serialized_plan;
-            j["variant_pack_uids"]  = variant_pack_uids;
+        // OSS-sentinel candidates (-2, -3) are NVRTC kernels that bypass the cuDNN backend; they
+        // have no cudnnBackendExecutionPlan and nothing to write — reject with a specific message.
```
tp5uiuc: condense to one liner

`// OSS-sentinel candidates bypass the cuDNN backend and have no cudnnBackendExecutionPlan to write`

include/cudnn_frontend/graph_interface.h
```diff
@@ -1683,14 +1683,24 @@ class Graph : public ICudnn, public INode {
...
 
+        auto const candidate = plans.candidate;
+        RETURN_CUDNN_FRONTEND_ERROR_IF(
+            candidate < 0 || candidate >= static_cast<int64_t>(plans.execution_plans.size()) ||
+                plans.execution_plans[candidate] == nullptr,
+            error_code_t::GRAPH_EXECUTION_PLAN_CREATION_FAILED,
+            "Graph::serialize: no built execution plan (candidate = " + std::to_string(candidate) + ")");
+
+        auto serialized_plan    = plans.execution_plans[candidate]->getJsonRepresentation();
+        j["cudnn_backend_data"] = serialized_plan;
```
tp5uiuc: Why no std::move(serialized_plan) if no one else uses it?

include/cudnn_frontend_ExecutionPlan.h
```diff
@@ -407,11 +407,17 @@ class ExecutionPlanBuilder_v8 {
 
 #if (CUDNN_VERSION >= 90400)
         if (m_execution_plan.kernel_cache) {
-            status = detail::set_attribute(m_execution_plan.pointer->get_backend_descriptor(),
+            // Copy the descriptor pointer by value under the KernelCache lock (defence in depth
+            // against a concurrent build() racing with this pool thread reading &desc). Mirrors the
+            // device_properties copy-then-take-address pattern at line ~513.
+            // IMPORTANT: use get_ptr_locked() here, not get_ptr() — plan build runs on the pool
+            // concurrently with KernelCache::build() on other threads.
```
tp5uiuc: Condense this comment to
```
Copy the descriptor pointer by value under the KernelCache lock. This ensures that the right pointer is passed to the backend, bypassing lazy materialization thread-safety hazards.
```

include/cudnn_frontend/backend/kernel_cache.h
```diff
@@ -147,5 +169,15 @@ class KernelCache : public detail::backend_descriptor {
 
    private:
     bool finalized = false;
+    // Protects finalized, and the desc/status members inherited from backend_descriptor.
+    // Precedent: cudnn_frontend_ExecutionPlanCache.h uses the same plain mutable mutex pattern.
```
tp5uiuc: Remove this precedent line.

include/cudnn_frontend/backend/kernel_cache.h
```diff
@@ -147,5 +169,15 @@ class KernelCache : public detail::backend_descriptor {
...
+    // KernelCache is always heap-allocated via shared_ptr and never moved after construction.
+    mutable std::mutex mutex_;
```
tp5uiuc, line 178: Note to reviewers : this lock is needed in multi-threaded code when two threads share a single kernel-cache.

Consider two application threads each building their own Graph for a different shape of the same topology, sharing a single std::shared_ptr<KernelCache>. Both reach the `!kernel_cache->is_finalized()` check in `build_operation_graph()` and both observe false, because finalized is a plain `bool` with no synchronization. Then this situation is possible

```
  T1  is_finalized() -> false, enters build()
  T2  is_finalized() -> false, enters build()      <- plain bool, no ordering
  T1  descriptor is null -> creates descriptor A
  T2  descriptor is null -> creates descriptor B   <- A leaked, handle replaced
  T1  sets operation-graph attribute, finalizes    <- on B, not the one it created
  T2  finalizes again                              -> CUDNN_STATUS_BAD_PARAM
```

The backend rejects such double sets and double finalizes with `CUDNN_STATUS_BAD_PARAM`, which the frontend surfaces as a failed `Graph::build_operation_graph()` on the losing thread. So one of the two graphs fails to build for reasons unrelated to the graph itself, and one backend descriptor is leaked for the process lifetime. Because it depends on how the two threads interleave, it shows up intermittently rather than reliably.

Note that this is frontend creating/accessing the backend pointer in a thread-unsafe manner, and no backend locking can address this. Hence we add a mutex here to prevent races (if its only finalized, an atomic would have sufficed, but we need to lock the bakcedn pointer accesses as well). With this change `build()` acquires the cache's mutex and re-checks finalized inside the critical section, so `create/attach/finalize` runs exactly once; the second thread observes the finalized cache and returns success without touching the descriptor. `to_json()` and `from_json()` take the same lock, so serialization can no longer observe a descriptor mid-publication. These functions are usually one-time setup code, so the performance impact of these mutexes are not expected to be minimal.

test/cpp/kernel_cache_thread_safety.cpp
```diff
@@ -0,0 +1,344 @@
...
+// pass the is_finalized() check and both call cudnnBackendFinalize(), causing
+// CUDNN_STATUS_BAD_PARAM.  With the mutex all N calls must return is_good().
+// ---------------------------------------------------------------------------
+TEST_CASE("KernelCache build() succeeds for all concurrent callers", "[kernel_cache][thread_safety]") {
+    if (cudnn_frontend::detail::get_backend_version() < 90500) {
+        SKIP("KernelCache build() requires cuDNN >= 9.5");
+    }
+
+    constexpr int N     = 16;
+    constexpr int ITERS = 40;
```
tp5uiuc: Reduce all the ITERS to 10 so as not to hog up CI resources

test/cpp/kernel_cache_thread_safety.cpp
```diff
@@ -0,0 +1,344 @@
...
+    // get_name_at_index with out-of-range index must return an error, not crash.
+    {
+        std::string name;
+        auto const count = graph->get_execution_plan_count();
+        REQUIRE(graph->get_plan_name_at_index(-1, name).is_bad());
+        REQUIRE(graph->get_plan_name_at_index(count, name).is_bad());
+    }
+
+    cudnnDestroy(handle);
+}
```
tp5uiuc: Move to `test/cpp/serialize`

include/cudnn_frontend/graph_interface.h
```diff
@@ -1684,13 +1684,23 @@ class Graph : public ICudnn, public INode {
...
         }
 
+        RETURN_CUDNN_FRONTEND_ERROR_IF(
+            candidate < 0 || candidate >= static_cast<int64_t>(plans.execution_plans.size()) ||
+                plans.execution_plans[candidate] == nullptr,
+            error_code_t::GRAPH_EXECUTION_PLAN_CREATION_FAILED,
+            "Graph::serialize: no built execution plan (candidate = " + std::to_string(candidate) + ")");
+
+        auto serialized_plan    = plans.execution_plans[candidate]->getJsonRepresentation();
+        j["cudnn_backend_data"] = std::move(serialized_plan);
```
tp5uiuc:
```suggestion
        j["cudnn_backend_data"] = plans.execution_plans[candidate]->getJsonRepresentation();
```

test/cpp/kernel_cache_thread_safety.cpp
```diff
@@ -0,0 +1,294 @@
...
+ *   - Subsequent build() calls are idempotent
+ *
+ * Measured unpatched failure rates on cuDNN 9.26.0.29, N=16, 40 iterations:
+ *   CUDNN_STATUS_BAD_PARAM: 11 of 640 build calls
+ * A small run can pass unpatched by luck; tests use N=16, 40 iterations.
+ *
+ * Note: TSan (the gold-standard witness) cannot run inside a Docker container
+ * by default (setarch -R is blocked by seccomp).  A bare-metal TSan run names
+ * kernel_cache.h and backend_descriptor.h as the contested locations.
+ */
```
tp5uiuc: Remove all of this.

include/cudnn_frontend/backend/kernel_cache.h
```diff
@@ -56,6 +67,7 @@ class KernelCache : public detail::backend_descriptor {
 
     error_t
     to_json(std::string &str_json) const {
+        cudnnBackendDescriptor_t desc = get_ptr_locked();
```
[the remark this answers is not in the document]
tp5uiuc: For the to_json() null-deref gap, it is fixed in the latest commit: `get_ptr_locked()` is now called inside the #if block and guarded with RETURN_CUDNN_FRONTEND_ERROR_IF(desc == nullptr, ...). The cuDNN attribute reads run only after a non-null descriptor is confirmed. I dont want to hold the mutex for the entire function since it is costly.

The error check is intentionally left as-is. Requiring is_good() on all 16 threads would make the test fail on unrelated GPU or driver errors (resource exhaustion, context loss, etc.) that are not the race condition under test. The `BAD_PARAM` check is precise: it rejects exactly the double-finalize symptom while tolerating noise the test cannot control.

tp5uiuc:
@cudnn-ci-bot run backend

Anerudhan (approved):

tp5uiuc:
@cudnn-ci-bot run backend

tp5uiuc:
CI results on `79f3272b` (`@cudnn-ci-bot run backend`):

| Job | Status |
|-----|--------|
| `san:cpp_test:sm80` | ✅ passed |
| `san:cpp_test:sm90` | ✅ passed |
| `san:cpp_test:sm100` | ✅ passed |

620 assertions, 35 test cases. Includes the 5 new `[kernel_cache]` tests and the updated `[serialize]` test. Merging.

PR #818, opened 2026-08-31
frost(sdpa): python-native validate() — defer C++ lowering when a python engine is a candidate

### LINKED ISSUE #704: frost(sdpa): pygraph.validate() lowers every SDPA graph to C++ and runs backend validation, even when a FROST engine will serve it
### Behavior

`pygraph.validate()` (python/cudnn/_pygraph.py, "Classic parity: C++ validation happens HERE" block) lowers the graph to the C++ binding and calls the backend `validate()` for **every backend-lowerable graph**. The only exemptions are ops with no backend lowering at all (GDN/KDA/...) and caller-registered engines. SDPA is backend-lowerable, so every SDPA pygraph pays C++ construction + backend validation — including graphs that `engines.manifest` routes to a pure-Python FROST engine that never touches the backend.

### Consequences

1. **Backend-version coupling for backend-free engines.** A graph the FROST engine fully serves is rejected at `validate()` when the installed backend (or the compiled binding headers — effective version is `min(lib, headers)`) predates the attribute. Concrete case from the PR #648 bringup: frost fp8/mxfp8 THD `cu_seq_len` graphs fail with `No suitable implementation for given SDPA_attributes` on backend < 9.25/9.26, even though the frost kernel executes them without any backend involvement. Similarly, declaring an `FP8_E5M3` tensor at all requires a backend/binding new enough to have the enum. Net effect: OSS-engine features are gated on backend releases, which defeats part of the point of frontend-only engines.

2. **Host cost per graph.** C++ graph construction + backend validation runs on every SDPA pygraph build even when the selected plan is a python engine. (Related: `backend_plan_entries()` documents that obtaining the backend's ranked list costs a real `create_execution_plans` — ~178 ms measured — so the C++ round-trips on the build path are not trivial.)

### What is deliberate about the current design

The comment marks this as *classic parity*: configs the backend rejects should raise from `validate()` where callers catch it to skip. That contract is worth keeping for the default path — this issue is a design request, not a bug report.

### Suggested direction

When `CUDNN_FRONTEND_ENABLE_FROST_ENGINES` is on and the manifest has a candidate python engine for the graph:

- defer the C++ lowering + backend validation until backend plans are actually needed (ranking includes backend engines, or every python candidate declines), **or**
- treat a backend `validate()` decline as non-fatal when a python engine remains eligible (record it, surface it only if python candidates also decline).

Either way, a FROST-served graph should not require a backend new enough to *validate* attributes the backend will never execute, and the classic behavior should be unchanged when frost engines are disabled.

vedaanta:
Fixes #704.

## Problem

`pygraph.validate()` lowers every backend-lowerable graph to C++ and runs the backend's `validate()`. SDPA is backend-lowerable, so every SDPA pygraph pays C++ construction + backend validation — including graphs a pure-python FROST engine fully serves. That couples frontend-only engines to the installed backend's version: a frost-served graph is rejected at `validate()` because the *backend* is too old to know an attribute it will never execute (e.g. fp8/mxfp8 THD `cu_seq_len` graphs on backend < 9.25, or merely declaring an `FP8_E5M3` tensor).

## Change

- **`python/cudnn/_sdpa_validate.py` (new):** python-native semantic validation for the SDPA node family (SDPA / SDPA_BWD / FP8 / MXFP8 flavors) — the version/arch-agnostic subset of the classic C++ pre-validation (rank-4 dim/stride + unit head-dim stride, GQA divisibility, alibi/causal/bottom-right/sliding-window combination rules, padding-mask ⇔ seq-len rules, dropout well-formedness, ragged/paged constraints, FP32 Stats, MXFP8 descale layout), with classic error types (`cudnnGraphNotSupportedError` for GRAPH_NOT_SUPPORTED parity; `ValueError` for ATTRIBUTE_NOT_SET / INVALID_VALUE parity) and classic messages. Backend-version gates (`get_backend_version() < X`) and device-arch gates are deliberately absent — those are support-surface answers each engine gives at planning.
- **`pygraph.validate()`:** when every node is SDPA-family *and* the manifest offers a python engine for the graph, run the python validator and skip the eager C++ lowering. The backend's verdict is deferred to planning, where a decline is already recorded (`backend_plan_entries()`) and surfaced by `plan()` only if no python engine proposes a plan either. Without a python candidate (frost disabled, or family unavailable) the classic eager-lowering timing is byte-for-byte unchanged.

## Testing

- New `test/python/sdpa/frost/test_sdpa_python_validate.py` (12 tests, device-free): no C++ lowering at `validate()` with a frost candidate, classic path still lowers without one, and semantic rejections (GQA, BR-causal+bias, padding/seq-len rules, dropout p=1, alibi w/o causal, non-FP32 Stats, head-dim multiple, sliding-window s_q>s_kv, bwd s_q=s_kv=1) raise from `validate()` with classic error types.
- `test/python/sdpa/frost/` integration + analyzer + heuristics suites: 117 passed on SM100 / 9.26.
- End-to-end smoke both ways (bf16 causal prefill, matches torch SDPA): frost off → lowered at validate, backend engine executes; frost on → not lowered at validate, `sdpa_fwd_prefill_sm100` serves.

### DIFF
--- a/python/cudnn/_pygraph.py
+++ b/python/cudnn/_pygraph.py
@@ -826,10 +826,37 @@ class pygraph:
         # rejects raises from validate() where callers catch it to skip. Skipped
         # only for a graph the backend has no lowering for (GDN/KDA/...), or when
         # the caller registered its own engine — the pre-existing exemption.
+        #
+        # SDPA-family graphs with a python engine candidate validate natively in
+        # python instead (issue #704): the eager C++ round-trip couples a graph
+        # a FROST engine fully serves to the installed backend's version (an
+        # attribute the backend is too old to *validate* but will never
+        # execute). Semantic validity is checked here by _sdpa_validate with
+        # classic error types; the backend's own verdict is deferred to
+        # planning, where a decline is recorded (backend_plan_entries) and
+        # surfaced by plan() only if no python engine proposes a plan either.
         if self._backend_lowerable() and self._lowered_graph is None:
-            self._lowered_graph = self._lower_to_cpp()
-            self._lowered_graph.validate()
-            self._verify_uid_ownership()
+            if self._python_native_validation():
+                from . import _sdpa_validate
+
+                for node in self._nodes:
+                    _sdpa_validate.validate_node(node)
+            else:
+                self._lowered_graph = self._lower_to_cpp()
+                self._lowered_graph.validate()
+                self._verify_uid_ownership()
+
+    def _python_native_validation(self) -> bool:
+        """Whether validate() may skip the eager C++ lowering: every node has a
+        python-native validator AND the manifest offers a python engine for this
+        graph (frost engines available and enabled). Without a candidate the
+        backend is the only possible server, so classic timing — raise its
+        rejection from validate() — must hold."""
+        from . import _sdpa_validate
+
+        if not self._nodes or any(n.node_type not in _sdpa_validate.COVERED_NODE_TYPES for n in self._nodes):
+            return False
+        return bool(self._candidate_engines())
 
     def build_operation_graph(self) -> None:
         """Validate the graph; lower to C++ when no python engines are registered.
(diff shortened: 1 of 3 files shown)


egilliam-nv:
Taking this over as discussed — continued in https://github.com/NVIDIA/cudnn-frontend/pull/869 (same commit, rebased onto current develop; also verified it fixes the A100 d=256 backward case from #864/#704). @vedaanta feel free to close this one.

vedaanta:
closed in favour of 869

PR #841, opened 2026-09-01
Complete SM100 D192/D128 SDPA feature support and performance tuning

adshen:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Complete the SM100 D192/D128 SDPA kernel feature set and improve performance for BF16/FP16, per-tensor FP8, and MXFP8 workloads.

## Why

The existing D192/D128 kernels did not have the same feature coverage as their sibling SM100 prefill kernels. The per-tensor FP8 implementation also used a separate E5M2+sink accumulation and correction path, which increased numerical and maintenance complexity.

This change completes the missing functionality, unifies sink handling, fixes edge-case synchronization and bounds issues, and retains only performance optimizations that passed correctness checks and repeated measurements.

## Changes

- Add THD packed-variable-length and sliding-window support for D192/D128 per-tensor FP8 and MXFP8.
- Add dense split-KV support for D192/D128 per-tensor FP8 and MXFP8.
- Support bottom-right causal masking and GQA where the corresponding data layout permits it.
- Remove the E5M2-only sink accumulation/correction path and use one sink flow for E4M3 and E5M2, including empty-KV and infinite-sink handling. The remaining dtype predicate only selects the validated first-real-KV row-max policy; it is not a separate sink algorithm.
- Fix FP8 edge bounds, empty-KV behavior, TensorMap ownership, and barrier handoffs.
- Tune the BF16, per-tensor FP8, and MXFP8 pipelines for the D192/D128 workloads.
- Consolidate repeated masked-softmax control flow while preserving the established dense and no-mask paths.

## Related issues

None.

No public API change.

The routing and scheduling changes are limited to the SM100 D192/D128 kernel family. Existing sibling-kernel defaults remain unchanged.

## Testing

- Full Frost suite: `996 passed, 314 skipped, 83 deselected`.
- Compute Sanitizer synccheck: `8 passed`, `ERROR SUMMARY: 0 errors`.
- Compute Sanitizer memcheck: `8 passed`, `ERROR SUMMARY: 0 errors`.
- `pre-commit` passed for every file changed by this PR.
- Ruff, `py_compile`, Black, and `git diff --check` passed.
- Tested with CUTLASS DSL `4.7.0a0` and cuDNN backend `9.26.0`.

Full Frost command:

```bash
python3 -m pytest -q test/python/sdpa/frost
```

Sanitizer coverage includes D192 BF16 causal, per-tensor FP8 THD+SWA, MXFP8 THD mask variants, and per-tensor FP8 split-KV.

## Performance

Local reproduction of the attention-training CI benchmark shapes on GB100. All rows use `B=2`, `Sq=Skv=8192`, `Dqk=192`, `Dv=128`, and top-left causal masking. Results are means from three interleaved runs with 10 warmups and 40 measured iterations per run.

Observed clocks were 847 MHz SM and 4000 MHz memory, using default clock-control behavior.

| Precision | CI model | Hq/Hkv | Upstream Frost (ms) | This PR (ms) | cuDNN 9.26 (ms) | Duration vs upstream | Duration vs cuDNN |
|---|---|---:|---:|---:|---:|---:|---:|
| BF16 | Kimi K2.6 | 64 | 3.363000 | 3.367333 | 3.364000 | +0.13% | +0.10% |
| BF16 | Kimi K3 | 96 | 5.034333 | 5.027667 | 5.029667 | -0.13% | -0.04% |
| BF16 | DeepSeek V3 | 128 | 6.711333 | 6.686667 | 6.695333 | -0.37% | -0.13% |
| Per-tensor FP8 E4M3 | Kimi K2.6 | 64 | 2.777000 | 2.481667 | 2.577333 | **-10.64%** | **-3.71%** |
| Per-tensor FP8 E4M3 | Kimi K3 | 96 | 4.127333 | 3.697667 | 3.861000 | **-10.41%** | **-4.23%** |
| Per-tensor FP8 E4M3 | DeepSeek V3 | 128 | 5.476000 | 4.922000 | 5.142000 | **-10.12%** | **-4.28%** |
| MXFP8 E4M3 | Kimi K2.6 | 64 | 2.844333 | 2.822333 | 2.783667 | -0.77% | +1.39% |
| MXFP8 E4M3 | Kimi K3 | 96 | 4.246667 | 4.175000 | 4.170000 | -1.69% | +0.12% |
| MXFP8 E4M3 | DeepSeek V3 | 128 | 5.645000 | 5.526333 | 5.553333 | -2.10% | -0.49% |

Negative percentages indicate lower duration.

Per-tensor FP8 reduces duration by **-10.12% to -10.64%** relative to upstream Frost and by **-3.71% to -4.28%** relative to cuDNN 9.26. BF16 remains at parity, while MXFP8 improves by **-0.77% to -2.10%** relative to upstream.

## Summary by CodeRabbit

- **New Features**
  - Added FP8, MXFP8, and F16 support for packed and variable-length sequences across SM100 head dimensions, including D192/D128.
  - Added split-KV execution with automatic scheduler and configuration selection.
  - Added support for causal, bottom-right causal, sliding-window, padded, and zero-length KV scenarios.
  - Added configurable output and log-sum-exp layouts, dynamic sequence lengths, attention sinks, and separate Q/K/V dimensions.
  - Improved ragged offsets and grouped-query attention support.
  - Added handling for padded MXFP8 scale-factor data.

- **Tests**
  - Expanded coverage for D192, masking modes, split-KV, variable-length inputs, and FP8/MXFP8 variants.


adshen:
@cudnn-ci-bot run frost

adshen:
@cudnn-ci-bot run frost

vedaanta (approved):

python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_fp8_sm100.py
```diff
@@ -1645,15 +2073,27 @@ def _softmax_warp_group(
...
 
-    eff_seqlen_kv = _resolve_seqlen_kv(seq_kv_lens_tensor, batch_idx, seqlen_kv)
-    bounds = _bounds_for_tile(q_super_idx, seqlen_q, eff_seqlen_kv, cta_in_pair, CFG.QH_PER_KH)
+    eff_seqlen_q, eff_seqlen_kv = _load_effective_lengths(True, sched, cutlass.Int32(0), batch_idx, seq_kv_lens_tensor, seqlen_q, seqlen_kv, n_batch)
+    if cutlass.const_expr(_PREDECODE_THD_SWA_SEGMENTS):
+        decoded_swa_left, decoded_swa_left_pad_start, decoded_swa_left_end, decoded_swa_right_start, decoded_swa_right = _load_swa_segment_bounds(
+            True, sched, cutlass.Int32(0)
+        )
+    else:
+        bounds = _bounds_for_tile_split(q_super_idx, eff_seqlen_q, eff_seqlen_kv, cta_in_pair, None, None, split_idx, CFG.QH_PER_KH)
```
vedaanta, line 2134: **Trace-time crash: `bounds` unbound when `_PREDECODE_THD_SWA_SEGMENTS` folds True.** `bounds` is only assigned in this `else` arm (and again on the scheduler back-edge at ~L2239), but the call to `_softmax_masked_kv_loops(_SWA_ONE_SIDED_GEOMETRY, bounds, segment_bounds, ...)` at ~L2204 evaluates `bounds` unconditionally. Since `cutlass.const_expr` only traces the taken branch, an E4M3 THD packed-varlen graph with top-left causal SWA wide enough to satisfy `_SWA_ONE_SIDED_GEOMETRY` (`WL+WR >= ~638` at cga2, which is what per-tensor E4M3 THD resolves to) raises `UnboundLocalError` at trace/compile time — exactly the config the predecode path (and the SCHEDULER_STAGES=3 tuning) was built for. CI never traces it because the THD-SWA tests use `swa_window=64`. Fix: assign `bounds` unconditionally, as the mxfp8 sibling does (~L2648 there); the one-sided branch of `_softmax_masked_kv_loops` never reads it, so this is free.

python/cudnn/sdpa/fwd/config_sm100.py
```diff
@@ -901,12 +900,139 @@ def _validate_cfg_d192(cfg: CfgD192) -> None:
...
+        mxfp8
+        and params.split_kv == 1
+        and not thd
+        and window_left is None
+        and window_right == 0
+        and not template_bottom_right
+        and 4096 < s_kv <= 8192
+        and (params.dtype_qkv == DTYPE_E5M2 or s_q >= 4096)
+    )
+    sched_policy = SCHED_NATURAL if mx_dense_mid_causal_cga1 else params.sched_policy
```
vedaanta: This silently overrides an **explicitly requested** `sched_policy` (LPT/LPT_L2 → NATURAL) for the `mx_dense_mid_causal_cga1` cell — and the comment above the params construction in `api_dsl.py` (~L1176) still says the heuristic's explicit policy "is honored verbatim, NATURAL included", which is now false for d192. It's internally consistent (the compile key, split helpers, and launch grid all use the resolved params), but it's the same silent-knob-drop this file's own `_validate_params` comment refuses for `split_kv`/`cta_mma`. Suggest either restricting the override to the derived-policy case (`sched_policy is None` at the api_dsl level, threaded through as a flag), or updating the stale api_dsl comment to document that d192 treats `sched_policy` as a hint.

python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_mxfp8_sm100.py
```diff
@@ -2687,7 +3183,30 @@ def _build_sf_desc(sf_tensor, num_tiles, sf_smem_size, num_rows_box, num_heads):
     q_clusters = (SQ + rows_per_cluster - 1) // rows_per_cluster
     grid_q_supers = q_clusters * CFG.CTA_MMA
     q_supers = grid_q_supers
-    grid_shape = (grid_q_supers, QH, B) if cutlass.const_expr(CFG.SCHEDULER_POLICY == SCHED_NATURAL) else (grid_q_supers * QH * B, 1, 1)
+    if cutlass.const_expr(CFG.THD_VARLEN):
+        # D192 K is rank-5, so its sequence extent is TensorMap dimension 1;
+        # V remains rank-4 with sequence extent at dimension 2.
+        _build_thd_meta_o_kv_descs_kernel(
```
vedaanta, line 3234: **THD packed-SF pad rows are an unenforced producer contract (0×NaN hazard).** The GLOBAL_DIM clamp here correctly zeroes the K/V *data* capacity tail, but the SF stream is per-sequence tile-padded and each sequence's tail SF tile is loaded in full — the pad-row V-SF blocks feed `mma_ts_step(..., tmem_sf_b=...)` for K-dim rows whose P is exactly 0. If a producer leaves `0xFF` (E8M0 NaN) in pad SF bytes, the block-scale MMA computes `0 × (V·NaN) = NaN` and poisons the whole O tile. Nothing in this kernel or `api_dsl._reshape_sf_packed` (which only validates byte size) guarantees finite pad bytes. Given the prior ragged NaN-pad-poison bugs in this codebase, please document/assert the pad-byte contract (e.g. fill `0x7F`) at the packing boundary.

python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_mxfp8_sm100.py
```diff
@@ -2723,75 +3243,103 @@ def compile(  # noqa: A001
...
+                assumed_align=4,
+            )
     else:
+        if lse_head_major or lse_head_stride:
+            raise ValueError("lse_head_major / lse_head_stride are THD-only")
         fake_lse = (
-            cute.runtime.make_fake_tensor(cutlass.Float32, (b, qh, sq), lse_stride, assumed_align=4)
-            if lse_stride is not None
+            cute.runtime.make_fake_tensor(cutlass.Float32, (_lse_batch, qh, sq), lse_stride, assumed_align=4)
+            if lse_stride is not None and SPLIT_KV == 1
```
vedaanta, line 3388: Silent drop: a caller-provided `lse_stride` is ignored when `SPLIT_KV > 1` (this falls through to the compact `(b·SPLIT, qh, sq)` layout). Dead in practice today — the split path allocates its own partial workspace — but if an adapter ever passes both, the kernel writes wrong addresses with no diagnostic. A `raise`, matching how `lse_head_major` misuse is handled, would be safer.

python/cudnn/sdpa/fwd/kernels/prefill_d192_d128_f16_sm100.py
```diff
@@ -105,6 +101,23 @@
     MASK_CAUSAL,
     MASK_SWA,
 )
+from cudnn.block_sparse_attention.csrc.utils.kernel_utils import ex2_emulation_2
```
vedaanta, line 106: This import chain (`kernel_utils` → `cute_dsl_utils` → module-level `import torch`) pulls torch in at kernel-module load time, even when `_DENSE_CAUSAL_PUBLIC_EXP2_MIX` is False and `ex2_emulation_2` is never traced. The chain pre-exists on develop (the fp8/mxfp8 siblings already import it), but this line newly extends the exposure to the bf16/fp16 flavor — the one a torch-free JAX process is most likely to load, where this raises ImportError at plan time. The import-boundary guard test doesn't cover lazily loaded kernel templates, so CI won't catch it. Suggest making `kernel_utils` lazy-import `sub_packed_f32x2` (fixes all three flavors) rather than moving this line.

python/cudnn/frost/tile_dsl/mma.py
```diff
@@ -119,6 +130,10 @@ def mma_ss(desc, desc_a_base, desc_b_base, tmem_c, tmem_sf_a=None, tmem_sf_b=Non
     enable_input_d = accumulate
     SF_CYCLE = 4
     SF_REGISTERS_PER_BLOCK = 4
+    if cutlass.const_expr(elect_once):
+        elected = nvvm.elect_sync()
+    else:
+        elected = cutlass.Boolean(True)
```
vedaanta: Dead code nit: when `elect_once=False`, the loop's `const_expr` branch calls `nvvm.elect_sync()` per step and `elected` is never read, so this `Boolean(True)` is only keeping the name defined — and the `True` value is misleading (it suggests all lanes issue). `elected = None` plus a one-line comment, or dropping the else entirely if the tracer allows it, would read better.

python/cudnn/sdpa/fwd/api_dsl.py
```diff
@@ -1197,47 +1193,34 @@ def compile(self) -> None:
...
-            lpt_q_tiles=lpt_q_tiles,
             thd_varlen=self.thd,
             pack_gqa=self.pack_gqa,
             qh_per_kh=int(self.q_desc.shape[1]) // int(self.k_desc.shape[1]),
             split_kv=self.split_kv,
             fused_ldtm_stat=fused_ldtm_stat,
             softmax_f16=self.softmax_precision == _cudnn_dtype.HALF,
         )
+        if self.flavor == (192, 128):
+            params = resolve_d192_template_params(
```
vedaanta: The comment at ~L1176 above ("the heuristic emits an explicit policy … and it is honored verbatim, NATURAL included") is now stale for d192: `resolve_d192_template_params` can flip an explicit LPT/LPT_L2 to NATURAL (`mx_dense_mid_causal_cga1`) and also rewrites `window_right`, `bottom_right`, and `cta_mma`. Worth updating the comment to say d192 canonicalizes/overrides these after this point (see the matching comment on `config_sm100.py`).

vedaanta (commented):
Reviewed the full diff (all 13 files), with deep passes over the three kernel files, the `resolve_d192_template_params` plumbing, and the `mma.py` elect changes. Overall this is a careful, well-tested change — the unified FP8 sink flow, split-KV chunk agreement across warp groups, THD TensorMap ownership, the SMEM stats mailbox / `mb_stats_read` protocols, SF/data tile lockstep, STAGES_KV halving under cga1, and the register-budget sums all check out. One trace-time crash found (inline comment on the per-tensor FP8 kernel) plus a few smaller items inline.

Two notes that don't anchor to a single diff line:

1. **Rubin direct-construction gate**: dropping the api_dsl "THD FP8 requires D_QK=D_V=128" NotImplementedError gate is right for SM100, but on cc10.7 a *direct* `SdpaFwdDslSm100` construction of THD FP8 d192 now falls through to the SM100 d192 kernel file (`_load_sm100_kernel_module`'s `rubin=` flag only reroutes the (128,128) flavor). The engines row still pins Rubin to d128, so the graph path is safe — but the removed gate explicitly existed to cover direct construction. Worth a replacement guard or a comment.
2. **One-sided SWA geometry is exactly tight and untested at scale**: `WL+WR >= MASK_TOKENS_PER_CGA + TILE_N - 2` holds with zero margin at equality (I verified the ceil/floor derivation), and the 4-region segment formulas are duplicated across the three kernel files plus `compute_kv_loop_bounds`. The added tests only exercise `swa_window=64`, which never reaches the one-sided threshold (~638 at cga2), so none of the region math — nor the crash flagged inline — is covered by CI. Please add a wide-window SWA test (dense and THD) at the `WL+WR == M+N-2` boundary with the seam landing mid-tile.

adshen:
@cudnn-ci-bot run frost

adshen:
@cudnn-ci-bot run frost

PR #844, opened 2026-09-01
frost gemm heur: score real M, densify the tile scan, never pad the cluster grid

brandonfzhang:
Follow-up to #781. This is the first revision informed by an exhaustive per-layer sweep
run *against* #781's logic (every buildable config timed per layer, CUPTI device time,
L2 flushed, median of repeats), and it fixes the three biggest things that sweep exposed:

**1. `_tile_score` was scoring a rounded-down M.** The scorer received
`_floor_pow2(M)` — a bucketing convenience inherited from the FlashInfer heuristic's
tactic cache, where power-of-2 M buckets keep the cache small. frost calls
`select_config` directly and never buckets, so the snap was pure information loss:
M = 7009 scored as 4096, misstating both the number of launched M-tiles and the
last-tile padding, which over-ranked short tiles for every non-power-of-2 M. Scoring
the real M fixes the whole mid-M fp8 cohort.

**2. The tile scan was sparser than the hardware.** The N axis gains the legal
intermediate widths (96, 160, 192, 224) alongside the powers of two, and the 64-tall
family — added for M ≤ 128 in #781 — extends to M ≤ 512, where the sweep shows it
winning well past one tile row. 128 is scanned first so a scoring tie keeps the taller
tile (fewer CTAs, same waves). Both extensions stay restricted to the plain single-GEMM
row-major-B path: block-scale needs `cta_n % 128 == 0`, multi-GEMM shares a
power-of-two budget, and N-major B re-rounds to swizzle groups.

**3. Cluster ties could pick a padded launch.** Under a single wave,
`_cluster_score`'s quantization and occupancy terms cancel, so a cluster shape wider
than the grid (launching padding CTAs) tied with an exact fit and pool order decided
the winner. Ties now break by fewest launched CTAs first — never pad — then by the
sweep's measured preference, A-multicast `(cta_group, 4)`, which wins the remaining
tie group in every cohort we could split (fp8/fp4, 1- and 2-CTA, block-scale or not).
This replaces #781's conditional preference rules with one universal one.

**4. Block-scale takes the 2-CTA MMA pair earlier.** Pairing halves each CTA's
B-operand footprint, which buys pipeline depth — a win even without a second M tile,
but only while the pair's padded M row still fits the machine in one wave
(`2 × n_tiles ≤ sm_count`); past that the padding costs a whole extra wave, so it
stays at 1.

## Measured results

248-layer golden set (GEMMs extracted via FlashInfer API logs of priority models;
fp8 per-tensor-scale and nvfp4 block-scale, batch 1), heuristic efficiency =
chosen-config time / best-swept time, every "after" silicon-retimed:

| arch | fp8 before → after | fp4 before → after |
|---|---|---|
| sm100 | 0.874 → 0.937 | 0.911 → 0.956 |
| sm103 | 0.884 → 0.936 | 0.917 → 0.955 |

"Before" is current develop, i.e. already including #781. The change was validated the
same way on every other architecture frost supports. Known >1% regressions: 20 layers
on sm100 / 29 on sm103, concentrated in fp8 M=192 with large N (the 64x192 tile edging
out 64x256 in score where the wider tile measures faster; worst case 0.99 → 0.78) —
the aggregate is net-positive on every measured axis.

## Notes for reviewers

- Still no fitted coefficients: the new thresholds are hardware-meaningful quantities
  (one wave, one tile row, `sm_count`), and the only ordering knowledge is tie-breaking
  among *score-equal* shapes.
- The `_floor_pow2` removal is the highest-leverage single line; if this PR needs
  splitting, that hunk stands alone.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved GEMM configuration selection across a broader range of matrix sizes and tile layouts.
  * Added support for more warp-grid and tile geometry combinations where hardware resources permit.
  * Improved handling of 16-bit and 8-bit input layouts while preserving compatibility for sub-byte workloads.
  * Refined block-scale and multi-GEMM configuration choices for more reliable performance.
  * Improved cluster selection to reduce unnecessary execution units and provide more consistent results.


yanqinz2 (approved):

brandonfzhang:
Pushed one follow-up commit (e506993e): `test_the_auto_path_never_picks_a_non_power_of_two_tile` pinned the old scan's power-of-two width ladder, which this PR deliberately retires — the densified scan makes 96/160/192/224 the engine's own picks on the plain path. The test now pins the new invariants instead: every auto width is a hardware-legal multiple of 32 in [32, 256], every pick resolves in the catalog, and the constrained paths (block-scale, multi-GEMM) still produce only powers of two. Verified against the patched heuristic over the test's own shape grid.

brandonfzhang:
@cudnn-ci-bot run frost

PR #845, opened 2026-09-01
frost(sdpa): per-tensor FP8 d512 forward kernel on SM100

RomanAnders90:
## What

Adds `prefill_d512_fp8_sm100.py` — the per-tensor FP8 (E4M3/E5M2) sibling of the d512 f16 kernel — and routes the SM100 per-tensor FP8 engine to it. Same cga4×1 role-split pipeline, Q∪K_ring SMEM alias, UTCCP(Q → TMEM) + `mma_ts` BMM1, and THD leg as the f16 flavor.

Head-dim coverage: the flavor serves **(256, 512] on both head dims** — the band no smaller FP8 flavor reaches — through the usual TMA zero-padding envelope, so `d_qk != d_v` is served too.

## Kernel deltas (all CFG-derived, no literals)

| | why |
|---|---|
| `STAGES_KV = XFER_STAGES = 3` | BPE=1 halves the K/V ring; FP8 packs 4 elements per 4-byte TMEM column, so the UTCCP'd Q costs 128 cols instead of 256 and a third S_acc parity fits: `3*128 + 128 = 512`, exactly at the Blackwell TMEM cap |
| `MMA_KIND = F8F6F4`, `TILE_K_HW = 32`, idesc `k_dim = 0` | the Blackwell K=32 QMMA step. Both `k_dim`s derive from `CFG.TILE_K_HW_*` so they cannot drift into the Rubin `k_dim=1` combination, which is **silently wrong** here (per-row scrambled accumulator, no crash) |
| `P_XFER_HALVES = (TILE_N*BPE)//Q_SWZ_BYTES` → 1 | a 128-column FP8 P row is a single swizzle atom, so the sg0→sg1 KV-split degenerates to one whole-P chunk; the BMM2 K-half unroll becomes a loop over it |
| Q∪O alias sized in **bytes**, `sO` viewed as `OUT_STORAGE_DTYPE` | an element-sized alias under-allocates at `BPE_O = 2` (FP8 in, BF16/FP16 out) and the wider O store overflows into the P-transfer ring |
| descales read from **device** | `descale_q*descale_k` → `scale_softmax_log2`, `descale_v*scale_o` → `o_scale_fused` → folded into `beta`. No host readback |
| THD via `build_thd_meta_o_kv_descs_kernel` | K/V load through packed-total-clamped runtime descriptors: the last sequence's tile steps past the packed total into the caller's capacity tail, and a NaN there would wipe the tile through BMM2's `P·V` (masking zeroes P, not V) |

`Amax_O` is an atomicMax over the fp32 pre-cast output, reduced with a **ternary tree** (`_abs_max_tree`). The obvious running `acc = max(acc, |e|)` builds a `TILE_O`-deep dependency chain on the epilogue's critical path and cost ~7% of kernel time at S=8192. A fully-masked row needs no per-element sanitizing — its O TMEM is written by the MMA with a zero P — so one select on the row's scalar result covers the dead-row case.

## Framework changes

- **`Capabilities.d_envelope_floors`** (new row data): a per-native-shape envelope floor. Without it the d512 shape would act as a plain upper bound and swallow e.g. a d256 graph at 2× zero-padding. The FP8 row floors d512 at 256; `api_dsl` mirrors the table and a test pins the two together.
- **`api_dsl._pick_flavor`** now takes a candidate set, so an FP8 graph can no longer land on the d256 flavor, which has no FP8 kernel.
- **`sdpa_support_surface.h`**: the graph-level surface rejected `d_qk > 128` for `sdpa_fp8` before any engine was consulted; it now admits the same (256, 512] band, mirroring the existing d192/d128 exception.

## Not in scope

Split-KV (needs a half-precision O and its own combine-side amax) and MXFP8 (there is no d512 block-scale kernel). Both stay pinned to `{(128, 128)}` via `split_d_shapes` / the MXFP8 row.

## Test

Blackwell (cc10.0). 34 new cases in `test_sdpa_fwd_fp8_sm100.py`: masks, all 8 in×out dtype pairs, sink, GQA/MQA, per-batch KV padding, LSE, multi-wave, the (256, 512] envelope band (`384×448`, `464×368`, `272`, `512×496`), and THD (self/cross attention, GQA, sink + ragged stats, zero-length KV, `cu_seq_len` form). Accept **and** reject coverage for the new floor in `test_sdpa_fp8_sm107.py`.

Full `sdpa/frost/` suite: **1003 passed**. The three `*_strided_stats` failures on that node pre-date this branch — they need a cuDNN backend ≥ 9.26 and the node has 9.25.1 (verified: they fail identically on the pre-change `.so`).

## Perf

Kernel-only GPU time (nsys, B=1 H=8 S=8192 d=512 dense E4M3): **530.8 µs** vs **521.4 µs** for the pre-upstream kernel this was ported from — **+1.8%**, which is the in-kernel `Amax_O` that kernel does not compute.

> **Note for reviewers:** this changes a header, so the pybind extension needs rebuilding. In-tree `pip install -e .` fails on boxes where `/usr/bin/cmake` is 3.22 (repo needs ≥ 3.23) — configure out-of-tree with a ≥3.23 cmake and copy the `.so` into `python/cudnn/`.

## Summary by CodeRabbit

* **New Features**
  * Expanded FP8 scaled dot-product attention support on SM100+ devices for head dimensions from 256 through 512 when aligned to multiples of 16.
  * Added support for FP8 d512 configurations, including dense and variable-length inputs.
  * Added support for masks, output types, grouped-query attention, padding, statistics, and multi-tile execution with d512.

* **Bug Fixes**
  * Improved validation to reject unsupported dimensions and prevent invalid kernel selection.
  * Preserved existing MXFP8 limitations and device-specific support boundaries.

### DIFF
--- a/include/cudnn_frontend/node/sdpa_support_surface.h
+++ b/include/cudnn_frontend/node/sdpa_support_surface.h
@@ -187,16 +187,29 @@ SDPA_attributes::validate_sdpa_support_surface(const detail::Context& context,
         // validate basic dimension requirements
         // d_qk=192 with d_v=128 is only supported starting from cuDNN 9.19
         bool const d192_v128_supported = (detail::get_backend_version() >= 91900);
+        // Head dims in (256, 512] on BOTH sides are served by the frontend-only
+        // FROST SM100 per-tensor FP8 prefill engine
+        // (sdpa_fwd_prefill_sm100_fp8, the d512 kernel flavor), which runs the
+        // band through its native d = 512 tile with TMA zero-padding. The floor
+        // is the engine's: below it the flavor would pad by more than 2x, and
+        // no smaller FP8 flavor reaches above 192/128. The cuDNN backend itself
+        // has no plan for the band, so a graph that does not select that engine
+        // still fails at plan creation.
+        bool const d512_supported =
+            (d_qk > 256) && (d_qk <= 512) && (d_v > 256) && (d_v <= 512) && (d_qk % 16 == 0) && (d_v % 16 == 0);
         if (prop_major >= 10) {
             RETURN_CUDNN_FRONTEND_ERROR_IF(
-                ((d_qk > 128) || (d_qk % 16 != 0)) && !(d192_v128_supported && d_qk == 192 && d_v == 128),
+                ((d_qk > 128) || (d_qk % 16 != 0)) && !(d192_v128_supported && d_qk == 192 && d_v == 128) &&
+                    !d512_supported,
                 error_code_t::GRAPH_NOT_SUPPORTED,
                 "hidden_dim d_qk should be less than or equal to 128 and hidden_dim d_qk "
-                "should be multiple of 16 unless d_qk == 192 and d_v == 128 (requires cuDNN 9.19+)");
+                "should be multiple of 16 unless d_qk == 192 and d_v == 128 (requires cuDNN 9.19+) "
+                "or both head dims are in (256, 512] and multiples of 16");
             RETURN_CUDNN_FRONTEND_ERROR_IF(
-                ((d_v > 128) || (d_v % 16 != 0)),
+                ((d_v > 128) || (d_v % 16 != 0)) && !d512_supported,
                 error_code_t::GRAPH_NOT_SUPPORTED,
-                "hidden_dim d_v should be less than or equal to 128 and hidden_dim d_v should be multiple of 16");
+                "hidden_dim d_v should be less than or equal to 128 and hidden_dim d_v should be multiple of 16 "
+                "unless both head dims are in (256, 512] and multiples of 16");
         } else {
             RETURN_CUDNN_FRONTEND_ERROR_IF(
                 (d_qk > 256) || (d_qk % 16 != 0) || (d_v > 256) || (d_v % 16 != 0),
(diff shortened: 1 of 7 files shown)


RomanAnders90:
@cudnn-ci-bot run frost

RomanAnders90:
Addressed in f058151f.

The `lse_ref` binding was indeed unused — but renaming it to `_` would have papered over the actual defect. `test_fp8_d512_thd_zero_len_kv` only asserted that the *dead* rows come back `-inf`; it never checked the live sequences' LSE, so a wrong-but-finite LSE on them would have passed silently.

So instead of dropping the binding I used it, matching what the d128 sibling (`test_fp8_thd_zero_len_kv`) already does:

```python
torch.testing.assert_close(lse, lse_ref, atol=2e-2, rtol=2e-2, equal_nan=False)
```

keeping the explicit dead-row assert alongside it, since that is the point of the case. While comparing against the sibling I also noticed the d512 copy had dropped its second shape — a zero-length **Q** sequence (`[126, 0, 60]`), which contributes no tile at all — so that is restored too.

Swept the rest of the d512 tests for the same pattern (AST check for stored-but-never-loaded names): no other unused bindings.

50 passed with `-k "d512 or thd"` on Blackwell cc10.0.

vedaanta (approved):

vedaanta:
@cudnn-ci-bot run frost

PR #858, opened 2026-09-02
feat(gemm): optional tanh soft-clamp (tanh_clamp_scale) for grouped GEMM srelu/dsrelu

wanyingw:
- [x] Labels: suggest `cat-enhancements`, `mod-cutedsl`, `orig-nv-eng` (no access to add).

## Affected area

CuTe DSL grouped GEMM: `srelu` (forward) and `dsrelu` (backward) SM100 kernels, their Python APIs, tests, and docs.

## Summary

Adds an optional tanh soft-clamp scale `tanh_clamp_scale` to the squared-ReLU epilogues of the SM100 block-scaled grouped GEMM kernels, forward and backward. With scale `s`, writing `t = tanh(relu(x)/s)` and `c = s·t`:

- **Forward (srelu):** `out = c² · prob` — plain srelu is the `s → ∞` limit.
- **Backward (dsrelu):** `dgrad = g · 2c(1−t²) · prob`, `dprob = Σₙ c²·g`, `d_srelu` regen `= c² · prob`, replacing the `relu(C)` / `relu(C)²` terms.

`s` is a trace-time constant folded into the compiled kernel and included in the compile-cache key (distinct `s` compile distinct kernels). `tanh_clamp_scale=None` (default) is bit-identical to today's unclamped path.

## Why

A new MoE activation (tanh soft-clamped squared ReLU) currently runs as a standalone elementwise kernel between the two grouped GEMMs — extra memory round-trips and no FP8-requant fusion. This PR makes it fusable into the grouped GEMM epilogue. The clamp bounds the activation by `s² · prob`, which also keeps the fused FP8 quantization well-behaved. Downstream TE and Megatron-Core integration follow in separate PRs; this is the first link.

- Purely additive: new optional keyword `tanh_clamp_scale: Optional[float] = None` on `GroupedGemmSreluSm100`, `GroupedGemmDsreluSm100`, and the `grouped_gemm_srelu_wrapper_sm100` / `grouped_gemm_dsrelu_wrapper_sm100` convenience wrappers. No existing parameter changes.
- Default (`None`) reproduces current kernels bit-for-bit; all existing tests pass unmodified.
- Validated finite and > 0 when set. These APIs remain experimental and subject to change.
- Deterministic dprob (PR #521) is unaffected: the clamp changes only the per-element value feeding the accumulation, never the slotting or reduction order.

## Testing

On GB300 (sm_103), full `test_grouped_gemm_srelu.py` + `test_grouped_gemm_dsrelu.py` suites green:

- Clamped cases: two scales × {fp8, fp4} × `vector_f32` × `use_dsrelu_reuse` × deterministic dprob.
- Additive-only: `None` vs argument-omitted compared with `torch.equal` for both kernels.
- Cache-key: distinct `s` values compile distinct kernels.
- Forward/backward consistency: `d_srelu` regen matches the forward output.
- fp64 autograd check that the closed-form backward is the forward's derivative.
- Saturation: deep-tail inputs give exactly-zero dgrad (and non-zero dprob, so the check can't pass vacuously).

## Review notes

- **Why relu before tanh:** `s·tanh(relu(x)/s) ≡ relu(s·tanh(x/s))` (ReLU commutes with any non-decreasing `g` fixing 0; derivatives agree too). ReLU-first matches the existing epilogue structure and the reference implementation.
- **`t` is capped at 1** (`fmin` after `tanh.approx.f32`): defensive — measured on sm_103 over [0, 64] plus saturated inputs, the approximation never exceeded 1. The cap makes `1−t² ≥ 0` (no gradient sign flip) and the `s²·prob` output bound structural rather than observed. One instruction per element.
- **Accuracy:** measured max abs error of the tanh path is 7.7e-6 vs an fp64 reference (worst-case spec bound 4.9e-4); end-to-end output error is dominated by output quantization, not tanh.
- **Naming:** `tanh_clamp_scale` distinguishes the tanh *soft* clamp from the existing *hard* clamp parameters (`glu_clamp_max`/`glu_clamp_min`) and aligns with the downstream Megatron config `activation_func_tanh_clamp_scale`.

## Summary by CodeRabbit

- **New Features**
  - Added optional tanh soft-clamping for grouped GEMM SReLU and dSReLU operations.
  - Positive scale values provide bounded activations while preserving existing behavior when unset.
  - Forward and backward operations support matching clamp scales for consistent gradients.
  - Invalid scale values are rejected, and compiled variants remain separately cached.

- **Documentation**
  - Documented soft-clamped activation formulas, deterministic behavior, precision considerations, and configuration requirements.

- **Tests**
  - Added coverage for clamping, saturation, gradients, cache separation, validation, quantization paths, and backward consistency.

### DIFF
--- a/docs/fe-oss-apis/gemm_fusions/grouped_gemm_dsrelu.md
+++ b/docs/fe-oss-apis/gemm_fusions/grouped_gemm_dsrelu.md
@@ -55,6 +55,28 @@ $$
 \mathrm{dprob}[m, 0, 0] = \sum_n C[m, n, 0] \cdot \mathrm{relu}(G[m, n])^2
 $$
 
+#### Tanh soft clamp (`tanh_clamp_scale`)
+
+`tanh_clamp_scale=s` differentiates the soft-clamped forward epilogue
+(see [grouped_gemm_srelu](grouped_gemm_srelu.md)) instead of the plain one. Writing
+$t = \tanh(\mathrm{relu}(G)/s)$ and $b = s\,t$, both equations keep their shape with
+$\mathrm{relu}(G)$ replaced by $b$, and the row output picks up the chain-rule factor
+$1 - t^2$:
+
+$$
+D\_{row}[m, n] = \mathrm{prob}[m, 0, 0] \cdot 2 \cdot C[m, n, 0] \cdot b[m, n] \cdot \left(1 - t[m, n]^2\right)
+$$
+
+$$
+\mathrm{dprob}[m, 0, 0] = \sum_n C[m, n, 0] \cdot b[m, n]^2
+$$
+
+Both kernels must be built with the same `s`, or the saved pre-activation is differentiated
+against the wrong nonlinearity. `s` is baked in at compile time, so distinct values compile
+distinct kernels. Determinism is unaffected: the clamp changes only the per-element value that
+feeds the dprob accumulation, never the slotting or reduction order, so
+[deterministic dprob](#deterministic-dprob) stays bit-exact for the clamped path too.
+
 `D_col` stores the companion column-quantized output used by the grouped kernel family. When FP8 output is enabled, the kernel also emits `SFD_row` and `SFD_col`. When fp16/bf16 output is used, the kernel can emit per-expert `Amax`.
 
 ### Diagram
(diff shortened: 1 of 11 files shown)


vedaanta:
@cudnn-ci-bot run oss

vedaanta (approved):

PR #865, opened 2026-09-02
docs(sdpa): add the FROST SDPA support matrix + Rule S2 to keep it current

RomanAnders90:
## Affected area

Documentation or samples (FE OSS kernels / CuTeDSL — FROST SDPA)

## Summary

Adds `docs/frost/SDPA_SUPPORT_MATRIX.md`: a per-architecture support matrix for the FROST SDPA engines. One table per arch (**SM100/SM103**, **SM107**, **SM120/SM121**, **SM80**); columns are the kernel flavor (native head-dim geometry, with the model class it was tuned for in brackets) crossed with FPROP/BPROP; rows are data types, layout/ragged, and masks/features.

Every cell is derived from a `Capabilities` field in `fwd/engines.py` / `bwd/engines.py` — the rows that actually admit or decline a graph — not from a kernel grep.

Plus a rule to keep it current, in the places review already looks (details below). **Docs only: no code, no tests, no behaviour change.**

## Why

What FROST SDPA serves is spread across five `Capabilities` rows. Answering something as ordinary as *"does FROST cover d=64 FP8 on SM100, and does THD work there?"* currently means reading all of them plus `_pick_flavor` and the envelope predicates in `mismatch()`. This makes it one page.

The table deliberately encodes three things a `Capabilities` row does not show on its face:

**1. Native vs envelope-served.** A head dim with no kernel of its own rides the smallest covering flavor with TMA zero-padding — correct, but it pays that flavor's MMA cost. This gets its own glyph (⚠️), because it is a perf cliff that is invisible from the engine row. The worked case, and the reason this distinction exists:

> There is no d=64 geometry on SM100. `_SM100_FLAVORS` is `((128,128), (192,128), (256,256), (512,512))` (`fwd/api_dsl.py:56`), so `_pick_flavor(64, 64)` returns `(128, 128)` and a GPT-OSS-class graph runs the **d128 Llama kernel** with the contraction zero-padded 64 → 128 — numerically exact, but roughly **2× the MMA work** on both BMM1 and BMM2. d=64 is native on SM80 (`gptoss`) and SM120 (64 is a supported head tile); SM100 and SM107 are the gap.

**2. Exact-only gates that the envelope does not cover.** MXFP8 sets `d_pad_multiple=0`, so d=64 MXFP8 is *declined*, not padded. And `thd_d_shapes` is exact tuple membership rather than an envelope (`engines.py:384`), so quantized THD at d=64 is declined while f16/bf16 THD works.

**3. Feature support varying by dtype inside one head-dim column** on SM100: `dense_flex` is f16-only, MXFP8 lacks the padded+stats LSE trim (`padded_stats=False`), and both FP8 and MXFP8 lack the dense padded-Q trim.

The doc closes with a gaps summary — no backward engine at all on SM100/SM103/SM107, no f16 forward on Rubin, no quantized or ragged backward on any arch.

### Keeping it current

A hand-maintained table rots silently, so this adds the rule where reviews already look, rather than adding a new CI gate:

- **`python/cudnn/sdpa/AGENTS.md` — Rule S2**: a change to a `Capabilities` row or an `EngineSpec` updates the matrix in the same commit; knob-domain-only changes are exempt. Numbered per that file's append-never-renumber convention so review can cite it.
- **`python/cudnn/frost/README.md` § "The rules" #14**, next to its existing "keep this document true".
- One pointer in the root `AGENTS.md` conventions, one index entry in `llms.txt`.

No new enforcement was introduced: the root AGENTS.md *"Reviewing a PR"* section already directs reviewers (human or agent) to the Hard Rules of every directory a diff touches, and the PR template already carries the matching checkbox. If the team would rather have this mechanically enforced, a path-based CI check (engines.py touched + matrix untouched ⇒ fail) is easy to add — deliberately left out of this PR as a separate decision.

## Related issues

None.

None. Documentation and agent-guide text only.

## Testing

No automated test applies — the change is markdown. What was done instead:

- Every cell was derived by reading the `Capabilities` rows, and the head-dim/THD envelope conclusions were checked by replaying the `mismatch()` / `_pick_flavor` predicates directly (confirming `(64,64) → (128,128)` for f16 and FP8, `(64,64)` declined for MXFP8, and `(64,64) ∉ thd_d_shapes`).
- Table column counts and footnote-marker/definition pairing verified by script (0 mismatches).
- `pre-commit run` is a **no-op** on this diff: all five hooks are typed `c++ / c / cuda / python / jupyter`, and every changed file is markdown.

### Review ask

The table is a first cut and the flavor→model-class naming (Llama / DSv3 MLA / Qwen / DSv4 / GPT-OSS) is my reading of the config and kernel docstrings — please correct anything mislabeled, and say if any cell overstates what the row really serves.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive FROST SDPA support matrix covering architecture, data type, layout, masking, feature support, forward/backward passes, fallback behavior, and known limitations.
  * Updated guidance and documentation links to direct contributors to the new support-matrix tracker for recording and reviewing SDPA coverage changes.
  * Clarified which capability changes require support-matrix updates and documented knob-only exceptions.
  * Documented validation caveats, head-dimension behavior, architecture-specific details, and major support gaps.


vedaanta (approved):

RomanAnders90:
Thanks — all four findings were valid and are fixed in 40aa15a8, along with a path move.

**Moved:** `docs/frost/SDPA_SUPPORT_MATRIX.md` → `python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md`. This is a maintainer-facing tracker of what the engine rows admit rather than published user documentation, and Rule S2 asks for it to be updated in the *same commit* as a `Capabilities` change — sitting it beside `fwd/engines.py` / `bwd/engines.py` makes that the path of least resistance instead of a cross-tree edit. References updated in `llms.txt`, root `AGENTS.md`, `frost/README.md` #14 and `sdpa/AGENTS.md` Rule S2.

**Line 125 — SM120 alignment inconsistency.** Correct catch, and the header was the wrong one. `SUPPORTED_HEAD_TILES` steps by 16, but that is the head *tile* granule, not the eligibility rule: an actual `D_QK`/`D_V` only has to satisfy `d_pad_multiple`, which is **8** on the f16 row (`engines.py`, sm120 spec) and **16** on the FP8 row. The header now quotes the head-dim rule, and a new footnote ᵃ spells out the distinction — a d=72 f16 graph *is* eligible and computes on the 80-wide tile, which the old header wrongly excluded.

**Line 41 — O-dtype row.** Agreed it over-advertised. Kept it one row but scoped the label explicitly (`O dtype ≠ QKV dtype — **quantized graphs only**`) rather than splitting, since the columns are head-dim flavors and a split would duplicate five of six cells. Footnote ¹ now states both halves: quantized graphs may emit FP16/BF16/E4M3/E5M2; on the f16 row there is no `out_dtypes` domain and `facts.uniform_dtype` requires `O == Q`.

**Line 107 — accepted-but-untested marked as declined.** This one was the most useful of the four, because it exposed a gap in the legend rather than just a wrong cell. The legend defined `❌` as "declined at plan time", but `softmax_precision=HALF` on a d=64 SM107 graph *passes* the probe — the domain gates on `flavor == (128, 128)` (`fwd/api_dsl.py:1100`) and a d=64 graph's flavor **is** `(128,128)`. What is untested is the f16x2 exponent arm over the zero-padded 64 → 128 region. Added a third status, `❔` = "accepted by the capability row but not validated on this path", so the table stops conflating *we reject this* with *we have not checked this*.

**Line 37 — MD056.** All 9 single-cell section-label rows now span their tables.

One note on the CodeRabbit summary's "several formatting and eligibility inaccuracies, which could mislead users about supported configurations": the SM120 header was a genuine eligibility error and is fixed. The other three were presentation — the underlying support claims were right, but the table said them ambiguously. Every cell is still derived from a `Capabilities` field rather than a kernel grep, and the flavor→model-class naming is still the part I would most like a human eye on.

RomanAnders90:
Valid, and fixed in 9a42c2bc — thanks, this one was a real drift hazard rather than a wording nit. Enumerating the triggering fields in three places guarantees the three lists diverge, and the root bullet had already lost `out_dtypes`, `padded_stats` and `decode`.

Took the suggested wording ("any `Capabilities` field that affects graph eligibility") in both the root `AGENTS.md` bullet **and** `frost/README.md` rule 14, which had the same partial-enumeration shape and was not flagged.

One deliberate departure from the proposed diff: **I kept the knob-domain exemption in the root bullet.** The suggestion drops it, which would make the root guide *broader* than canonical Rule S2 — it would imply a `tile_ms` / `sched_policies` change needs a tracker edit, which S2 explicitly exempts because the matrix does not track knobs. Trading a false negative for a false positive is still drift, just in the other direction.

Both pointers now name `python/cudnn/sdpa/AGENTS.md` **Rule S2** as canonical for the exact scope, so there is one list to keep true instead of three. Also rewrapped rule 14 to the file's 76-column prose width, which my earlier path swap had overrun.

PR #816, opened 2026-08-31
make sm120 matmul configurable and support all m/n/k-major combinations

yihuawei:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: one `cat-*`, one or more `mod-*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).
- [x] I set the **Milestone** and **Projects** fields in the sidebar (required to merge; maintainers can set these for external contributions).

## Summary by CodeRabbit

* **New Features**
  * Expanded SM120 matrix multiplication support across configurable tile sizes, warp layouts, K widths, and operand memory layouts.
  * Added support for 8-bit operands with additional major-layout combinations and transposed loading.
  * Improved configuration validation and pipeline selection for supported geometries.

* **Bug Fixes**
  * Registry configurations without CTA group settings now use a safe default.
  * Improved memory-layout and swizzle validation across supported configurations.

* **Tests**
  * Expanded correctness and rendering coverage across SM120 configurations, layouts, geometries, and edge cases.

PR #871, opened 2026-09-02
frost gemm heur: block-scale early pairing only when M is exact, not a MoE group average

brandonfzhang:
Fixes the `cta_group` assert failure in `test_planner_uses_total_m_not_unobservable_moe_row_distribution` introduced by #844 (first seen in `frost-gemm:cutlass-rel:sm100`).

**Root cause.** #844's block-scale early-pairing rule takes the 2-CTA MMA pair without a second M tile while the padded pair row fits the machine in one wave. That wave bound is arithmetic over M — valid only when M is an exact problem dimension. `plan_config` hands MoE chains a **per-group average** M (the only plan-time bound on runtime-ragged expert groups), and for `M=2816, moe_groups=23` the averaged `tile_m=123` sits below the terminal-quantizer force's total-row bound lapse point, so the ordinary heuristic decides — and the new disjunct paired a chain the planner tests pin at 1-CTA. The pairing win was measured on dense block-scale GEMMs where M is exact; an average of ragged groups is outside that envelope.

**Fix.** `select_config` grows an append-only `m_is_group_average: bool = False` parameter; the early-pairing disjunct additionally requires it false. `plan_config` sets it from `chain.moe is not None`. The precondition lives on the rule that depends on it — no duplication of tile-selection internals in the planner, and `force_cta_group` precedence is untouched.

Verified against all pinned expectations: MoE `groups=23` → 1-CTA, `groups=2` (`tile_m=1408 > cta_m`) → 2-CTA, dense block-scale M-starved keeps the measured pairing win, forced 1-CTA still overrides, dense batch cases unchanged.

(The `test_sm120_warp_grid_axis` failures in the same job are unrelated — TMEM-column overflow from #816.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic GEMM configuration selection for mixture-of-experts workloads using per-group average sizes.
  - Prevented inappropriate 2-CTA selection for block-scaled workloads when the reported size represents a group average.

### DIFF
--- a/python/cudnn/gemm/frost/compiler.py
+++ b/python/cudnn/gemm/frost/compiler.py
@@ -3484,6 +3484,7 @@ def plan_config(chain: FusionChain) -> TileConfig:
         b_n_major=chain.matmul.b_major == "n",
         b_elem_bytes=DTYPE_BYTES[chain.matmul.b_dtype],
         force_cta_group=force_cta_group,
+        m_is_group_average=chain.moe is not None,
     )
     # Re-target at the preferred family and MMA-inst K width; cta_group rides
     # the geometry and only moves when the family cannot serve it (sm120 is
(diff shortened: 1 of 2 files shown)


brandonfzhang:
@cudnn-ci-bot run frost

yanqinz2 (approved):

PR #887, opened 2026-09-03
frost(sdpa): SM100 d=512 backward (f16/bf16) — three-stage chain, masks, GQA, any seqlen

RomanAnders90:
## SM100 d=512 SDPA backward (f16 / bf16)

Adds `sdpa_bwd_sm100`, a FROST backward engine for the large-head-dim band
**d ∈ (256, 512]** on Blackwell. Before this, every `sdpa_backward()` on SM100
fell through to the cuDNN backend, which does not serve that band.

It is a **three-stage chain, not one fused kernel**. A fused d=512 backward
needs 512 TMEM columns for dV and 512 more for dK against 512 per CTA, so it
does not fit: S and dS go to a GMEM workspace and the gradients become three
batched GEMMs over it (`do_dot` → `bprop_d512_f16_sm100` → `bprop_matmul_sm100`).
Stage 2 forks the *forward* d512 SM100 kernel's cga4x1 role split, which gives
each sub-group its own 227 KiB of SMEM and its own 512 TMEM columns.

Two consequences a user can see, both documented in the support matrix:
the workspace is `2·B·H_chunk·S_q·S_kv·2 B` (the host loops over head chunks to
hold it under 4 GiB), and the whole band is **envelope-served** — tiles are
fixed at 512, so d=264 pays d=512's MMA cost.

### What it serves

f16/bf16 · d ∈ (256, 512] in multiples of 8 · GQA/MQA · causal (top-left and
bottom-right) · SWA · right-band widening · arbitrary S_q/S_kv (not just tile
multiples) · BSHD and arbitrary dense stride order (`dense_flex`).

Declined, each with an asserted reject test: THD/ragged, per-batch seq_len,
bias/dbias, sinks, deterministic mode, FP8/MXFP8, and d outside the band.

### Commits

| | |
|---|---|
| `3e62b316` | Bring-up: the kernel chain, engine row, manifest slot, GQA, causal/SWA/band masks |
| `a407a668` | Arbitrary seqlen, the new `d_envelope_floor` capability, and the 33-case test module |
| `43d51836` | Drop padding from the workspace memset (~3 % at B1 H128 S8192 causal) |
| `1ce1a474` | Stage-3 GEMM 2x1 → 2x2 cluster |
| `d1edf029` | Honor FP16 in the stage-3 GEMM |

### Notes for review

**`d_envelope_floor` is a new `Capabilities` field.** An envelope with no floor
silently claimed every small head dim and padded it onto the 512-wide kernel —
a d=128 graph would have taken the whole 512 MMA with a d256 flavor sitting
right there. Only the adapter's `check_support` caught it, i.e. *after*
eligibility, so the row was lying. It defaults to 0, so the sm120/sm80 rows are
unchanged. `SUPPORT_MATRIX_TRACKER.md` is updated in the same commit (Rule S2).

**Causal correctness rests on the zero-fill, not the trim.** At the 2x2 cluster
config the stage-3 M tile (512) is wider than stage 2's write block (256), so a
512-row tile straddles two blocks and no per-tile K range can exclude the
skipped region. The trim is an optimization; `_zero_ws` is what makes the causal
path correct. Two docstrings previously claimed otherwise and are corrected —
please do not narrow that zero-fill later without re-reading them.

**The stage-3 cluster change was not a pure constants swap.** Beyond the
tile/cluster/stage constants it needed the epilogue's SMEM staging widened
*and* the matching output TMA descriptor (`box_dims` 32 → 64, `s64b` → `s128b`).
Missing only the descriptor half leaves the epilogue writing a 128 B swizzle
through a 64 B descriptor: no crash, no launch failure, just `cos ≈ 0.006` on
all three GEMMs. The epilogue store and its descriptor are one unit.

**The FP16 fix is worth a look** (`d1edf029`): the row claimed `{HALF, BFLOAT16}`
while the stage-3 template hardcoded `cutlass.BFloat16` in seven places, so every
fp16 graph was accepted and then failed at execute. All 33 tests passed over it
because all 33 ran bf16. The tests are now parametrized per dtype.

### Performance

Stage-3 GEMM at 2x2, measured B=1 H=128 S=8192 d=512 bf16, paired alternating
A/B/B/A so clock drift cannot pick a winner:

| mask | 2x2 | 2x1 | |
|---|---|---|---|
| no_mask | **732.3** TFLOPS | 707.8 | +3.5 % (4/4 rounds) |
| causal | **653.5** TFLOPS | 606.0 | +7.8 % (2/2 rounds) |

The three products are within 1 % of each other at ~41.4 % of the boost-clock
SOL *bound* (a lower bound — `pynvml` is unavailable on the node, so this is not
a sampled-clock figure). An earlier isolated-GEMM A/B read the same change as a
coin flip; it was measuring the GEMM out of context, on boosted clocks against a
warm cache, and is superseded by the paired end-to-end numbers.

Broader perf tuning and THD support are the planned follow-ups.

### Testing

`test/python/sdpa/frost/test_sdpa_bwd_dsl_sm100.py` — **38 cases**, accept AND
reject for every claim, marked `L0`. Reject tests ask the row's own `mismatch()`
rather than walking the ranked plan list, which is confounded by the backend
(it serves d=128/256 backward and raises outright on some graphs). The THD and
padding-mask rejects are asserted on real graphs so they exercise the analyzer,
and both invert when those features land rather than being deleted.

Validated on Blackwell (cc 10.0, cuDNN 9.27): 38/38 over 3 consecutive runs,
plus the full `sdpa/frost/` suite (1100 passed) to check for regressions.

Two bugs the tests found during development, both listed here because the
signatures are reusable:

* `do_dot` writes `delta` with a row stride of `ceil(S_q/128)*128`, not `S_q`,
  so the buffer and stage 2's view of it must round identically. They coincide
  at every tile-multiple S_q, which is why only a ragged length exposed it —
  and the signature was exact: **dV correct, dQ and dK wrong**, because dV is
  the one gradient that does not consume `do_dot`.
* The envelope floor above, found by a reject test rather than by review.

## Summary by CodeRabbit

- **New Features**
  - Added SDPA backward support on Blackwell SM100–SM103 GPUs for FP16/BF16 head dimensions greater than 256 and up to 512.
  - Supports causal and sliding-window masking, GQA/MQA, dense layouts, and irregular sequence lengths.
  - Omitted attention scales now default based on the query/key head dimension.

- **Bug Fixes**
  - Benchmark diagnostics now distinguish unsupported configurations, missing clock samples, and peak lookup failures.
  - Unsupported backend plans now report clear “graph not supported” results.

- **Documentation**
  - Updated the support matrix with SM100/SM103 capabilities and limitations.

### DIFF
--- a/benchmark/attention_training/benchmark_single_sdpa.py
+++ b/benchmark/attention_training/benchmark_single_sdpa.py
@@ -1978,8 +1978,29 @@ else:
     except Exception:
         pass
 
-    fwd_sol_str = f", {fwd_tflops / _peak_mma_tflops * 100:.1f}% SOL" if _peak_mma_tflops and fwd_tflops > 0 else ""
-    bwd_sol_str = f", {bwd_tflops / _peak_mma_tflops * 100:.1f}% SOL" if _peak_mma_tflops and bwd_tflops > 0 else ""
+    def _sol(tflops):
+        """SOL% against the dense-MMA peak, or an explicit note when we cannot
+        compute it.
+
+        The peak needs a sampled boost clock, which needs ``pynvml``. When that
+        import is missing the SOL suffix used to vanish entirely and an
+        impossible TFLOPS number printed unremarked -- the only cross-check this
+        harness has on its own arithmetic, silently disabled by a missing
+        optional dependency. Say so instead, and shout when a number exceeds the
+        hardware: >100% SOL means the FLOP model or the timing is wrong, never
+        that the kernel is fast.
+        """
+        if tflops <= 0:
+            return ""
+        if not _peak_mma_tflops:
+            return ", SOL n/a (no clock sample -- pip install pynvml)"
+        pct = tflops / _peak_mma_tflops * 100
+        if pct > 100.0:
+            return f", {pct:.1f}% SOL -- ABOVE HARDWARE PEAK, MEASUREMENT IS WRONG"
+        return f", {pct:.1f}% SOL"
+
+    fwd_sol_str = _sol(fwd_tflops)
+    bwd_sol_str = _sol(bwd_tflops)
 
     if args.format_output:
         print(
--- a/benchmark/attention_training/benchmark_single_sdpa.py
+++ b/benchmark/attention_training/benchmark_single_sdpa.py
@@ -1969,31 +1969,40 @@ else:
     # Compute MMA SOL% using the per-arch FLOPs/clk/SM table and the actual
     # sampled boost clock observed during the benchmark window.
     _peak_mma_tflops = None
+    # Why the peak is unavailable, when it is -- the two causes need different
+    # advice and conflating them sends the user after the wrong one.
+    _peak_unavailable = "unavailable"
     try:
         _flops_per_clk_per_sm = _peak_flops_per_clock_per_sm(args.data_type)
         _num_sms = torch.cuda.get_device_properties(torch.cuda.current_device()).multi_processor_count
         _sampled_mhz = _clock_sampler.peak_mhz()
-        if _flops_per_clk_per_sm is not None and _sampled_mhz is not None:
+        if _flops_per_clk_per_sm is None:
+            # No dense-MMA peak modelled for this dtype / arch (e.g. --data_type float).
+            _peak_unavailable = f"no MMA peak modelled for {args.data_type} on this arch"
+        elif _sampled_mhz is None:
+            _peak_unavailable = "no clock sample -- pip install pynvml"
+        else:
             _peak_mma_tflops = _flops_per_clk_per_sm * _num_sms * _sampled_mhz / 1e6
-    except Exception:
-        pass
+    except Exception as _e:
+        _peak_unavailable = f"peak lookup failed ({type(_e).__name__})"
 
     def _sol(tflops):
         """SOL% against the dense-MMA peak, or an explicit note when we cannot
         compute it.
 
-        The peak needs a sampled boost clock, which needs ``pynvml``. When that
-        import is missing the SOL suffix used to vanish entirely and an
-        impossible TFLOPS number printed unremarked -- the only cross-check this
-        harness has on its own arithmetic, silently disabled by a missing
-        optional dependency. Say so instead, and shout when a number exceeds the
-        hardware: >100% SOL means the FLOP model or the timing is wrong, never
-        that the kernel is fast.
+        The peak needs both a modelled dense-MMA rate for the dtype and a
+        sampled boost clock (which needs ``pynvml``). When either is missing the
+        SOL suffix used to vanish entirely and an impossible TFLOPS number
+        printed unremarked -- the only cross-check this harness has on its own
+        arithmetic, silently disabled. Say so instead, naming WHICH of the two
+        is missing, and shout when a number exceeds the hardware: >100% SOL
+        means the FLOP model or the timing is wrong, never that the kernel is
+        fast.
         """
         if tflops <= 0:
             return ""
         if not _peak_mma_tflops:
-            return ", SOL n/a (no clock sample -- pip install pynvml)"
+            return f", SOL n/a ({_peak_unavailable})"
         pct = tflops / _peak_mma_tflops * 100
         if pct > 100.0:
             return f", {pct:.1f}% SOL -- ABOVE HARDWARE PEAK, MEASUREMENT IS WRONG"
--- a/include/cudnn_frontend/backend/plan_helpers.h
+++ b/include/cudnn_frontend/backend/plan_helpers.h
@@ -5,6 +5,7 @@
 
 #pragma once
 
+#include <sstream>
 #include <vector>
 
 #include "cudnn.h"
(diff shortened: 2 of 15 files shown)


RomanAnders90:
Thanks — went through all five. Two were real and are fixed in 42a37c5d; three I checked and am leaving as-is, with reasons.

**Fixed**

1. **`attn_scale` default (`api_dsl.py`)** — real bug, and worse than the comment suggested. `attn_scale` is optional on the graph, so `scale_softmax` arrives `None`; SM120 and SM80 default it to `1/sqrt(d_qk)` and SM100 did not. The row admitted the graph, `check_support()` passed, and execute died with `TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'`. Reproduced, fixed, and covered by `test_default_attn_scale` — which was seen RED against the unfixed adapter. It compares against a reference that always uses `1/sqrt(d)`, so a *wrong* default fails too, not just a missing one.

2. **Benchmark SOL message** — agreed. `_peak_mma_tflops` is unavailable for two different reasons (no modelled MMA rate for the dtype/arch, vs. no sampled clock) and the note blamed `pynvml` for both, which sends the reader after the wrong thing for e.g. `--data_type float`. It now names which.

**Not changed**

3. **`d512_supported` permits `d_qk != d_v`** — the Python side already declines it: `mismatch()` hits `elif facts.d_qk != facts.d_v` for this row (`bwd/engines.py:191`), since it does not set `dqk_ge_dv`. The header being permissive only means such a graph falls through to plan creation, which is the behaviour its own comment documents for the whole band. I did add `test_reject_rectangular_head_dims` in 42a37c5d, because that decline was unasserted and it is the only thing keeping a rectangular graph away from an adapter that raises on it.

4. **clang-format on `scaled_dot_product_flash_attention.h`** — the `pre-commit` CI job runs the repo's pinned clang-format 21 hook over this diff and passes.

5. **Move the sweeps off `L0`** — the sibling SM120 backward suite runs 119 cases at `L0`; this module is 40 and runs in ~52 s on the Blackwell node, so it is well within the established bar for this directory.

Re-validated on Blackwell (cc 10.0, cuDNN 9.27): 40/40.

RomanAnders90:
Pushed 24d3fdc2 — widening `test_sdpa_random_bwd_L0`'s head-dim sweep to 512 found a real defect in this branch, worth calling out for review.

**What happened.** Setting `d_qk_max`/`d_v_max` to 512 turned **91 of 384** cases red (baseline: 289 passed, 0 failed). None was a numerics bug — all 91 were this branch reporting *"nobody serves this graph"* as the **wrong kind of error**.

Admitting d ∈ (256, 512] at the node's validation (so the FROST engine can claim the band) means such a graph now reaches plan creation. If the caller also asks for something no engine serves — deterministic, say, which `sdpa_bwd_sm100` declines and the backend has no plan for in this band — then `override_heuristics_query()` pins a backend engine id (5/17) and **bypasses the heuristics query entirely**. That pinned config fails to finalize with `CUDNN_STATUS_NOT_SUPPORTED`, which `_CUDNN_CHECK_CUDNN_ERROR` folds into `CUDNN_BACKEND_API_FAILED` → a bare `RuntimeError` in Python.

The type is load-bearing: every SDPA harness here skips on `cudnnGraphNotSupportedError` and **fails** on anything else, so "unsupported" became "the backend broke" for any caller trying to feature-detect.

**Fixes** (narrow, then general):

1. `override_heuristics_query()` returns `{-1, {}}` for the band. That override exists to pin a specific backend bprop kernel, and the backend has no plan in (256, 512] at all — pinning one there cannot succeed. Declining it lets heuristics run, return no configs, and report `GRAPH_NOT_SUPPORTED`, which is already what band graphs that did *not* trip the override were doing.
2. `create_engine_config` maps a finalize `CUDNN_STATUS_NOT_SUPPORTED` to `GRAPH_NOT_SUPPORTED` rather than `CUDNN_BACKEND_API_FAILED`; any other status stays an API failure. **This one is beyond the SDPA path and is the piece most worth a second opinion** — I believe NOT_SUPPORTED from the backend means exactly that at every call site, but it is a shared header.

`test_unserved_band_graph_declines_as_not_supported` asserts the exception *type*, not its message, and was seen RED against the unfixed headers reproducing the original `RuntimeError` exactly.

**Coverage trade to be aware of** in the widened sweep: the generator now spends part of its 384-case budget on a band this suite cannot execute (FROST engines are opt-in and it does not enable them), so those cases skip — **200 passed / 184 skipped**, against 289 / 95 before. It buys frontend admission-and-decline coverage for the band; executing coverage stays in `sdpa/frost/test_sdpa_bwd_dsl_sm100.py` (41 cases).

Validated on Blackwell (cc 10.0, cuDNN 9.27): widened sweep 91 failed → 0; full `test_mhas_v2.py` L0 **2090 passed, 796 skipped, 0 failed**; `test_sdpa_bwd_dsl_sm100.py` 41/41; C++ build clean under `-Werror -Wall -Wextra -Wpedantic`.

RomanAnders90:
@cudnn-ci-bot run frost

RomanAnders90:
On the line-183 formatting nitpick: black itself disagrees. `black -l 160 --fast --check test/python/test_mhas_v2.py` reports the file unchanged, both with and without this edit, and the `pre-commit` CI job (which runs that same pinned hook) passes. The line is 219 chars, but its untouched siblings in the same file are 209 — this is one of the pre-existing not-formatter-clean files AGENTS.md warns against reformatting, and splitting only the line I touched would leave it inconsistent with the fifteen identical `RandomHiddenDimSize(...)` calls around it. Leaving it matching its neighbours.

vedaanta (approved):

PR #892, opened 2026-09-03
sdpa: consume the persistent THD scheduler in the last two FP8 kernels

### LINKED ISSUE #618: frost(sdpa): THD envelope grid dead-tile tax under far-oversized declarations — capped persistent grid with device-read live-unit count
Follow-up from #552 (resolved by #543 + #606 + #608): the THD forward grid is now the plan-time envelope — `B * ceil(S_q_declared / tile) * QH` units on SM100, `(ceil(S_q_declared / q_tile), B, H)` on SM120 — with dead tiles exiting via the kernel contract. That removed the last execute-time D2H and made THD capture-legal, but a caller whose declared `S_q` is far above the live token totals pays for the dead tiles.

Measured on SM100 (B200, d128, B=8 QH=8, S_decl=16k, CGA_TILE_M=512):

| declaration vs live | exact grid | envelope | tax |
|---|---|---|---|
| full/near-full | 31.5 ms | 31.5 ms | ~0% |
| half-length | 8.64 ms | 8.66 ms | +0.2% |
| 3% live (64/2048 units) | 44 µs | 330 µs | +648% |

Dead units absorb at ~20 ns each while live work runs, ~145 ns each when exposed. Realistic THD prefill declarations (S_decl ≈ max length in batch) pay <0.2%; the pathological case is a fat static declaration with tiny live batches — the same class of tax the C++ backend's `s_max` grid has always paid.

**Proposed fix (SM100)**: cap the grid at the resident-cluster count, have the setup kernel (which already builds the metadata per execute) also write the live unit total `Σ_b ceil(s_b/tile) * QH` into a metadata slot, and replace the scheduler warp's `clusterlaunchcontrol_try_cancel` id source with a grid-stride sequence bounded by that device-read total (`scheduler_warp_loop` in `frost/tile_dsl/scheduler.py`). Dead cost then scales with #SMs instead of the envelope, and capture-legality is preserved (the bound is read on device). SM120's rectangular grid would need the analogous per-CTA early-out against a device-read `ceil(max_live_sq/q_tile)`.

Worth doing only with a motivating workload — an end-to-end capture-mode benchmark against the C++ backend / FA varlen would decide.

yanzhuo607:
prefill_d512_fp8_sm100 and prefill_fp8_sm120 sized their THD grid from the declared S_q envelope while already launching the setup kernel that publishes the live unit total and the claim counter. Both now consume it: the launch caps at the resident cluster/CTA count and units come off the device-bounded counter, so no unit past the live total is ever handed out and no tile past a sequence's real length is scheduled.

The SM120 FP8 kernel gains the per-unit split its f16 sibling already carries: the K/V mbarrier parity continues across units rather than restarting, and the load warp drains the last tile's consumer arrivals, so a CTA can re-enter the pipeline for a second unit. Both kernels gain a case where a CTA claims more than one unit, which nothing covered before.

Closes #618.

## Before submitting

## Summary by CodeRabbit

* **New Features**
  * Added paged key-value cache support for FP8 attention on supported SM100 GPUs.
  * Added padded per-batch statistics output for variable-length attention on SM100 and SM120 GPUs.
  * Added support for dynamic batch and head dimensions in supported variable-length workloads.

* **Bug Fixes**
  * Improved FP8 attention scheduling and work distribution for ragged workloads.
  * Improved support for larger head dimensions, varied CTA configurations, and padded layouts.
  * Added validation for padded statistics buffers and dense query-length inputs.


yanzhuo607:
@cudnn-ci-bot run frost

vedaanta (approved):

yanzhuo607:
@cudnn-ci-bot run frost

YangXu1990uiuc (approved):

Reviewed `b3ec7cf84b6e` against base `a3f82cdf9a7b`.

The persistent THD conversion makes sense. I found no new actionable correctness issue in this head. The SM120 phase count follows actual K/V loads, the final consumer arrivals are drained, and the claim helper separates shared-memory use between units. SM100 d512 uses the cluster size when capping the launch and the matching CGA stride when decoding claimed units.

Validation: SM100 exact-head d512 THD selection: 9 passed, including forced multi-unit cluster reuse, both FP8 input types, causal/noncausal, GQA, sinks/Stats, empty KV and cu_seqlens. Two additional D512/Hq=8/Hkv=2 bottom-right+sink workloads passed the reference, sync-debug check and changed-V CUDA Graph replay after poisoning output/Stats/workspace. Q/KV lengths were [128,64,0,1]/[128,0,64,65] and [1024,128,16,0]/[256,512,64,0]. Baseline/head/head/baseline on the same SM100: warmed GPU us/call were 23.072/23.586/23.590/23.072 for the short case and 37.162/32.708/32.677/37.214 for the ragged case. CPU execute medians were 139/130/135/123 us and 128/131/138/122 us respectively; host variation is too large for a firm CPU regression claim. The existing [SM120 re-entry barrier concern](https://github.com/NVIDIA/cudnn-frontend/pull/892#discussion_r3993628060) is covered in this head by the leading full-CTA barrier inside [thd_claim_next](https://github.com/NVIDIA/cudnn-frontend/blob/b3ec7cf84b6eec048e7b9fc764cec4837f91df27/python/cudnn/frost/tile_dsl/thd.py#L270); another barrier at the caller is unnecessary. The test also now pins four CTAs and replaces the CTA-count cache.

Limitations: No independent SM120 execution or timing in this review. The owner already started current-head FROST CI, including SM100 and RTX 5080/SM120; those lanes are still running, so their results are pending. Local measurements use cuDNN 9.25.1, torch 2.13/CUDA 13.0 and CuTe DSL 4.8 development, and cover two sampled workloads only.

PR #894, opened 2026-09-03
frost(sdpa): align the CLC scheduler with the canonical CUTLASS pattern

yanzhuo607:
Four ordering deviations from the canonical PipelineClcFetchAsync contract. None of them caused a known hang, but each is a real window that today is closed only by timing:

- read_clc_payload() decodes a try_cancel response with one atomic 128-bit load instead of three independent 32-bit loads. The credit that lets the scheduler refill a slot is arrived at the top of the same loop iteration that reads it, so three loads leave a partially-decoded slot exposed to a refill; one indivisible load removes that state. All SM100/SM107 prefill kernels decode through the helper.

- The helper ends in a generic->async cross-proxy fence, ordering the read before the scheduler's next write into the slot (the DSL's documented requirement).

- read_tile_id_arrive issues a release rather than a relaxed arrive, so the payload reads are ordered before the credit that gates the refill.

- The first CTA's scheduler warp arms arrive+expect_tx on every CTA's barrier, program-ordered before it issues the multicast try_cancel. Arming per-CTA locally left a peer's expect_tx unordered against the leader's issue. scheduler_warp_loop takes a cga_size fan-out parameter.

The helper takes a slot base word rather than a stage index: the predecode kernels size tile_id_smem by SCHED_PAYLOAD_WORDS, which is not always 8.

The THD paths hand payloads over with explicit per-peer DSMEM stores rather than a multicast try_cancel, so they keep their per-CTA arming; they pick up the decode and arrive changes, which are shared.

## Before submitting

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved persistent scheduling reliability across supported attention kernels.
  - Added consistent cluster-aware scheduling for standard and variable-length execution.
  - Improved scheduler barrier coordination, including local and peer synchronization.
  - Standardized atomic scheduler response decoding across tile loading, computation, storage, and correction workflows.
  - Preserved tile validity checks, metadata decoding, and warp-uniform behavior while improving synchronization consistency.

### DIFF
--- a/python/cudnn/frost/tile_dsl/scheduler.py
+++ b/python/cudnn/frost/tile_dsl/scheduler.py
@@ -98,7 +98,12 @@ def read_tile_id_arrive(mb, cga_size: int):
             target_lane = i * lane_stride
             if lane == cutlass.Int32(target_lane):
                 peer_mb = nvvm.mapa(mb, cutlass.Int32(i))
-                nvvm.mbarrier_arrive(peer_mb, scope=nvvm.MemScope.CLUSTER, relaxed=True)
+                # RELEASE (not relaxed) arrive: this signal is the credit that
+                # lets the scheduler refill the very slot this warp decodes, so
+                # the payload loads must be ordered-before the arrive becomes
+                # visible.  A relaxed arrive imposes no such order and lets the
+                # refill race a still-in-flight decode.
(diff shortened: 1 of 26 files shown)


yanzhuo607:
@cudnn-ci-bot run frost

vedaanta (approved):

vedaanta:
@cudnn-ci-bot run frost

yanzhuo607:
@cudnn-ci-bot run frost

PR #897, opened 2026-09-04
sdpa: admit the (256, 512] FP8 head-dim band for per-tensor FP8 only, not MXFP8

vedaanta:
## Problem

The attention_training nightly reports the **libcudnn backend failing with `CUDNN_STATUS_INTERNAL_ERROR_COMPILATION_FAILED` on 16 cases each on GB200, GB300 and VR200** (plus 8 of the same under the OSS runner's backend fallback). They are the `deepseek_v4` fp8/mxfp8 forward cases (d=512, 1 KV head, SWA 128): 2 models × 4 seqlens × {fp8, mxfp8} = 16.

Reproduced on B200 with develop-TOT frontend + libcudnn 9.25.1 (the public `apt install cudnn` the nightly container gets):

```
mxfp8 d=512 fwd: 1 plan: eng10_k24=1_k27=0_k38=0_k40=3_k41=2
  [0] BUILD FAILED: ... Encountered runtime kernel compilation failure
      cudnn_status: CUDNN_STATUS_INTERNAL_ERROR_COMPILATION_FAILED
```

## Root cause

Two stacked defects:

1. **Frontend (#845).** The Blackwell FP8 forward head-dim gate in `sdpa_support_surface.h` was opened to `(256, 512]` for the FROST per-tensor fp8 d512 kernel, on the assumption that "the cuDNN backend itself has no plan for the band, so a graph that does not select that engine still fails at plan creation". It does not fail cleanly: the backend's runtime-compiled unified SDPA engine has **no head-dim upper bound for SM100/SM107 prefill** (only `d % 16 == 0`; the d≤128/192/256 caps are gated on `s_q == 1 || BPROP`), so it claims the graph at check-support and NVRTC-compiles a `TILE_K = 512` kernel its fp8/mxfp8 emitters weren't designed for. Per-tensor fp8 happens to compile; mxfp8 does not. And since `sdpa_fp8` and `sdpa_mxfp8` both set `mma_core_mode = FP8_E4M3`, the widened gate admitted **MXFP8 d512** as well — which #845 itself lists as out of scope (no d512 block-scale kernel anywhere; the FROST engine row already declines it).

2. **Backend.** The engine should return `NOT_SUPPORTED` for that shape instead of failing at compile time. That is a separate backend fix (shape gate in `attentionDescriptors.cpp`), tracked outside this PR.

## Fix (frontend side)

- `sdpa_support_surface.h`: the `(256, 512]` band now requires **per-tensor** descales. MXFP8 is recognised by its `FP8_E8M0` scale-factor `Descale_Q` and is declined at `validate()` with `GRAPH_NOT_SUPPORTED` — the caller sees a support gate, not a plan-build compile failure. Per-tensor fp8 d512 is unchanged. Error messages now say "for per-tensor FP8 only (not MXFP8)".
- `benchmark_single_sdpa.py`: a frontend `GRAPH_NOT_SUPPORTED` at `validate()` now maps to the runner's unsupported exit code, so the nightly CSV marks such cases *skipped* (support gate) rather than *failed*. Errors past `validate()` (heuristics, check_support, plan build) still fail loudly — a runtime-compile error there is a library bug worth seeing.
- `test_sdpa_fwd_fp8_sm100.py::test_fp8_d512_graph_surface_admits_per_tensor_only`: builds the d512 graph in both forms; per-tensor validates, MXFP8 raises `cudnnGraphNotSupportedError`.
- `python/cudnn/sdpa/AGENTS.md`: **Rule S3** — opening the graph-level surface for a FROST-only shape admits exactly the forms that engine serves, with the runnable plan-list/build-each detector to run before merging such a widening.

## Verification (B200 cc 10.0, libcudnn 9.25.1, FE at develop TOT)

- Probe: mxfp8 d512 fwd → rejected at `validate()`; per-tensor fp8 d512 fwd → still 1 plan (`eng10…`), builds OK.
- `pytest test_sdpa_fwd_fp8_sm100.py -k "not strided_stats"` green (`test_fp8_strided_stats` is the known 9.26-only failure noted in #845).
- Harness: `benchmark_single_sdpa.py --data_type mxfp8 --head_dim 512 …` now exits `UNSUPPORTED_CONFIG` (rc 42) instead of `COMPILATION_FAILED`.

### DIFF
--- a/benchmark/attention_training/benchmark_single_sdpa.py
+++ b/benchmark/attention_training/benchmark_single_sdpa.py
@@ -906,7 +906,18 @@ else:
             if args.verbose:
                 print(f"[INFO] cudnn_oss: pinned FROST engine '{eligible[0]}' for {pass_name}")
 
-        graph_fwd.validate()
+        def validate_or_skip(graph, pass_name):
+            """A frontend-level GRAPH_NOT_SUPPORTED is a support gate (shape / dtype /
+            feature the library declines up front), not a failure: report it as a
+            skip so the runner's CSV marks the case unsupported. Anything past
+            validate() (heuristics, check_support, plan build) still fails loudly,
+            since e.g. a runtime-compile error there is a library bug."""
+            try:
+                graph.validate()
+            except cudnn.cudnnGraphNotSupportedError as e:
+                exit_unsupported(f"{pass_name} graph rejected by cuDNN frontend validation: {e}")
+
+        validate_or_skip(graph_fwd, "fwd")
         graph_fwd.build_operation_graph()
         graph_fwd.create_execution_plans([cudnn.heur_mode.A, cudnn.heur_mode.FALLBACK])
         if args.sdpa_backend == "cudnn_oss" and run_fwd:
@@ -1146,7 +1157,7 @@ else:
                 dK_bwd.set_output(True).set_dim(dKey.size()).set_stride(dKey.stride())
                 dV_bwd.set_output(True).set_dim(dValue.size()).set_stride(dValue.stride())
 
-            graph_bwd.validate()
+            validate_or_skip(graph_bwd, "bwd")
             graph_bwd.build_operation_graph()
             graph_bwd.create_execution_plans([cudnn.heur_mode.A, cudnn.heur_mode.FALLBACK])
             if args.sdpa_backend == "cudnn_oss":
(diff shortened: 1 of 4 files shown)


vedaanta:
Backend counterpart (the half that turns the compile failure into `NOT_SUPPORTED` at check-support): cuDNN `dev` commit `18bbf0bd1` on branch `vagarwalla/sdpa-fp8-fprop-sm100-d-cap` (worktree `cudnn_sdpa-varlen-ima/.claude/worktrees/sdpa-fp8-fprop-d-cap`; patch at `scratch.vagarwalla_gpu/sdpa_fp8_fprop_sm100_d_cap.patch`). It caps fp8/mxfp8 **fprop** at `d_qk, d_v <= 256` on SM100/SM107 in `extractAttentionParameters()` (matching the SM90 fprop cap); decode, backward and f16 prefill are untouched. Not yet pushed to gitlab (token expired on the dev box).

With both halves in: MXFP8 d512 → `GRAPH_NOT_SUPPORTED` at validate (frontend); per-tensor FP8 d512 with FROST off → backend `NOT_SUPPORTED` at check-support instead of a plan that compiles into an unverified 4% SOL kernel.

PR #905, opened 2026-09-04
frost mxfp8 d=256 bwd: split the Q-head group across dK/dV clusters (GQA occupancy); write skipped causal-tail tiles

vedaanta:
## Summary

Follow-up to #904. The fused dK/dV kernel of `sdpa_bwd_sm100_mxfp8` launches one 2-CTA cluster per (KV tile, KV head, batch) and that cluster walks every Q head of its KV head. With few KV heads at small batch — the GQA shapes this engine exists for — most of the GPU idles: 2 KV heads at S=8192 is 64 clusters on 148 SMs (43%), at S=2048 it is 16 (11%). On the 32/2-head causal d=256 shape the MXFP8 backward was therefore only 1.1–1.3x the bf16 backend at 32k and *slower* than bf16 below 8k.

This PR splits the dK/dV work of one KV tile across clusters and folds the partials:

- **Kernel** (`bprop_dkdv_d256_mxfp8_sm100.py`): two new constructor parameters, `h_r_split` (Q-head group slices) and `s_q_split` (Q-tile-range chunks), both default 1 = unchanged behaviour. The grid becomes (KV tiles, KV heads × split, B); cluster `bidy` handles KV head `bidy // split`, a Q-head slice and a Q-tile chunk of that KV tile's trip range, and stores its dK/dV into partial slot `bidy` of a `[B, S_kv, H_kv·split, D]` buffer (the `dK`/`dV` operands are viewed with `h_k·split` heads). The load warp gets a Q-head start offset and a shorter trip range; the compute warps' mask classification stays relative to the full range start (`iter_start_global`), so causal/diagonal handling is unchanged. Persistent mode rejects splits > 1 (not needed here).
- **Skipped causal-tail tiles are now written.** An in-range KV tile that no query row attends (causal with S_kv > S_q) used to skip the mainloop *and* the epilogue, leaving the caller's dK/dV rows with whatever was in the buffer. A new `epilogue_zero` stores zeros for those tiles (`test_causal_kv_tail`; the test harness now NaN-poisons outputs so this class of bug fails loudly).
- **Adapter** (`api_dsl_mxfp8_sm100.py`): `_dkdv_splits()` picks the head split as the smallest divisor of the GQA group that gives about one cluster per SM (or the whole group); `dkdv_head_split=` / `dkdv_seq_split=` override it. With a split > 1 it carves two partial buffers (gradient dtype) from the workspace and folds them with the existing fixed-order fp32 `dkv_reduce` kernel (shared with the half-precision GQA backward), so `use_deterministic_algorithm` still holds. MHA (group = 1) is bit-for-bit the old path: no partials, no fold.
- **Sequence split is opt-in only.** It is wired through the same partial/fold path and tested (`test_dkdv_seq_split_forced[2|3|8]`, including more chunks than Q tiles), but on every shape measured it was neutral at best and up to 30% slower otherwise (each chunk re-pays the per-cluster K/V load and pipeline setup, and the fold reads `seq` partials), so the heuristic never selects it. Numbers below.
- Also fixes a latent strictness bug in the adapter's BSHD check: extent-1 dims (MQA K/V head dim) now wildcard their stride, as the analyzer already does.
- Tests: `test_dkdv_head_split_heuristic` (shape-only, nothing computed: chosen split on a 32/2-head shape, (1,1) for MHA, non-divisor rejected), `test_dkdv_head_split_forced[1|2|8]` and `test_dkdv_seq_split_forced[2|3|8]` (each split matches the reference and the unsplit result to cosine > 0.99999) drive the adapter directly; the existing GQA/MQA graph tests run with a split > 1 by construction of their shapes. Docstrings added to every function this diff touches (CodeRabbit's coverage check).
- Tracker + module docstring note the split and its cost (2·split·|dK| bytes of workspace, one extra small launch).

## Numbers

B200 (SM clock pinned at 847 MHz on this box, so absolute numbers are ~0.42x of a 2 GHz part; compare relatively), b=1, 32 Q / 2 KV heads, d=256, causal, backward only, `cudnn_oss`, `benchmark/attention_training`:

| S | before (#904) | after | speedup | split chosen |
|---|---|---|---|---|
| 2048 | 1.116 ms · 154 TF | 0.735 ms · 234 TF | 1.52x | 8 |
| 4096 | 2.305 ms · 298 TF | 1.819 ms · 378 TF | 1.27x | 4 |
| 8192 | 5.350 ms · 514 TF | 5.275 ms · 521 TF | 1.01x | 2 |
| 16384 | 18.06 ms · 609 TF | 17.97 ms · 612 TF | 1.01x | 1 |
| 32768 | 66.22 ms · 664 TF | 65.87 ms · 668 TF | 1.01x | 1 |

MHA regression guard (b=4, 16/16 heads, S=8192, dense; split stays 1): 18.42 ms before, 18.41 ms after.

Why "one cluster per SM" and not more — device time (dQ + dK/dV + fold, torch profiler) with forced factors, same box:

| shape (causal) | CTA waves unsplit | unsplit | head split 2 / 4 / 8 / 16 |
|---|---|---|---|
| 32/2 @2k | 0.43 | 1.028 ms | 0.723 / 0.667 / **0.655** / 0.700 |
| 32/2 @4k | 0.86 | 2.174 ms | 1.820 / 1.705 / **1.693** / – |
| 32/2 @8k | 1.73 | 5.126 ms | **5.087** / 5.523 / 5.132 / – |

Once the grid covers the machine (8k+) more slices only add fold traffic and per-cluster setup; targeting two waves (the first version of this PR) cost 8% at 8k. The sequence split on MHA shapes, where there is no head group to slice:

| shape (causal) | CTA waves unsplit | unsplit | seq split 2 / 3 / 4 / 8 |
|---|---|---|---|
| 8/8 @2k | 1.73 | **0.252 ms** | 0.288 / 0.333 / 0.347 / – |
| 8/8 @4k | 3.46 | **0.589 ms** | 0.662 / 0.752 / – / – |
| 2/2 @2k | 0.43 | 0.180 ms | **0.169** / – / 0.184 / 0.211 |
| 2/2 @8k | 1.73 | **0.512 ms** | 0.561 / – / 0.631 / 0.750 |

At S=2048 the remaining gap to the bf16 backend is the adapter's host-side launch path (~0.9 ms per execute), tracked separately.

## Summary by CodeRabbit

* **New Features**
  * Improved MXFP8 SDPA backward performance for grouped and multi-query attention through parallel Q-head splitting.
  * Added opt-in sequence-range splitting for dK/dV computation with deterministic partial-result reduction.
  * Added automatic head-split selection and explicit configuration for head and sequence splitting.
  * Added support for extent-one dimensions with flexible strides.

* **Bug Fixes**
  * Ensured empty or fully masked causal-attention chunks produce correctly initialized dK/dV results.

* **Documentation**
  * Updated support documentation with split behavior and workspace requirements.

### DIFF
--- a/python/cudnn/sdpa/bwd/api_dsl_mxfp8_sm100.py
+++ b/python/cudnn/sdpa/bwd/api_dsl_mxfp8_sm100.py
@@ -31,6 +31,14 @@ upstream; it costs ~1-2% of the backward (SF bytes are 1/32 of the payload)
 and is the first thing to remove once the kernels' SF path reads canonical
 atoms.
 
+Occupancy: the fused dK/dV kernel's grid is (KV tiles, KV heads, B) and one
+2-CTA cluster walks every Q head of its KV head, so with few KV heads (GQA /
+MQA at small batch) most SMs idle -- 2 KV heads at S=8192 is 64 clusters on
+148 SMs. The adapter therefore splits the Q-head group across clusters
+(``_dkdv_head_split``: the smallest divisor of the group that reaches about two
+waves of clusters); each slice writes a partial dK/dV and the shared
+fixed-order ``dkv_reduce`` fold sums them, so the result stays deterministic.
+
 Lives in its own module (not ``api_dsl.py``) so the MXFP8 lowering's imports
 stay out of the half-precision adapters' way; ``engines._adapter`` resolves it
 by name.
@@ -88,6 +96,7 @@ class SdpaBwdDslSm100Mxfp8(SdpaBwdDsl):
         sample_sf_do,
         sample_sf_do_T,
         p_scale_log2: int = 8,
+        dkdv_head_split: Optional[int] = None,
         **kwargs,
     ) -> None:
         # Stashed raw; descs are built in _initialize_implementation, which the
@@ -106,13 +115,19 @@ class SdpaBwdDslSm100Mxfp8(SdpaBwdDsl):
             sf_do_T=sample_sf_do_T,
         )
         self.p_scale_log2 = int(p_scale_log2)
+        # None = pick from the shape (see _dkdv_head_split); an int forces it.
+        self._dkdv_head_split_override = None if dkdv_head_split is None else int(dkdv_head_split)
         super().__init__(sample_q, sample_k, sample_v, sample_o, sample_do, sample_stats, sample_dq, sample_dk, sample_dv, **kwargs)
 
     # --- geometry --------------------------------------------------------------
     @staticmethod
     def _bshd_physical_ok(desc: TensorDesc) -> bool:
+        """Compact BSHD storage under a logical BHSD desc. Extent-1 dims are
+        wildcards: their stride is unobservable, and torch reports an arbitrary
+        one for e.g. the head dim of an MQA K/V tensor."""
         b, h, s, d = (int(x) for x in desc.shape)
-        return tuple(int(x) for x in desc.stride) == (s * h * d, d, h * d, 1)
+        want = (s * h * d, d, h * d, 1)
+        return all(sz == 1 or int(st) == w for sz, st, w in zip((b, h, s, d), desc.stride, want))
 
     def _sf_plan(self):
         """The eleven kernel-side SF buffers: (name, graph SF, rows, k_groups, planes, layout, plane_major).
@@ -178,7 +193,41 @@ class SdpaBwdDslSm100Mxfp8(SdpaBwdDsl):
         if self.scale_softmax is None or self.scale_softmax == 0.0:
             self.scale_softmax = 1.0 / math.sqrt(self.head_dim_qk)
         self._gqa_group = self.h_q // max(self.h_kv, 1)
+        self._dkdv_split = self._dkdv_head_split()
         self._compiled = None
+
+    # dK/dV cluster count below which the Q-head group is split across clusters:
+    # two waves of 2-CTA clusters over the SMs.
+    _DKDV_TARGET_CLUSTERS_PER_SM = 2
+
+    def _dkdv_head_split(self) -> int:
+        """How many clusters share one (KV tile, KV head): the smallest divisor of
+        the GQA group that gives the fused dK/dV kernel about two waves of
+        clusters, or the whole group if none does.
+
+        The kernel's grid is (KV tiles, KV heads, B) and one cluster walks every
+        Q head of its KV head, so with few KV heads most SMs idle: 2 KV heads at
+        S=8192 is 64 clusters on 148 SMs. Each split writes a partial dK/dV
+        (one slot per Q-head slice) that a fixed-order fold sums, so the
+        result stays deterministic; the price is the partial buffers
+        (2 * split * |dK| bytes) and the fold's read of them.
+        """
+        if self._dkdv_head_split_override is not None:
+            split = self._dkdv_head_split_override
+            if split < 1 or self._gqa_group % split != 0:
+                raise ValueError(f"SM100 MXFP8 bwd: dkdv_head_split={split} must divide the GQA group {self._gqa_group}")
+            return split
+        h_r = self._gqa_group
+        kv_clusters = _cdiv(self.s_k_max, 128) * self.h_kv * self.batch_size
+        target = self._DKDV_TARGET_CLUSTERS_PER_SM * torch.cuda.get_device_properties(torch.cuda.current_device()).multi_processor_count
+        split = 1
+        for cand in range(1, h_r + 1):
+            if h_r % cand:
+                continue
+            split = cand
+            if kv_clusters * cand >= target:
+                break
+        return split
 
     # --- capability backstop ---------------------------------------------------
     def check_support(self) -> bool:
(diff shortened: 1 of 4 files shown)


YangXu1990uiuc:

I am requesting FROST test CI for the current head (`d19785846aba`) with the next standalone command while validating the new backward split paths locally. Style has already passed.

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc (changes requested):

Reviewed `d19785846aba` against base `0ed2710bc6db`.

The occupancy improvement is measurable, and the split indexing, empty slices and deterministic fold passed the focused checks. However, the new FP16 partial buffers can overflow before cancellation across head slices: the default heuristic turns a valid all-zero gradient into all-NaN dV. This is a high-risk P1 that needs addressing before merge. There is also a device-selection issue in the new heuristic.

**[P1 · blocks approval] Keep split gradients in FP32 until the final fold** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/d19785846aba64d5f1eed7c7f4070f1ac36433d1/python/cudnn/sdpa/bwd/api_dsl_mxfp8_sm100.py#L649))

Carving partials in self.out_dtype rounds each cluster accumulator to FP16 before the FP32 reduction. A finite final gradient can require larger, cancelling partials. Repro on SM100 with 148 SMs: B=3, Hq=32, Hkv=4, Sq=Sk=1024, D=256, dense attention, FP16 gradients, Q=K=V=0. Within each eight-head GQA group, use dO=+32768 for four heads and -32768 for the other four. All inputs are finite and every gradient should be zero. The default heuristic selects split=2; its dV partials require +/-131072, beyond FP16 range, and all 3,145,728 final dV elements become NaN. The pre-PR base and this head forced to split=1 return exact zeros. Keep partial storage and the mainloop epilogue in FP32 until the final output cast, or disable/decline FP16 splitting until that is supported. This blocks approval because an automatically selected, supported path silently corrupts gradients.

Evidence: Independent exact-head GPU reproduction: default split=2 gives 0 finite / 3,145,728 NaN dV elements; both the current-head unsplit control and base 2a0343f0 give exact-zero dQ/dK/dV. A smaller B=1,Hq=8,Hkv=1,S=128 forced-split reproduction gives 32,768 NaN dV elements versus exact zeros unsplit. The oracle is analytic uniform attention with cancelling dO; the repository reference also returns exact zeros. Both failing probes exited nonzero and were excluded from timing. Attribution: Introduced by the new split partial-storage path. The independently tested pre-PR base handles the same default-policy input correctly.

**[P2] Choose split factors using the operand device** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/d19785846aba64d5f1eed7c7f4070f1ac36433d1/python/cudnn/sdpa/bwd/api_dsl_mxfp8_sm100.py#L237))

The target SM count comes from torch.cuda.current_device(), although the plan already has the Q tensor device in q_desc.device. Constructing a plan for one GPU while another GPU is current therefore changes the selected split and workspace according to the unrelated GPU. For the documented B=1,Hq=32,Hkv=2,S=2048 geometry, a 148-SM target selects 8 while a 68-SM current device selects 4. Query the input/plan device explicitly so multi-GPU callers get the intended occupancy and workspace tradeoff.

Evidence: Source-level contract check: _dkdv_splits uses ambient current_device for the only hardware input to the new heuristic, rather than the captured operand device. The split choices above follow directly from the divisor loop; no multi-GPU throughput measurement is claimed. Attribution: Introduced by the new plan-time heuristic.

Validation: SM100, CuTe DSL 4.8, cuDNN 9.25.1: 15 selected tests passed, covering forced head/sequence splits, dense BF16/FP16, GQA/MQA and causal-tail writes. A separate combined head=2, sequence=3 case (B=2,Hq/Hkv=8/2,Sq/Sk=257/640) passed reference checks, matched unsplit with cosine above 0.99999 and repeated bit-exactly. The default-heuristic FP16 failure and its current-head unsplit/pre-PR controls were run independently on a 148-SM SM100 with isolated source and compilation caches. The regular random-input suite did not cover this partial-range failure. Performance spot check on a separate 68-SM SM100, using passing BF16 cases only, fresh-process ABBA against 2a0343f0: B=1,Hq/Hkv=32/2,S=1024 selected split=8. Whole-backward GPU time fell from 443.83–443.87 us to 340.31–340.35 us (about 23%), while CPU submission rose from 676.3–678.8 us to 753.3–756.5 us (about 11%). Workspace rose from 36.47 to 52.47 MiB. MHA Hq/Hkv=4/4,S=256 stayed near 48.1 us GPU and 0.65 ms CPU. The GPU gain therefore has a host-overhead tradeoff; these are not end-to-end training measurements. Every timed arm passed the repository gradient-reference gate first. Negating dO and both quantized dO orientations, poisoning all output gradients, then replaying the captured chain produced exactly negated gradients. No timing was taken for the failing FP16 inputs. Read the complete four-file diff and traced head/slice indexing, mask classification, partial layout and workspace accounting. The prospective merge into develop 0ed2710b was clean. The causal-tail zero epilogue is already present in the reviewed base after #960; this diff adapts its output geometry for split slots. Style passed. I started the matching-head FROST pipeline and verified its full GitHub SHA. Other GPU lanes remain pending/running; the Rubin job failed during runner preparation with no space left on device, before tests. The text-scan failure is in an unchanged convolution docstring.

Limitations: Local GPU evidence is SM100 only. The full pipeline is still pending, and the Rubin runner needs its disk-space problem resolved before a useful retry. Performance was sampled on two small backward workloads, not the full B200 sweep in the PR description.

YangXu1990uiuc:

CI follow-up for `d19785846aba`: SM80/SM100/SM103/SM120 SDPA and SM100 GEMM/linear passed. The SM100 SDPA lane reports 4,778 passed and 1,806 skipped, but its routing summary contains only one FROST backward graph; most backward coverage used the native backend. This does not cover the FP16 split/cancellation failure reproduced in the review, which still needs fixing.

Rubin did not reach tests because the runner ran out of disk space. The other failure is the text scan of an unchanged convolution comment. There is no new Rubin correctness result from this run.

PR #906, opened 2026-09-04
 Reduce instruction number for tanh soft clamp srelu/dsrelu grouped GEMM epilogues

wanyingw:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL; Benchmarks or performance.

## Summary

Follow-up to the `tanh_clamp_scale` epilogues (#79418e29). The clamped backward cost +12-14% over
the unclamped kernel at k=3072 while the clamped forward cost ~+3.5%. Nsight Compute shows the
dSReLU epilogue is the backward's critical path (one latency-bound epilogue warp per scheduler,
MUFU at 4%, registers/occupancy unchanged), so each of the 8 instructions the clamp added per
element was wall time.

Both clamped branches are rewritten in `t = tanh(relu(x)/s)`:

* `s` folded into per-tile scalars (`s^2*w`, `2*s*w`, `s^2`); no per-element multiply by `s`
* relu and the `t <= 1` cap fused into one saturating multiply (`tanh_clamp_unit`, new helper)
* backward: `t^2` computed once and shared by dgrad, dprob and the d_srelu regen; `t - t^3` is one
  FFMA; the whole branch is a single per-element pass
* forward and backward use the same `tanh_clamp_unit`, so the regen stays bit-identical to the
  forward when C is stored in fp32

Unclamped path unchanged; `tanh_clamp_scale=None` stays bit-identical (existing tests).

## Why

Extra instructions per element: backward 8.0 -> 2.2, forward 4.5 -> 2.5. Kernel time over the
unclamped kernel, GB300 (sm_103), n=8192/16384, k=3072, MXFP8, same node back to back:

| | before | after |
|---|---|---|
| backward, dynamic scheduler (TE default) | +13.4% to +13.7% | +2.1% to +2.4% |
| backward, static scheduler | +12.4% to +12.6% | +3.6% to +3.8% |
| forward | +2.0% to +3.9% | +1.1% to +3.7% |

Numerics on identical inputs, old vs new kernels: unclamped outputs bitwise identical (forward and
backward, deterministic dprob). Clamped, s=16: forward D, d_srelu and dprob bitwise identical;
d_row/d_col differ in ~3e-7 of fp8 codes (rounding sequence of the dgrad expression). Clamped
backward compiles to 255 registers/thread (unclamped 230), no spills.

## Related issues

Follow-up to the tanh_clamp_scale change (79418e29).

None. No API change; `tanh_clamp_scale=None` output is bit-identical to before; clamped outputs
differ from the previous clamped kernel only by fp32 rounding order.

## Testing

GB300 (sm_103), cuDNN FE test suite unmodified, from `test/python`:

```
pytest fe_api/grouped_gemm/test_grouped_gemm_srelu.py    # 81 passed, 80 skipped
pytest fe_api/grouped_gemm/test_grouped_gemm_dsrelu.py   # 121 passed, 80 skipped
```

Includes the clamped cases (two scales x fp8/fp4 x vector_f32 x reuse x deterministic), the
`None`-vs-omitted bitwise tests, the saturated-gradient tests and the regen-matches-forward test.
Performance: interleaved clamped-vs-unclamped bench (101 samples x 20 launches) and Nsight Compute
instruction/cycle counts, numbers above.

### DIFF
--- a/python/cudnn/gemm/cutedsl/grouped/moe_kernel_helpers.py
+++ b/python/cudnn/gemm/cutedsl/grouped/moe_kernel_helpers.py
@@ -138,6 +138,32 @@ def fmax(a: Union[float, Float32], b: Union[float, Float32], *, nan=True, loc=No
     )
 
 
+def tanh_clamp_unit(u: Float32, *, loc=None, ip=None) -> Float32:
+    """``t = sat(tanh(u))`` for ``u = x/s``: the soft-clamp unit of the srelu/dsrelu epilogues.
+
+    One MUFU.TANH plus one FMUL.SAT (``mul.rn.sat.f32`` by 1.0 clamps to [0, 1]; NaN -> +0).
+    The saturate does two jobs. It is the relu: tanh is monotone with tanh(0) = 0, so u < 0
+    lands on t = 0, and NaN lands on 0 exactly like ``where(NaN > 0, NaN, 0)`` did. And it is
+    the defensive t <= 1 cap the backward needs for t - t^3 >= 0, since tanh.approx.f32's
+    range is not assumed.
+
+    Both kernels must go through this one function: with C stored in fp32 the backward's
+    d_srelu regen has to reproduce the forward bit for bit.
+    """
+    t = cute.math.tanh(u, fastmath=True)
+    return Float32(
+        llvm.inline_asm(
+            T.f32(),
+            [Float32(t).ir_value(loc=loc, ip=ip)],
+            "mul.rn.sat.f32 $0, $1, 0f3F800000;",
+            "=f,f",
+            has_side_effects=True,
+            is_align_stack=False,
+            asm_dialect=llvm.AsmDialect.AD_ATT,
+        )
+    )
+
+
 def fmin_bf16x2(
     a0: BFloat16,
     a1: BFloat16,
--- a/python/cudnn/gemm/cutedsl/grouped/dsrelu/moe_blockscaled_grouped_gemm_dsrelu_quant.py
+++ b/python/cudnn/gemm/cutedsl/grouped/dsrelu/moe_blockscaled_grouped_gemm_dsrelu_quant.py
@@ -77,8 +77,8 @@ from ..moe_sched_extension import (
     ContiguousAndConsistentGroupedGemmSchedExtension,
 )
 from ..moe_kernel_helpers import (
-    fmin,
     fmax,
+    tanh_clamp_unit,
     atomic_add_float32,
     atomic_max_float32,
     compute_stages,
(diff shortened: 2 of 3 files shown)


vedaanta:
@cudnn-ci-bot run oss

vedaanta (approved):

python/cudnn/gemm/cutedsl/grouped/moe_kernel_helpers.py
```diff
@@ -138,6 +138,32 @@ def fmax(a: Union[float, Float32], b: Union[float, Float32], *, nan=True, loc=No
     )
 
 
+def tanh_clamp_unit(u: Float32, *, loc=None, ip=None) -> Float32:
+    """``t = sat(tanh(u))`` for ``u = x/s``: the soft-clamp unit of the srelu/dsrelu epilogues.
+
+    One MUFU.TANH plus one FMUL.SAT (``mul.rn.sat.f32`` by 1.0 clamps to [0, 1]; NaN -> +0).
+    The saturate does two jobs. It is the relu: tanh is monotone with tanh(0) = 0, so u < 0
+    lands on t = 0, and NaN lands on 0 exactly like ``where(NaN > 0, NaN, 0)`` did. And it is
```
shdetect, line 146: A good idea to do the clamp.
One concern for "NaN lands on 0 exactly like ``where(NaN > 0, NaN, 0)`` did", may I know whether this NaN is for tanh result  `t` ? if so, probably, we need to propagate this NaN result to down stream

wanyingw: The NaN is not from the tanh result, it is a NaN input to the clamp, not what the clamp produces. 
The clamped `where(NaN > 0, NaN, 0)` stays consistent to the behavior of the unclamped path, see [unclamped srelu](https://github.com/NVIDIA/cudnn-frontend/blob/7d67dc7b10fa0f6fdbf25a09b8820f1847ddf254/python/cudnn/gemm/cutedsl/grouped/srelu/moe_blockscaled_grouped_gemm_srelu_quant.py#L1970)

So in our current production code, the NaN does not propagate in both the original unclamped path, and the newly added tanh clamp path. This is also consistent with the TE behavior.

However I think we should get back to this in a followup PR: currently the fused path does not propagate NaN, and the unfused Megatron fallback propagates NaN. We should make sure there is consistency, and if we want to change the fused path into propagating NaN, we should test that for both the clamped and unclamped paths.

shdetect: Yep, that make sense.

Issue #745, opened 2026-08-25
fp8/mxfp8 SDPA: output mismatches on rows with deeply negative attention scores (fp8 fwd O mismatch; mxfp8 kernel/emulator exact-match disagreements)


vedaanta:
## Summary

The negative-score test-data injection in #743 (rows whose post-scale attention scores land at -200..-490 — the regime real models like Qwen2.5 actually reach, and the one behind pytorch/pytorch#193893) surfaces fp8-family mismatches on B200 that classic sparse-int data never exercised:

1. **fp8 fwd** — `test_sdpa_fp8_fwd_L0[test110]` fails `torch.testing.assert_close` on O (an MQA config: h_q=9, h_k=1, s=3770, d_qk=192/d_v=128). 1 of 256 sampled configs.
2. **fp8 bwd** — 3 of 128 sampled configs fail gradient tolerances.
3. **mxfp8** — 11 fwd + 2 bwd tests fail their **exact-match** comparison against the mxfp8 emulator (e.g. `test_sdpa_mxfp8_fwd_L0[test213]`: "output mismatch: 43 elements differ"), i.e. kernel and reference emulation disagree bit-wise only in the negative-score regime. Needs triage into kernel-vs-emulator: either the kernel's exp/rescale path deviates for large-magnitude scores, or the emulator's modeling does.

## Notes

- All observed with the SM100 decode fwd fix already applied, so these are independent of that bug.
- fp16/bf16 forward suites are fully green on the same data, and the fp16/bf16 **backward** counterpart finding is tracked separately (fp16-rounded score recompute signature).
- Injected values are exactly representable in fp8 e4m3/e5m2 (|k| <= 4, |q| a power of two), so quantization of the inputs themselves is exact; the divergence is downstream in the softmax/rescale path.

## Repro

Check out #743, point at a cuDNN with the decode fwd fix, then:

```
pytest test/python/test_mhas_v2.py::test_sdpa_fp8_fwd_L0   -k test110
pytest test/python/test_mhas_v2.py::test_sdpa_mxfp8_fwd_L0 -k "test213 or test214"
```

Each failing test prints a self-contained `test_repro` dict.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Adnios:
Fixed. Closed this issues.

PR #910, opened 2026-09-04
test_mhas_v2: flash-style SDPA references (~16x less memory), seq up to 8192, more L0 configs

vedaanta:
## Why

The fp16/fp8/mxfp8 SDPA references in `test/python/sdpa/*_ref.py` materialized `(b, h_q, s_q, s_kv)` fp32 tensors: the forward built a full `combined_bias` for the masks and the backward held S, P, dP and dS at once. At the largest test_mhas_v2 configs (b=8, h=8, 4096x4096) that is several 4 GiB tensors per test, which OOMs sibling pytest-xdist workers sharing one GPU and caps `-n`.

## What

**Commit 1 — blocked (flash-attention style) references.** Both passes loop over 128-wide KV blocks. Masks (padding, causal / sliding window with TOP_LEFT or per-batch BOTTOM_RIGHT alignment, ALiBi, block mask, bias) are produced per block from index arithmetic (`_ScoreMask` in `fp16_ref.py`), GQA/MQA use grouped-head einsums so K/V are never expanded, and the backward is two-pass (row LSE, then per-block P/dQ/dK/dV/dBias/dSink). The fp8 and mxfp8 refs share the helpers; the mxfp8 dS quantization stays exact per block because its 32-element scale groups tile the 128-column block. Public signatures/returns are unchanged (the frost tests that import these refs are unaffected).

Peak memory, b=8 h=8 GQA causal 4096x4096 (B200, torch allocator), reference only:

| pass | before | after |
|---|---|---|
| fwd | 8.39 GiB | 0.88 GiB |
| bwd | 26.64 GiB | 1.63 GiB |

No runtime change.

Equivalence vs the previous references (old-vs-new A/B over random configs covering GQA/MQA, padding, both alignments, sliding window, ALiBi, bias, block mask, dropout, sink): fp16 with `torch_type=float32` <= 1e-5 rel over 150 configs; native fp16/bf16 within 1-ulp rounding over 200; fp8 outputs within 1 e4m3 ulp over 120 (stats/amax to 1e-7); mxfp8 O bit-identical, grads within bf16 rounding over 60.

**Commit 2 — widen the suites and two small fixes.**
- L0 suites (fp16 fwd/bwd, ragged, decode, paged, block-mask, fp8, mxfp8) now draw sequence lengths up to 8192 (the two bias suites stay at 4096: the bias tensor itself is `(b, h, s_q, s_kv)` on the device) and run 1.5-2x more configs (fwd 256->512, bwd 384->512, ragged/decode 128->256, fp8/mxfp8 256->384).
- `random_config.py`: `band_around_diag` did `rng.randint(1, s_kv // 2)`, an empty range when `s_kv == 1` (ValueError before any kernel ran); clamp like `left_window_only` already does.
- `torch.exp2` on fp32 CUDA tensors raises `CUDA driver error: invalid argument` on some torch nightlies (2.14.0.dev20260710+cu130) for ANY size, not just large tensors as the #904 comment said (`pow`/`exp`/`log2`/fp16 exp2 work). Keep `torch.pow(2, x)` in `mxfp8_ref.py` with the corrected explanation (sdpa-only; other `torch.exp2` uses in the repo are left alone here).

## Validation

B200, cuDNN 9.26.0.33, this branch's FE, `pytest -n 8` on the widened suites: no OOM, no mem-gate waits. Largest whole-test peak 4.2 GiB (b=8 h=7 8192x8012 d=264 bwd). fp16 fwd 91/92, fp16 bwd 34/34, ragged fwd 32/32, ragged bwd 26/26, fp8 29/30, mxfp8 27/27 (rest skipped as unsupported). Every remaining failure reproduces identically with the OLD references, i.e. is a backend issue:
- decode (`sq1`/`lean_attn`) ~50%: the stock 9.26.0.33 decode row-max bug (cuDNN MR !4303, fixed for the next 9.26.x).
- fp8 bwd test7: the known fp8 backward gap (#745).
- fwd test29 (bf16, d=256, s=5705, b=h=1): NEW cuDNN 9.26 SM100 fprop bug — the 2-CTA d256 kernel is routed onto the prefill split-K path it does not implement (s_kv >= 4096, under-filled grid); fixed on cuDNN dev by !4279, backport MR for release/9.26.0 opened alongside this PR. The FROST `sdpa_fwd_prefill_sm100` engine serves the same graph correctly.

CI note: the xdist worker count for test_mhas_v2 lives in the gitlab-only `ci/` scripts (`-n 4`); a companion gitlab MR raises it to `-n 8`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an edge case in sliding-window attention configurations when the key/value sequence contains a single token.
  * Improved numerical compatibility for sigmoid calculations across grouped and quantized operations.

* **Attention Improvements**
  * Updated FP16, FP8, and MXFP8 reference attention paths to support blockwise processing, masking, grouped heads, padding, and sink tokens more consistently.

* **Tests**
  * Expanded attention test coverage, including longer sequences up to 8,192 tokens and additional quantized scenarios.

### DIFF
--- a/python/cudnn/gemm/cutedsl/discrete_grouped/discrete_kernel_utils.py
+++ b/python/cudnn/gemm/cutedsl/discrete_grouped/discrete_kernel_utils.py
@@ -356,9 +356,7 @@ def sigmoid(x):
     import torch
 
     LOG2_E = 1.4426950408889634
-    # torch.pow(2, y) rather than torch.exp2: exp2 on fp32 CUDA tensors raises
-    # "CUDA driver error: invalid argument" on some torch nightlies (2.14.0.dev2026xx).
-    exp_x = torch.pow(2.0, x * (-LOG2_E))
+    exp_x = torch.exp2(x * (-LOG2_E))
     ret = 1.0 / (exp_x + 1.0)
     return ret
 
--- a/python/cudnn/gemm/cutedsl/discrete_grouped/discrete_kernel_utils.py
+++ b/python/cudnn/gemm/cutedsl/discrete_grouped/discrete_kernel_utils.py
@@ -356,7 +356,9 @@ def sigmoid(x):
     import torch
 
     LOG2_E = 1.4426950408889634
-    exp_x = torch.exp2(x * (-LOG2_E))
+    # torch.pow(2, y) rather than torch.exp2: exp2 on fp32 CUDA tensors raises
+    # "CUDA driver error: invalid argument" on some torch nightlies (2.14.0.dev2026xx).
+    exp_x = torch.pow(2.0, x * (-LOG2_E))
     ret = 1.0 / (exp_x + 1.0)
     return ret
 
--- a/python/cudnn/gemm/cutedsl/grouped/glu_hadamard_quant/moe_kernel_helpers.py
+++ b/python/cudnn/gemm/cutedsl/grouped/glu_hadamard_quant/moe_kernel_helpers.py
@@ -464,9 +464,7 @@ def sigmoid(x):
     import torch
 
     LOG2_E = 1.4426950408889634
-    # torch.pow(2, y) rather than torch.exp2: exp2 on fp32 CUDA tensors raises
-    # "CUDA driver error: invalid argument" on some torch nightlies (2.14.0.dev2026xx).
-    exp_x = torch.pow(2.0, x * (-LOG2_E))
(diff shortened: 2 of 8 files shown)


vedaanta:
@cudnn-ci-bot run frost

jhjpark (approved):

Anerudhan (approved):

PR #914, opened 2026-09-04
Optimize Rubin split-KV combine

scottyokim:
## Summary

- distribute split LSE loads and softmax-weight calculation across warp lanes, then broadcast the results with warp shuffles
- reduce the recombined-output amax once per warp before publishing it instead of issuing one atomic per lane
- use a 32-thread combine block for SM107 D128 at split factors 2 through 6; retain the 128-thread block for larger factors and all other architectures/head dimensions
- retain the original scalar combine for partial-warp output widths and split counts above 32

The combine owns one output row per block. On Rubin D128, the original 128-thread block made four warps independently repeat the same row-uniform LSE and weight work. The context workloads have thousands of output rows, so one warp per row preserves enough grid parallelism while removing that duplication. Split-8 measured slower with one warp, hence the split-6 cutoff.

## GR100 performance

NVIDIA Graphics Device, SM107, 208 SMs. Times are medians from 7 repeats of 50 iterations after 5 warmups. Baseline and prototype were run on the same idle GPU with identical settings.

```text
Shape                 Split  Combine baseline  Prototype  Gain     E2E baseline  Prototype  Gain
--------------------  -----  ----------------  ---------  -------  ------------  ---------  ------
Llama TP8 context       4       0.037313 ms     0.027920   25.17%    0.400018 ms   0.394310    1.43%
Llama TP4 context       2       0.065644 ms     0.028798   56.13%    0.775188 ms   0.744993    3.90%
Llama TP2 context       4       0.130337 ms     0.087505   32.86%    1.183657 ms   1.160637    1.94%
AR-DiT context          4       0.041142 ms     0.028924   29.70%    0.218333 ms   0.207395    5.01%
Llama TP8 8K x 8K       1       0.000000 ms     0.000000    0.00%    0.132817 ms   0.132844   -0.02%
```

The saturated 8K x 8K case remains unsplit, so this change does not alter its kernel path; the measured -0.02% is noise.

## Validation

- `test/python/sdpa/frost/test_sdpa_fp8_sm107.py`: 15 passed on SM107
- direct D128 split-8 combine with dead splits: passed
- direct D16 split-3 scalar fallback: passed
- adapter-level SM107 split-4 (32 threads) and split-8 (128 threads): passed
- no NaNs; output and LSE stayed within existing tolerances
- recombined amax stayed within 0.3% of the output reference
- Black, `py_compile`, and `git diff --check`: passed

## Summary by CodeRabbit

* **Performance**
  * Improved split-KV attention recombination performance for select Rubin hardware configurations.
  * Added optimized warp-based processing for eligible workloads, including more efficient output scaling and maximum-value reduction.
* **Compatibility**
  * Existing configurations continue using the standard 128-thread execution path.
  * Supported thread configurations are limited to validated options.

### DIFF
--- a/python/cudnn/sdpa/fwd/api_dsl.py
+++ b/python/cudnn/sdpa/fwd/api_dsl.py
@@ -1398,6 +1398,11 @@ class SdpaFwdDslSm100(SdpaFwdDsl):
                 has_lse=self.lse_desc is not None,
                 has_amax=self._fp8,
                 lse_stride=self._lse_stride,
+                # Rubin's D128 combine has thousands of row blocks at the
+                # admitted prefill shapes.  Up through split-6, one warp keeps
+                # that grid full while avoiding four copies of the row-uniform
+                # LSE/weight work.  Larger factors favor the wider block.
+                threads=32 if self._device_cc == (10, 7) and self.head_dim_v == 128 and self.split_kv <= 6 else 128,
             )
         self._logger.debug("compile completed")
 
--- a/python/cudnn/sdpa/fwd/kernels/split_combine_sm100.py
+++ b/python/cudnn/sdpa/fwd/kernels/split_combine_sm100.py
@@ -10,8 +10,7 @@ A split whose range came out empty ends with total_sum == 0, which the epilogue
 turns into ``O := 0 / lse := -inf`` — the identity here, so empty splits need no
 special case.
 
-One block per (q_row, head, batch); the block's threads stride over d_v, and
-each thread walks the split axis in registers.
+One block per (q_row, head, batch); the block's threads stride over d_v.
 """
 
 from typing import Callable, Optional, Tuple
@@ -25,9 +24,8 @@ from cutlass.base_dsl.typing import Pointer
 from cutlass.experimental import primitives as nvvm
 import cuda.bindings.driver as _cuda_driver  # noqa: F401  (cute.compile pulls cuda)
 
-# One block per (q_row, head, batch); 128 lanes stride over d_v.  d_v <= 128 for
-# every flavor that uses this pass today, so the stride loop runs once.
-THREADS = 128
+WARP_SIZE = 32
+DEFAULT_THREADS = 128
 
 NEG_INF = float("-inf")
 
@@ -47,6 +45,7 @@ def _combine_kernel(
     n_batch: cutlass.Int32,
     n_splits: cutlass.Int32,
     d_v: cutlass.Int32,
+    threads: cutlass.Constexpr[int],
 ) -> None:
     tidx, _, _ = cute.arch.thread_idx()
     q_row = cute.arch.block_idx()[0]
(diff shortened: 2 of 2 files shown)


YangXu1990uiuc (approved):

Reviewed `6113b4495478` against base `2a0343f079b0`.

The lane-distributed weights and warp-level amax reduction make sense, and the focused correctness checks passed on both SM100 and SM107. The unconditional 32-thread selection covers more workloads than the measured winning cases.

**[P2] Include row count and amax usage in the 32-thread selection** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/6113b449547817d8a98c3e12f7b9e77a8d2c8a47/python/cudnn/sdpa/fwd/api_dsl.py#L1405))

This condition also selects one warp for small grids and ordinary f16/bf16 inference without amax, although those workloads have a different trade-off. On SM107, split-4 `(B,H,SQ,D)=(1,4,4,128)` with LSE/amax took 5.15 us on the baseline versus 9.51 us on this selection; the same new kernel forced to 128 threads took 4.29 us. For `(1,4,2048,128)` without LSE/amax, the three times were 23.95 / 26.29 / 18.70 us. Please retain 128 threads outside a benchmarked row-count/output-feature domain rather than applying the cutoff solely by architecture, D and split count.

Evidence: Repeated interleaved three-arm measurements on SM107 (208 SMs), with 100 combine calls per CUDA Graph; both reported regressions also appeared in a preceding independent run. Every measured arm passed the reference and changed-input/poisoned-output graph-replay checks. These are combine-only timings, not full-attention regressions. Attribution: PR-introduced thread-selection regression; the 128-thread control retains the new kernel optimization.

Validation: 20 direct baseline/head correctness cases passed on each of SM100 and SM107: D16/40/96/128/256/512, split factors through 32 and the >32 scalar fallback, dead splits with NaN partials, fully dead rows, strided LSE, optional outputs, fp16/bf16 and changed-input CUDA Graph replay. For the large D128 grid with amax/LSE, SM107 combine time improved at splits 2/4/6/8 (approximately 34.14->17.19, 38.88->28.14, 40.19->37.01 and 44.10->39.48 us). Six SM100 component spot checks showed no regression. Steady-state compiled-call CPU enqueue medians stayed around 3.0-3.3 us on the SM107 host and 5.0-5.4 us on the SM100 host.

Limitations: Current develop has a conflict in the relocated `kernels/sm100/split_combine.py`. On rebase, preserve its compiled-cache path, FP32 partials, FP8 output scaling and pre-scale amax contract, and append `threads` after the newer compile arguments. The measurements above cover the PR head against its branch baseline, not a resolved merge or full attention. I approved the pending fork [Style workflow](https://github.com/NVIDIA/cudnn-frontend/actions/runs/33899158969); it passed. GPU CI remains deferred until rebase. Afterwards, a maintainer can request `@cudnn-ci-bot run frost`; fork Actions may separately need maintainer approval.

PR #854, opened 2026-09-02
frost: prefix GEMM and SDPA kernel symbols with cudnn

adshen:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and the changes comply.
- [x] I added GitHub labels: one `cat-*`, one `area:*`, and one `orig-*`.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

- Apply the public CuTe DSL `set_name_prefix("cudnn", remove_cutlass_symbol=True)` API to all 53 GEMM and 32 SDPA Frost kernels, including auxiliary kernels and generated GEMM templates.
- Preserve descriptive generated GEMM names without duplicating the cuDNN ownership prefix.
- Add an L0 AST guard for future GEMM/SDPA kernels and document the requirement in `python/cudnn/AGENTS.md`.

## Why

Frost kernels currently appear with CUTLASS-owned prefixes in profiler and debug traces. A consistent cuDNN prefix makes their ownership immediately identifiable while retaining the existing mangled function and trace-specific suffixes.

## Related issues

Fixes #820

## API and compatibility impact

No public API or kernel-computation changes. Generated GPU symbol names intentionally change from `cutlass_kernel...` to `cudnn_kernel...`. Kernel behavior and performance are unchanged.

## Testing

- `pre-commit run --files <all changed files>`: passed.
- `pytest -q test/python/test_frost_kernel_name_prefix.py`: 2 passed with cuDNN 9.26.
- `pytest -q test/python/test_frost_kernel_name_prefix.py test/python/gemm/frost/test_matmul.py::test_sm120_render_smoke`: 3 passed with cuDNN 9.26.
- SDPA graph-API smoke under nsys: 1 passed; generated symbol starts with `cudnn_kernel__kernel_...`.
- GEMM block-scale epilogue smoke under nsys: 1 passed; generated symbol starts with `cudnn_kernel_frost_sm100_matmul_...` with no duplicate cuDNN prefix.

## Summary by CodeRabbit

* **Enhancements**
  * Standardized generated CUDA kernel symbols with a `cudnn` prefix across GEMM and scaled dot-product attention operations.
  * Removed CUTLASS-specific components from generated kernel names for clearer, more consistent identification.

* **Tests**
[...]

PR #917, opened 2026-09-04
Gate the CuTe DSL version at runtime (AGENTS.md Rule 7)

YangXu1990uiuc:
## Before submitting

- [x] I set the **Milestone** and **Projects** fields in the sidebar (Projects needs a maintainer).

## Affected area

FE OSS kernels / CuTeDSL — dispatch seams and agent guide only; no kernel changes.

## Summary

Rebased onto #921. `pyproject.toml` now floors `nvidia-cutlass-dsl` at **`>=4.6.2`** — the **downstream** floor (vLLM and SGLang inherit quack-kernels' `==4.6.2`), which is **below** what the FROST-derived kernels need (`CUTEDSL_MIN_VERSION = 4.7.0`). So an installed DSL that satisfies pip can still be below what a kernel needs, and the package has to say so at runtime instead of tracing back from inside the DSL. #799 showed it assuming otherwise:

- `causal_conv1d_update` imported `cudnn.frost.tile_dsl` (→ `cutlass.experimental`, 4.7+) from a route with no version check and failed on the `oss:cutlass-4.6` (4.6.2) lane on 09-04 with a bare `ModuleNotFoundError: cutlass.experimental`. The bulk route next to it already gated on `cutedsl_too_old()` and declined cleanly.
- Earlier, #854's module-scope `set_name_prefix(..., remove_cutlass_symbol=True)` failed the same way on the since-dropped 4.5.x lane, and the lazy-import wrapper reported it as *"requires optional dependencies — pip install ...[cutedsl]"*, which sends the user to a `pip install` that changes nothing.

This PR adds the rule and applies it at the shared seams:

- **`python/cudnn/AGENTS.md` Rule 7** — gate the DSL version at runtime, never assume the installed DSL satisfies your kernel: where the gate lives, the known API→version floors, "tests skip, not fail", and the incidents above as the why. Plus a Rule 6 note (its module-scope call is legal only behind the gate), checklist item 6, and a root `AGENTS.md` convention bullet pointing at Rule 7.
- **`frost.buffers.cutedsl_requirement_error(what)`** — builds the version-naming message from the existing `cutedsl_state()` / `cutedsl_too_old()` / `CUTEDSL_MIN_VERSION`; returns `None` when the DSL is absent, at/above the floor, or an internal RC.
- **Lazy-import wrappers** (`cudnn/__init__.py`, `cudnn/ops/__init__.py`) now say `X requires nvidia-cutlass-dsl >= 4.7.0; found 4.6.2. Upgrade with: pip install -U 'nvidia-cutlass-dsl>=4.7.0'` when the DSL is installed but too old; the existing install hint is kept for a genuinely missing dependency.
- **`_validated_native_update`** declines with `NotImplementedError` naming the version before importing the kernel module (mirrors the bulk route).
- **Tests**: `test_causal_conv1d_update_sm100.py` and `test_fla_short_conv_compat.py` skip below the floor; new host-only `test/python/test_cutedsl_version_gate.py` pins the wrapper and route behavior (too-old / at-floor / prerelease / internal-RC / not-installed).

Kernels are unchanged and the kernel floor stays 4.7.0 — downstream is expected to bump; until then 4.6.2 users get a clean decline instead of a DSL-internal traceback.

No API change. Error text on a too-old DSL changes from the install hint to a version message. Behavior at/above 4.7.0 is unchanged.

## Testing

- Host-only: `test_cutedsl_version_gate.py` 7 passed; `test_import_boundaries.py` 9 passed; conv1d/FLA host suites pass (57 total in the touched set).
- B200 (DSL 4.7.0): `test_causal_conv1d_update_sm100.py::test_nonzero_state_across_consecutive_steps` passes — the gate is transparent at the floor.
- Expected CI effect: on `oss:cutlass-4.6` the 18 conv1d-update failures become skips. Note the GitLab `.oss_tests:template` still pins the `cutlass-rel` lane to 4.5.1, which the new `>=4.6.2` pyproject floor will refuse at install time — that pin needs the GitLab-side follow-up to 4.6.2 regardless of this PR.

<sub>note to self: claude::c21097e1-70b0-4dab-b38c-343d85ba2e22 — "codex/kernel-factory PR batch design audit (797–814 + 835)" · cwd /home/scratch.yanxu_libs/cudnn_frontend · worktree scratchpad/wt-agents</sub>

## Summary by CodeRabbit

- **New Features**
  - Added runtime checks for the supported CuTe DSL version.
  - Clearer errors now distinguish missing dependencies from installed but outdated versions.
  - Unsupported CuTe DSL versions are blocked before affected kernels are loaded.

- **Documentation**
  - Clarified Python 3.10+ support requirements.
  - CuTe DSL is now included in the default editable installation.
  - Documented that CuTe DSL is required, while tensor frameworks remain optional.
  - Added guidance for version-aware runtime checks and testing.


YangXu1990uiuc:
@cudnn-ci-bot run python_tests,oss

egilliam-nv (approved):
Small feedback from Claude (doesn't block approval):

```(1) Rule 7 lists the gate points as the semantic op's route check, an engine's check_support, or the family __init__'s lazy import, but the FROST wrapper entry points (sdpa_*_wrapper_*) are a fourth caller-facing entry that reaches kernel modules on their own, and none of them gate today; the rule text should name them or the shared APIBase should. (2) The two test files it fixes use the pytestmark idiom while the rule text says "pytest.skip"; saying "skipif marker" in the rule avoids two idioms appearing. Neither blocks approval. No file overlap with #869.```

egilliam-nv:
One factual nit on the Rule 7 text, checked against the FlashInfer repo rather than memory: FlashInfer's floor is no longer 4.5.0.

- `main` `pyproject.toml`: `cu12 = ["nvidia-cutlass-dsl>=4.7.0a0"]`, `cu13 = ["nvidia-cutlass-dsl[cu13]>=4.7.0a0"]` (its `requirements.txt` still says `>=4.6.2a0`).
- latest release **v0.6.18** `pyproject.toml`: `>=4.6.2a0`.

So the downstream picture is: vLLM/SGLang `==4.6.2`, FlashInfer release `>=4.6.2a0`, FlashInfer main `>=4.7.0a0`. The conclusion (no floor in our `pyproject.toml`, gate at runtime) still holds because of the `==4.6.2` pins, but "flashinfer `>=4.5.0`" in the rule text and PR body is stale and worth updating so the rule doesn't send someone chasing a 4.5 constraint that no longer exists.

Separately: #919 applies Rule 7 to the SM80 SDPA wrapper entries and tests (stacked on this PR); it also notes that the FROST wrapper entry points are a fourth caller-facing gate location the rule text could name.

YangXu1990uiuc:
@cudnn-ci-bot run python_tests,oss

egilliam-nv:
Follow-up for the two CodeRabbit findings that merged unaddressed here — the prerelease-blind `cutedsl_too_old()` (`4.6.2rc1` read as not too old) and `_optional_dependency_message` blaming the DSL for a missing `torch` — is in #919 (commit `2bf1f0848`), with regression tests; the RUF043 raw-string nit too.

PR #921, opened 2026-09-04
build(python): make CuTeDSL a required dependency, floor 4.6.2, require Python >=3.10

Anerudhan:
## What

Moves the CuTeDSL runtime out of the `cutedsl` optional-dependency group and into the required `[project] dependencies`, raises the DSL floor from `>=4.5.0` to `>=4.6.2`, and raises `requires-python` to `>=3.10`.

The OSS kernel engines (FROST SDPA, GEMM fusions, sparse/linear attention) are Python kernels that JIT through CuTeDSL. They are a headline part of this package rather than an add-on, so the runtime they need should install by default.

```diff
-requires-python = ">=3.9"
+requires-python = ">=3.10"

+dependencies = [
+    "nvidia-cutlass-dsl[cu13]>=4.6.2",
+    "apache-tvm-ffi>=0.1.11",
+]

 [project.optional-dependencies]
-cutedsl = [
-    "nvidia-cutlass-dsl[cu13]>=4.5.0",
-    "cuda-python",
-    "apache-tvm-ffi>=0.1.11",
-]
+cutedsl = [
+    "cuda-python",
+]
```

Resulting wheel metadata:

```
Requires-Python: >=3.10
Requires-Dist: nvidia-cutlass-dsl[cu13]>=4.6.2
Requires-Dist: apache-tvm-ffi>=0.1.11
Requires-Dist: cuda-python; extra == "cutedsl"
Requires-Dist: cuda-tile>=1.4; extra == "cutile"
```

## Notes for review

**Python 3.9 is dropped.** `requires-python` was `>=3.9` while the classifiers already only listed 3.10–3.14, and every `nvidia-cutlass-dsl` release is `Requires-Python >=3.10` — so with the DSL required, 3.9 could not have resolved anyway. Raising the floor makes that explicit instead of encoding it in a marker. It also makes the `python_version >= '3.10'` marker on `cuda-tile` redundant, so that is removed too. `README.md` and `AGENTS.md` now state Python ≥ 3.10.

**`cuda-python` stays in the `cutedsl` extra rather than moving up.** It arrives either way: the required `nvidia-cutlass-dsl` pulls `nvidia-cutlass-dsl-libs-base`, which itself declares `cuda-python>=12.8`, so a plain `pip install nvidia-cudnn-frontend` still has it. Naming it in the extra pins it to this package instead of to a transitive detail of the DSL's own packaging, and keeps `pip install nvidia-cudnn-frontend[cutedsl]` resolving for everyone who still writes it that way.

**`CUTEDSL_MIN_VERSION` is unchanged at `(4, 7, 0)`.** The packaging floor deliberately sits below the version the FROST engines want, for the same reason as before: pinning the hard floor at 4.7 would make cudnn-frontend uninstallable next to anything holding the DSL back. The engines version-check at support time and decline, so an older DSL costs those engines and nothing else.

**One compatibility call worth flagging:** the previous comment justified the low floor partly by "quack-kernels pins `==4.6.0`". A 4.6.2 floor now hard-conflicts with that pin, and since the dependency is required rather than opt-in, the conflict can no longer be sidestepped by skipping the extra. Flagging in case quack co-installation matters to anyone.

**The lazy-import boundary is untouched.** `import cudnn` must still not eagerly import torch or cutlass; torch and jax remain opt-in via the PEP 735 dependency groups. `python/cudnn/__init__.py`'s `_OPTIONAL_DEPENDENCY_INSTALL_HINT` is left as-is, so `test/python/test_import_boundaries.py` — which hard-asserts that hint string — is unaffected.

## Docs

Install snippets across `docs/fe-oss-apis/`, `AGENTS.md`, `CONTRIBUTING.md`, and the benchmark READMEs drop the now-redundant extra, and the stale "package floor is 4.5.0" / "optional dependency" claims are corrected (`bsa.md`, `csa/compressor/compressor_sm100_r128.py`, `block_sparse_attention/_interface.py`, `python_graph_and_execution_backends.md`).

## Verification

- Generated the wheel `METADATA` via the PEP 517 `prepare_metadata_for_build_wheel` hook and confirmed the two required packages appear as unconditional `Requires-Dist` with no `extra ==` guard, and that `cuda-python` appears only under `extra == "cutedsl"` (output quoted above).
- Parsed every requirement with `packaging.Requirement` and asserted name / specifier / extras, that no marker remains on either the DSL or `cuda-tile`, that `requires-python == ">=3.10"`, and that no 3.9 classifier survives.
- Confirmed on PyPI that `nvidia-cutlass-dsl` 4.6.2 is published, that every release carries `Requires-Python >=3.10`, and that `nvidia-cutlass-dsl-libs-base` 4.6.2 declares `cuda-python>=12.8` — which is what makes demoting `cuda-python` safe.
- `black --line-length 160` clean on the touched Python files; pre-commit (black, SPDX) passed.

Not run: the Python test suite, which needs a compiled extension and a GPU. CI should cover it.

## Summary by CodeRabbit

* **Breaking Changes**
  * Python 3.10 or newer is now required; Python 3.9 is no longer supported.

* **Installation**
  * The standard installation now includes CuTeDSL and related kernel dependencies.
  * Existing installation instructions have been updated to use the base package rather than the optional CuTeDSL extra.
  * The CuTeDSL extra remains available as a compatibility alias for `cuda-python`.

* **Documentation**
  * Updated setup, testing, benchmark, and integration guidance to reflect the revised dependencies and requirements.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,8 +43,8 @@ The C++ build uses `-Werror` (`/WX` on MSVC) with `-Wall -Wextra -Wpedantic` —
 Python (editable; compiles the pybind11 extension via CMake):
 
 ```bash
-pip install -e .              # core graph API only
-pip install -e ".[cutedsl]"   # + OSS CuTeDSL kernels (nvidia-cutlass-dsl, cuda-python, tvm-ffi; framework-neutral)
+pip install -e .              # graph API + the OSS CuTeDSL kernels (nvidia-cutlass-dsl, cuda-python, tvm-ffi; framework-neutral)
+                              # ".[cutedsl]" still resolves -- it is now an empty back-compat alias for the line above
 pip install -e ".[cutile]"    # + the cuTile linear-attention engines (cuda-tile; needs a system tileiras)
 pip install --group torch      # + torch for the CuTeDSL APIs (torch, torch-c-dlpack-ext)
 pip install --group jax        # + jax for the CuTeDSL APIs (jax >= 0.5; XLA entry points via cutlass.jax)
@@ -66,7 +66,7 @@ cd test/python
 pytest                        # default is -m L0 (smoke level) per pytest.ini
 pytest -m L1                  # deeper levels: L0..L4
 pytest test_conv_fprop.py     # one file (still filtered by -m L0 — pass -m "L0 or L1" to widen)
-pytest fe_api/                # OSS kernel tests; require ".[cutedsl]" + `--group torch` (and `--group jax` for the *_jax tests) + SM90/SM100 GPU
+pytest fe_api/                # OSS kernel tests; require `--group torch` (and `--group jax` for the *_jax tests) + SM90/SM100 GPU
 ```
 
 Read [test/AGENTS.md](test/AGENTS.md) before touching tests — `test/python/conftest.py` has import-order and env-var requirements that are easy to break.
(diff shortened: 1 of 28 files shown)


egilliam-nv:
The doc/README nits CodeRabbit left here (stale `4.5` floors in `fla.md`/`bsa.md`, the truncated sentence in `benchmark/dsa/README.md`, the remaining `[cutedsl]`-extra hints in the flex-attention benchmark README and `python/cudnn/AGENTS.md`) are fixed in #919 as a separate commit.

Issue #943, opened 2026-09-08
test(sdpa): random ragged sweeps lack the cuDNN >= 9.25 gate for zero-length sequences (NaN rows on 9.22)


yuweih205:
## Summary

On an H200 with **cuDNN 9.22.0** (torch 2.13.0+cu132, frontend built from `develop`
2cc71fbf), the ragged L0 sweeps in `test_mhas_v2.py` fail for every configuration in
which some batch has `seq_len_kv == 0`: the engine returns **NaN** for the query rows
of that batch, the reference returns zeros, and the comparator counts NaN-vs-finite
as a mismatch.

This looks like a version-gate gap rather than an engine or reference regression:

- `test_sdpa_edge_cases.py` already covers zero-length THD sequences
  (`test_thd_zero_seqlen`, `test_cu_seqlen_zero_seqlen`) and gates them with
  `cudnn.backend_version() < 92500 -> skip, "zero sequence length SDPA requires
  cuDNN >= 9.25.0"`.
- The random sweeps draw zero-length sequences with ~10% probability per batch
  (`seq_len_q` since v1.18, `seq_len_kv` since a5ce256b / #5xx on 2026-08-15) but
  `exec_sdpa` has no matching `< 9.25` waiver, so on cuDNN 9.22–9.24 every such draw
  fails.
- The reference returned zeros for zero-KV rows both before and after #910, so the
  flash-style rewrite did not change the expectation; it only coincided with my run.

| sweep | total | with `seq_len_kv == 0` somewhere | failing | zero-length Q only | failing | no zero-length | failing |
|---|---|---|---|---|---|---|---|
| `test_sdpa_random_fwd_ragged_L0` | 256 | 75 | 67 | 50 | 0 | 131 | 0 |
| `test_sdpa_random_bwd_ragged_L0` | 384 (104 waived) | 263 | 176 | 97 | 1 | 24 | 0 |

The counts are identical with the ragged stride-gap fuzz disabled and on pristine
`develop`, so this is independent of the stride fuzzing. The pre-#910 test tree (75a3f430^) fails the same way on this engine (41 / 128 fwd, 112 / 256 bwd, all with a zero-length sequence and NaN from the engine; 0 failures without one), so #910 is not the cause. Under cuDNN 9.25.1.1 (same H200, same frontend build) the failing configs pass: `test1` passes as ragged, padded, COMPOSITE, UNIFIED, fp16 and bf16, and `test61` passes in every variant tried.

[...]

PR #944, opened 2026-09-08
test(sdpa): waive zero-length sequences below cuDNN 9.25 in the random sweeps

yuweih205:
## What

Waive configurations that contain a zero-length sequence when the backend is older
than cuDNN 9.25, in `exec_sdpa`, mirroring the gate that `test_sdpa_edge_cases.py`
already applies to its zero-seqlen cases (`zero sequence length SDPA requires
cuDNN >= 9.25.0`). Six lines; on cuDNN >= 9.25 nothing changes.

## Why

The random sweeps draw a zero-length sequence for a batch with ~10% probability
(`seq_len_q` since v1.18, `seq_len_kv` since a5ce256b). Zero-length sequences are a
9.25 feature: on cuDNN 9.22 the engine returns NaN for the query rows of a zero-KV
batch while the reference returns zeros, so every such draw fails —
`test_sdpa_random_fwd_ragged_L0` 67 / 256 and `test_sdpa_random_bwd_ragged_L0`
177 / 384 on an H200, identical before and after #910 and with the ragged stride-gap
fuzz on or off. Details and repro in #943.

## Verification

H200, frontend built from `develop`:

| cuDNN | | fwd ragged L0 | bwd ragged L0 |
|---|---|---|---|
| 9.22.0 | without this change | 67 failed / 189 passed | 177 failed / 103 passed / 104 skipped |
| 9.22.0 | with this change | 0 failed / 131 passed / 125 skipped | 0 failed / 22 passed / 362 skipped |
| 9.25.1.1 | (gate inactive) | 0 failed / 256 passed | 0 failed / 280 passed / 104 skipped |

Every config without a zero-length sequence is unaffected (131 / 131 and 22 / 22
still run and pass). `black` and the SPDX hook pass.

Refs #943.

### DIFF
--- a/test/python/sdpa/fp16.py
+++ b/test/python/sdpa/fp16.py
@@ -130,6 +130,12 @@ def validate_config(cfg):
         print("@@@@ Overall result: WAIVED, mixed-form sequence lengths require cuDNN 9.25.0 or higher.")
         pytest.skip("mixed-form sequence lengths (cumulative on one side only) require cuDNN 9.25.0 or higher")
 
+    # Same gate as test_sdpa_edge_cases.py: the random sweeps draw zero-length
+    # sequences (~10% per batch), which older engines answer with NaN rows.
+    if cudnn_version < "9.25.0" and (0 in cfg.seq_len_q or 0 in cfg.seq_len_kv):
+        print("@@@@ Overall result: WAIVED, zero sequence length SDPA requires cuDNN 9.25.0 or higher.")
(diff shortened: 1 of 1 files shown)


egilliam-nv (approved):

egilliam-nv:

egilliam-nv:
Thanks for the careful diagnosis and the before/after table on 9.22 vs 9.25.1, and for filing #943 alongside it. The gate matches the neighbouring version waivers in `validate_config`, and CI's only failures are a develop-wide conv test unrelated to this change. Merging; we'll mark #943 fixed once this lands.

PR #952, opened 2026-09-08
test_mhas_v2 refs: torch.matmul instead of torch.einsum (SM107 CI worker crashes in cuBLAS)

vedaanta:
Replaces the three `torch.einsum` contractions in the `test_mhas_v2` SDPA references (`fp16_ref.py`, `fp8_ref.py`, `mxfp8_ref.py`) with `torch.matmul` on the same grouped 5-D views. Same fp32 math; only the cuBLAS call shape changes.

### Why
`test_mhas_v2.py -n 8` on the SM107 CI (aarch64 VR200, `dl/dgx/pytorch:rubin-py3-devel`: torch 2.14.0a0 nvinternal, CUDA 13.5, cuBLAS 13.8.0.1023, driver 615.62.07) loses xdist workers to `CUBLAS_STATUS_EXECUTION_FAILED` in `cublasSgemmStridedBatched` / `unspecified launch failure`, or hangs with the GPU at 100 %, inside the **PyTorch reference** (e.g. gitlab job 430534451, after the stream fix in #947). A cuDNN-free reproducer that only runs the reference einsums shows the same thing:

| workload on one VR200 GPU, 8 processes | result |
|---|---|
| reference `einsum`s (`"bhgqd,bhkd->bhgqk"` etc.) | 4 of 4 runs hit: 5 crashes + 1 hang across 32 workers |
| same products via `torch.matmul` on the 5-D views | 0 of 3 runs |
| einsum, 1 or 2 processes, or `CUDA_LAUNCH_BLOCKING=1` | clean |

The difference is the GEMM layout: einsum folds the GQA group into M (one bmm with M = g·s_q up to ~10^5, N = the 128-wide KV block, batch = b·h_k); matmul keeps the group as a batch dimension (M = s_q). Only the einsum shape faults on that platform. B200 and x86-hosted Rubin are clean either way. The cuBLAS/driver issue is being reported separately with the standalone reproducer; this PR just removes our exposure so the CI can stay at 8 workers.

### Change
- `fp16_ref.py`: `_qk` / `_pv` rewritten with `torch.matmul`; new `_kv_reduce(x, y, h_kv)` = x^T @ y per head summed over the GQA group (the old `"bhgqk,bhgqd->bhkd"`), used for dK and dV. Comment at the definitions records the reason.
- `_mm(a, b)`: `torch.matmul`, except a contraction of length 1 (a 1-column KV block in P·V, or s_q = 1 in dSᵀ·Q / Pᵀ·dO) is computed as the broadcast product `a * b`. Found on x86 Rubin with the torch 2.14.0.dev nightly: its experimental `torch._native` layer reroutes `aten::bmm` with K = 1 to a Triton outer-product kernel, and Triton's bundled ptxas has no `sm_107a` target (`PTXASError`, and in some cases a process abort that takes the xdist worker down). einsum never hit that path; the plain matmul form did in 32 cases (`*_sq1_*`, random shapes with s_kv ≡ 1 mod 128). The product is exact either way.
- `fp8_ref.py`, `mxfp8_ref.py`: the dP / dK / dV einsums call `_qk` / `_kv_reduce`; the now-unused pre-grouped views are gone.
- Helper parity vs the old einsums (fp32, several GQA shapes): Q·Kᵀ and P·V bit-identical; dK/dV differ only by fp32 summation order (≤ 2e-6 relative).

### Verification (`test_mhas_v2.py -n 8`, cudnn 9.26 release build)
| GPU | result |
|---|---|
| B200 | **3243 passed / 0 failed** |
| x86 Rubin (cc 10.7) | **3243 passed / 0 failed** |

No test tolerances or expected values change.

## Summary by CodeRabbit

* **Tests**
  * Updated FP16, FP8, and MXFP8 attention reference calculations to use revised grouped contraction paths.
  * Improved coverage of grouped-query and multi-query attention backward calculations, including gradient computations.
  * Avoided affected contraction paths for single-column operations in FP16 reference testing.

### DIFF
--- a/test/python/sdpa/fp16_ref.py
+++ b/test/python/sdpa/fp16_ref.py
@@ -93,17 +93,42 @@ def _grouped(x, h_kv):
     return x.view(b, h_kv, h_q // h_kv, s, d)
 
 
+# The three contractions below are written with torch.matmul on the grouped 5-D views rather
+# than torch.einsum. Same fp32 math, but einsum lowers "bhgqd,bhkd->bhgqk" to one bmm with the
+# GQA group folded into M (M = g*s_q up to ~10^5, N = the 128-wide KV block); on aarch64 VR200
+# (SM107) that cublasSgemmStridedBatched shape crashes or hangs under multi-process GPU
+# sharing (xdist -n 8), while the matmul form keeps the group as a batch dimension and does
+# not. See /home/scratch.vagarwalla_gpu/sm107_cublas_concurrency_repro.
+
+
+def _mm(a, b):
+    # matmul, except that a contraction of length 1 (a 1-column KV block in P @ V, or s_q = 1 in
+    # dS^T @ Q / P^T @ dO) is a plain broadcast product. torch nightlies (2.14.0.dev, torch._native)
+    # reroute aten::bmm with K = 1 to a Triton outer-product kernel, and Triton's ptxas has no
+    # sm_107a target (PTXASError / process abort on Rubin). The product is exact either way.
+    if a.shape[-1] == 1:
+        return a * b
+    return torch.matmul(a, b)
+
+
 def _qk(q, k_block, h_k):
     # q: (b, h_q, s_q, d), k_block: (b, h_k, n, d) -> (b, h_q, s_q, n) without expanding K.
     b, h_q, s_q, _ = q.shape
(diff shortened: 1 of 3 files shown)


vedaanta:
@cudnn-ci-bot run frost,backend

YangXu1990uiuc (approved):

PR #954, opened 2026-09-09
frost(sdpa): Rubin (SM107) prefill forward — 9 kernels, 3 engine rows, f16/FP8/MXFP8

RomanAnders90:
## What

Ports the **SM107 (Rubin) prefill forward** family onto the FROST engine surface
and registers it end to end. Nine new kernels join the already-shipped d128 FP8
sibling behind three opt-in engine rows:

| | f16 / bf16 | per-tensor FP8 | MXFP8 |
|---|---|---|---|
| **d128** | new | shipped | new |
| **d192 × d128** | new | — (rides d256) | — (declined) |
| **d256** | new | new | new |
| **d512** | new | new | new |

Engine rows: `sdpa_fwd_prefill_sm107` (slot **15**, new), `sdpa_fwd_prefill_sm107_fp8`
(slot 14, widened to d256/d512), `sdpa_fwd_prefill_sm107_mxfp8` (slot **16**, new).
All `opt_in=True`.

## Why a separate `config_sm107.py`

Rubin is a different kernel *lineage*, not the Blackwell kernels recompiled — the
two share flavor NAMES and almost nothing else. Routing the SM107 d256 FP8 kernel
through the Blackwell `make_cfg_d256` gave a scheduler mbarrier init count of **15
against 11 actual arrivers**: an unreachable barrier that hung at *every* shape,
down to `B=1 H=1 S=128`, with no fault and nothing for a sanitizer to find. The
new module carries that and two more geometry constraints as raising validators.

## Kernel fixes this port required

Each is a silent-wrong-answer class — no crash, no error, wrong numbers:

- **V scale factors are D-plane-major in GMEM** (d256/d512 MXFP8). The columnwise
  quantization applies the F8_128x4 atom rule to the *transposed* `[D, S/32]`
  matrix, so the D-block is the OUTER index and the planes sit a whole plane of
  atoms apart — a stride that **grows with S**. Reusing the Q/K per-tile
  descriptor builder was correct at one KV tile and degraded from there
  (cos 0.9997 → 0.980 → 0.969). d=128 has a single plane, so the passing sibling
  could not catch it.
- **Version-1 tcgen05 SMEM descriptors past 256 KiB.** A version-0 descriptor
  truncates `start_address` to 14 bits and wraps to the bottom of SMEM. The d512
  MXFP8 scale-factor tiles start near 292 KiB, so the UTCCP copied **Q data** into
  the SF TMEM columns and BMM1 multiplied by garbage E8M0 exponents: `LSE=+inf`,
  `O=NaN` at every shape. Applied per buffer, only where the line is crossed —
  setting it everywhere regressed d128/d256 MXFP8.
- **`SCHED_LPT` is not honored by the ported kernels.** A heuristics fallback bug
  hid it: when a row's domain excluded the primary policy, ranking fell back to
  NATURAL (and listed it twice) instead of walking the preference order. Fixing
  that let causal graphs reach LPT for the first time — where it returns NaN
  (d512 FP8) or `max|O-ref| ≈ 1.9` (d128 MXFP8). All three rows now declare
  `sched_policies={SCHED_NATURAL}`, and the standalone wrapper's own derivation
  is arch-gated to match.
- **`has_lse=False` specialization** on all nine kernels: a stats-less graph binds
  no dummy buffer at any level and reports `get_workspace_size() == 0`. The FP8
  row already claimed `lse_optional`, which was untrue for the wide kernels it
  was widened to.
- **Empty-KV guard** (d256/d512): an empty mainloop leaves `total_sum == 0`, where
  the `1e-30` division floor leaks into the output — `LSE = log(1e-30) = -69.08`
  and `residue * inf = NaN`. Substituted with a SELECT, never `* 0`.

## Coverage

The SM100 forward suites now run on cc 10.7 with an arch-aware engine pin
(`test_sdpa_fwd_dsl_sm100.py`, `..._fp8_sm100.py`, `..._mxfp8_sm100.py`), which is
how the f16/bf16 row gets its first automated numerics — **273 executing tests
that previously had none**. Plus `test_sdpa_fwd_dsl_sm107.py` (device-independent
row/config assertions).

Every feature a Rubin row declines — THD, PackGQA, split-KV, dense padded-Q trim,
d192 MXFP8, wide-shape strided Stats — is a `skipif(_SM == 107, ...)` whose reason
names the row field that declines it, so the skip **inverts** when the gap closes
rather than being deleted. `SUPPORT_MATRIX_TRACKER.md` updated per SDPA **Rule S2**.

## Validation

`pytest sdpa/frost/ -m "L0 or L1"` on each arch:

| arch | result |
|---|---|
| **SM107** Rubin (`w2u1g-lc-0030`) | **746 passed**, 10 failed — all 10 pre-existing |
| **SM100** Blackwell (`4u4g-gen-0150`) | **1371 passed**, 9 failed — the same pre-existing module |
| **SM80** A100 | **363 passed, 0 failed** |

**No failure is attributable to this PR:**

- **9 ×** `test_tile_dsl_q_loop_bounds.py` — arrived with #948 and fails on
  **both** SM100 and SM107. *Measured, not argued:* checked out pure
  `origin/develop` (c541e716) on the Rubin node and it fails 9/9 there, with none
  of this branch applied. It imports only `tile_dsl.mask`, which this PR does not
  touch. Worth its own issue.
- **1 ×** `test_fp8_thd_sliding_window[False-e5m2-d128]` (Rubin only) — the known
  pre-existing E5M2 case: 1 element in 3600 at LSE diff 0.0337 against a 0.02
  tolerance, on the *shipped* d128 FP8 kernel. Not fixed and **not widened**;
  it needs its own look.

## Not in this PR

- **Scheduler (`SCHED_LPT` / `LPT_L2`) on the Rubin line.** Recovering it for the
  shipped d128 FP8 kernel — which does honor both and loses them here because
  `sched_policies` is row-wide — needs a per-shape domain
  (`sched_policies_by_d_shape`, mirroring `cgas_by_d_shape`).
- **THD/varlen** on the ported kernels (7-arg vs 14-arg setup contract, 3B+2 vs
  4B+4 metadata). Declined honestly by every row.
- **d512 MXFP8 test module.** The kernel is correct (cos 0.9997, LSE exact, SQ ∈
  {128…512}) but SM100 has no d512 MXFP8 sibling, so the shared suite carries no
  d512 case to widen — it needs new cases, not a widened gate. Marked ⚠️ in the
  tracker rather than ✅.
- Perf. Functional only.

## Labels

`cat-feature` · `area:frost` · `op:sdpa` · `orig-nvidia`

## Summary by CodeRabbit

- **New Features**
  - Added SDPA forward support for SM107/Rubin GPUs across FP16, BF16, per-tensor FP8, and MXFP8 formats.
  - Expanded supported head dimensions, including d128, d192, d256, and d512 where applicable.
  - Added support for masking, GQA, LSE outputs, stream-aware execution, and architecture-specific scheduling.

- **Bug Fixes**
  - Improved scheduler fallback behavior and configuration validation.
  - Corrected empty-KV results to return zero outputs and appropriate LSE values.

- **Documentation**
  - Updated the SM107/Rubin support matrix with supported features and limitations.

### DIFF
--- a/python/cudnn/engines/manifest.py
+++ b/python/cudnn/engines/manifest.py
@@ -206,6 +206,8 @@ MANIFEST: Tuple[EngineFamily, ...] = (
             "sdpa_fwd_prefill_sm100_mxfp8": EngineSlot(12, opt_in=True),
             "sdpa_fwd_prefill_sm100_fp8": EngineSlot(13, opt_in=True),
             "sdpa_fwd_prefill_sm107_fp8": EngineSlot(14, opt_in=True),
+            "sdpa_fwd_prefill_sm107": EngineSlot(15, opt_in=True),
+            "sdpa_fwd_prefill_sm107_mxfp8": EngineSlot(16, opt_in=True),
         },
         analyzer=("cudnn.sdpa.graph_analyzer", "analyze"),
         heuristics=("cudnn.sdpa.fwd.heuristics", "recommend"),
--- a/python/cudnn/frost/tile_dsl/handles.py
+++ b/python/cudnn/frost/tile_dsl/handles.py
@@ -158,6 +158,9 @@ class SmemTile:
     def desc(self):
         from cutlass.experimental import primitives as prims
 
+        if self.desc_version not in (0, 1):
+            raise ValueError(f"SmemTile.desc_version must be 0 or 1 (got {self.desc_version}); there is no other tcgen05 SMEM-descriptor format")
+
         if self.desc_version == 0:
             return prims.Tcgen05SmemDesc.build(
                 self.base,
@@ -173,6 +176,15 @@ class SmemTile:
         import cutlass
         from cutlass._mlir.dialects import llvm
         from cutlass.experimental.primitives import nvvm_wrapper as _nvvm_wrap
+
+        if not hasattr(_nvvm_wrap, "_tcgen05_mma_smem_desc_v2"):
+            raise RuntimeError(
+                "desc_version=1 needs the versioned tcgen05 SMEM-descriptor intrinsic "
+                "(cutlass.experimental.primitives.nvvm_wrapper._tcgen05_mma_smem_desc_v2), "
+                "which this cutlass-dsl build does not expose.  It is required on Rubin "
+                "(SM107) for any MMA operand at or above 256 KiB; there is no kernel-side "
+                "workaround -- upgrade the DSL."
+            )
 
         addr = cutlass.Int32(llvm.ptrtoint(cutlass.Int32.mlir_type, self.base.ir_value()))
         return prims.Tcgen05SmemDesc(
--- a/python/cudnn/frost/tile_dsl/handles.py
+++ b/python/cudnn/frost/tile_dsl/handles.py
@@ -127,6 +127,24 @@ class SmemTile:
     tma_granu_elems: int = 0
     tma_subtile_stride_elems: int = 0
     stages: int = 1
+    # tcgen05 SMEM-descriptor version for :meth:`desc`.
+    #
+    # 0 = the public ``Tcgen05SmemDesc.build`` path (SM100 format).  Its
+    # ``start_address`` field is bits [0:15) with bit [14] *reserved on SM100*,
+    # i.e. only 14 usable bits = a 256 KiB addressable window.
+    #
+    # 1 = the extended format, which is what the C++ ``SmemTile::make_desc``
+    # always emitted.  REQUIRED on Rubin (SM107) for any MMA operand whose SMEM
+    # buffer sits at or above 256 KiB — the arch raises the per-CTA SMEM cap to
+    # 327 KiB, but a version-0 descriptor still truncates the address to 14 bits
+    # and silently wraps it to the bottom of SMEM.
+    #
+    # Symptom of getting this wrong: the MMA runs, both operands are provably
+    # correct in SMEM, the accumulator TMEM is the one the epilogue reads — and
+    # the accumulator is EXACTLY zero, because the wrapped address landed on an
+    # untouched buffer.  No crash, no launch error.  (Found 2026-09-04 on the
+    # d512 f16 SM107 port: sP_xfer_raw began at exactly 262144.)
+    desc_version: int = 0
 
     def __getitem__(self, stage):
         off = stage * self.elems_per_stage
@@ -140,11 +158,34 @@ class SmemTile:
     def desc(self):
         from cutlass.experimental import primitives as prims
 
-        return prims.Tcgen05SmemDesc.build(
-            self.base,
-            leading_byte_offset=self.leading_byte_offset,
-            stride_byte_offset=self.stride_byte_offset,
-            layout=self.layout,
+        if self.desc_version == 0:
+            return prims.Tcgen05SmemDesc.build(
+                self.base,
+                leading_byte_offset=self.leading_byte_offset,
+                stride_byte_offset=self.stride_byte_offset,
+                layout=self.layout,
+            )
+
+        # Hand-rolled because the public ``Tcgen05SmemDesc.build`` takes no
+        # ``version`` argument — it calls the non-versioned intrinsic.  The
+        # versioned one exists in the same wheel, just module-private, and the
+        # argument list below mirrors ``build``'s defaults exactly.
+        import cutlass
+        from cutlass._mlir.dialects import llvm
+        from cutlass.experimental.primitives import nvvm_wrapper as _nvvm_wrap
+
+        addr = cutlass.Int32(llvm.ptrtoint(cutlass.Int32.mlir_type, self.base.ir_value()))
+        return prims.Tcgen05SmemDesc(
+            _nvvm_wrap._tcgen05_mma_smem_desc_v2(
+                addr >> cutlass.Int32(4),
+                self.leading_byte_offset >> 4,
+                self.stride_byte_offset >> 4,
+                self.desc_version,
+                0,  # base_offset
+                0,  # leading_dim_mode
+                0,  # k_segment_offset
+                self.layout,
+            )
         )
 
 
--- a/python/cudnn/frost/tile_dsl/tmem.py
+++ b/python/cudnn/frost/tile_dsl/tmem.py
@@ -11,9 +11,12 @@ import cutlass.cute as cute
 def tmem_alloc(tmem_ptr_i32, num_cols: int, cta_group_kind, is_exclusive: cutlass.Constexpr[bool] = False):
     # is_exclusive is the sm_107/sm_109 >512-column mechanism (GR100's 576-col
     # TMEM; cute.arch.tmem.is_tmem_allocation_exclusive) — pass True only
-    # there.  It is fenced out of the PUBLIC cutlass-dsl tcgen05_alloc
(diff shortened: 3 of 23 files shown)


RomanAnders90:
## Review response — 21 of 24 findings applied, 3 declined

Thanks — several of these were real, and two led somewhere more interesting than
the finding itself. Validated on all three arches; details at the end.

### Declined

**`k_dim=1` on the f16 `Tcgen05InstrDesc`s (flagged 🟠 Major).** This is wrong and
would break three validated kernels. `k_dim` is the **FP8/MXFP8 K=64 fast-path
selector**, and it is *arch-opposite*: `k_dim=1`+`TILE_K_HW=64` on Rubin,
`k_dim=0`+`TILE_K_HW=32` on Blackwell. **f16/bf16 use `TILE_K_HW=16` and leave
`k_dim` at its default `0` on both arches** — the SM107 f16 idescs are
byte-identical to their validated SM100 siblings
(`prefill_d128_f16_sm100.py:1104-1118`), and `config_sm107._validate_cfg`
*raises* if `TILE_K_HW != 16` at a half dtype. The premise ("a 16-element
hardware K-step requires `k_dim=1`") does not hold. Setting it would produce the
per-row-scrambled accumulator that this bit is famous for — no crash, passes a
1-tile shape.

**Drop `(192, 128)` from `_SM107_KERNEL_FILES`** and **make the adapter's FP8
floor table arch-aware.** Both correctly identified a row/adapter contradiction;
both resolve the *other* way — see below. Hiding a shipped kernel, or widening
the adapter to match a row that overclaims, would each have locked in the wrong
half.

**Sort `__all__` (RUF022) / RUF005.** There is no ruff in this repo — no
`ruff.toml`, and nothing in `pyproject.toml` or `.pre-commit-config.yaml` (the
hooks are clang-format, black, and the SPDX check). No lint gate to keep green,
and the list is grouped semantically. I took the RUF005 one-liner anyway since
it reads better; skipped the `__all__` re-sort.

### Applied, wider than reported

**Empty-KV epilogue guard.** Real, and the family was in worse shape than the two
files flagged. A survey found the guard present in **six** SM107 kernels under
**three different names** — `_kv_empty` (d256), `_row_empty` (d512), `row_dead`
(d128 FP8, and every `*_sm100.py`) — and **absent in three**: d128 f16, d192×d128
f16, and **d128 MXFP8**, which the review missed. All three now guard it.

Worth recording because it defeats the obvious grep: a single-name search reports
"clean" on a file with no guard at all. The family sweep is
`grep -L 'row_dead\|_kv_empty\|_row_empty' python/cudnn/sdpa/fwd/kernels/prefill_*.py`.

Two corrections to the finding's reasoning, both of which changed the fix:
- The accumulator is **not** the only problem, and `total_sum` is **not** residue.
  The softmax warp publishes `(total_max = -inf, total_sum = 0)` even on a
  zero-iteration KV loop, so **LSE is already exact on both arms** in these three
  kernels — `-inf + log(0) = -inf` (they put no `1e-30` floor *inside the log*;
  their d256 siblings do, and that is why d256 needed the LSE select and these
  do not). What is wrong is **O**: `inv_sum = 1/max(0, 1e-30) = 1e30` multiplying
  a TMEM accumulator that BMM2 never overwrote. So the O select is the load-bearing
  half; I kept a (no-op) LSE select for uniformity and documented the coupling.
- On the MXFP8 kernel the select must land **before the amax fold**, not just
  before the store — `atomicMax` only grows, so one dead row's NaN would inflate
  `Amax_O` for the whole tensor.

**The `pack_gqa_d_shapes` / `d_envelope_floors` pair.** The exact-vs-envelope
finding was right, and chasing it surfaced a worse sibling: the Rubin FP8 row
declared `d_envelope_floors=()` while `api_dsl._SM100_FP8_ENVELOPE_FLOORS` is
arch-**independent**. So `mismatch()` *admitted* `(512, 256)` and `(192, 128)` on
Rubin and `check_support` then raised a bare `ValueError` — a plan that enters
the ranked list only to die in the lowering, and an untyped escape (contract
rules 1 and 2). Two tests asserted the wrong half out loud, next to a comment
asserting the right one.

Fixed by giving the row the floors rather than widening the adapter: the
rationale is *kernel geometry* (the Rubin d512 FP8 kernel is the same cga4×1
role-split; its d256 padded envelope is no more validated than Blackwell's), so
it transfers with the port. `test_fp8_envelope_floor_matches_engine_row` existed
and still missed this because it pinned **one** row; it now pins both, deriving
the expectation from the row's own `d_shapes`.

**`_sm107_spec.d_shapes` was missing `(192, 128)`.** Resolved by widening the row,
not by deleting the kernel: `prefill_d192_d128_f16_sm107.py` ships in this PR and
`_pick_flavor` already lands d=192 f16 graphs on it. The row said "there is no
d192×d128 sibling" while the adapter used one. The quantized rows genuinely lack
it and now *decline* d=192 rather than riding an unvalidated padded envelope.

**`DESC_VERSION`.** The suggestion to assert a module constant instead of grepping
source was right, and the old test could match its own prose. Every SM107 kernel
now declares one `DESC_VERSION` next to its `CFG` and passes
`desc_version=DESC_VERSION` at *every* `SmemTile`, with two tests: one pinning the
value against the >256 KiB boundary (now across f16 **and** the quantized
families — the d512 MXFP8 kernel is where a missing v1 descriptor actually shipped
`LSE = +inf, O = NaN`), one counting `SmemTile(` against wired call sites so a
re-literalled `desc_version=0|1` fails. Without the second test the constant is
decorative and the drift returns.

### Applied as reported

`handles.py` `desc_version` validation (+ a named error when the versioned
intrinsic is absent) · `Capabilities.pack_gqa_d_shapes` moved to the end of the
dataclass (append-only) · the SM107 THD guard narrowed to FP8/MXFP8 (it repeated
the dtype set above it and declined nothing) · both kernel docstrings (the THD
claim, the d192 title) · the superseded `sched_policies` comment blocks and the
MXFP8 "optional stats" docstring · `SUPPORT_MATRIX_TRACKER.md` (MXFP8 row, d192
sibling, envelope row, gap list) · the three inert `skipif(False)` markers and
their 25 references · the stale THD comment and d512 docstring · the tautological
`rubin_pool` loop (now checks the module is on disk) · both misleading test names
· `_sched_points` multi-element fallback coverage.

Also fixed while in there: the d512 MXFP8 amax atomic sat inside the **dense** arm
of the THD branch, so a THD graph folded no amax at all.

### Validation

| arch | result |
|---|---|
| SM80 (A100, local) | **378 passed**, 0 failed |
| SM100 (Blackwell) | **1311 passed**, 9 failed |
| SM107 (Rubin) | **731 passed**, 10 failed |

Every failure is **pre-existing and unrelated** — 9 ×
`test_tile_dsl_q_loop_bounds` (a DSL internal compiler error, and that test only
imports `tile_dsl.mask`, which this diff does not touch) plus one
`test_fp8_thd_sliding_window[False-e5m2-d128]` (1 of 3600 LSE elements at 0.034
vs a 0.02 tol). Confirmed rather than assumed: I stashed the diff and re-ran
exactly those node IDs on Rubin — **all 10 reproduce on the baseline tree.**
`black -l 160` clean.

vedaanta (approved):

RomanAnders90:
@cudnn-ci-bot run frost

RomanAnders90:
@cudnn-ci-bot run frost

PR #961, opened 2026-09-09
[FROST] Add sm100 conv forward

zach-ye0:
## Before submitting

- [X] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [X] I ran `pre-commit run` and committed any formatting changes.
- [X] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches (see [root AGENTS.md § Reviewing a PR](https://github.com/NVIDIA/cudnn-frontend/blob/develop/AGENTS.md#reviewing-a-pr-human-or-agent)) and my changes comply, or I explain the exception below.
- [X] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

- FE OSS kernels or CuTeDSL

## Summary

Add FROST convolution forward kernel for sm100. Supported input and output dtypes are: FP32, BF16, FP16, FP8 and NVFP4. An optional unary epilogue function is also supported. All the operations are fused into one kernel.

The basic examples demonstrating the usage of the API can be found in `test/python/conv/frost/examples`. Each of the example file is a standalone script that can be directly run with `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1 python <script_path>`, although these examples fail if the cutlass DSL version is lower than required (see below for minimal version requirements).

Test cases and benchmark files are also added. See the "Benchmark" section below for the benchmark results.

## Why

This PR adds initial support of convolution kernels for FROST engine.

* For NVFP4 input/output dtype, a minimal version of `nvidia-cutlass-dsl` is 4.8.0.
* For the other input dtypes, a minimal version of `nvidia-cutlass-dsl` is 4.9.0, which is not released at the time this PR is created.

## Testing

Test cases are added to `test/python/conv/frost` which should be automatically run with CI, or run manually with:
```
pytest -m "L0 or L1 or L2 or L3 or L4" -- test/python/conv/frost
```

### Setup and methodology

- Benchmarked two sets of dtypes: BF16 inputs and output; NVFP4 inputs with E4M3 block-16 scales and BF16 output. No bias or activation fusion.
- GPU configuration: B200, frequency locked to 1005MHz.
- Environment: cuDNN 9.24.0, frontend 1.29.0, PyTorch 2.14.0+cu132 (CUDA 13.2), CuTeDSL 4.9.0 development build (which is not released for now), Nsight Systems 2026.1.3.
- Fixed configuration for both dtypes: `CONFIG_sm100_128x256x128_128x256x32_cluster2x1_2ctamma`.
- Lower padding `(2,1,1)`, upper padding `(0,1,1)`, stride and dilation `(1,1,1)`.
- 10 warmups and 100 measured iterations, with automatic tensor-buffer rotation for cold-L2 timing. Nsight measures GPU kernel time, excluding graph construction, JIT compilation, allocation, and warmup.
- The cuDNN baseline uses the first supported native plan from heuristic mode A. Speedup is cuDNN time divided by FROST time.
- FLOPs: `2 × (N × Z × P × Q) × K × (C × T × R × S)`. SOL uses 1218 TFLOPS for BF16 and 4874 TFLOPS for NVFP4 at this clock.

### Shapes

| ID | Input N,C,D,H,W | Filter K,T,R,S |
|---|---|---|
| ID1 | (1,1024,1,44,80) | (1024,3,3,3) |
| ID2 | (1,1024,2,88,160) | (1024,3,3,3) |
| ID3 | (1,1024,1,88,160) | (1024,3,3,3) |
| ID4 | (1,1024,4,176,320) | (512,3,3,3) |
| ID5 | (1,1024,1,176,320) | (512,3,3,3) |
| ID6 | (1,512,4,176,320) | (512,3,3,3) |
| ID7 | (1,512,1,176,320) | (512,3,3,3) |
| ID8 | (1,512,4,352,640) | (256,3,3,3) |
| ID9 | (1,512,1,352,640) | (256,3,3,3) |
| ID10 | (1,256,4,352,640) | (256,3,3,3) |
| ID11 | (1,256,1,352,640) | (256,3,3,3) |

### BF16 → BF16

| ID | FROST µs | FROST TFLOPS | FROST SOL | cuDNN µs | cuDNN TFLOPS | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| ID1 | 235.0 | 848.18 | 69.64% | 364.4 | 547.01 | 1.55× |
| ID2 | 1360.4 | 1172.09 | 96.23% | 1570.7 | 1015.18 | 1.15× |
| ID3 | 685.5 | 1163.04 | 95.49% | 794.4 | 1003.62 | 1.16× |
| ID4 | 5413.4 | 1178.19 | 96.73% | 5381.4 | 1185.19 | 0.99× |
| ID5 | 1359.8 | 1172.61 | 96.27% | 1360.3 | 1172.21 | 1.00× |
| ID6 | 2709.6 | 1176.94 | 96.63% | 2705.2 | 1178.86 | 1.00× |
| ID7 | 685.7 | 1162.77 | 95.47% | 691.3 | 1153.31 | 1.01× |
| ID8 | 5422.3 | 1176.27 | 96.57% | 5371.8 | 1187.33 | 0.99× |
| ID9 | 1359.6 | 1172.77 | 96.29% | 1361.4 | 1171.20 | 1.00× |
| ID10 | 2709.4 | 1177.03 | 96.64% | 2656.7 | 1200.38 | 0.98× |
| ID11 | 685.1 | 1163.75 | 95.55% | 691.2 | 1153.48 | 1.01× |

### NVFP4 → BF16

| ID | FROST µs | FROST TFLOPS | FROST SOL |
|---|---:|---:|---:|
| ID1 | 66.4 | 3003.34 | 61.62% |
| ID2 | 355.7 | 4483.12 | 91.98% |
| ID3 | 182.3 | 4372.83 | 89.72% |
| ID4 | 1386.8 | 4599.11 | 94.36% |
| ID5 | 355.7 | 4482.93 | 91.98% |
| ID6 | 709.6 | 4494.34 | 92.21% |
| ID7 | 186.4 | 4277.90 | 87.77% |
| ID8 | 1415.0 | 4507.53 | 92.48% |
| ID9 | 362.9 | 4394.11 | 90.15% |
| ID10 | 752.8 | 4236.13 | 86.91% |
| ID11 | 193.4 | 4121.66 | 84.56% |

The native cuDNN FP4 baseline was unavailable for all 11 shapes; graph planning exposed only the FROST implementation.

### Reproduction

Run from the repository root in an environment with the dependencies above, with an idle B200 locked to 1005 MHz. Confirm that `cudnn` imports from the checkout under test (`python -c 'import cudnn; print(cudnn.__file__)'`). Select the GPU through `CUDA_VISIBLE_DEVICES` if needed.

The successful measurements disabled CPU sampling and context-switch tracing by adding `--sample=none --cpuctxsw=none` to `nsys profile`. To reproduce that setting, add these two arguments to `profile_command` in `benchmark/conv/frost/benchmark_utils.py`; CUDA tracing remains enabled. This avoided a profiler capture-start stall on the benchmark host.

The following commands run all 22 cases directly through the existing benchmark entry points:

```bash
set -euo pipefail
export CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1
config=CONFIG_sm100_128x256x128_128x256x32_cluster2x1_2ctamma

# Each shape is N,C,D,H,W,K,T,R,S, ordered ID1 through ID11.
shapes=(
    1,1024,1,44,80,1024,3,3,3
    1,1024,2,88,160,1024,3,3,3
    1,1024,1,88,160,1024,3,3,3
    1,1024,4,176,320,512,3,3,3
    1,1024,1,176,320,512,3,3,3
    1,512,4,176,320,512,3,3,3
    1,512,1,176,320,512,3,3,3
    1,512,4,352,640,256,3,3,3
    1,512,1,352,640,256,3,3,3
    1,256,4,352,640,256,3,3,3
    1,256,1,352,640,256,3,3,3
)
for benchmark in benchmark_conv.py benchmark_block_scale_conv.py; do
    for shape in "${shapes[@]}"; do
        python -u "benchmark/conv/frost/$benchmark" \
            --shape "$shape" \
            --pre-padding 2,1,1 --post-padding 0,1,1 \
            --stride 1,1,1 --dilation 1,1,1 \
            --configs "$config" \
            --warmup 10 --iters 100 --rotate-buffers auto
    done
done
```

## Summary by CodeRabbit

- **New Features**
  - Added FROST 3D convolution support for dense and NVFP4 block-scaled workloads.
  - Added configurable tile selection, unary epilogues, asymmetric padding, strides, and dilation.
  - Added automatic execution planning for supported convolution graphs.
  - Added examples for standard, fused-epilogue, and block-scaled convolutions.

- **Benchmarking**
  - Added Nsight Systems benchmarks comparing FROST configurations with native cuDNN execution.
  - Added optional correctness checks before timing benchmark configurations.

- **Validation**
  - Added coverage for shapes, data types, epilogues, tile configurations, and block-scaled operations.
  - Added validation for unsupported convolution modes and configurations.


zach-ye0:
@cudnn-ci-bot run frost

python/cudnn/conv/frost/engine.py
```diff
@@ -0,0 +1,642 @@
...
+    def build_plan(self, graph: "pygraph", plan: PlanConfig, ctx: ExecutionContext = None) -> CompiledPlan:
+        analysis = analyze(graph)
+        node = analysis.conv_node
+        image, weight, output = analysis.image, analysis.weight, analysis.output
+        ncdhw = tuple(image.dim)
+        ktrs = (weight.dim[0], weight.dim[2], weight.dim[3], weight.dim[4])
+        pre_padding = _tuple_param(node, "pre_padding", _ZERO)
+        post_padding = _tuple_param(node, "post_padding", _ZERO)
+        stride = _tuple_param(node, "stride", _ONE)
+        dilation = _tuple_param(node, "dilation", _ONE)
```
Adnios, line 613: We should check conv mode here. Current implementation only support cross-correlation mode. Need reject convolution mode.

test/python/conv/frost/examples/00_conv.py
```diff
@@ -0,0 +1,116 @@
...
+    workspace = torch.empty(max(g.get_workspace_size(), 1), device="cuda", dtype=torch.uint8)
+    g.execute({X: X_gpu, K: K_gpu, Y: Y_actual}, workspace)
+    torch.cuda.synchronize()
+
+    torch.testing.assert_close(Y_actual, Y_ref, atol=1e-4, rtol=1e-3)
+    print(f"[00] PASS    conv {shape}")
+
+
+def main():
+    # _FpropCTMKernel currently uses a 256x256 implicit-GEMM output tile and a
```
Adnios:
```suggestion
    # _Fprop Primitive Kernel currently uses a 256x256 implicit-GEMM output tile and a
```
fprop only supports 256x256 tile config?

test/python/conv/frost/examples/00_conv.py
```diff
@@ -0,0 +1,116 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Example 00: 3D convolution forward with BF16 inputs and outputs.
+
+Exercises multiple input shapes, asymmetric padding, and spatial strides.
+Also covers anisotropic dilation; bias is omitted.
+
+Inputs:
+    X: input, in bp16, NDHWC format
+    K: filter, in bp16, NDHWC format
```
Adnios:
```suggestion
    W: filter, in bp16, NDHWC format
```
W is perfered.

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# CTM-style FP16/BF16 implicit-GEMM fprop convolution kernel for Blackwell B200.
```
Adnios:
```suggestion
# sm100 implicit-GEMM fprop convolution kernel.
```
Not this PR, I think this kernel can extend to support fp32/fp8.

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
...
+def _c_smem_swizzle(c_dtype: Type[Numeric], subtile_n: int) -> "cutlass.Swizzle":
+    """Select the epilogue C-SMEM swizzle from the output dtype and subtile width.
+
+    The store packs ``subtile_n`` output channels per row, so the swizzled
+    region spans ``subtile_n * c_dtype.width // 8`` bytes. The tcgen05 SMEM
+    swizzle family is named by that byte width: 32B/64B/128B. Keying off bytes
+    (not a fixed 2-byte assumption) keeps fp32 (4B) and fp8 (1B) outputs
+    correct, not just fp16/bf16. The host C TMA descriptor and the device
+    ``store_swizzled`` both derive from this one value, so they stay in
+    lockstep (a mismatch silently transposes the store; see descriptors.py).
```
Adnios: descriptors.py is inside cutlass dsl. Maybe remove

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
...
+        # Setup A/B/C stage count in shared memory and ACC stage count in tensor memory
+        self.num_acc_stage, self.num_ab_stage, self.num_c_stage = self._stage_counts(self.a_dtype, self.c_dtype, self.smem_capacity)
+
+        # SMEM A/B/C layouts are not materialized on the host: the device
+        # kernel allocates flat cutlass.Array byte buffers sized from
+        # a/b/c_bytes_per_stage and expresses swizzle via the TMA
+        # descriptors + Tcgen05SmemDesc.
+
+
+class _FpropCTMKernel:
```
Adnios: Remove CTM

python/cudnn/conv/frost/engine.py
zach-ye0: Do we also need to export the enum `convolution_mode` to cudnn pakcage? It is now only accessible via `cudnn._pybind_module.convolution_mode`.

test/python/conv/frost/examples/00_conv.py
zach-ye0: No, this comment is outdated. Should have removed it.

python/cudnn/conv/frost/templates/sm100_conv.py
zach-ye0: Indeed. This kernel already supports FP32 and FP8 as the input dtype.

python/cudnn/nodes.py
```diff
@@ -53,6 +53,8 @@ def validate(self) -> None:
 
         if self.node_type == NodeType.MATMUL:
             self._validate_matmul()
+        elif self.node_type == NodeType.CONV_FPROP:
+            self._validate_conv_fprop()
```
Adnios, line 57: Will cudnn-backend reach here? If so, `if x.dim[1] != k.dim[1]:` will reject grouped conv which is supported in cudnn-backend.

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
...
+                # warp_m_idx==0 picks one leader per N-band: cfgA/cfgB (4,1) ->
+                # warp 0 only, stores all mma_tiler_per_cta_m rows; cfgC (2,2) ->
+                # warps 0 and 2, each stores its 64-row band from a per-band SMEM
+                # offset (warp_n_idx*64 rows) to a k_off shifted by the N-band
+                # width (warp_n_idx*128). The (q,p,z,n) pixel coords are shared by
+                # both bands (the (2,2) split is on the K-out channel axis only).
+                if warp_m_idx == 0:
+                    k_off = k_off_base + warp_n_idx * (mma_tiler[1] // warp_n_count) + subtile_idx * subtile_n
+                    band_smem_base = smem_tile_base.subview(warp_n_idx * mma_tiler_per_cta_m * subtile_n)
+                    prims.cp_async_bulk_tensor_global_shared_cta(
```
Adnios: Only one thread issues tmaldg.
```
if prims.elect_sync():
   prims.cp_async_bulk_tensor_global_shared_cta(
```

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
...
+_DEFAULT_LOWER_PAD: Tuple[int, int, int] = (0, 0, 0)
+_DEFAULT_STRIDE: Tuple[int, int, int] = (1, 1, 1)
+_DEFAULT_DILATION: Tuple[int, int, int] = (1, 1, 1)
+_DEFAULT_DTYPE: Type[Numeric] = cutlass.BFloat16
+
+
+def _dense_config_violation(tile_config: ConvTileConfig, channel_bytes: int) -> str | None:
+    """Return why a tile cannot implement dense convolution, if anything."""
+    if tile_config.cta_tile_k_bytes not in (64, 128):
+        return f"CTA K must be 64 or 128 bytes, got {tile_config.cta_tile_k_bytes}"
```
Adnios: This kernel hard code tma swizzle to s128B. If cta_tile_k_bytes = 64, the kernel will hit error? Can you double check？

python/cudnn/conv/frost/templates/sm100_conv.py
```diff
@@ -0,0 +1,1766 @@
...
+                epi_stage_idx = (epi_stage_idx + 1) % num_c_stage
+
+                # TMEM -> RMEM: per-warp LDTM. The column origin advances ONLY by
+                # subtile_n and carries NO warp_n term: tcgen05_ld routes by the
+                # executing warp's sub-partition, so each warp lands on its own
+                # N-band without an explicit column add. An explicit warp_n column
+                # offset would over-shoot into unwritten TMEM (reading back zero).
+                # The global k128..255 offset is applied on the STORE side
+                # (band_smem_base + k_off), not here.
+                tmem_ctm = prims.make_tmem_ptr_from_warp_row_col(
```
Adnios: remove ctm

python/cudnn/conv/frost/engine.py
Adnios: For this PR, we can leave `TODO` comments, and address in following PRs.

python/cudnn/conv/frost/templates/sm100_conv.py
zach-ye0: Fixed. The kernel does not hit an error with k = 64, but the results are incorrect. Test cases did not reveal this bug because heuristic selection always chooses k = 128.

Adnios (approved):

python/cudnn/conv/frost/engine.py
```diff
@@ -0,0 +1,669 @@
...
+            except KeyError as exc:
+                raise ValueError(f"frost_conv: tensor uid {exc} is bound by the kernel but is not an operand of this graph") from exc
+
+        operands = variant_pack.operands(self._indices)
+        for operand, (role, shape, stride, dtypes, required_bytes) in zip(operands, self._expected):
+            actual_shape = tuple(int(x) for x in operand.shape)
+            actual_stride = tuple(int(x) for x in operand.stride())
+            if actual_shape != shape or actual_stride != stride:
+                raise ValueError(
+                    f"frost_conv: {role} must have physical shape {shape} and stride {stride}; " f"got shape {actual_shape}, stride {actual_stride}"
```
YangXu1990uiuc, line 443: [P1] Reconcile the scale-buffer checks with current develop's variant-pack normalization

This fails after merging current develop (`b2712de2d`, including #1035) into this head. `_normalize()` now calls native `describe_from()`, which redescribes dense buffers using their graph declarations. With example 02's metadata, the resulting SFA is `(1, 8, 1, 12, 16)`, whereas this check requires `(192, 8, 1)`; SFB changes from `(55296,)` to `(192, 8, 3, 3, 3)`, and SFD is also redescribed to rank 5. Thus the NVFP4 example raises before launching.

I reproduced this without a GPU using the actual `variant_pack.cpp` from the merge and this plan's actual `execute()` checks, with only the compiled launch stubbed: the original metadata passes, then `describe_from()` changes slots 2/3/5 and execution raises at SFA.

Please test the NVFP4 graph against the merged develop path and reconcile the ABI: ordinary unswizzled scales can be rebound through checked zero-copy views, while reordered SFB needs its actual physical capacity preserved for the padded blob check. Merely weakening this shape check would leave the compiled callable's argument layouts inconsistent.

python/cudnn/conv/frost/templates/epilogue.py
```diff
@@ -0,0 +1,178 @@
...
+    upper_clip = kwargs.get("upper_clip")
+
+    def op(x):
+        zero = _full_like(x, 0.0)
+        if negative_slope is None:
+            out = cute.math.max(x, _full_like(x, 0.0 if lower_clip is None else lower_clip))
+        else:
+            out = cute.math.max(x, zero) + negative_slope * cute.math.min(x, zero)
+            if lower_clip is not None:
+                out = cute.math.max(out, _full_like(x, lower_clip))
```
YangXu1990uiuc, line 123: [P2] Apply the ReLU slope relative to lower_clip

With both `negative_slope` and a nonzero `lower_clip`, this computes leaky ReLU around zero and clamps afterward. cuDNN instead applies the slope below `lower_clip`, around that threshold: `lower_clip + slope * (x - lower_clip)` for `x < lower_clip`, then applies the upper clip. The Python binding forwards these attributes directly to that backend definition ([reference](https://docs.nvidia.com/deeplearning/cudnn/backend/latest/api/cudnn-graph-library.html#cudnn-backend-pointwise-descriptor)).

For the existing test's `lower_clip=-0.05, negative_slope=0.2, upper_clip=0.10`, `x=-0.25` should produce `-0.09`, but this produces `-0.05`; `x=-0.025` should remain `-0.025`, but becomes `-0.005`. Both dense and block-scaled convolution use this helper. I evaluated the exact closure with CPU arithmetic to verify these examples.

Please fix the epilogue and the independent reference: `_relu_reference()` currently repeats the same formula, so the existing `relu-leaky-clipped` case cannot detect this mismatch.

python/cudnn/conv/frost/engine.py
```diff
@@ -0,0 +1,669 @@
...
+            )
+
+        _check_output_shapes(expected_output, (("output", output),))
+
+        if node.compute_data_type not in (None, cudnn.data_type.FLOAT):
+            raise NotImplementedError("frost_conv requires FP32 compute")
+        if analysis.epilogue_node is not None and analysis.epilogue_node.compute_data_type not in (None, cudnn.data_type.FLOAT):
+            raise NotImplementedError("frost_conv requires FP32 unary epilogue compute")
+        _check_fp32_compute_type(analysis.context)
+        if analysis.epilogue_node is not None and analysis.context.intermediate_data_type != cudnn.data_type.FLOAT:
```
YangXu1990uiuc, line 286: [P2] Validate the convolution intermediate tensor's resolved dtype

`context.intermediate_data_type` is a default, not the dtype of every virtual tensor. A graph with a FLOAT context can explicitly call `conv.set_data_type(BFLOAT16)` before `graph.ceil(conv)`; the actual analyzer and `check_support()` accept this graph, but the template applies the epilogue directly to its FP32 accumulator and skips the declared BF16 rounding.

A deterministic example uses BF16 1x1x1-convolution operands with two nonzero products, `1 + 1/512`. The accumulator is `1.001953125`: rounding the declared intermediate to BF16 before `ceil` yields **1**, while the current fused calculation yields **2**. This is an observable semantic difference, not an output tolerance issue.

Please validate `node.outputs["Y"].data_type == FLOAT` for unary fusion (as the block-scale path already does), and add a case where a tensor explicitly overrides the graph default. Source-level acceptance and the CPU numerical counterexample were verified; no GPU kernel execution was performed.

YangXu1990uiuc (approved):
The overall convolution integration and separation of graph analysis, plan validation, and templates make sense. Approving the direction, with three concrete follow-ups below to address before landing: compatibility with current develop's scale-buffer normalization, clipped/leaky ReLU semantics, and explicit intermediate dtype overrides.

Reviewed head `57b8bd4bd`, including an isolated merge with develop `b2712de2d`. Validation was source review plus CPU/native metadata probes; I could not run SM100 kernels locally. Please rerun the relevant NVFP4 and epilogue cases on the merged head. Existing review threads are not repeated here.

YangXu1990uiuc:

The recorded FROST run is for the older `3672c24` head. I am starting one for `57b8bd4bd`; the following standalone `@cudnn-ci-bot run frost` command is also a Codex bot action.

The current FROST target does not select `test/python/conv/frost`, so its result will cover shared-engine regressions, not these new convolution kernels. Please also run `pytest -m "L0 or L1" test/python/conv/frost` on SM100 with the appropriate DSL version (4.8 for NVFP4; 4.9 for dense convolution), and retain that coverage in CI. The L1 selection matters for the new NVFP4 correctness cases.

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc (approved):

Reviewed `57b8bd4bdadc` against base `a3f82cdf9a7b`.

Incremental follow-up on unchanged head 57b8bd4bd. The three previously reported findings remain unresolved; linked below rather than duplicated. The NVFP4 integration failure is now confirmed through the public graph API on SM100. No P0 found; owner retains responsibility for addressing the comments and deciding when to merge.

- **[P1] Still open:** [Reconcile the scale-buffer checks with current develop's variant-pack normalization](https://github.com/NVIDIA/cudnn-frontend/pull/961#discussion_r4007707202).

- **[P2] Still open:** [Apply the ReLU slope relative to lower_clip](https://github.com/NVIDIA/cudnn-frontend/pull/961#discussion_r4007707221).

- **[P2] Still open:** [Validate the convolution intermediate tensor's resolved dtype](https://github.com/NVIDIA/cudnn-frontend/pull/961#discussion_r4007707229).

Validation: Isolated merge with develop a3f82cdf9 succeeded. The normalization, graph and dispatch contracts relevant to the prior review are unchanged from develop b2712de2d. Same SM100 device and dependencies, two existing public-graph cases: test_block_scale_conv_correctness[implicit_identity-bf16] and [implicit_identity-fp4-e2m1] both PASS on the unmerged head and both FAIL after merging current develop. Both failures are at execute-time SFA validation: expected (192, 8, 1), received (1, 8, 1, 12, 16). L1 was explicitly selected. Four selected host version-gate cases passed. Native metadata probe reconfirmed that normalization changes SFA/SFB/SFD layouts; the plan rejects the normalized SFA before launch. CPU probes on the current merged source also reconfirmed the clipped/leaky-ReLU formula mismatch and acceptance of explicit BF16/HALF convolution intermediates despite applying ceil directly to FP32. Current-head Style check passed. Started and verified a current-head FROST test pipeline; it is still running. Posted the additional convolution test command and DSL requirements in the CI-assistance comment.

Limitations: No successful merged-head GPU correctness or performance claim: the NVFP4 graph stops before launch, so timing was not performed. Dense convolution GPU execution requires DSL 4.9 and was not run locally. The current FROST CI target does not select test/python/conv/frost. A green result there would not validate these new kernels; explicit SM100 convolution coverage with L0 or L1 and the applicable DSL is still required.

YangXu1990uiuc:

CI follow-up for `57b8bd4bd`: the started SM80/SM100/SM103/SM120 SDPA, linear/GEMM, build and C++ sanitizer jobs passed. The remaining failures are the guardword scan (`nv-` matched inside `Conv->GEMM`) and the Rubin runner running out of disk space during setup, before tests ran.

This still does not cover `test/python/conv/frost`; please retain the explicit convolution and current-develop integration checks discussed in the review. No duplicate pipeline was started.

zach-ye0:
@cudnn-ci-bot run frost

PR #974, opened 2026-09-10
frost(sdpa): Rubin DSv3 quantized flavors, THD on every f16/FP8 flavor, and arch-grouped forward kernels

RomanAnders90:
## Summary

Rubin (SM107) SDPA forward: adds the **DSv3 (d_qk=192 / d_v=128) quantized flavors**, brings
**THD/varlen to every f16 and per-tensor FP8 flavor**, groups the forward kernels by
architecture, and closes two coverage gaps found on the way. Eleven commits, each
independently reviewable.

Branched from `develop` @ `60cb6cda` (the SM107 port, #954).

## What changed

**1. `kernels/sm<arch>/` — the forward kernel directory is grouped by architecture.**
26 flat files whose only grouping was an `_sm###` filename suffix become four packages:
`kernels/prefill_d128_fp8_sm107.py` → `kernels/sm107/prefill_d128_fp8.py`. Arch is the one
coverage axis that earns a directory — the sets are disjoint (no two arch packages ever run on
the same device) and it makes ownership physical. Every other axis stays in the filename.

Modules shared ACROSS arch lines stay at `kernels/` level so the directory a file sits in always
names its only owner: `_common_sm100.py` → `_common_blackwell.py` (of its 21 importers, 10 are
SM107 — the old name was already a lie), `thd_helpers.py` unchanged.

No kernel body changed. `frost/README.md` rule 10 said coverage axes go in filenames "never in
directory depth", so per rule 13 the contract moves with the code.

**2. DSv3 quantized on Rubin.** Two kernels that are the d128 bodies with a different config
factory — the relationship `prefill_d192_d128_f16` already has to its own d128 sibling. Two
things were not config:

- The row-sum "ones tile" borrowed **Q's** swizzle. At d_qk=192 the FP8 Q swizzle drops to 64 B
  (192 is not a multiple of 128), narrower than the sum-MMA's 128 B K extent. It now derives its
  own constants; re-swizzling is free because an all-ones tile is invariant under byte
  permutation, so only the WIDTH is load-bearing.
- `DESC_VERSION` was audited by summing slabs rather than copied. That found MXFP8 at cga1 puts
  all four scale-factor tiles past the 256 KiB version-0 descriptor window — the configuration
  that shipped `LSE=+inf`/`O=NaN` on the d512 sibling. (192,128) therefore inherits `cgas={2}`.

**3. THD/varlen on all eight f16 and per-tensor FP8 Rubin flavors.** The premise that this needed
a *port* was wrong: Blackwell f16 THD and Rubin FP8 THD already worked and share every helper the
Rubin f16 kernels use. Six pieces of the FROST THD contract had simply never been adopted, and
one was doing the damage — the prologue K load passed raw `batch_idx` instead of the
THD-collapsed `tma_batch`, with no `kv_seq_off`. The packed batch extent is 1, so `batch_idx=1`
is OOB and TMA zero-fills: K came back all zeros, and `LSE = log(S_kv)`. For sequence 0 both
forms coincide, so it was exactly invisible there. The other five (dynamic packed extents, the
persistent claim-counter scheduler, the dead-unit O-store guard, the clamped runtime K/V
descriptors, per-sequence Q lengths) are each real and each described in their commit.

The mechanical part is `frost_dev/port_thd_contract.py`; every edit asserts its anchor, so a body
whose idioms differ fails loudly instead of half-porting.

**4. A descriptor-window hazard found in review, and dense GQA/MQA qualified.**
d192 FP8 at cga1 with a half-precision O put the ones tile at 272 KiB; the guard is now *derived*
from the allocator's own constants rather than argued in prose. Separately,
`graph_analyzer.expand_gqa_heads` turned out to be dead code claiming dense GQA was unqualified —
GQA is native (`qh_per_kh` is threaded to the kernel) and now has ratio × shape coverage
including MQA across all three dtype families.

## Validation

- **Rubin (`w2u1g-lc-0030`, cc 10.7):** whole suite **1111 passed / 10 failed**, against
  **829 / 11** on unmodified `develop`. The 10 are pre-existing (9 ×
  `test_q_loop_bounds_match_reference`, 1 × `test_torch_ops::test_padded_seq_lens`) and were
  verified to reproduce by running the same selection at `60cb6cda`.
- The SM100 f16 THD suite runs on Rubin for the first time: **215 passed** (previously skipped in
  its entirety). Per-tensor FP8 THD: **43 passed / 0 failed**.
- Per-flavor THD attribution sweep (B=1/2/3 × none/causal/bottom-right): **9/9 per flavor**.
- Dense GQA/MQA: **33/33** (3 ratios × 4 shapes × 3 dtype families).
- **A100:** 456 passed / 0 failed across `sdpa/`, `fe_api/sdpa/`, `test_import_boundaries.py`,
  `test_dispatch.py`.

## Performance

Full 121-case sweep on the locked-clock Rubin node (2376 MHz), single commit.

**DSv3 quantized went from `unsupported` to served** — all ten FROST cells previously reported
*no FROST OSS engine serves this fwd graph*:

| dtype | 2K | 4K | 8K | 16K | 32K |
|---|--:|--:|--:|--:|--:|
| FROST fp8 | 2596 | 4547 | 5850 | 6515 | **6454** |
| backend fp8 | 4017 | 5554 | 6311 | 6755 | **6455** |
| FROST mxfp8 | 2428 | 3992 | 5075 | 5720 | 5867 |
| backend mxfp8 | 3929 | 5446 | 6217 | 6661 | 6385 |

TFLOPS, B=1 H=128 D=192/128 no_mask. FP8 reaches parity at 32K and closes monotonically with S;
the deficit is concentrated at small S, the fixed-overhead-dominated regime.

**The arch move is perf-neutral, with a control.** Across the 100 cells present in both this
sweep and the pre-refactor one: FROST median **+0.00 %** (min −1.09 %, max +0.90 %, n=50). The
closed backend, whose code this PR never touches, spans −1.48 %…+0.73 % over the same cells — so
FROST's movement is inside the rig's own noise.

## One tolerance change, characterised not widened

`test_fp8_thd_sliding_window[e5m2-d128]` was red on Rubin before this branch. Measured over 5
seeds: the error scales ~2× from e4m3 to e5m2, exactly the one-mantissa-bit difference (3 bits →
1 ulp ≈ 6.25 % relative, 2 bits → ≈ 12.5 %), and is deterministic — quantization, not a kernel
bug. e4m3 was also over the raw `atol` and survived only on `rtol×|ref|`. The bound is now
per-format and single-sourced (`_LSE_ATOL_RTOL`), with the measured table beside it; only the one
site that needed it moved, since a passing tight test is free information.

## Not in this PR

**MXFP8 THD.** Its blocker is different in kind from the other two lines: the scale-factor
tensors need their own packed-THD layout, a data-layout problem rather than the ABI plumbing f16
and FP8 needed. The row declines it rather than half-serving, and `SUPPORT_MATRIX_TRACKER.md`
states it.

Also unchanged: `sdpa/bwd/kernels/` keeps its arch suffixes (out of scope for the move), and the
Rubin scheduler domain stays NATURAL-only.

## Summary by CodeRabbit

- **New Features**
  - Expanded Rubin/SM107 SDPA support for D192×D128 FP16/BF16, FP8, and MXFP8 workloads.
  - Added THD/varlen execution for supported Rubin FP16/BF16 and FP8 shapes, including packed layouts and dynamic sequence lengths.
  - Added KV-split result combination with optional LSE and amax outputs.
  - Added support for additional GQA/MQA ratios and native quantized kernel shapes.

- **Bug Fixes**
  - Improved handling of masked, empty, and padded sequences for more reliable outputs.

- **Documentation**
  - Updated kernel layout, naming, and usage references to match the architecture-based organization.


RomanAnders90:
@cudnn-ci-bot run frost

vedaanta (approved):

RomanAnders90:
Bot review addressed in 2025e1af. Dispositions, one per finding:

**Fixed — one real bug, and it was mine (Major: `check_support` ignores `supported_cgas`).**
Correct. When I extracted `supported_cgas_for()` in the previous round I replaced both the block that
computed the value *and* the `_value_error_if` that consumed it, and only restored the assignment. So an
explicit `cga=1` on the Rubin quantized d192 path passed support validation and died inside `compile()` —
exactly the [Rule 8b'](AGENTS.md) standalone-wrapper failure the helper was extracted to prevent. Restored,
plus a test that asserts `check_support` still *consumes* the domain: the value itself was never wrong, so a
value-only assertion would not have caught this.

**Hardened — the two `eff_seqlen_q` criticals (d512 f16 / d512 fp8).**
The structural reading is right: in `_compute_warp_group` the only binding is inside the masked `else`, so at
`MASK_FLAGS == 0` the name is formally unbound where the unmasked sg0 loop passes it. It is **not reproducible
as filed** — dense no-mask d512 compiles on Rubin today and `test_fp8_d512_masks` is 8/8 including both `none`
variants, so the DSL preprocessor is binding it. Depending on how a `const_expr` branch's locals are treated is
not something this kernel should rely on, so both arms now bind it explicitly. At `MASK_FLAGS == 0` the
per-sequence length *is* the scalar, so the code now says that rather than depending on grace.

**Fixed — three doc/comment findings.** Footnote v still called f16 THD unported while footnote z said it was
served everywhere, and the gaps table still listed f16/bf16 forward THD as missing (my drift from widening the
rows in place). The `SM107_FP8_THD_SHAPES` comment claimed d256/d512 do not qualify, contradicting the constant
beside it. And the arch-directory rename left two pre-upstream filenames in benchmark-path comments; the bot
flagged one, the other was the same class, and a check that every `python sm*/prefill_*.py` comment resolves to
a real file now passes.

**Declined — ruff E741 on `O` in `test_sdpa_fwd_mxfp8_sm100.py`.** No ruff config is tracked in this repo, and
the file uses `O, O_ref` in 33 places. Renaming only the line I added would break the file's own convention for
a rule nothing enforces here.

---

**Re: the SM80 CI job — it is not from this PR.**

The job URL is not readable from here (GitLab returns 302 to SSO), so I reproduced it locally on the A100 and
A/B'd it:

- The failures at 30–32 % map to `test_conv_fuzzer.py`. A band sample gives two:
  `test_conv_random_L0[t6_N1_C8x144_g2_16x64_f16_2d_wg_none]` and
  `test_conv_random_L0[t34_N1_C144x256_g8_40x8_f32_2d_wg_none]`.
- Checked out unmodified `develop` (60cb6cda) and ran the same two: **both fail identically.** Pre-existing.
- Diff scope corroborates: this PR touches 46 files under `python/cudnn/sdpa`, 2 docs under `python/cudnn/frost`,
  8 under `test/python/sdpa`, and `test_import_boundaries.py` — **zero** conv or matmul files.

Full FROST validation on this branch: A100 388 passed / 0 failed; Rubin d256+d512 FP8 77 passed / 14 skipped.

Issue #981, opened 2026-09-10
sdpa fp8 forward at d=256 returns wrong output instead of declining (Qwen3.5 geometry)


vedaanta:
## Summary

fp8 SDPA **forward** at `d_qk = d_v = 256` (Qwen3.5 head geometry: 32 q-heads, 2 kv-heads, d=256) returns **wrong output** rather than declining: ~0.1–0.3% of the O elements are off by up to ~1.9 absolute (tolerance 0.08), deterministically, on SM100 and SM103.

```
Mismatched elements: 58323 / 20185088 (0.3%)
Greatest absolute difference: 1.875 (up to 0.08 allowed)
Greatest relative difference: inf
```

## Context

The generic fp8 context suite caps `d` at 192; d=256 fp8 forward was previously only reachable via this model preset. fp8 **backward** does draw d=256 elsewhere, so the forward-only nature is notable.

## Expected

If d=256 fp8 forward is unsupported, `check_support` should reject the graph (the harness then skips). Producing several-percent-wrong O silently is a correctness bug.

## Workaround in the test suite

The `sdpa/suites` model fp8 flavor is now gated to `d <= 192`, so Qwen3.5 gets no fp8 flavor until this is resolved.

Issue #980, opened 2026-09-10
frost(sdpa): f16 THD decode at d_qk=192 throws tvm-ffi 'Out of bound k_tensor.strides[0]' instead of serving/declining


vedaanta:
## Summary

f16 **THD (ragged) decode** with `d_qk=192` (DeepSeek-V3 / Kimi-K3 head geometry: 96–128 heads, d_qk=192, d_v=128) under FROST opt-in fails at execute with a tvm-ffi stride bound error from the decode kernel:

```
E   ValueError: Out of bound k_tensor.strides[0] on argument #1 when calling:
    `_host(q_tensor: Tensor([1, n0, 128, 192], float16), k_tensor: Tensor([1, n1, 128, 192], float16), ...)`
```

(`n0`/`n1` are packed token counts; s_q=1 decode.)

## Impact

Reddens the model generation suites for the d=192 presets (dsv3, kimi_k3) whenever the decode draws a packed-THD layout. Padded (dense) decode at the same d=192 is fine — the fault is specific to the ragged/THD path's stride into `k_tensor`.

## Expected

Either serve d=192 packed-THD decode, or decline in `validate()` so the native path serves it — not a mid-execute `ValueError`.

## Workaround in the test suite

The `sdpa/suites` model generation knobs now draw padded-only decode for presets with `d_qk > 128` (mirrors the existing fp8 THD gate), so this combo is not exercised until the kernel is fixed. Removing that gate re-exposes it.

Issue #979, opened 2026-09-10
frost(sdpa): sm107 prefill kernel raises NotImplementedError on padded seq_len_q instead of declining in validate()


vedaanta:
## Summary

On SM107 (Rubin), with the FROST engines opted in (`CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`), the just-merged prefill forward kernel (`sdpa_fwd_prefill_sm107`, #954) **raises `NotImplementedError` at execute time** for any graph with per-batch (padded / varlen) `seq_len_q` shorter than the padded `S_q`, instead of declining during validate so the native backend serves it.

```
cudnn/sdpa/fwd/engines.py:1296: in _execute
    raise NotImplementedError(
E   NotImplementedError: sdpa_fwd_prefill_sm107: per-batch seq_len_q shorter than S_q=1108 is not plumbed
    (no dense padded-Q trim in this kernel); got min 231
```

## Impact

Because the engine is a candidate (opted in) and throws at execute rather than declining, the whole graph fails — even though the native backend supports padded/varlen prefill on SM107. In the new `sdpa/suites` CI lane this reddened ~75 configs across `context.f16`, `generation.f16`, and the model suites on SM107.

## Expected

A FROST engine that cannot serve a shape should **decline in `validate()`** (return not-eligible), so the plan list falls through to an engine that can (native, or another FROST row). Raising `NotImplementedError` from `_execute` violates the "decline cleanly, never throw from a selected engine" contract (python/cudnn/sdpa/AGENTS.md).

## Repro

Any prefill config on SM107 with a padding mask (per-batch `seq_len_q < S_q`) under FROST opt-in. Minimal: `b=2`, `S_q=1108`, `seq_len_q=[231, 900]`, f16, FROST enabled.

PR #971, opened 2026-09-09
test(sdpa): budget fp8 midpoint flips per d-row; widen e5m2 d512 envelope (sm103 lane)

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and my changes comply.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`.

## Affected area

- Tests (`test/python/sdpa`, `test/python/sdpa/frost`), `AGENTS.md`

## Summary

`frost-sdpa:cutlass-rel:sm103` (GB300) has failed the same three fp8 tests on every run since 2026-09-04 while `sm100` (B200) passes on the same cuDNN 9.27 build: `test_fp8_d512_thd[True-e5m2]` (0.0718 > 0.07, #879), `test_sdpa_fp8_fwd_L0[test69]` (1 element of O off by 0.225) and `test_sdpa_fp8_bwd_L0[test310]` (all 192 elements of one dK row off by 0.25).

None is an sm103 kernel defect. torch's CUDA Philox lays draws out by grid size, which follows the SM count, so the same seed produces a **different dataset** on GB300 (152 SMs) than on B200/B300 (148). Feeding the GB300 tensors into the B300 bwd kernel reproduces test310 bit-for-bit. Each failure is an fp8 rounding-boundary flip that dataset happens to contain:

- **test310**: one `dS*dP_scale` within 1e-5 relative of an e4m3 midpoint. The `inject_negative_score_rows` q rows all share one dense q vector, so their `dS(i, j)` terms into dK row j add coherently and the flip moves all 192 elements of that row (the reference recomputed with `P*(1+1e-5)` reproduces the kernel row exactly), blowing the 49-element budget. `assert_close_fp8_grad` now also budgets affected d-rows (1e-5 of rows, at least 1).
- **test69**: e5m2 `P*s_scale = 1407.57`, 0.498 of an e5m2 step (one step is 25% of P) times |v| = 2. The forward O compare now uses the same budget.
[...]

PR #982, opened 2026-09-10
test(sdpa/suites): fix CI failures from the first full MR !2367 run (mxfp8 amax, d-envelope gates, decode OOM)

vedaanta:
Follow-up to #846: fixes the failures surfaced by the first full run of the GitLab `sdpa` CI lane (MR !2367) on the blocking docker arches (sm100/sm120). The sm103/sm107 cluster cells are already `allow_failure` (new-arch bringup), so only these were pipeline-blocking.

## Suite-side fixes

- **mxfp8 THD output-amax tolerance 0.02 → 0.05** (`mxfp8.py`), matching the dense mxfp8 path (`compare_amax(rtol=0.05)`). Block-scaled fp8 inputs give the peak O element a few-percent spread between the kernel's fp32 accumulation and the reference; the tighter 2% was inconsistent with the dense path and flaked ~11 `context.mxfp8.thd` configs on sm100/sm103.
- **Model THD decode gated to d≤128** (`knobs.py`), mirroring the existing fp8 `thd_ok` gate — dsv3/kimi_k3 (d=192) decode is padded-only now. d=192 THD decode trips a FROST decode stride bound (#980).
- **Model generation batch capped at 8** (was 32) — a 96–128-head d=192 padded decode's dense KV cache is >20 GiB at b=32, s_kv=8192 and OOMed the 40 GiB A100 (sm80). Real decode batch for models this size is small; the tail added no coverage.
- **fp8 model flavor gated to d≤192** (`registry.py`) — Qwen3.5's d=256 fp8 forward returns wrong output instead of declining (#981), so it gets no fp8 flavor.

## Backend / FROST bugs the suite surfaced (filed, not fixed here)

- #979 — sm107 prefill kernel raises `NotImplementedError` on padded `seq_len_q` instead of declining in `validate()`.
- #980 — f16 d=192 THD decode throws a tvm-ffi `strides[0]` bound error.
- #981 — fp8 d=256 forward returns wrong output instead of declining.

Collection 7,621 → 7,597 (Qwen3.5's three fp8 model suites removed). COVERAGE.md regenerated.

## CI result (GitLab pipeline 67094154, dev backend, with these fixes synced)

Blocking docker arches: **sm80 4/4 green, sm120 9/9 green, sm100 8/9** — the remaining sm100 red was `context:mxfp8` at **2 tests** (down from 11): THD `test41`/`test147` at 5.4% / 6.4% amax deviation, deterministic (identical on sm103). e4m3 P carries 3 mantissa bits, so one half-ULP of the dominant P term already moves the peak O element by ~6%; the per-element compare right above allows `rtol=0.20`, so the amax (their max) is now bounded by the same tolerance, THD and dense alike (the dense path had the same 5% bound and the same failure class on sm107). Cluster-cell reds (sm103/sm107, `allow_failure`) are the known classes: #979 sm107 prefill decline, fp8 gradient midpoint flips (#971), cuBLAS-concurrency launch failures.

## Summary by CodeRabbit

- **New Features**
  - Added broad SDPA test coverage for FP16, FP8, and MXFP8 across dense, ragged, chunked, paged, context, generation, and backward scenarios.
  - Added model-oriented coverage for LLaMA, Qwen, GPT-OSS, DeepSeek, and Kimi attention configurations.
  - Added single-configuration reproduction support for easier issue investigation.

- **Documentation**
  - Added a generated coverage guide cataloging 56 suites and 7,592 configurations.

- **Bug Fixes**
  - Improved handling and validation of ragged sequence capacities, zero-length sequences, and declared sequence totals.


YangXu1990uiuc (approved):

Reviewed `8cff5a49f3ae` against base `2a0343f079b0`.

Capping decode test memory and revisiting MXFP8 reference tolerance are reasonable follow-ups to #846. Two nonblocking points remain: independently check the written amax, and retain regression coverage for the dimensions fixed since this branch was created.

**[P2] Check the amax output independently of reference error** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/8cff5a49f3ae21cf8a18a6bd3ccc1bd8e1c7c8ef/test/python/sdpa/mxfp8.py#L895))

`compare_amax(o_gpu, o_ref)` reduces the two O tensors; it never reads the kernel-written `amax_o_gpu`. This means even a missing or NaN amax output passes this assertion. While adjusting the tolerance, add an independent comparison between `amax_o_gpu` and the peak of the actual O values, allowing for output rounding; keep the wider reference-error bound separate. The THD path can use the same invariant over live rows.

Evidence: On SM100, dense `test373` passed normally. In a separate negative control, replacing only `amax_o_gpu` with NaN immediately before this assertion still produced 1 passed. The route was the native MXFP8 backend. Attribution: Pre-existing harness blind spot, exposed while reviewing the assertion changed by this PR; not a demonstrated kernel defect.

**[P2] Retain coverage for dimensions whose defects are now fixed** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/8cff5a49f3ae21cf8a18a6bd3ccc1bd8e1c7c8ef/test/python/sdpa/suites/knobs.py#L69))

The unconditional D192 THD-decode exclusion here and D256 FP8 exclusion in `registry.py` now outlive their original bugs: [#984](https://github.com/NVIDIA/cudnn-frontend/pull/984) fixes #980, and [#985](https://github.com/NVIDIA/cudnn-frontend/pull/985) fixes #981; both are in current develop. On rebase, retain those model/regression cases for the fixed routes. If older backends or other architectures still need exclusions, scope them to that version/route instead of removing the entire dimension family.

Evidence: Verified both merged fix commits are ancestors of current develop. The new gates remove these cases for every architecture/backend, including the fixed SM100 routes. Attribution: PR-added coverage exclusions became stale after their runtime fixes merged.

Validation: Collected 7,597 tests under the default L0 selection. Generated and round-tripped all 7,592 randomized configurations across 56 suites; generated COVERAGE.md matches. 14 focused SM100 tests passed on cuDNN 9.25.1, including THD test41/test147, empty-Q, bottom-right/band/sink, slack capacities and f16 backward. Of 15 executed graphs, 13 used FROST and 2 used native backends. The separate NaN negative control demonstrates the amax test gap above.

Limitations: The current develop merge has conflicts in mxfp8.py, random_config.py, COVERAGE.md, knobs.py and registry.py. Please rebase and rerun the affected CI lanes; I did not start CI on the conflicting merge. No full cross-architecture sweep or new Rubin run. These are test-only changes, so CPU hot-path and GPU performance comparisons are not applicable.

PR #985, opened 2026-09-10
frost(sdpa): fix d256 fp8 + mxfp8 masked-softmax TMEM race — aliased P store vs the other warpgroup's S load (#981); mhas_v2 fp8/mxfp8 fwd draw d=256

### LINKED ISSUE #981: sdpa fp8 forward at d=256 returns wrong output instead of declining (Qwen3.5 geometry)

vedaanta:
Fixes #981.

## Bug

`sdpa_fwd_prefill_sm100_fp8` at d=256 returned wrong output under any causal mask (TOP_LEFT or BOTTOM_RIGHT, e4m3 or e5m2, MHA or GQA, b≥1): ~0.1% of O elements off by up to ~2.0, sporadic and data-dependent. No-mask configs were always correct.

`sm100/prefill_d256_fp8.py` packs fp8 **P into the tail columns (96–127) of the same TMEM score slot** it was computed from (`P_EVEN_OFF = 96` inside the 128-column S slot — there is no spare TMEM at d=256). That is fine for a single owner (the d256 f16 kernel and this kernel's no-mask fast path load all of S before storing P). The two **masked** variants, however, split each row's 128 keys across the two softmax warpgroups: half 0 loaded everything, published the row max, and immediately stored *its* P (keys 0–63) into cols **96–111 — half 1's still-unread S for keys 96–111**. Whenever half 1 lagged (it can still be finishing the previous tile), it read clobbered scores and produced garbage weights for exactly keys 96–111.

Diagnosed by recovering the corrupted keys from `O_gpu − O_ref` against the sparse-integer `V`: the bad contributions cluster at `key mod 128 ∈ [96, 111]`.

## Fix (fp8)

Add an explicit release, mirroring `mb_softmax_max`: a new per-parity mbarrier `mb_softmax_hi_loaded`. Half 1 arrives it once its S chunk (keys 64–127) has landed in registers (`tcgen05.wait::ld` first — the softmax paths otherwise never wait on loads explicitly), and half 0 waits on it immediately before its first P store, in both masked variants. The fast path, the MMA warp and the correction warp are untouched.

Why not just swap which half owns which keys (so the half storing into cols 96–111 is the one that already waits on the max)? Because alpha/stats is aliased too — `stats_addr` is **S column 0** — so the half that stores alpha must own key 0. The original ownership is forced; only the ordering was missing. (Tried the swap first: it moved the corruption to key 0 instead of removing it.)

## Same hazard in the d256 **mxfp8** kernel (second commit)

Audit of every kernel that aliases P into the score slot (the six d256 kernels): sm107 d256 fp8/mxfp8 and both d256 f16 kernels are single-softmax-owner, the d512 kernels run one softmax warpgroup, and sm107 d128/d192 split the softmax but have dedicated P columns — none can race. `sm100/prefill_d256_mxfp8.py` can, and hides the split behind a different role: its `FUSED_CORR_SPLIT_P` schedule (mxfp8 + strict top-left causal, including padded and left-window) hands keys 64–127 to the **correction** warpgroup (`_fused_p1_step`), and half 0 stored P for keys 0–63 into cols 96–111 with nothing ordering it after that warpgroup's S-hi load. (The kernel's dense split-P schedule is safe: both halves meet at a 256-thread named barrier after their loads.)

Random data hides it: the fused warpgroup only lags when it rescales O, and `RESCALE_THRESHOLD` (4 log2 units) makes that rare — 8 random causal/SWA/BR configs (~90M elements, S up to 4096) were clean. Scaling K by 8 per 128-key tile forces a rescale every iteration, and then **every run** produced inf/NaN O and `max|O−ref| ≈ 2` (12/12 fused-causal runs), while the dense control stayed at that input's ~0.13–0.16 quantization floor.

Fix: the release already exists — `_fused_p1_step` arrives `mb_stat_empty` right after `tcgen05.wait::ld` on S-hi — so under `FUSED_CORR_SPLIT_P` half 0 now waits on it immediately before its first P store instead of at the end of the iteration (all three loop variants). No new barrier; non-fused schedules untouched.

## Verification (B200, cuDNN 9.26, FROST opted in)

fp8:
- All previously failing repros pass: exact CI config family (Qwen3.5 geometry, BR causal, s=1013 padded strides), packed strides, s=384, s=1024, e4m3, TOP_LEFT causal, MHA (was 20,091 bad elements). No-mask control still passes.
- Regression sample (96 tests): `test_mhas_v2::test_sdpa_fp8_fwd_L0[test1..48]` — now drawing d=256 — plus `sdpa/suites` `context.fp8.dense`/`context.fp8.thd` `[test1..24]` each: **56 passed, 39 skipped (envelope waives), 1 failed**.
- The one failure, `fp8_fwd_L0[test42]` (d=256, e5m2 in/out, band window 78/423, BOTTOM_RIGHT): **8 of 1.85M elements**, max abs 0.159 vs 0.125, one query row — and it fails **identically on the pre-fix kernel** (A/B by stashing only the kernel file). That is the known e5m2 P-rounding midpoint class (the forward compare has no mismatch budget yet; see #971 / #947), unrelated to this fix.

mxfp8:
- Forced-rescale causal runs (e4m3/e5m2, MHA, S=1024, 4 seeds each): pre-fix `max|O−ref|` 1.75–2.6 with inf/NaN → post-fix 0.10–0.19, all finite (dense control on the same data: 0.13–0.16).
- Random probes (8 configs, ~90M elements) and the full direct `sdpa/frost/test_sdpa_fwd_mxfp8_sm100.py` module (L0+L1) pass unchanged.
- New `test_mxfp8_d256_causal_rescale_every_tile[e4m3|e5m2]` fails on the pre-fix kernel and passes on the fixed one (A/B by stashing only the kernel file).

## Tests

- `test_sdpa_fp8_fwd_L0` now draws `d_qk/d_v` up to **256** (was 192; only the backward sweep drew 256), with `(256, 256)` in the high-probability list — this is how the fp8 defect went unexercised there; the `sdpa/suites` Qwen3.5 fp8 preset found it. `test_sdpa_mxfp8_fwd_L0` gets the same widening.
- `sdpa/frost/test_sdpa_fwd_mxfp8_sm100.py`: `_run(..., k_tile_growth=)` hook + the deterministic forced-rescale regression above.

## AGENTS.md

**Rule S3** (python/cudnn/sdpa/AGENTS.md): when a kernel aliases P into a score slot's tail, the warpgroup that stores into columns another warpgroup still has to read must be ordered after that read. The second commit records the fused-role variant and widens the detector — it keyed on `mb_softmax_max`, which is exactly why it missed the mxfp8 kernel; it now keys on `P_EVEN_OFF` aliasing + any config that can select `SOFTMAX_WARPGROUPS == 2`, and gives the forced-rescale recipe for surfacing the race.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved FP8 and MXFP8 attention stability by preventing synchronization-related errors during probability computation.
  - Fixed edge cases involving masked-tail processing, fused corrections, and repeated score rescaling.
  - Improved reliability for ragged packed sequences, including very large batch strides.

- **Testing**
  - Expanded coverage for single-query, lean-attention, FP8, and MXFP8 scenarios across larger hidden dimensions, including D512.
  - Added regression coverage for causal attention, ragged packed sequences, varied masks, and large-memory stride boundaries.
  - Improved test compatibility checks for supported GPU architectures.

### DIFF
--- a/python/cudnn/sdpa/AGENTS.md
+++ b/python/cudnn/sdpa/AGENTS.md
@@ -64,7 +64,7 @@ ordered after that read.**
 
 - Some kernels have no spare TMEM and pack P into the tail of the S slot it
   was computed from (`P_EVEN_OFF`/`P_ODD_OFF` inside the `S_ACC_*` column
-  range — e.g. `prefill_d256_fp8_sm100.py`, `P_EVEN_OFF = 96` in a 128-col
+  range — e.g. `sm100/prefill_d256_fp8.py`, `P_EVEN_OFF = 96` in a 128-col
   slot). With a single softmax owner that is safe: all of S is in registers
   before any P store. It is a race the moment a row's keys are split across
   two softmax warpgroups: the half storing into the aliased columns
@@ -76,17 +76,34 @@ ordered after that read.**
   after its `tcgen05.ld` has landed (`tcgen05.wait::ld` first — the
   softmax paths otherwise never wait on loads explicitly), and the storing
   half waits on it right before the aliased store (`mb_softmax_hi_loaded`
-  in `prefill_d256_fp8_sm100.py`, mirroring `mb_softmax_max`). Do NOT try
+  in `sm100/prefill_d256_fp8.py`, mirroring `mb_softmax_max`). Do NOT try
   to fix it by swapping which half owns which keys: the alpha/stats value
   is aliased too (`STATS_*_OFF == S_ACC_*_OFF`, i.e. S column 0), so the
   half that stores alpha must also be the one that owns key 0 — the
   original ownership is forced, only the ordering was missing.
-- Detector: a kernel that matches both
-  `grep -l "mb_softmax_max" python/cudnn/sdpa/fwd/kernels/*.py` and
-  `grep -l "P_EVEN_OFF: int = " python/cudnn/sdpa/fwd/kernels/*.py` with
-  `P_EVEN_OFF < S_ACC_EVEN_OFF + 128` has this hazard class. For each P
-  store, name the warpgroup that still reads the target columns and the
-  barrier that orders the store after that read; if there is none, it is
-  a bug. The symptom in a fuzz suite is `ref == 0, gpu != 0` elements on
-  masked configs with `key mod TILE_N` in the aliased range — recover the
-  leaked keys by matching `O_gpu - O_ref` against `V` rows.
+- The same kernel family can hide the split behind a different role name.
+  `sm100/prefill_d256_mxfp8.py` has no `mb_softmax_max`: its
+  `FUSED_CORR_SPLIT_P` schedule (mxfp8 + strict top-left causal) gives
+  keys 64–127 to the *correction* warpgroup (`_fused_p1_step`), and half 0
+  stored P into cols 96–111 with nothing ordering it after that
+  warpgroup's S-hi load. There the release already existed —
+  `mb_stat_empty` is arrived right after that load — so the fix is to wait
+  on it *before* the aliased store instead of at the end of the iteration.
+  The dense split-P schedule of the same kernel is safe only because both
+  halves meet at a 256-thread named barrier (`barrier_id=8`) after their
+  loads; if that barrier ever moves below the P store, this comes back.
+- Detector: every kernel with `P_EVEN_OFF: int = ` inside the `S_ACC_*`
+  range (`grep -l "P_EVEN_OFF: int = " python/cudnn/sdpa/fwd/kernels/*/*.py`,
+  today the six d256 kernels) whose config can select
+  `SOFTMAX_WARPGROUPS == 2` (`grep -n "SOFTMAX_WARPGROUPS=2\|split_p\|fused_corr_split_p" python/cudnn/sdpa/fwd/config_*.py`)
+  — regardless of which role runs the second half. For each P store, name
+  the warpgroup that still reads the target columns and the barrier that
+  orders the store after that read; if there is none, it is a bug. Random
+  data can hide it: the lagging half usually lags only when it rescales O
+  (alpha ≠ 1), which `RESCALE_THRESHOLD` makes rare — force it with K
+  scaled by `growth ** (key // 128)` (`k_tile_growth` in
+  `test/python/sdpa/frost/test_sdpa_fwd_mxfp8_sm100.py`; the pre-fix mxfp8
+  kernel then fails every run with inf/NaN O). The symptom in a fuzz suite
+  is `ref == 0, gpu != 0` elements on masked configs with `key mod TILE_N`
+  in the aliased range — recover the leaked keys by matching
+  `O_gpu - O_ref` against `V` rows.
--- a/python/cudnn/sdpa/AGENTS.md
+++ b/python/cudnn/sdpa/AGENTS.md
@@ -57,3 +57,36 @@ head-major, never dense-padded.**
 - Reviewing: if the diff touches `fwd/engines.py`, `bwd/engines.py` or
   `cudnn/engines/manifest.py` and `python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md` is
   untouched, ask why before approving.
+
+**Rule S3 — When P is aliased into a score slot's TMEM tail, the warpgroup
+that stores into columns another warpgroup still has to read must be
+ordered after that read.**
+
+- Some kernels have no spare TMEM and pack P into the tail of the S slot it
+  was computed from (`P_EVEN_OFF`/`P_ODD_OFF` inside the `S_ACC_*` column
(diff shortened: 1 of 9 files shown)


vedaanta:
@cudnn-ci-bot run frost

adshen (approved):

vedaanta:
**Mirror pipeline 67118556 triage** (`frost-sdpa:cutlass-rel:sm100`, job 433340407): 3 failed / 4557 passed — `test_sdpa_fp8_fwd_L0[test42|test136|test314]`, all e5m2 d=256, 12/158/6 elements at 0.13–0.375 over atol with reference ≈ 0. Each reproduces **bit-for-bit with the pre-fix kernel** (A/B by swapping in develop's `sm100/prefill_d256_fp8.py`), so they are the known e5m2 P-midpoint rounding class, newly reachable because this PR widens the fp8 forward sweep to d=256 — not a regression from the fix. They are covered by the calibrated mismatch budget on #971 (rows × keys row budget + 4·atol cap; the #981 corruption itself is still rejected there). **Merge #971 first**, then re-run this mirror.

The other reds in that pipeline are pre-existing and unrelated: `frost-sdpa:cutlass-rel:sm80` fails on develop and every recent PR mirror (the runner hands out an sm_75 GPU — `KeyError: 'sm_75'`, "SDPA FP16/BF16 requires SM80"), and `san:cpp_test:sm80` fails on develop too (this PR touches no C++).

vedaanta:
Rebased onto develop `eb88cd6fe` (includes #971's mismatch budget and #989/#991). Head `363b63b03`, three commits: fp8 fix, mxfp8 fix, sm80 skip-hardening.

Re-verified on B200 after the rebase: the three e5m2 d=256 `test_sdpa_fp8_fwd_L0` midpoint cases from the previous mirror run (test42/136/314) now pass under the merged budget (12 / 158 / 6 elements, 8 / 79 / 6 rows — inside the rows×keys budgets of 61 / 669 / 535, max diff ≤ 0.375 ≤ cap 0.5); the #981 corruption repros pass on the fixed kernel; the mxfp8 d256 regression + mask tests pass (10/10).

Rule S3 detector re-run on the kernels merged since this branch's base: `sm100/prefill_d512_mxfp8.py` (#989) aliases P at `P_EVEN_OFF = 96` and carries the fused split-P schedule, but already orders the aliased store (waits `mb_stat_empty` before it; `_fused_p1_step` does `tcgen05.wait::ld` before arriving) — no change needed. `sm120/prefill_d512_f16.py` (#991) does not alias P.

vedaanta:
@cudnn-ci-bot run frost

vedaanta:
**Mirror pipeline 67265843 (head `363b63b03`) triage**

- `frost-sdpa:cutlass-rel:sm100`: **1 failed / 4573 passed** — the three e5m2 d=256 midpoint cases from the previous run are gone (covered by #971's budget). The one failure, `test_fp8_sm100_execute_lse_contract` → `TypeError: _ref() missing 4 required positional arguments`, is a develop breakage from #971's `_ref` signature change (one caller left behind; this file is identical on develop and here). Fix: #1004 (one line). develop's nightlies had not reached the #971 merge yet.
- `frost-linear:cutlass-rel:sm100`: 2 `torch.OutOfMemoryError` in `test_la.py::test_piece_chain_bwd_matches_uncut_and_reference` with 44 MiB free of 178 GiB and other processes resident on the GPU — shared-runner contention, not this PR (linear attention is untouched here; the lane is green on develop's nightlies).
- `frost-sdpa:cutlass-rel:sm80` and all `san:cpp_test:*` lanes are green on this head (the sm80 skip-hardening did its job on the Turing runner). `analysis:guardwords_scan` is red on every concurrent pipeline (allow_failure).

Once #1004 lands, a rebase (or a re-run after merge) should make the sm100 SDPA lane green.

PR #990, opened 2026-09-10
Propagate shared memory limit during engine selection

scottyokim:
## Summary

- retain an explicitly requested `deselect_shared_mem_greater_than()` limit in the graph plan list
- pass the value through `CUDNN_ATTR_ENGINEHEUR_SHARED_MEMORY_LIMIT` before the heuristic query
- pass the value through `CUDNN_ATTR_ENGINE_SHARED_MEMORY_LIMIT` for forced-engine and knob-query flows
- preserve the existing post-query engine-config filter based on `CUDNN_ATTR_ENGINECFG_SHARED_MEMORY_USED`
- leave backend heuristic generation, engine lists, ordering, timing inputs, and generic filtering unchanged
- leave behavior unchanged when the user did not request a limit
- document the ordering, default, and cuDNN version behavior and update the restricted-shared-memory sample

The companion backend returns the same engine-config list and ranking whether or not a limit is supplied. The only backend behavior change is inside FORT/native: its dynamic pipeline-stage calculation uses the propagated budget. Other engines ignore the budget internally. The frontend's existing `deselect_shared_mem_greater_than()` post-query filter remains in place to deselect configs whose reported shared-memory use exceeds the limit.

To affect backend pipeline-stage selection, callers must set the limit before `create_execution_plans()`, or before `get_knobs_for_engine()` / `create_execution_plan()` for direct-engine flows. The backend descriptor snapshots the value at query time: changing it afterward does not regenerate existing engine configurations or resize their stages, though the existing post-query filter still uses the current value. The restricted-shared-memory matmul sample now demonstrates this ordering.

## Backend dependency

This requires the backend attributes added by merged cuDNN backend MR https://gitlab-master.nvidia.com/cudnn/cudnn/-/merge_requests/4430.

## Validation

- `pre-commit run` (clang-format and SPDX checks) passed
- warnings-as-errors C++ `tests` target built against the companion backend headers
- CPU regression tests passed: 51 assertions covering 16 heuristic/direct-engine forwarding cases across unset, zero, positive, and oversized limits on old and new runtime versions, plus retained plan-list state
- the complete C++ `samples` target built, including the updated restricted-shared-memory matmul sample
- standalone umbrella-header syntax checks passed against installed cuDNN 9.25.0 headers with `-Wall -Wextra -Werror`, both with exceptions enabled and with `NV_CUDNN_DISABLE_EXCEPTION`
- documentation pre-commit checks and local Fern navigation YAML parsing passed; the repository Fern check runs in CI

The CPU tests validate frontend forwarding and compatibility gates; GPU engine-selection behavior is covered with the companion backend change.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring maximum shared memory during heuristic queries, engine creation, and execution-plan selection.
  * Added a builder option for setting shared-memory limits.
  * Limits are applied on supported cuDNN versions and safely handled on older versions or when values exceed supported ranges.

* **Bug Fixes**
  * Ensured shared-memory restrictions are applied before execution-plan creation for restricted matrix multiplication scenarios.

* **Documentation**
  * Added guidance on configuring shared-memory limits and when to apply them.

* **Tests**
  * Added coverage for unset, zero, positive, oversized, and version-dependent limits.


scottyokim:
The repository requires PR labels, but this fork account does not have permission to apply them. Maintainers: please add `cat-feature`, `mod-backend`, `mod-frontend`, and `orig-nv-eng`.

include/cudnn_frontend/graph_interface.h
```diff
@@ -2839,8 +2840,12 @@ Graph::create_execution_plans(std::vector<HeurMode_t> const &mode) {
...
     if (!cudnn_modes.empty()) {
         EngineConfigList op_graph_to_configs;
-        CHECK_CUDNN_FRONTEND_ERROR(detail::query_cudnn_heuristics_impl(
-            operation_graph, op_graph_to_configs, cudnn_modes, context.get_target_sm_count(), device_properties));
+        CHECK_CUDNN_FRONTEND_ERROR(detail::query_cudnn_heuristics_impl(operation_graph,
+                                                                       op_graph_to_configs,
+                                                                       cudnn_modes,
+                                                                       context.get_target_sm_count(),
+                                                                       device_properties,
+                                                                       plans.get_max_shared_mem_allowed()));
```
YangXu1990uiuc, line 2857: Non-blocking: document that the limit must be set before the query to affect pipeline-stage selection.

This snapshots the current limit into the heuristic/engine descriptors. However, the existing "Matmul with restricted shared memory" sample calls `create_execution_plans()` first and `deselect_shared_mem_greater_than()` afterward (`samples/cpp/matmul/matmuls.cpp:582-584`). With that order, the new attribute is never sent for those configurations; the budget only participates in the existing post-query filter.

Please update the example and API guidance to set the limit before `create_execution_plans()` (and before `get_knobs_for_engine()` / `create_execution_plan()` for those flows). It would also help to state that changing the limit after querying does not regenerate existing configs or resize their pipeline stages. Keeping the legacy filter-only ordering valid is fine; callers need to know how to opt into the new behavior.

YangXu1990uiuc (approved):
The propagation makes sense. The explicitly-set flag preserves the default behavior; the positive-only forwarding, INT32 clamp, compile/runtime gates, and retained engine-config filter are consistent across heuristics and direct-engine creation. I found no blocking correctness issue in the frontend change.

One non-blocking documentation/sample follow-up is inline. Please also add repository regression coverage for unset/zero/positive limits, the old-runtime path, and both heuristic and forced-engine/knob-query flows when validating the companion backend. Syntax-only C++ checks against cuDNN 9.25.1 headers passed with `-Wall -Wextra -Werror`, with exceptions both enabled and disabled. No GPU runtime validation was performed locally. A separate CPU call-recording probe passed 16 cases across heuristic/direct-engine paths, old/new runtime-version responses, and negative/zero/positive/oversized limits. Backend descriptor calls and the new header declarations were mocked, so this checks frontend forwarding rather than the backend runtime ABI.

include/cudnn_frontend/graph_interface.h
scottyokim: Addressed in 2db828bb. The API guidance now says to set the limit before create_execution_plans(), or before get_knobs_for_engine()/create_execution_plan() for direct-engine flows, and documents that changing it later does not regenerate configs or resize stages. The restricted-shared-memory matmul sample now sets the limit before plan creation. I also added repository CPU regression coverage for 16 old/new-runtime and unset/zero/positive/oversized forwarding cases across heuristic and direct-engine/knob-query descriptor paths.

include/cudnn_frontend/graph_interface.h
```diff
@@ -2354,9 +2354,18 @@ class Graph : public ICudnn, public INode {
...
+     * selecting pipeline stages. For direct-engine flows, call it before get_knobs_for_engine()
+     * or create_execution_plan(). Changing the limit after a query does not regenerate existing
+     * engine configurations or resize their pipeline stages; it only affects subsequent queries
+     * and the existing post-query engine-config filter.
+     */
     Graph &
-    deselect_shared_mem_greater_than(int64_t const workspace) {
-        plans.set_max_shared_mem_allowed(workspace);
+    deselect_shared_mem_greater_than(int64_t const shared_memory_limit) {
+        plans.set_max_shared_mem_allowed(shared_memory_limit);
```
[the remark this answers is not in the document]
scottyokim: Verified and addressed in f176d2b8. I added a public Utilities page for Graph::deselect_shared_mem_greater_than(), including the required pre-query ordering, unset/default behavior, positive-limit forwarding, the cuDNN 9.27 header/runtime gates, legacy filter-only behavior, and a link to the updated matmul sample. The page is also registered in the Fern navigation. Local pre-commit checks, diff checks, and YAML parsing pass; Fern validation is now running in PR CI.

YangXu1990uiuc:

I am requesting backend CI for `f176d2b8` using the standalone command immediately below. This change crosses the FE/backend descriptor boundary, so local CPU forwarding tests should be supplemented by the supported backend CI.

For the next update, post a comment containing only `@cudnn-ci-bot run backend` once the intended revision is ready; if your account lacks trigger permission, ask a maintainer to issue it. The command must be a separate comment without a signature or trailing explanation.

YangXu1990uiuc:
@cudnn-ci-bot run backend

YangXu1990uiuc (approved):

Reviewed `f176d2b8ae0c` against base `344a9f578e2f`.

Incremental review of f176d2b8 after c0d42f4d: the sample now sets the limit before querying plans, the public documentation explains ordering and version behavior, and the shared helper preserves the prior positive-limit forwarding and INT32 clamp. The earlier documentation follow-up is addressed. No new actionable findings; approving this revision.

Validation: Built and ran the new shared_memory_limit.cpp tests against cuDNN 9.25.1 headers: 51 assertions in 2 test cases passed. These exercise the forwarding helper and retained plan-list state. Separately rebuilt/reran the descriptor-call probe against this head: 16 cases passed through EngineHeuristicsBuilder and create_engine with old/new runtime responses and unset/zero/positive/oversized limits. Backend calls and the new attribute declarations were mocked; this verifies frontend plumbing, not the backend ABI. Umbrella-header checks passed with GNU C++17, -Wall -Wextra -Wpedantic -Werror against cuDNN 9.25.1 headers, with exceptions enabled and NV_CUDNN_DISABLE_EXCEPTION. Fern navigation parses and resolves the new page. I approved the pending fork Style/Docs workflows for this head; both have now passed. Requested backend CI for f176d2b8; the CI bot confirms its pipeline is running. For future updates, use a standalone comment containing only @cudnn-ci-bot run backend; fork workflow approvals may still require a maintainer.

Limitations: Backend CI is still pending. No independent GPU runtime or performance claim: the local checks cover frontend descriptor forwarding, and this increment does not change the execute hot path.

YangXu1990uiuc:

CI follow-up for `f176d2b8ae0c`: both build configurations fail before runtime tests because their cuDNN headers do not declare `CUDNN_ATTR_ENGINEHEUR_SHARED_MEMORY_LIMIT` (`cudnn_frontend_Heuristics.h:266`) or `CUDNN_ATTR_ENGINE_SHARED_MEMORY_LIMIT` (`backend/plan_helpers.h:98`). The version guard is entered, so the backend header dependency still needs alignment with a build that supplies these attributes; please also ensure the guard matches their actual availability. My earlier 9.25 guarded-path and mocked forwarding checks do not cover this real-header combination. Please rerun CI after aligning it. This is additional evidence from the requested CI, not a passing backend validation.

PR #991, opened 2026-09-10
frost(sdpa): add a specialized SM120 D512 F16 prefill kernel

Aneureka:
## Before submitting

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Add a dedicated D=512 prefill kernel for SM120, with the following design to allocate register/smem appropriately and improve the performance:
- Tile size is fixed at (64, 32).
- Two warps share each 16-row Q (each owns half of D), sum their partial QK scores before softmax, and accumulate separate 256-column output halves.
- Use TMA Q/K/V loads, partial Q staging in shared memory, persistent dense/THD execution, next-Q prefetch, and output staging that reuses the Q/score-exchange storage.
- Prefer PackGQA for SWA cases (e.g., DeepSeek V4) in the heuristic: all Q rows within a Q tile then share the same K/V tile window, which reduces the number of K/V tiles processed.
- Register the D512 flavor with matching envelope checks, tile selection, shared-memory accounting, windowed-GQA ranking, support documentation, and forward/split-KV tests.

image

## Why

The existing SM120 FP16/BF16 forward kernels stop at head dimension 256. D512 attention is needed by workloads such as DeepSeek-V4 and Gemma4.

Extending the D256 warp mapping directly would require 256 FP32 registers per thread for the output accumulator alone. Splitting output ownership across a warp pair reduces that to 128 registers per thread while retaining the existing attention pipeline and feature handling.

## Related issues

Related to #381.

- No public API signature or dependency changes.
- FP16/BF16 forward now accepts independent QK and V head dimensions in `(256, 512]`, in multiples of eight. Smaller dimensions in this band are zero-padded to the fixed 512×512 head tiles and retain their computation/storage cost. Cross-band pairs such as `(256, 512)` remain unsupported.
- Existing dimensions at or below 256 keep their general/D256 paths. FP8 and backward head-dimension limits remain 256.
- Preserve causal/bottom-right/window masks, padding, sinks, MHA/GQA/MQA, optional LSE, THD, and split-KV support. D512 accepts only the 64×32 CTA tile.

## Performance

```
python -m benchmark.attention_training.runner \
    --config deepseek_v4 \
    --backend cudnn,cudnn_oss \
    --dtype bfloat16 \
    --profile-pass fwd
```

image

## Testing

See the pipeline.

## Summary by CodeRabbit

- **New Features**
  - Added SM120 FP16 support for attention head dimensions above 256 through 512, including native 512-dimensional execution.
  - Added support for independent Q/K and V/O head dimensions within the supported range.
  - Added zero-padding for smaller dimensions and expanded support for masks, sinks, sliding windows, GQA/MQA, causal workloads, and split-KV execution.
  - Documented the new D512 kernel coverage and supported dimension ranges.

- **Bug Fixes**
  - Improved tile selection and validation for supported head dimensions and execution configurations.

### DIFF
--- a/python/cudnn/frost/README.md
+++ b/python/cudnn/frost/README.md
@@ -322,6 +322,7 @@ python/cudnn/
         sm107/prefill_d128_fp8.py     Rubin siblings (dense K=64 MMA, desc v1)
         sm120/prefill_f16.py          general SM120 template (any head dim)
         sm120/prefill_d256_f16.py     d256 flavor
+        sm120/prefill_d512_f16.py     d512 flavor
         sm120/_common.py              SM120-only warp-level primitives
         _common_blackwell.py      SHARED by sm100/ + sm107/ (cc 100-119), so it
                                   sits ABOVE both rather than inside either
--- a/python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md
+++ b/python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md
@@ -358,15 +358,21 @@ Engines: `sdpa_fwd_prefill_sm120`, `sdpa_fwd_prefill_sm120_fp8`,
 kernel picks Q/K and V head tiles independently (f16/bf16 head dims it would
 tile at 256 on both sides run a dedicated template, `sm120/prefill_d256_f16.py` with the same support; fp8 has no such flavor).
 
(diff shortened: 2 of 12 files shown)


Aneureka:
@cudnn-ci-bot run frost

vedaanta (approved):

PR #993, opened 2026-09-10
Fix SM100 2CTA Flex Attention mask slot synchronization

jiayus-nvidia:
## Summary

Fix intermittent incorrect mask reads in the SM100 2CTA Flex Attention forward kernel. On B200, a valid row can consume an all-zero mask, producing zero output and `LSE=-inf`, followed by large gradient errors.

Count all 128 mask-reading threads in each slot's empty barrier and issue `fence.proxy.async.shared::cta` after the shared loads and before releasing the slot. The existing 2CTA configuration and backward kernels are preserved.

## Why

Each N-direction stream reuses a single shared-memory mask slot. The original pipeline released once per consumer warp, while individual lanes read different mask words. It also lacked ordering between those generic-proxy shared loads and the next async-proxy bulk copy that overwrites the slot. Waiting for the previous bulk copy to complete covers the producer-to-consumer handoff; safe slot reuse also needs the reverse handoff.

A diagnostic captured `0x03000000` becoming zero for a row with valid keys. The softmax statistics were already incorrect at their producer, and matched the values read by the correction stage. This localizes the observed failure to the forward mask path.

## Affected area

FE OSS kernels / CuTeDSL: Flex Attention shared-mask consumption and SM100 2CTA forward pipeline construction.

No public API, supported architecture, or dependency-floor changes. The shared-mask consumer helper also gains the fence for its existing callers. GPU validation was performed on B200; SM90 was not tested.

## Testing

B200, CuTe DSL 4.6.2, CUDA 13.2 PyTorch build. Imports were verified to resolve to the modified checkout.

- Fixed-input reproducer: BF16 Q/K `[1,4096,4,192]`, V `[1,4096,4,128]`; 2000 rounds, two complete forward/backward calls per round. Zero accuracy failures or NaN/Inf failures; all 4000 forward outputs were identical. Each call checked output, dQ, dK, and dV against the supplied FP64 reference.
- All 25 existing Flex Attention L0/L1/L2 tests passed, using `cd test/python && pytest fe_api/flex_attention/ -m 'L0 or L1 or L2' -q` with `PYTHONPATH` pointing to this checkout's `python/` directory.
- Standalone mask handoff diagnostic under `compute-sanitizer --tool racecheck --error-exitcode 99`: the original pipeline reports a read/write race; the fixed handoff reports zero errors and warnings. Diagnostic files and the fixed-input data are not part of this PR.
- `pre-commit run --files python/cudnn/flex_attention/kernels/sm100/fwd/forward.py python/cudnn/flex_attention/plan/kernels/packed_mask.py` passed.

## Related issues

None linked.

## Before submitting

- [x] I reviewed the Hard Rules in the applicable `AGENTS.md` files and the changes comply.
- [x] I added labels for change type, affected area, and originator.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved mask processing synchronization during attention operations.
  - Prevented shared-memory data from being overwritten before all reads complete.
  - Improved coordination across participating threads for mask payload transfers.

### DIFF
--- a/python/cudnn/flex_attention/kernels/sm100/fwd/forward.py
+++ b/python/cudnn/flex_attention/kernels/sm100/fwd/forward.py
@@ -25,6 +25,7 @@ import cuda.bindings.driver as cuda
 import cudnn.flex_attention.kernels.common.pipeline as pipeline_custom
 from cudnn.flex_attention.kernels.common import device_utils as utils
 from cudnn.flex_attention.plan.kernels.packed_mask import (
+    create_mask_payload_pipeline,
     softmax_arbitrary_forward_sm100,
     softmax_arbitrary_forward_qstage1_n_direction_sm100,
 )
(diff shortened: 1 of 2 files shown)


jiayus-nvidia:
@cudnn-ci-bot run oss

vedaanta (approved):

PR #995, opened 2026-09-10
frost(sdpa_bwd_sm80): serve THD / ragged ports at their own token stride

egilliam-nv:
## Summary

`sdpa_bwd_sm80` serves THD / ragged ports at **their own token stride**. A port declared as a view into a wider per-token record — a K or V slice of an interleaved `[T, 2, H_kv, D]` buffer (token stride `2*H_kv*D`, the fused-KV slicing layout), or any Q/O/dO/dQ/dK/dV in a record of its own width — no longer has to be compact BSHD.

Follow-up to PR #950, which landed the engine-path THD backward with a compact-only rule.

## Why

In `test_sdpa_random_bwd_ragged_L0` the harness draws a per-tensor token-stride gap on Q/K/V/O (0–3 extra rows of `H*D`). On A100 the cuDNN backend declines every ragged backward it is offered (`hidden_dim <= 128`, deterministic THD, packed LSE), and with the compact rule the SM80 row declined 382 of 384 draws — so those graphs ran nowhere. With this change the whole sweep routes to `sdpa_bwd_sm80`.

## How

- **Strides are plan-time and ride the compiled fakes.** `compile(..., thd_token_strides=(Q, K, V, O, dO, dQ, dK, dV))` builds each THD operand's fake as `[1, T_sym, H, D]` with strides `(T_sym*ts, ts, D, 1)`; the kernels read `.stride[1]` and address a row as `token * token_stride + head * D + col`. A compact port keeps the compact fake, so the dense path and compact THD graphs are byte-identical.
- **Every kernel that touches a caller buffer** honours it: the main kernel's Q/K/V/dO loads and direct-bound dK/dV stores, the O·dO dot, the bounded dQ cast and the bounded dK/dV fold. Ports the host stages (head dim inside the flavor envelope) and the GQA per-query-head partials stay compact inside `compile()`.
- **Rule** (`mismatch()` and the adapter, same words): head stride `D`, element stride 1, token stride `>= H*D` and a multiple of 8 elements (16-byte rows for the `cp.async` loads). Overlapping strides are already declined by the layout envelope.
- The adapter's execute view check compares the load-bearing strides against the plan and rebuilds the view (never a copy) so size-1 axes carry the compiled stride.

## Testing

`test_sdpa_bwd_thd_sm80.py` builder takes `gaps={role: extra_elements}`; bound inputs are re-homed in records with **NaN gap columns** (a read would poison the result), gradient records are NaN-filled and their gap columns asserted untouched.

- `test_graph_thd_gapped_token_strides[kv_interleaved|q_side|all_ports|gqa|mqa|padded_d|causal_det]`
- `test_graph_thd_gapped_capacity_tail_left_untouched` (tail + gap columns on gapped records)
- probes: `test_accept_thd_gapped_token_strides`, `test_reject_thd_misaligned_token_stride`, `test_reject_thd_token_stride_below_the_row`

Local, A100 (`CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`):

| suite | result |
|---|---|
| `test_sdpa_bwd_thd_sm80.py` | 67 passed |
| SM80 bwd / integration / stream / gate / q-loop-bounds | all passed |
| `test_mhas_v2.py::test_sdpa_random_bwd_ragged_L0` | 384 passed, all 384 on `frost:sdpa_bwd_sm80` (was 2 of 384) |

Tracker Rule S2: footnote ᵏ updated.

## Labels

`cat-feature` · `area:frost` · `area:global_attention` · `orig-nv-eng`

### DIFF
--- a/python/cudnn/AGENTS.md
+++ b/python/cudnn/AGENTS.md
@@ -131,11 +131,13 @@ none is precedent:
   declaration would drop the KV-only-padding population these kernels serve
   correctly. It goes away when the FP8 kernels get the epilogue trim; until
   then the read is what keeps a short length from being silently ignored.
-- `cu_seqlens_{q,k}.to(dtype=..., device="cpu")` in the SM80 packed-THD backward
-  (`sdpa/bwd/kernels/bprop_f16_sm80.py`). Reachable only through the standalone
-  wrapper: the registered `sdpa_bwd_sm80` spec declares `thd=False`, so
-  `graph.execute()` does not route here. Still a violation, and it is the one to
-  fix first if that spec ever gains THD.
+- `cu_k.to(dtype=..., device="cpu")` in the SM80 packed-THD WRAPPER path
+  (`_sm80_thd_backward` in `sdpa/bwd/api_dsl.py`), taken only when the caller
+  passes no `max_s_kv` hint. Reachable only through the standalone wrapper: the
+  `sdpa_bwd_sm80` engine path bounds its kv-tile grid and relay counter from
+  the graph's envelope `S_max` and turns the per-batch lengths into
+  `cu_seqlens` on device, so `graph.execute()` never reads a length. Still a
+  violation on the wrapper surface (a caller contract, documented there).
 
 When auditing this list, grep for the ARGUMENT, not the call shape:
 `device="cpu"` finds `to(dtype=..., device="cpu")`, which `to(device="cpu")`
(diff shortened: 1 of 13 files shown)


egilliam-nv:
Folded into PR #950 as its fourth commit (e21712fb1) so the THD rule is reviewed once in its final, stride-aware form; the two PRs touched the same paragraphs of the probe, the adapter and tracker footnote ᵏ.

PR #1001, opened 2026-09-10
frost(sdpa): SM107 honours SCHED_LPT, and STAGES_KV is a knob again

RomanAnders90:
Two independent Rubin (SM107) SDPA fixes, both correctness-first. They are in one PR because they touch the same kernel and the same capability row.

## 1. SM107 honours `SCHED_LPT` — a dropped argument, not a broken decode

Every SM100 kernel calls `make_sdpa_helpers(CFG, lpt_q_tiles_in_cga_units=True)`. The SM107 port omitted that argument on **9 of 11 flavors**. Without it the LPT linearization walks a row range `CTA_MMA`× too large, no tile is ever claimed, and the kernel writes **nothing** — cosine `0.0000`, dense and causal, which reads like a dead kernel rather than a scheduling bug.

The two flavors that kept it (d128 FP8, d192 FP8) are exactly the ones `engines.py` already singled out as working. That comment was describing this bug's symptom without its cause.

Restored on all seven 2-CTA SM107 flavors. **It is a no-op under `SCHED_NATURAL`** — that decode branch never reads `q_tiles` — so the shipped path is unchanged.

`sched_policies` is row-wide while LPT support is per-kernel, so this adds **`sched_policies_by_d_shape`**, mirroring the existing `cgas_by_d_shape`. Only what is validated is advertised: the f16 row serves `SCHED_LPT` at `(256, 256)`.

### Measured

Validated on d256 f16: cos `1.0000` at n_kv 2/3/4/8, dense **and** causal. Causal SOL on Rubin (212 SMs, 2376 MHz, 4126 TFLOP/s bf16 peak):

| S | NATURAL | LPT | gain | dense ceiling | gap recovered |
|---|---|---|---|---|---|
| 4096 | 53.0% | 62.6% | **+18.2%** | 77.1% | 40% |
| 8192 | 71.1% | 79.3% | **+11.6%** | 87.3% | 51% |
| 32768 | 76.7% | 77.5% | +1.1% | 79.5% | 29% |

Dense is neutral, which is the right control: with no mask there is no imbalance to fix. The decay with S is the signature of scheduler imbalance, which is what LPT exists to fix.

### Deliberately not claimed

- **d128 / d512 f16 and every FP8/MXFP8 row** — unvalidated under LPT, not known-broken. The argument is restored so they should work; each needs a numerics sweep before its row claims it.
- **d512** — cga4×1 role-split, a different scheduler shape; the older NaN report there is still unexplained and I did not assume the same cause.
- **`SCHED_LPT_L2`** — declined by every flavor. Its decode needs `qh_per_kh` and `seqlen_kv`, which the SM107 call sites do not pass, so it *raises* rather than miscomputes. Threading those two arguments is a mechanical follow-up.

## 2. `STAGES_KV` is a knob again — the old pin blamed the wrong thing

`config_sm107` pinned the d256 `STAGES_KV` to 2, blaming a body that "conflates the KV ring index with the 2-slot `S_acc` parity". The body in fact derives parity from the absolute `kv_loop & 1` and rides separate K/V `PipelineState`s, so it is depth-agnostic.

The real constraint was the **>256 KiB tcgen05 descriptor wrap**. Buffers are laid out `sQO | sK[stages] | sV[stages]`; at depth 4 the last two V stages start at 256 and 288 KiB, and a version-0 descriptor truncates `start_address` to 14 bits, so those stages alias the bottom of SMEM. The answer breaks from the KV iteration that first touches a wrapped stage — which is why it looked like a ring bug and why it only shows up at `S_kv > 256`.

| config | n_kv=2 | 3 | 4 | 8 |
|---|---|---|---|---|
| depth 4, desc v0 | 1.0000 | 0.6806 | 0.4980 | 0.4640 |
| **depth 4, desc v1** | **1.0000** | **1.0000** | **1.0000** | **1.0000** |
| depth 3, desc v0 | 1.0000 | 1.0000 | 1.0000 | 1.0000 |

`DESC_VERSION` is now **derived** from the layout rather than a literal, so every depth in 2..4 is correct by construction, and the validator accepts that range bounded by the 320 KiB carveout.

**The default stays 2.** Depth is perf-neutral — dense SOL moves 0.3 points across all three depths — and 2 leaves 135 KiB free for a future fused epilogue.

## Testing

- New: accept **and** reject for the per-shape domain, a honored-or-ineligible knob test, and a parametrized guard that `DESC_VERSION` tracks `STAGES_KV`.
- `SUPPORT_MATRIX_TRACKER.md` updated in the same commit (Rule S2).
- On Rubin: `sdpa/frost` **1069 passed, 634 skipped**; `test_sdpa_fwd_dsl_sm107.py` **61 passed**.

## Summary by CodeRabbit

- **New Features**
  - Expanded SM107 SDPA d256 configurations to support KV pipeline depths from 2–4.
  - Added validated LPT scheduling for supported d256 f16 workloads.
  - Improved scheduling selection based on the active attention shape.
  - Added automatic descriptor version selection for larger pipeline configurations.

- **Bug Fixes**
  - Corrected query-tile interpretation for LPT scheduling across supported SM107 kernels.
  - Preserved NATURAL scheduling behavior and rejected unsupported LPT_L2 configurations.

- **Tests**
  - Added coverage for scheduler eligibility, descriptor versions, and invalid pipeline-depth values.

### DIFF
--- a/python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md
+++ b/python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md
@@ -332,15 +332,44 @@ PackGQA and THD families this line declines per-feature, not per-shape.
 kernel's KV loop bound is a floor division, so an un-synthesized ragged `S_kv`
 would silently drop the tail tile.
 
-**Scheduler policy — NATURAL only on every Rubin row.** `SCHED_LPT_L2` was
-dropped first (the ported decode sites never thread `qh_per_kh` / `seqlen_kv`,
-so `_common_blackwell._decode_initial` raises); plain `SCHED_LPT` followed on
-2026-09-08 once a heuristics fallback fix let a causal graph actually reach it —
-it is silently WRONG on the ported kernels (causal d512 FP8 → NaN, masked d128
-MXFP8 → max|O-ref| ≈ 1.9). A knob is honored or the engine is ineligible. The
-SHIPPED d128 FP8 kernel does honor both and loses them here too, because
-`sched_policies` is row-wide; recovering that needs a per-shape domain
-(`sched_policies_by_d_shape`, mirroring `cgas_by_d_shape`).
+**Scheduler policy — `SCHED_LPT` now served on the f16 (256, 256) Rubin
+flavor; NATURAL elsewhere.**
+
+The earlier reading — "the ported decode does not honor SCHED_LPT" — had the
+symptom right and the cause wrong. Every SM100 kernel calls
+`make_sdpa_helpers(CFG, lpt_q_tiles_in_cga_units=True)`; the SM107 port omitted
+that argument on 9 of 11 flavors. Without it the LPT linearization walks a row
+range `CTA_MMA`× too large, no tile is ever claimed and the kernel writes
+**nothing** (cosine 0.0000, dense and causal). The two flavors that kept it,
+d128 FP8 and d192 FP8, are exactly the ones previously described as working.
+
+The argument is restored on every 2-CTA SM107 flavor. It is a **no-op under
+`SCHED_NATURAL`** (that decode branch never reads `q_tiles`), so the shipped
+path is unchanged.
+
+Only what is validated is advertised, via the new per-shape
+`sched_policies_by_d_shape` (mirroring `cgas_by_d_shape`): the f16 row serves
+`SCHED_LPT` at **(256, 256)** only. Validation: cos 1.0000 at n_kv 2/3/4/8,
+dense and causal. Measured causal SOL on Rubin at S = 4096/8192/32768:
+53.0/71.1/76.7 % under NATURAL → **62.6/79.3/77.5 %** under LPT
+(+18.2/+11.6/+1.1 %), recovering 40/51/29 % of the causal-vs-dense gap; dense is
+neutral. The decay with S is the signature of scheduler imbalance.
+
+Still declined, and why: d128/d512 f16 and every FP8/MXFP8 flavor are
+**unvalidated** under LPT rather than known-broken (d512 is cga4×1 role-split,
+a different scheduler shape). `SCHED_LPT_L2` is declined by **every** flavor —
+its decode needs `qh_per_kh` and `seqlen_kv`, which the SM107 call sites do not
+pass, so it raises rather than miscomputes. Both are follow-ups.
+
+**`STAGES_KV` on the d256 flavors is 2..4, not pinned to 2.** The old pin blamed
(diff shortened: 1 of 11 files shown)


RomanAnders90:
@cudnn-ci-bot run frost

vedaanta (approved):

RomanAnders90:
Pushed `201f87f6` addressing the review comment and the CI guardword failure.

### 1. CodeRabbit: `stages_kv=0` bypassed the 2..4 validation

Valid, and it is a contract issue rather than a nit, so it is worth stating why. The truth test mapped an explicit `stages_kv=0` onto the default 2, which means the engine would run a depth the caller did not request. That is knob **substitution**, and the engine contract is honored-or-ineligible. Fixed with `is not None`, as suggested.

One qualification for the record: this is **latent, not live**. The shared forward `TemplateParams` has no `stages_kv` field, which is why the read is a `getattr`, and nothing on the shipped path sets the attribute. Today the branch always takes the default. It stops being latent the day the knob is declared, which is the direction this PR points.

Pinned by tests rather than left to the diff:

- `test_sm107_d256_rejects_an_out_of_domain_stages_kv[0,1,5]` -- reject side.
- `test_sm107_d256_stages_kv_defaults_when_the_attribute_is_absent` -- accept side, so the fix cannot break the only path that ships.

Red-then-green checked against the old spelling: only the `[0]` case flips. `1` and `5` raise either way, so `0` is the single value a truth test swallows, and it is the one that would have regressed silently.

### 2. CI guardword failure

Two of the four hits are mine, both prose in comments:

| file | before | after |
|---|---|---|
| `SUPPORT_MATRIX_TRACKER.md:359` | `known-broken` | `known-incorrect` |
| `engines.py:948` | `not a broken decode` | `not an incorrect decode` |

Meaning is unchanged in both.

The other two hits in that report, `test/python/api_index/README.md:62` (`GitLab`) and `test/python/api_index/test_api_index.py:108` (`"def broken(:"`, a deliberate parser fixture), are **pre-existing on `develop`** at `8059fdf4` and are not touched by this PR. Left them alone rather than reaching outside this branch's scope. The fixture one in particular looks like it wants an allowlist rather than a rename.

### Gate

- `test_sdpa_fwd_dsl_sm107.py`: **65 passed** (61 before, plus the 4 new).
- `sdpa/frost`: **341 passed, 1366 skipped** -- this run is on the A100 box, so the kernel tests skip; the SM107 **config** tests that cover the changed line do execute there. The 1069-passed full gate quoted earlier in this PR was the Rubin box and is unchanged by a comment reword plus a one-operator config fix.
- `black -l 160` clean on all changed files.

PR #674, opened 2026-08-19
benchmark: add block_sparse_attention (BSA fwd+bwd vs FA4)

vedaanta:
## What

New `benchmark/block_sparse_attention` microbenchmark: block-sparse attention forward and backward through the public `cudnn.block_sparse_attention_forward`/`_backward` APIs, with an optional FA4-lineage CuTe DSL arm (`flash_attn.cute`) on identical block masks.

- Sparse-block granularities 64 / 128 / 256, plus a dense peak-reference bar per case
- Two mask families: `topk` (VSA-style learned, per-head data-dependent) and `frame_causal` (autoregressive video: frame window + anchor, variable per-row counts, ~90% sparse defaults)
- Default cases: Wan2.1 1.3B/14B at 480P/720P and MiniMax-H3 (open-weights config, 56 heads x d128, ~31k/~91k tokens); all head_dim 128, bf16, batch 1
- TFLOPS count selected blocks only (2 matmuls fwd, 5 bwd); both arms attend exactly the same token set, with FA4 granularity-floor aggregation documented in the README
- `--check` validates fwd + dq/dk/dv of every arm/family/granularity against an fp32 autograd reference (40 checks)
- `--csv` / `--plot` outputs following the repo chart palette

## Testing

All 40 parity checks pass on B200 (SM100). Full grids (both families, fwd+bwd) run end-to-end; example numbers in the README-described format were collected on a clock-locked B200.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #442, opened 2026-07-28
[Roadmap]: cuDNN Roadmap Q3 2026 (July – September)


Anerudhan:
# [Roadmap]: cuDNN Roadmap Q3 2026 (July – September)

This issue tracks planned cuDNN development and releases for **Q3 2026 (July – September)**.
Plans are subject to change as the team iterates and receives feedback.
If you have suggestions for features, please open a feature request or comment below.

This roadmap covers both the **cuDNN backend library** (9.x) and the open-source **cuDNN Frontend** (1.x). Frontend release notes: <https://github.com/NVIDIA/cudnn-frontend/releases>. Backend release notes: <https://docs.nvidia.com/deeplearning/cudnn/backend/latest/release-notes.html>.

---

### cuDNN Backend — 9.23.0 & 9.24.0

**9.24.0 — Highlights**

- **Expanded causal 1-D convolution support:** Causal conv1d is expanded through the `SubquadraticOps` API with NVRTC runtime compilation, offering multiple depthwise variants.
- **Ubuntu 26.04 support:** cuDNN 9.24.0 adds support for Ubuntu 26.04.
- **CUDA Toolkit discovery via environment variables:** `CUDA_HOME`, `CUDA_PATH`, and `CUDA_ROOT` can now be used to locate CUDA Toolkit components at runtime.
- **Stability fixes:** Convolution forward plans with very large filter tensors are now supported (previously returned `CUDNN_STATUS_NOT_SUPPORTED`); fixed incorrect depthwise convolution results when multiple operations with different spatial dimensions executed sequentially; fixed GEMM fusions (`ENGINE_GLOBAL_INDEX=11`) producing incorrect results with unpacked tensors.

**9.23.0 — Highlights**

- **CUDA Toolkit 13.3 support:** cuDNN 9.23.0 adds support for NVIDIA CUDA Toolkit 13.3.
- **SDPA backward `d=256` on Blackwell:** Scaled dot-product attention backward now supports embedding dimension `d=256` with bfloat16/float16 using the 2-CTA MMA path on Blackwell-architecture GPUs.
- **Stability fixes:** Large-filter convolution forward plans now return `CUDNN_STATUS_NOT_SUPPORTED` before launch instead of failing/illegal-memory-access; fixed false cache hits in backward normalization passes producing incorrect gradients; fixed `RMSNorm`/`LayerNorm` cache-hit issues with large row counts causing CUDA errors.

[...]

PR #1010, opened 2026-09-11
BSA: support logical blk128 forward on SM120

tiffany940107:
## Before submitting

- [x] I reviewed the Hard Rules in the `AGENTS.md` files for the Python and test directories touched by this PR.
- [x] Please apply `cat-feature`, `mod-cutedsl`, `mod-frontend`, and `orig-external`; fork authors do not have permission to add labels.

## Affected area

FE OSS kernels or CuTeDSL, Python API, tests, and BSA documentation.

## Summary

- Add explicit `sparse_block_size=128` forward support on SM120 for BF16 Q/K/V with QK=V=128.
- Lower logical blk128 sparse metadata on the current CUDA stream to the existing native SM120 blk64 ABI; Q, K, and V are not copied.
- Cover fixed odd Top-K, variable Top-K, MHA/GQA, BHSD/BSHD, Q tails, rank-1/2/3 `block_sizes`, and validation failures.
- Document the support matrix and current restrictions.

## Why

SM120 already has a tuned blk64 BSA forward kernel. A logical metadata adapter reuses that implementation, keeps one kernel source of truth, and exactly preserves each selected 128-token KV block as two adjacent 64-token blocks. It also avoids host reads and device synchronization.

The target workload is BF16 `BHSD=[1, 8, 142720, 128]` at 15% or 20% logical-block density. The public `cudnn.BSA.block_sparse_attention_forward` path exceeds the requested 4.5x speedup at 20% density.

## Related issues

Related to #442. The broader BSA benchmark suite is already being developed in #674, so this PR keeps the functional change focused.

There is no signature change, and the SM120 default remains blk64. Callers opt in with `sparse_block_size=128`.

The new path is forward-only and currently requires:

- SM120, BF16, QK=V=128;
- `S_kv` divisible by 128;
- unpacked per-query-head metadata (`pack_gqa=None` or `False`);
- `kv_splits=1` and `use_clc=None`.

It supports MHA, GQA, and MQA; fixed or per-query-block counts; and rank-1/2/3 `block_sizes`.

## Testing

- `pre-commit run --from-ref upstream/develop --to-ref HEAD`: passed.
- Full `test/python/fe_api/bsa` suite on SM120: 37 passed, 19 skipped.
- The regression test was observed failing on the base revision with `SM120 only provides a blk64 forward path`, then passing after the change.

Performance was measured through the public cudnn-frontend wrapper with five warmups and the median of 21 CUDA-event samples. Dense is PyTorch cuDNN SDPA; sparse patterns contain unique logical block IDs. Conversion is `measured_speedup / (1 / actual_density)`.

| Pattern | Requested density | Actual density | Dense (ms) | Sparse (ms) | Speedup | Ideal | Conversion |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Strided | 15% | 14.9776% | 224.3058 | 36.0429 | 6.2233x | 6.6766x | 93.2100% |
| Local | 15% | 14.9776% | 224.3058 | 36.0147 | 6.2282x | 6.6766x | 93.2830% |
| Strided | 20% | 20.0000% | 224.3058 | 48.1593 | 4.6576x | 5.0000x | 93.1516% |
| Local | 20% | 20.0000% | 224.3058 | 48.0503 | 4.6681x | 5.0000x | 93.3628% |

Sampled first/last-row FP32 reference checks passed for every case. The largest observed output error was `8.17e-5`; the largest LSE error was `1.91e-6`.

## Provenance and scope

This adapter was first developed and validated in [`tiffany940107/Block-Sparse-Attention`](https://github.com/tiffany940107/Block-Sparse-Attention/tree/sm120_blk128_bf16), based on the original [`NVIDIA-JerryChen/Block-Sparse-Attention`](https://github.com/NVIDIA-JerryChen/Block-Sparse-Attention) implementation. The standalone benchmark scripts and experimental multi-GPU communication/compute-overlap prototype remain in that repository and are intentionally outside this focused PR.

## Summary by CodeRabbit

- **New Features**
  - Added support for SM120 block-sparse attention with 128-token logical blocks.
  - Added BF16 support for this execution path when key sequence lengths are divisible by 128.
  - Added support for regular SM120 batched `block_sizes` shapes and variable-count attention scenarios.

- **Bug Fixes**
  - Improved handling of sparse metadata, including variable block sizes, query tails, and odd block counts.

- **Documentation**
  - Clarified caller-visible KV block sizing and documented SM120 logical block-size behavior and constraints.

### DIFF
--- a/docs/fe-oss-apis/bsa.md
+++ b/docs/fe-oss-apis/bsa.md
@@ -100,7 +100,7 @@ is consumed. Values in the inactive suffix are ignored.
 - With fixed counts, `block_sparse_num` must be in `[1, K_max]`. The
   SM100/SM103 blk128 path additionally requires an even value, i.e. an even
   `block_sparse_num` in `[2, K_max]`.
-- The `block_sizes` entry for every physical KV block referenced by an active
+- The `block_sizes` entry for every caller-visible KV block referenced by an active
   `q2k_block_index` value must be in `[1, sparse_block_size]`. Entries for
   unreferenced physical KV blocks are ignored. A zero-sized referenced block is
   not supported; use `q2k_block_nums` (or the sparse index prefix) to drop the
@@ -121,10 +121,18 @@ Provide `block_sizes` whenever a referenced final block is only partially
 valid.
 
 `sparse_block_size=None` chooses blk64 on SM90/SM120 and blk128 on
-SM100/SM103. Passing `sparse_block_size=64` explicitly selects the SM100/SM103
-blk64 CuTe DSL path, whose shape support is narrower. `kv_splits` is available
-on SM90 and the explicit Blackwell blk64 path; `use_clc` applies only to the
-explicit Blackwell blk64 path.
+SM100/SM103. Passing `sparse_block_size=128` explicitly on SM120 selects the
+logical-blk128 BF16 path. It requires QK/V dimensions of 128 and `S_kv` to be
+a multiple of 128. Its logical metadata is expanded on the current CUDA stream
+and dispatched through the native SM120 blk64 kernel; Q, K, and V are not
+copied. The path supports MHA, GQA, and MQA with unpacked per-query-head
+metadata, so `pack_gqa` must be `None` or `False`.
+
+Passing `sparse_block_size=64` explicitly selects the SM100/SM103 blk64 CuTe
+DSL path, whose shape support is narrower. `kv_splits` is available on SM90 and
+the explicit Blackwell blk64 path; `use_clc` applies only to the explicit
+Blackwell blk64 path. SM120 logical blk128 requires `kv_splits=1` and
+`use_clc=None`.
 
 `kv_splits=2..256` computes FP32 partial outputs and combines them, with
 workspace growing linearly in the split count. SM90 accepts an explicit integer
@@ -225,6 +233,7 @@ therefore requires full physical KV blocks and `block_sizes=None`.
 | SM100/SM103 | 64 (explicit) | BF16 | QK=128, V=128 | MHA |
 | SM100/SM103 | 64 | BF16 / FP8 E4M3 | QK=128, V=128 | MHA |
 | SM120 | 64 | FP16, BF16 | QK=128, V=128 | MHA, GQA, MQA |
+| SM120 | 128 (logical, explicit) | BF16 | QK=128, V=128 | MHA, GQA, MQA |
 | SM120 | 64 | BF16 / FP8 E4M3 | QK=128, V=128 | MHA |
 
 SM90 currently requires `S_q` to be a multiple of 64. Its fixed count may be
@@ -235,6 +244,8 @@ defaults to `False`; when it is `True`, empty rows (`q2k_block_nums == 0`)
 produce `O = 0` and `LSE = -inf`. SM90 selects the empty-row handling as a
 compile-time specialization, so the default non-empty configuration keeps its
 branch-free fast path. Split-KV execution therefore excludes empty rows.
+Regular SM120 forward accepts `block_sizes` shaped `(N_kv,)`, `(B, N_kv)`, or
+`(B, H_q, N_kv)` for both sparse block sizes.
 
 ### Backward
 
--- a/python/cudnn/block_sparse_attention/api.py
+++ b/python/cudnn/block_sparse_attention/api.py
@@ -208,7 +208,7 @@ def block_sparse_attention_forward(
         if isinstance(kv_splits, str) or not 1 <= int(kv_splits) <= 256:
             raise ValueError("SM90 kv_splits must be an integer in [1, 256]")
 
-    if arch_family in {9, 12} and sparse_block_size != 64:
+    if arch_family == 9 and sparse_block_size != 64:
         raise NotImplementedError(f"SM{arch} only provides a blk64 forward path")
     if arch_family == 9:
         if head_dim not in {64, 96, 128} or value_dim not in {64, 96, 128}:
@@ -218,6 +218,10 @@ def block_sparse_attention_forward(
     elif arch_family == 12:
         if head_dim != 128 or value_dim != 128:
             raise NotImplementedError("SM120 forward requires QK and V dimensions of 128")
+        if sparse_block_size == 128 and q_tensor.dtype != torch.bfloat16:
+            raise NotImplementedError("SM120 blk128 forward requires BF16")
+        if sparse_block_size == 128 and seqlen_k % 128:
+            raise NotImplementedError("SM120 blk128 forward requires seqlen_k to be a multiple of 128")
     elif sparse_block_size == 64:
         if q_tensor.dtype != torch.bfloat16 or head_dim != 128 or value_dim != 128:
             raise NotImplementedError("SM100/SM110 blk64 forward requires BF16 and QK=V=128")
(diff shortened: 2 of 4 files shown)


tiffany940107:
Maintainers: could you please apply `cat-feature`, `mod-cutedsl`, `mod-frontend`, and `orig-external`, then add the appropriate milestone/project? Fork authors do not have permission to set these fields. Local pre-commit and the full BSA test suite pass; details are in the PR description.

Anerudhan:
@cudnn-ci-bot run oss

YangXu1990uiuc (approved):

Reviewed `1d5ed9d596d5` against base `2ed6458726c0`.

Reusing the native blk64 kernel through logical metadata lowering makes sense. I found one P2 at the newly enabled zero-sized physical-half boundary below. A non-blocking follow-up would let callers reuse prepared metadata when the sparse pattern is unchanged, since the per-call lowering cost is visible on small work. The existing documentation clarification at https://github.com/NVIDIA/cudnn-frontend/pull/1010#discussion_r3986399854 is also worth retaining.

**[P2] Skip the empty physical half before consuming V** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/1d5ed9d596d51087bcc2f81cc44a1f7ea260189f/python/cudnn/block_sparse_attention/csrc/fwd/sm120_blk128/metadata.py#L30))

A valid logical block_sizes=[64] becomes physical sizes [64, 0], but both physical blocks remain in the active index prefix. The native kernel still loads V and executes P@V for the zero-sized half; masking its scores to -inf does not prevent 0 * NaN. For BF16 B=H=1, S_q=S_kv=D=128, Q=K=0, one selected logical block 0, V[:64]=1 and V[64:]=NaN, all output elements become NaN although the valid-token result is exactly 1 and LSE=log(64). The equivalent native blk64 call selecting only physical block 0 stays finite. Please make zero-sized physical halves skip their V contribution (and corresponding work safely), or omit them from the active physical metadata, and add this poisoned-padding regression. Finite random padding does not exercise it.

Evidence: Source trace: lower_sm120_blk128_metadata doubles every selected index/count; bsa_fwd_sm120.py applies the score mask but unconditionally executes the V MMA. Directly compiling and running that same SM120 kernel source on an actual SM100 GPU reproduced 16,384 NaN output elements with the new lowering; both the finite-tail case and equivalent native blk64 control returned exact ones and log(64). This is source plus SM100 execution evidence, not a native SM120 run. Attribution: Newly enabled combination in this PR. The native kernel behavior predates it, but the lowering now turns a valid positive caller-visible block size into an actively consumed zero-sized physical block.

Validation: The five new metadata tests passed; the three SM120-only API tests were collected and skipped on the available SM100 GPU. The reviewed Python source used bindings checked against the same C++ sources. Eight additional finite-input O/LSE reference checks passed using the SM120 kernel source directly on SM100 with CuTe DSL 4.8: B=2, Q tails 129/191/193/255, MHA/GQA/MQA, fixed and variable counts, and absent/rank-1/rank-2/rank-3 block_sizes. These tests used a token mask constructed from the original logical metadata; maximum O error was below 0.00374. A component spot check compared fresh lowering plus the native kernel against the same kernel with prelowered metadata, at BF16 B1/H2/S_q193/S_kv512/D128 and three selected logical blocks. ABBA runs on a 68-SM SM100 device gave approximately 23.69 vs 13.74 us GPU time and 240.5 vs 107.8 us warmed CPU call/enqueue time. GPU timing used 16 calls per captured graph and seven batches; changed-V and poisoned-output replay checks passed. This measures adapter overhead on that device, not a regression of an existing blk128 path. A disposable merge onto develop 2ed6458726c0 had no conflicts. The new base commit changes only SDPA backward kernels/tests; the native SM120 BSA kernel and launcher remain unchanged. Existing exact-head CI was inspected; no duplicate run was requested.

Limitations: No native SM120 GPU was available for independent execution. The SM120 public API cases, target-device performance, and native-device compilation still need target-hardware validation; the direct private-launcher checks above do not substitute for that. The performance numbers are a small component comparison with prepared physical metadata as the control, not an end-to-end workload result or a baseline/head SM120 benchmark.

jiayus-nvidia:
This is not the sol implement of blk128 forward kernel on SM120. Close for now, and feel free to reopen it once there is a dedicated high-performance kernel implementation.

tiffany940107:
## Progress update: dedicated native KV128 implementation

The follow-up implementation is now on the independent
[`bsa-sm120-native-blk128-fa4-style` branch](https://github.com/tiffany940107/cudnn-frontend/tree/bsa-sm120-native-blk128-fa4-style)
at `97102524` (kernel checkpoint `e0602027`). This comment does **not** change
this PR's legacy adapter implementation or reopen it.

The new implementation directly consumes blk128 sparse metadata and loads
complete physical KV128 blocks. It does not expand metadata into blk64 or
call a blk64 kernel.

Adopted work:

- General native KV128 forward kernel and a fixed-top-k FA4-style specialization.
- Register-resident Q, dedicated K/V TMA load warp, independent K/V synchronization,
  early sparse-address preparation, and shared-memory reuse for the O epilogue.
- Two-fold, then four-fold producer/compute sparse-loop unrolling.
- Warpgroup-uniform register donation (correctness hardening).
- Underfilled final-wave scheduling: only final Q128 work may become two Q64
  CTAs; **each CTA still uses KV128 and the original parent-Q sparse metadata**.
  This is not the KV128-to-KV64 lowering discussed in this PR. It uses one
  kernel launch, no additional scratch, and no SM-count query on cache hits.
- Reusable paired benchmark, accuracy coverage, and an explicit 1.05x gate.

### Current paired performance

One RTX PRO 6000 Blackwell **Server Edition**, SM120; BF16 Q/K/V BHSD
`[1, 8, 142720, 128]`; ten warmup pairs and 101 timed A/B pairs per case.
The baseline `9869b9b6` already includes FA4-style execution and two-fold
unrolling. Compilation and reference checks are excluded from timing.

| Actual density | Pattern | Baseline | Current | Speedup |
| ---: | --- | ---: | ---: | ---: |
| 14.9776% | strided | 31.5949 ms | 31.3348 ms | 1.0083x |
| 14.9776% | local | 31.7997 ms | 31.5128 ms | 1.0091x |
| 20.0000% | strided | 43.1518 ms | 42.8118 ms | 1.0079x |
| 20.0000% | local | 44.2680 ms | 43.7693 ms | 1.0114x |

**The requested additional 1.05x has not been achieved.** These are cumulative
ratios against that baseline. The isolated final-wave improvement is
1.0033x-1.0050x. Earlier Workstation Edition measurements of the initial
native-to-FA4 change showed a 3.77% latency reduction at 20%; their absolute
times and dense baseline must not be mixed with this Server Edition table.

### Validation and scope

- September 15 refresh: `cd test/python && CUDA_VISIBLE_DEVICES=0 python -m pytest -q fe_api/bsa`:
  **53 passed, 19 skipped**. This validates the SM120 CuTe BSA path, not other
  GPU architectures or backend Graph API CI.
- `pre-commit run --from-ref upstream/develop --to-ref HEAD`: passed applicable
  Black/Black Jupyter/SPDX checks; no changed C++ files.
- Separate full-target comparisons found O and LSE bitwise identical to the
  baseline at both densities and both address patterns. The public benchmark
  itself checks representative FP32-reference rows, not full-tensor equality.
- Q/K/V remain BF16, with FP32 accumulation and BF16 probabilities. No
  FP8/INT8 quantization or accuracy-tolerance relaxation is enabled. A stricter
  exploratory reference-bound exceedance reproduced by the unchanged kernel
  is explicitly recorded in the screening log, not counted as a passing test.
- Variable counts, explicit valid block sizes, and partial KV use the general
  native kernel. The SM120 default remains blk64; no public signature change.
- Multi-GPU communication/compute-overlap prototypes remain in the separate
  Block-Sparse-Attention repository and are outside this cuDNN branch.

Review entry points:

- [Adopted changes, results, limitations, and reproduction commands](https://github.com/tiffany940107/cudnn-frontend/blob/97102524/benchmark/bsa/SM120_REVIEW_SUMMARY.md)
- [Detailed screening and profiler evidence, including rejected experiments](https://github.com/tiffany940107/cudnn-frontend/blob/97102524/benchmark/bsa/SM120_OPTIMIZATION.md)
- [Native specialized kernel](https://github.com/tiffany940107/cudnn-frontend/blob/97102524/python/cudnn/block_sparse_attention/csrc/fwd/sm120_blk128/bsa_fwd_sm120_fa4.py)

Machine-specific launch files, raw profiler reports, and rejected prototype
sources were not pushed. Shared results omit machine IPs, hostnames, and GPU UUIDs.

PR #1017, opened 2026-09-11
linear_attention: add a Hopper (sm90) KDA prefill path

Anerudhan:
## What

Adds a Hopper (sm90) path for KDA (Kimi Delta Attention) prefill, so
`cudnn.linear_attention.kimi_delta_attention` works on H100/H200 instead of
falling back to a much slower engine — and makes it **faster than FlashKDA**,
the CUTLASS kernel the Kimi team ship and vLLM adopted.

## Why

KDA is shipped by Kimi K3 (69 of 93 layers) and GLM-5.3-Flash (34 of 45), so
Hopper deployments of those models care about this path.

FROST declines sm90 (`linear_attention/frost/engine.py` gates
`100 <= sm <= 103 or sm == 107`) because its KDA kernel is built on Blackwell-only
hardware — 42 `tcgen05`, 84 `tmem`, 30 `make_tmem_ptr` references across 3,108
lines of prefill. **This is not a port**: Hopper has neither Tensor Memory nor
`tcgen05` MMA, so relaxing the arch gate cannot work. sm90 needs its own
schedule. That leaves cuTile, which requires the `cuda.tile` runtime and is
~2.1–2.6× slower on GPU time than what this PR adds.

## How

A CuTe DSL kernel: a chunk-parallel PREP pass over **16-token** chunks feeding a
**segmented affine scan**, with the `[Dv, Dk]` state held transposed in fp32
`wgmma` accumulators.

The numerics are the hard part and drove the design. The chunked UT/WY form puts
`k / exp(cumulative decay)` in the inner loop, and production gates
(`gate_lower_bound = -5`, mean log-decay ≈ −2.5) make a 64-token chunk span ~118
in the exponent — fp32 overflows at 88. This kernel arranges **every exponent
reaching `exp2` to be ≤ 0**, so the exponential can only underflow to zero:
overflow is structurally impossible rather than merely budgeted. The segmented
scan buys back the serial-chain length that a 16-token chunk would otherwise
cost.

`BT ≤ 16` is forced, not conventional. Anchoring per 16-token sub-block inside a
larger chunk does **not** work: with per-token anchors the UT product picks up a
residue

```
(Wn Uᵀ)_ij = Σ_d −k_i[d]k_j[d]β_j · exp(cs_i[d]−cs_j[d]) · exp(r_j[d]−r_i[d])
```

and because KDA's decay is **per-key-channel**, `exp(r_j[d]−r_i[d])` sits inside
the channel sum where no post-matmul scalar can remove it. Every token in one
matmul must share an anchor — the same reason FlashKDA and cuDNN's own Blackwell
KDA kernel both use a 16-token chunk.

**Provenance:** produced by a Kernel Factory campaign (`cute_dsl`,
`gpu_spec h100`, campaign `m58kq4q63h0zn893p6sxywe9g4`, solution
`kda_sm90_segscan_leader_checkpoint`) and vendored under
`linear_attention/hopper/kernel/`. It is machine-generated and reformatted to
repo style. It is included because it is measurably correct and fast, **not**
because it was reviewed line by line — reviewers should treat it accordingly.

## Performance

All four paths measured in **one job on one node** (`ipp2-0161`, H100 80GB
HBM3), same harness, same inputs, production gate (`gate_lower_bound = -5`),
non-zero `initial_state`, median of 20 iterations. FlashKDA is re-measured in
each section as a control and came out at **427.1 / 427.1 / 428.7 µs** — within
0.4%, so the sections are directly comparable.

| engine | per-call | vs FlashKDA | pipelined | vs FlashKDA |
|---|---|---|---|---|
| **`kda_hopper` (this PR)** | **392.4 µs** | **1.09× faster** | **267.1 µs** | **1.45× faster** |
| FlashKDA | 427.1 µs | — | 387.0 µs | — |
| `kda_cutile` (in tree) | 758.2 µs | 0.56× | 603.4 µs | 0.64× |
| fla (Triton `chunk_kda`) | 902.1 µs | 0.48× | 797.3 µs | 0.48× |

Against the engine it would replace on Hopper, `kda_hopper` is **1.93× faster
per-call and 2.26× pipelined** than `kda_cutile`.

Per shape, per-call (µs):

| T, H, N | `kda_hopper` | `kda_cutile` | FlashKDA | fla |
|---|---|---|---|---|
| 2048, 12, 1 | **313.2** | 525.4 | 289.5 | 787.7 |
| 2048, 16, 1 | **320.5** | 521.4 | 291.4 | 857.7 |
| 4096, 8, 1 | **346.9** | 601.4 | 482.5 | 824.0 |
| 4096, 12, 1 | **365.4** | 623.3 | 501.7 | 790.8 |
| 8192, 12, 4 | **414.2** | 843.4 | 346.1 | 822.2 |
| 8192, 16, 1 | **504.2** | 981.2 | 974.7 | 926.4 |
| 1024, 4, 1 | **276.2** | 539.0 | 159.2 | 820.7 |
| 3072, 10, 1 | **331.9** | 572.9 | 380.5 | 860.3 |
| 8192, 24, 1 | **603.6** | 1398.0 | 1016.6 | 1151.4 |
| 16384, 16, 8 | **589.0** | 1710.3 | 476.0 | 1309.0 |

`kda_hopper` beats `kda_cutile` on **all ten shapes**. It beats FlashKDA on six
and loses on four, all short-GPU-phase cases (see below).

Accuracy is equivalent across all four: 4.7e-03 … 9.3e-03 against an fp64
oracle, versus FlashKDA's own 6.6e-03 … 9.8e-03 on the same inputs.

### GPU time, so the wall-clock is not misread

Wall clock understates the kernel gap, because every cuDNN engine pays the same
op-layer dispatch on top of its kernel:

| T=2048, H=12, N=1 | wall | **GPU** | host |
|---|---|---|---|
| `kda_hopper` | 343.5 µs | **97.3 µs** | 246.2 µs |
| `kda_cutile` | 553.2 µs | **203.4 µs** | 349.8 µs |

On GPU time `kda_hopper` is **2.1× faster than cuTile** (2.6× at T=4096: 151.8
vs 402.0 µs). The remaining wall-clock difference is dispatch, which both share.

That also explains the four shapes where FlashKDA still wins: at 2048/12/1 the
GPU phase is only 97 µs, and the ~209 µs of CuTeDSL launch cost cannot be
amortised against it. FlashKDA is a C++ extension paying ~45 µs of host. Where
the GPU phase is long the ordering flips decisively — 8192/16/1 at **1.93×**,
8192/24/1 at 1.68×.

Reproducibility: an independent two-pass run on a different node of the same
GPU model (`ipp2-0177`) gave 408.1 µs per-call / 278.9 µs pipelined for
`kda_hopper` against FlashKDA's 441.9 / 399.5 — 1.08× / 1.43×. FlashKDA moved by
the same ~4%, so the ratios are stable across nodes.

### Dispatch overhead

Reaching the kernel through `kimi_delta_attention` was originally **0.72×
per-call** — GPU time was identical through both paths, but ~250 µs of Python
sat on top. cProfile found two pieces of pure waste:

- the kernel re-converted **its own cached workspace** on every launch (11 of 20
  DLPack conversions per call, on buffers whose pointers never change);
- every operand was **converted twice** — the engine materialised torch tensors
  from the `OperandBuffer` views purely so the kernel could convert them again
  to CuTe.

`_ws` now caches converted CuTe tensors, and the engine converts straight to
CuTe via a new `run_cute()` entry point, passing the stream explicitly instead
of pushing a `torch.cuda.stream` context for the kernel to read back.
**211 µs removed**; cuDNN's dispatch now costs **+6 µs** over calling the kernel
directly at T=4096 (+37 µs at T=2048), down from +251 µs.

What remains is the CuTeDSL runtime's own JIT launch marshalling plus nine
genuinely unavoidable per-call operand conversions — paid identically by a
standalone caller, so reducing it further means changing `nvidia_cutlass_dsl`.

## Correctness

Validated against an fp64 chunked oracle that is itself bit-exact against
`test/python/linear_attention/reference_kda.py`, at the production gate with a
non-zero `initial_state`:

**5.0e-03 … 1.0e-02**, against **FlashKDA's own 7.5e-03 … 1.0e-02** on identical
inputs — i.e. at least as accurate as the incumbent, on the gate distribution
that broke earlier attempts.

## Scope — and what is refused

Deliberately narrow. Everything outside it **declines** rather than being
silently mis-served:

| case | behaviour |
|---|---|
| `initial_state` | **supported** (V-major `[N, H, V, K]`); a graph that omits it gets a zero seed |
| `initial_state` dtype | fp32 only — the kernel holds the state in fp32 `wgmma` accumulators, so a bf16 pool would be reinterpreted, not converted |
| backward | declines — no Hopper KDA bwd kernel yet |
| `checkpoint_every_n_tokens` | declines |
| `safe_gate` / `a_log` / `dt_bias` | declines |
| `use_beta_sigmoid_in_kernel` | declines |
| `use_qk_l2norm_in_kernel` | declines — q/k must be pre-normalised |
| grouped heads | declines — q/k/v head counts must match |
| non-128 head dim, non-bf16 tokens, non-fp32 g/beta | declines |
| int64 `cu_seqlens` | declines |
| non-default `scale` | declines |

Several of these exist **because the expanded test matrix caught them**: adding
`hopper` to the backend fixture surfaced `test_fwd_scale` (the kernel bakes in
`1/sqrt(D)` and silently ignored a custom scale — rms ratio 0.91) and
`test_cu_seqlens_int64`; enabling `initial_state` later surfaced the fp32
requirement. All decline instead of returning a wrong answer.

## Test coverage

`hopper` joins the `backend` fixture in `test/python/linear_attention/test_la.py`.
`pytest -m L0` on H100 SXM:

| selection | result |
|---|---|
| all backends | **420 passed, 3554 skipped, 1 xfailed, 0 failed** |
| `-k hopper` | **56 passed, 1020 skipped, 0 failed** |
| `-k "hopper and kda"` | **47 passed, 216 skipped, 0 failed** |

One test changed: `test_bwd_split_initial_state` called `autograd.grad` *outside*
its `waive_unsupported` block, so a forward-only backend could not waive the
backward once its forward began being served. Moved inside, matching its sibling
`test_bwd_split_d_final_state`, which already does exactly that.

## Engine selection — follow-up, deliberately not in this PR

Selection on sm90 is **not** a fall-through to this engine. Asking each KDA
engine directly on an sm90 graph (H100 80GB HBM3):

| engine | slot | sm90 |
|---|---|---|
| `kda_frost` | 0 | declines |
| `kda_cutile` | 1 | **accepts** |
| `kda_summary_frost` | 2 | declines |
| `kda_cake` | 3 | declines |
| `kda_hopper` | 4 | **accepts** |

The offered set is exactly `['kda_cutile', 'kda_hopper']`. The three FROST/cake
engines are Blackwell-gated and do drop out, but cuTile is not — it has a
working sm90 KDA path (correct at the production gate, 205.4 µs of GPU time at
2048/12/1 against this kernel's 97 µs). Since the KDA family declares
`heuristics=None`, the first offered engine wins, so **an unpinned
`kimi_delta_attention()` gets cuTile**: verified by profiling a default call,
which launches `chunk_gated_delta_rule_fwd_kernel_h_*` /
`chunk_kda_fwd_kernel_inter_*`, not this kernel.

**This is conditional on the `cuda.tile` runtime being installed.** cuTile is an
optional extra (`pip install -e ".[cutile]"`, needs a system tileiras). Without
it cuTile declines, the offered set collapses to `['kda_hopper']`, and selection
*does* fall through to this engine with no heuristic needed. With it — including
the NGC container every number above was measured in — cuTile wins.

So preferring `kda_hopper` on sm90 is justified on the numbers, but it only
changes anything for cuTile-enabled installs, it is a default-behaviour change
for those users, and it belongs in the family `heuristics` hook
(`recommend(kind, facts, offered) -> [PlanConfig]`) with its own review.

## Revision note

Earlier revisions of this PR vendored a different kernel and quoted a ~3.88×
figure against fla. Both were wrong and have been corrected in the thread: that
kernel returned **100% NaN at `gate_lower_bound = -5`** (and was silently wrong
from ≈ −2.5), and the benchmark behind the figure called `kimi_delta_attention`
without `plan_name`, so it was measuring **cuTile** rather than this engine. All
numbers above are measured with the plan pinned, at the production gate.


Anerudhan:
## ⚠️ Do not merge — the kernel produces NaN on the production gate configuration

Follow-up benchmarking against [FlashKDA](https://github.com/MoonshotAI/FlashKDA)
surfaced a correctness failure that the original validation missed. Posting it
here rather than quietly fixing, because it invalidates the merge case.

### What happens

The vendored kernel anchors each `BT = 64` chunk at `r = cs_31` to keep every
`exp` argument within ±40 (its own docstring says so). The half-chunk cumulative
gate sum is roughly `32 · mean(g)`, so that budget is consumed in direct
proportion to how negative the gate is. Sweeping gate magnitude at
`T=2048, H=12, D=128`, error measured against the fp64 oracle:

| gate | `32·\|mean g\|` | this PR | fla Triton |
|---|---|---|---|
| the campaign's own generator, `log(sigmoid(randn))` | 25.8 | 6.21e-03 ✓ | 9.32e-03 |
| `safe_gate lower_bound = -1.0` | 16.0 | 6.29e-03 ✓ | 6.29e-03 |
| `safe_gate lower_bound = -2.0` | 32.0 | 6.25e-03 ✓ | 6.25e-03 |
| `safe_gate lower_bound = -3.0` | 48.0 | **6.90e-02** ✗ | 6.25e-03 |
| `safe_gate lower_bound = -5.0` | 80.0 | **NaN** ✗ | 6.25e-03 |

Degradation begins precisely where the ±40 budget is exceeded, and becomes a
hard NaN at −5.0. fla is stable across the whole range.

### Why it matters

**`-5.0` is cuDNN's own `DEFAULT_GATE_LOWER_BOUND`, and it is what Kimi K3 and
GLM-5.3-Flash ship** (`gate_lower_bound: -5.0` in both configs). So the failing
row is the production configuration for the two models this work targets, not an
exotic corner.

There is no safe way to gate around it: the failure is **data-dependent**, and
`check_support` runs at graph-build time with no access to tensor values.

### Root cause — mine, not the kernel's

The operation definition I wrote for the campaign generated the gate as
`log(sigmoid(randn))`, mean ≈ −0.8. Production KDA with `lower_bound = -5`
averages ≈ −2.5. The campaign therefore optimised, and its correctness gate
validated, against a gate distribution roughly 3× milder than the models
actually use. The kernel is a correct solution to the problem I specified; the
problem I specified was wrong.

The earlier 3.88× figure was measured on that same mild distribution, so it is
also not representative. Re-measured on safe-gate inputs the margin narrows to
**2.20× geomean over 10 shapes** (cuDNN 411 µs vs fla 904 µs) — but those runs
are the ones producing NaN at `-5.0`, so the number should not be quoted either.

### What needs to happen

1. Re-run the campaign with a realistic gate in `generate_inputs`
   (`lower_bound = -5` safe-gate), so the search optimises against, and the
   correctness gate rejects on, the distribution that matters.
2. The fix is likely bounded — either a smaller `BT`, or a per-chunk rather than
   mid-chunk anchor. The existing FROST KDA kernel uses `B_T = 16` for exactly
   this reason.
3. Re-benchmark and re-verify before this is considered again.

Converting to draft.

### Unrelated note on FlashKDA

FlashKDA builds for `sm90a` but fails at runtime on ipp2-1949 with
`cudaErrorInvalidResourceHandle`, which looks like a CUDA version mismatch on
that host (torch `cu130`, venv nvcc 13.3, staged toolchain 13.2) rather than a
FlashKDA defect. So there is no FlashKDA column above. Worth noting it *does*
support `initial_state`, which this kernel does not.

Anerudhan:
## Blackwell regression: clean

The sm100 regression I promised above has completed on a stable tree:

```
$ pytest test/python/linear_attention/test_la.py -k kda
422 passed, 549 skipped, 3003 deselected, 1 xfailed  in 2038s (0:33:58)
```

No regression from adding the `kda_hopper` engine or the third `backend`
fixture parameter: the engine declines on any non-sm90 device, and `hopper`
waives on Blackwell through the existing `waive_unsupported` path.

Note this run was repeated deliberately. The first attempt produced the same
numbers but overlapped with a branch checkout that rewrote files mid-run, so it
was discarded rather than reported.

This does **not** change the `[DO NOT MERGE]` status — the blocker is the
data-dependent NaN at `gate_lower_bound = -5.0` described in the previous
comment, which is a Hopper correctness issue, not a Blackwell regression.

Anerudhan:
## Correction: the NaN was a toolchain artifact — and the speedup was too

Re-ran everything inside `nvcr.io/nvidia/pytorch:26.08-py3` on an H100 PCIe,
with cuDNN FE and FlashKDA both built against that container's single coherent
toolchain (gcc 13.3 / nvcc 13.4 / torch CUDA 13.4). Two things I reported
earlier do not survive.

### 1. The NaN does not reproduce — retracting the blocker

| gate | `32·\|mean g\|` | cuDNN | fla |
|---|---|---|---|
| campaign generator | 25.8 | 5.24e-03 | 5.24e-03 |
| `safe_gate lower_bound = -1.0` | 16.0 | 7.73e-03 | 5.15e-03 |
| `safe_gate lower_bound = -2.0` | 32.0 | 7.85e-03 | 5.24e-03 |
| `safe_gate lower_bound = -3.0` | 48.0 | 7.89e-03 | 5.26e-03 |
| **`safe_gate lower_bound = -5.0`** | **80.0** | **5.43e-03** | 5.43e-03 |

Every gate magnitude is finite and correct, including the `-5.0` production
setting that previously produced NaN. The earlier failure came from a cuDNN FE
module built against CUDA 13.2 headers while torch ran `cu130` on that host —
the same class of mismatch that made FlashKDA fail there with
`cudaErrorInvalidResourceHandle`. My apologies for the false alarm; the
`[DO NOT MERGE]` is withdrawn.

### 2. The 3.88× speedup was also an artifact

Same host, same toolchain, all three contenders, 10 shapes:

| T | H | N | this PR | fla Triton | FlashKDA |
|---|---|---|---|---|---|
| 2048 | 12 | 1 | 494.8 µs | 694.6 | **350.9** |
| 2048 | 16 | 1 | 778.5 | 712.5 | **334.0** |
| 4096 | 8 | 1 | 589.2 | 746.2 | **549.6** |
| 4096 | 12 | 1 | 675.6 | 759.4 | **567.8** |
| 8192 | 12 | 4 | 1051.5 | 932.0 | **400.7** |
| 8192 | 16 | 1 | 1382.7 | 1181.6 | **1075.6** |
| 1024 | 4 | 1 | 433.8 | 725.1 | **173.8** |
| 3072 | 10 | 1 | 545.3 | 724.4 | **439.8** |
| 8192 | 24 | 1 | 1815.3 | 1513.7 | **1165.7** |
| 16384 | 16 | 8 | 2177.3 | 1912.5 | **721.7** |

**Geomean: this PR 854.9 µs · fla 927.2 µs · FlashKDA ≈ 504 µs.**

- vs fla Triton: **1.08×** — not 3.88×, and this PR is *slower* on 5 of 10 shapes.
- vs [FlashKDA](https://github.com/MoonshotAI/FlashKDA): **0.59×**, i.e. FlashKDA
  is about **1.7× faster** than this kernel.

Because the same bad toolchain produced both the phantom NaN and the inflated
speedup, I am treating every measurement from that host as void rather than
salvaging the ones that happened to look right.

### What this means for the PR

The correctness case stands: the kernel is correct across the gate range and
across 12 shapes including six outside the campaign's tuning set, and Blackwell
is unaffected (422 passed / 0 failed). The *performance* case largely does not.
A ~8% geomean edge over fla, while losing to FlashKDA by 1.7×, is a weak reason
to vendor ~2.3k lines of machine-generated kernel into the repo.

Honest options, in the order I would rank them:

1. **Re-run the campaign against FlashKDA as the baseline solution** rather than
   the fp64 reference. The campaign optimised against a slow Python reference, so
   it had no pressure to beat a real kernel; supplying FlashKDA via
   `--baseline-solution` would put the search under the right constraint. This is
   the fix most likely to produce something worth landing.
2. **Close this PR** and instead point Hopper KDA users at FlashKDA, which is
   faster, already supports `initial_state` (which this kernel does not), and is
   what vLLM is moving toward (vllm-project/vllm#55737).
3. Land it anyway for the narrow benefit of an in-graph cuDNN path with no extra
   dependency — but I do not think an 8% margin justifies the maintenance.

I would not merge this as it stands. Leaving it as a draft pending a call on
which of the above to pursue.

### Caveat on hosts

The earlier numbers came from an H100 SXM-class box, these from an H100 PCIe.
Some gap is expected from HBM bandwidth, and notably fla barely moved across the
two hosts (904 → 927 µs) while this kernel doubled (411 → 855 µs), which would be
consistent with it being more bandwidth-bound. But since that host's toolchain
also produced a phantom NaN, I am not treating its timings as evidence either
way.

Anerudhan:
## Confirmed on H100 SXM: the conclusion holds, and my bandwidth hypothesis was wrong

Re-ran the same container, toolchain and harness on an **H100 80GB HBM3 (SXM)**
(computelab `ipp2-0177`), so the only variable versus the previous run is the
GPU. I had suggested the PCIe numbers might understate this kernel because it
looked bandwidth-bound while fla looked overhead-bound. That is not what
happened.

| | this PR | fla | FlashKDA | vs fla | vs FlashKDA |
|---|---|---|---|---|---|
| H100 PCIe (HBM2e) | 854.9 µs | 927.2 | ≈504 | 1.08× | 0.59× |
| **H100 SXM (HBM3)** | **752.0 µs** | **870.6** | **≈421** | **1.16×** | **0.56×** |

Per-shape on SXM:

| T | H | N | this PR | fla | FlashKDA |
|---|---|---|---|---|---|
| 2048 | 12 | 1 | 514.6 | 768.4 | **278.6** |
| 2048 | 16 | 1 | 533.2 | 772.4 | **287.4** |
| 4096 | 8 | 1 | 585.4 | 768.1 | **462.8** |
| 4096 | 12 | 1 | 591.2 | 742.5 | **481.9** |
| 8192 | 12 | 4 | 860.0 | 802.0 | **344.8** |
| 8192 | 16 | 1 | 989.1 | 933.3 | **937.7** |
| 1024 | 4 | 1 | 478.2 | 758.4 | **153.0** |
| 3072 | 10 | 1 | 551.4 | 822.7 | **377.6** |
| 8192 | 24 | 1 | 1509.9 | 1153.9 | **1003.2** |
| 16384 | 16 | 8 | 1797.7 | 1371.8 | **503.0** |

Both clean hosts agree to within ~8%, which also confirms that the 411 µs I
originally reported from ipp2-1949 was a bad measurement rather than a
PCIe-versus-SXM effect.

The gate sweep is clean on SXM too (all magnitudes finite, including
`lower_bound = -5.0`), so the NaN retraction is now confirmed on two independent
hosts.

### Where that leaves this PR

Correct on both GPUs, across the full gate range and 12 shapes including six
outside the campaign's tuning set; Blackwell unaffected. But ~1.16× over fla
while losing to FlashKDA by 1.8× does not justify vendoring ~2.3k lines of
machine-generated kernel.

I am pursuing the fix rather than arguing for this as-is: re-running the campaign
with **FlashKDA as the baseline solution** instead of the fp64 reference, so the
search is actually constrained to beat a real kernel, and with **`initial_state`
added to the operation definition** (this kernel seeds only from zero, so it
cannot serve chunked-prefill continuation — FlashKDA can, which is part of why it
wins).

This PR stays a draft until that produces something worth landing, or until it
is clear it will not, in which case the honest outcome is to close it and point
Hopper KDA users at FlashKDA.

Anerudhan:
## Correction: the benchmark was measuring the wrong engine, and my NaN retraction was wrong

Two things I posted earlier on this PR are wrong, and they have the same root
cause. Retracting both, with the measurements.

### The benchmark never ran this PR's kernel

`bench3_hopper.py` — which produced every number I posted here — calls
`kimi_delta_attention(...)` without a `plan_name`. The test suite pins the
engine (`plan_name="kda_hopper"`). **Plain dispatch on sm90 selects
`kda_cutile`.** Confirmed under CUPTI by the kernel names that actually launch:
`chunk_gated_delta_rule_fwd_kernel_h_*`, `chunk_kda_fwd_kernel_inter_*` — cuTile
mangling, not the vendored CuTeDSL kernel.

(The import resolved correctly into the worktree and the Hopper engine module
was present, so this is not the editable-install trap in AGENTS.md. It was
default engine *selection*.)

Pinning the engine, H100 PCIe, identical inputs:

| T, H, N | default (cuTile) | `kda_hopper` |
|---|---|---|
| 8192, 12, 4 | 1065.4 µs | **448.8 µs** |
| 8192, 16, 1 | 1387.4 µs | **642.7 µs** |

### Corrected numbers

10 shapes, H100 PCIe, median of 20, `gate_lower_bound = -1` (see below for why
not -5):

| | geomean | |
|---|---|---|
| **this PR, pinned `kda_hopper`** | **397.2 µs** | — |
| FlashKDA | 492.5 µs | this PR is **1.24× faster** |
| Triton `fla` | 928.5 µs | this PR is **2.34× faster** |

Wins 7 of 10 shapes; loses `(8192,12,4)` 0.91×, `(1024,4,1)` 0.75×,
`(16384,16,8)` 0.70× — large-batch / short-sequence, where the serial state scan
is not the bottleneck and FlashKDA's fully-parallel prepare kernel dominates.

So my earlier claim that this kernel loses to FlashKDA by 1.8× was a measurement
of cuTile, not of this PR.

### The NaN is real and I should not have retracted it

I reported a NaN at `gate_lower_bound = -5`, then retracted it when a clean
container "failed to reproduce" it. That reproduction attempt used the same
unpinned harness, so it exercised cuTile and never touched this kernel. The
retraction was wrong.

Pinned to `kda_hopper`, clean container, H100 PCIe:

| gate lower bound | result |
|---|---|
| **−5.0** (what the shipped models use) | **100% NaN** — every element of `o` and `final_state`, at 1024/4/1, 2048/12/1, 4096/12/1, 8192/12/4, 8192/16/1 |
| −1.0 | correct — rel(o) 3.8e-03…1.1e-02, rel(state) 3.1e-03…4.1e-03 |

Total, not sporadic, and not data-dependent.

**Root cause.** The kernel uses `BT = 64` with a mid-chunk anchor `r = cs_31`,
and its docstring asserts that this "keeps every exp argument inside ±40". That
holds for a gate with mean ≈ −0.6. Production gates have mean log-decay ≈ −2.5,
so a 64-token chunk spans ~160 in the exponent and ~±80 even after anchoring;
`exp(88)` already overflows fp32. FlashKDA and cuDNN's own Blackwell KDA kernel
both use a 16-token chunk for precisely this reason.

The 397.2 µs above is therefore measured at `lb = −1`, where the output is
correct. At `lb = −5` the kernel is fast and useless.

### Where this leaves the PR

Better and worse than I said. The kernel is genuinely faster than FlashKDA —
1.24× geomean, not 0.56× — but it breaks on the gate distribution the models
actually ship with, so it cannot merge as-is. This stays a draft.

The work item is now a numerical fix, not a performance one. I have also fixed
the Kernel Factory operation definition, which was generating
`g = log(sigmoid(randn))` (mean ≈ −0.69) instead of the production
`g = −5·sigmoid(…)` (mean ≈ −2.5) — a gate 3.6× milder than reality, which is
exactly why the search produced a kernel that overflows and never noticed.

Apologies for the noise on the earlier numbers; the engine-pinning mistake
invalidated both the performance table and the retraction built on top of it.

Anerudhan:
## Complete picture: production gate, pipelined — and a second defect

Follow-up to the correction above, with the measurement I should have led with.
Two changes in method:

- **Production gate.** All numbers at `gate_lower_bound = -5`, what the shipped
  models use — not the mild setting where this kernel happens to work.
- **Pipelined timing.** My earlier harness synchronised every iteration, which
  charges cuDNN's ~200 µs of Python dispatch in full while FlashKDA (a thin C++
  extension) pays ~45 µs. Serving issues calls back to back, so host work for
  call *n+1* overlaps GPU work for call *n*. Both are reported below; the
  pipelined column is what a user actually pays.

H100 PCIe, 4 shapes, error against an fp64 oracle:

| engine | per-call geo | **pipelined geo** | correct at lb = −5 |
|---|---|---|---|
| `kda_hopper` (this PR) | 415.7 µs | **267.7 µs** | **no — 100% NaN** |
| FlashKDA | 533.2 µs | **482.5 µs** | yes (8.2e-03…1.1e-02) |
| `kda_cutile` (already in tree) | 830.6 µs | 672.8 µs | yes (5.5e-03…1.0e-02) |
| Triton `fla` | 886.7 µs | 755.3 µs | yes (5.0e-03…6.9e-03) |

Per shape, pipelined (µs):

| T, H, N | this PR | FlashKDA | cuTile | fla |
|---|---|---|---|---|
| 2048, 12, 1 | 201.5 | 274.0 | 407.8 | 680.8 |
| 4096, 12, 1 | 220.0 | 523.5 | 472.4 | 693.0 |
| 8192, 12, 4 | 256.6 | 361.2 | 885.6 | 709.7 |
| 8192, 16, 1 | 451.6 | 1046.5 | 1201.0 | 971.9 |

**The state of play: the fast path is broken and the correct path is slow.**
cuTile — already in tree — is correct at production gates and 1.39× slower than
FlashKDA. This PR's kernel would be **1.80× faster** than FlashKDA if it were
correct (NaN arithmetic costs the same, so the timings stand; only the answers
don't). That is the size of the prize, and it is why I think this is worth
fixing rather than abandoning.

### Why the overflow is not a tuning knob

Sweeping gate steepness and recording the largest `|cs_i − anchor|` the kernel
actually forms (T=2048, H=12, BT=64, midpoint anchor):

| gate lb | max span | % NaN | rel error |
|---|---|---|---|
| −1.0 | 23.6 | 0% | 5.95e-03 ok |
| −2.5 | 59.1 | 0% | 1.63e-02 marginal |
| −3.0 | 70.9 | **0%** | **7.51e-02 wrong** |
| −4.0 | 94.5 | 42% | — |
| −5.0 | 118.1 | 100% | NaN |

Note it goes **silently wrong well before it goes NaN** — at −3.0 the error is
~4× the tolerance with not a single NaN.

The natural fix — keep `BT = 64`, anchor each 16-token sub-block separately —
**does not work**. With per-token anchors the UT product picks up a residue:

```
(Wn Uᵀ)_ij = Σ_d −k_i[d]k_j[d]β_j · exp(cs_i[d]−cs_j[d]) · exp(r_j[d]−r_i[d])
```

KDA's decay is **per-key-channel**, so `exp(r_j[d]−r_i[d])` sits inside the
channel sum and no post-matmul scalar can remove it. I implemented it to check:
the exponent bound improves exactly as predicted (56.4 → 16.4 at lb = −3) and
the answer is still completely wrong. Every token in one matmul must share one
anchor, which is why FlashKDA and cuDNN's own Blackwell KDA kernel both use a
16-token chunk. `BT ≤ 16` is forced by the per-channel gate, and the ~4× longer
serial chain it implies is the real difficulty.

### Second defect: this engine is never selected

Independent of the numerics. `engines_for()` returns a family's engines in slot
order and the KDA family declares `heuristics=None`, so the first supporting
engine wins. Slots are `kda_frost 0, kda_cutile 1, kda_summary_frost 2,
kda_cake 3, kda_hopper 4`, and on sm90 the offered set is
`['kda_cutile', 'kda_hopper']` — **cuTile always wins**. A plain
`kimi_delta_attention(...)` call never reaches this kernel; that is exactly how
my first numbers here ended up being cuTile's.

`manifest.py` is explicit that slots are "FIXED FOREVER … never reorder", so the
fix is the family's `heuristics` hook
(`recommend(kind, facts, offered) -> [PlanConfig]`), preferring `kda_hopper` on
sm90 + bf16 + THD. **It must not land before the numerics fix** — preferring it
today would hand every Hopper user NaN by default.

(Scope note: cuTile is an optional extra, so in an environment without it
`kda_hopper` would be selected today. The defect bites environments that have
cuTile — including the NGC container used for all of these measurements.)

### Plan

Still a draft. Two Kernel Factory campaigns are running against a corrected
operation definition — the old one generated `g = log(sigmoid(randn))`
(mean ≈ −0.69) instead of the production `−5·sigmoid(…)` (mean ≈ −2.5), a gate
3.6× milder than reality, which is why the search produced a kernel that
overflows and never noticed. They are scored at the production gate and briefed
with the analysis above. If one lands a correct kernel near this schedule's
speed, I will re-verify it here and add the selection heuristic in the same
change.

Anerudhan:
## New kernel: correct at production gates, and faster than FlashKDA at the kernel level

Short version: the kernel this PR now vendors is **correct at the production
gate** (the old one returned 100% NaN there) and **1.11× per-call / 1.53×
pipelined / ~2.6× on GPU time** versus FlashKDA. Reached through
`kimi_delta_attention`, however, the op is **still 0.72× per-call** because of
~250 µs of linear-attention dispatch overhead. Both halves are documented below;
please don't quote the first without the second.

This replaces the kernel this PR originally vendored. Summary of where it
landed, measured on **H100 80GB HBM3 (SXM)** at the production gate
(`gate_lower_bound = -5`) with a non-zero `initial_state`, geomean over ten
shapes, two independent passes each:

| | per-call | pipelined | correct at lb = −5 |
|---|---|---|---|
| **this PR (new kernel)** | **398.1 / 397.1 µs** | **260.8 / 264.9 µs** | **yes** |
| FlashKDA | 442.4 / 442.3 µs | 401.0 / 401.5 µs | yes |
| | **1.11× faster** | **1.53× faster** | |

Accuracy against an fp64 oracle is **5.0e-03 … 1.0e-02**, against FlashKDA's own
**7.5e-03 … 1.0e-02** on identical inputs — i.e. at least as accurate as the
incumbent, on the gate distribution that broke every earlier attempt.

Per shape, per-call (µs):

| T, H, N | this PR | FlashKDA | vs |
|---|---|---|---|
| 2048, 12, 1 | 305.9 | 296.4 | 0.97× |
| 2048, 16, 1 | 328.0 | 302.4 | 0.92× |
| 4096, 8, 1 | 342.5 | 500.3 | **1.46×** |
| 4096, 12, 1 | 363.6 | 514.1 | **1.41×** |
| 8192, 12, 4 | 404.6 | 362.9 | 0.90× |
| 8192, 16, 1 | 521.4 | 986.1 | **1.89×** |
| 1024, 4, 1 | 274.2 | 160.6 | 0.59× |
| 3072, 10, 1 | 330.5 | 398.4 | **1.21×** |
| 8192, 24, 1 | 646.4 | 1048.1 | **1.62×** |
| 16384, 16, 8 | 647.6 | 519.0 | 0.80× |

The win concentrates on long sequences, which is where prefill cost actually
lives. It still loses on 1024/4/1 and 16384/16/8, where the serial chain is short
and per-call launch overhead dominates — the pipelined column shows most of that
gap is launch overhead, not the kernel.

### Important caveat: end to end the op is still slower than FlashKDA

The table above is the kernel at its own entry point. Reached through
`kimi_delta_attention(..., plan_name="kda_hopper")` on the same allocation, with
identical numerics:

| | per-call | pipelined |
|---|---|---|
| kernel direct | 398.1 µs | 260.8 µs |
| via cuDNN engine | **619.5 / 621.2 µs** | **447.9 µs** |
| FlashKDA | 443.1 µs | 401.8 µs |

So end to end the op is **0.72× per-call / 0.90× pipelined** versus FlashKDA.
Please don't read the kernel table as an end-to-end claim.

Profiling says where it goes, and the answer is clean:

| T=2048, H=12, N=1 | wall | **GPU** | host |
|---|---|---|---|
| raw kernel | 308.7 µs | **97.1 µs** | 211.7 µs |
| via cuDNN `kda_hopper` | 551.0 µs | **96.7 µs** | 454.3 µs |
| via cuDNN `kda_cutile` | 593.9 µs | **205.4 µs** | 388.4 µs |

**GPU time is identical through both paths** (96.7 vs 97.1 µs) — cuDNN runs
exactly the same kernel work, and 100% of the regression is host-side. The
hottest frame is `cudnn::kimi_delta_attention_fwd` itself at 470–526 µs of
*self* time per call. It is not a graph rebuild: `fprop_cache` is keyed on
shapes/dtypes/flags and is hitting.

Two things worth separating:

- **The kernel is much better than the wall-clock suggests.** On GPU time it is
  **2.1–2.6× faster than the in-tree cuTile engine** (96.7 vs 205.4 µs;
  151.8 vs 402.0 µs) and roughly 2.6× faster than FlashKDA.
- **The overhead is only partly pre-existing.** cuTile pays 388/273 µs of host
  cost, so the linear-attention op layer is expensive for every engine — but
  `kda_hopper` pays ~200 µs more, which is the CuTeDSL kernel's own Python
  launch (the raw path alone shows 212–232 µs of host).

Reducing that dispatch cost is the remaining work, and it is a change against
shared op-layer code — I'd rather not rush it into this PR.

### Tests

`pytest -m L0 linear_attention/test_la.py` on H100 SXM:

| selection | result |
|---|---|
| all backends | **420 passed, 3554 skipped, 1 xfailed, 0 failed** |
| `-k hopper` | **56 passed, 1020 skipped, 0 failed** |
| `-k "hopper and kda"` | **47 passed, 216 skipped, 0 failed** |

Enabling `initial_state` surfaced three real failures on the first run (3 failed
/ 420 passed). Two root causes, both fixed here rather than papered over:

- `initial_state` must be fp32 — the kernel holds the state in fp32 wgmma
  accumulators, so a bf16 state pool would be reinterpreted rather than
  converted. Now declined (`test_fwd_state_dtype`, `test_dtype_cache_separation`
  waive cleanly).
- `test_bwd_split_initial_state` called `autograd.grad` *outside* its
  `waive_unsupported` block, so a forward-only backend could not waive the
  backward once its forward started being served. Moved inside, matching its
  sibling `test_bwd_split_d_final_state`, which already does exactly that.

`pre-commit run` on the staged files is green (black reformatted the vendored
kernel; SPDX header check passes).

### Engine selection — still a follow-up, now better evidenced

`kda_hopper` sits at manifest slot 4 behind `kda_cutile` at slot 1, and the KDA
family declares `heuristics=None`, so the first supporting engine wins and a
plain `kimi_delta_attention()` never reaches this kernel. On the evidence above
preferring it on sm90 is justified — it is faster than cuTile end to end
(551.0 vs 593.9; 621.6 vs 675.3) on top of the 2.1–2.6× GPU-time win. But it is
a default-behaviour change for every Hopper KDA user, it belongs in the family
`heuristics` hook, and it does not by itself reach "faster than FlashKDA". Left
for its own reviewed change.

### What changed, and why the old kernel had to go

The previously vendored kernel used `BT = 64` with a single mid-chunk anchor and
returned **100% NaN at `gate_lower_bound = -5`** — the setting the shipped models
use. A 64-token chunk spans ~118 in the exponent there; fp32 overflows at 88. It
was also *silently wrong* from about −2.5 upward (7.5e-02 relative error at −3.0
with zero NaNs), which is worse than failing loudly.

Keeping `BT = 64` and anchoring per 16-token sub-block does not rescue it. With
per-token anchors the UT product picks up a residue:

```
(Wn Uᵀ)_ij = Σ_d −k_i[d]k_j[d]β_j · exp(cs_i[d]−cs_j[d]) · exp(r_j[d]−r_i[d])
```

KDA's decay is **per-key-channel**, so `exp(r_j[d]−r_i[d])` sits inside the
channel sum and no post-matmul scalar removes it. Every token in one matmul must
share an anchor, which forces `BT ≤ 16` — the same reason FlashKDA and cuDNN's
own Blackwell KDA kernel use a 16-token chunk.

The new kernel goes further: it arranges **every exponent reaching `exp2` to be
≤ 0**, so the exponential can only underflow to zero. Overflow is structurally
impossible rather than merely budgeted. It pairs that with a segmented affine
scan, which buys back the serial-chain length that `BT = 16` would otherwise
cost.

### `initial_state` is now supported

The old kernel seeded only from zero, so it could not serve chunked-prefill
continuation in vLLM/SGLang. The new one takes `initial_state` (V-major
`[N, H, V, K]`), the engine's `has_initial_state` decline is removed, and a graph
that omits it is handed a zero seed.

### Diff shape

Net **−1,769 lines**: the replacement kernel is 570 lines against the 2,357 it
replaces (`kda_prefill_sm90.py` + the now-orphaned `kda_direct_sm90.py`).

### Provenance

Produced by a Kernel Factory campaign on a `kda_chunked_prefill_sm90` operation
definition (campaign `m58kq4q63h0zn893p6sxywe9g4`, solution
`kda_sm90_segscan_leader_checkpoint`). The definition originally generated
`g = log(sigmoid(randn))` — mean ≈ −0.69 against production's ≈ −2.5, a gate
3.6× milder than reality — which is precisely why the earlier search produced a
kernel that overflows and never noticed. Fixing the gate distribution produced a
correct kernel in a single round.

All measurements above are mine, taken with the same harness on the same
allocation for both kernels, not campaign-reported numbers.

Anerudhan:
## Dispatch overhead fixed — the op now beats FlashKDA too

Follow-up to the caveat in my previous comment. The kernel was already faster
than FlashKDA but `kimi_delta_attention` was **0.72× per-call** because ~250 µs
of Python sat on top. That is fixed: **211 µs removed from per-call dispatch**,
and the op now wins on both metrics.

H100 80GB HBM3 (SXM), production gate (`gate_lower_bound = -5`), non-zero
`initial_state`, geomean over ten shapes, two passes:

| | per-call | vs FlashKDA | pipelined | vs FlashKDA |
|---|---|---|---|---|
| baseline | 619.5 µs | 0.72× | 447.9 µs | 0.90× |
| + workspace CuTe caching | 520.7 µs | 0.85× | 369.7 µs | 1.08× |
| **+ no double conversion** | **408.1 µs** | **1.08×** | **278.9 µs** | **1.43×** |
| FlashKDA | 441.9 µs | — | 399.5 µs | — |

Correctness is untouched: 5.0e-03 … 1.0e-02 against an fp64 oracle, identical to
before the change, zero tolerance failures across all ten shapes.

### What the overhead actually was

cProfile, 200 iterations. The op was making 179,401 Python calls, dominated by
DLPack conversion:

| call | per launch | |
|---|---|---|
| `cute.runtime.from_dlpack` | **20** | for an op with **9** operands |
| `torch.utils.dlpack.from_dlpack` | 9 | the engine's own conversion |
| `torch._tensor.__dlpack__` | 20 | |
| `cutlass base_dsl.dsl.__call__` | 20 | one per CuTe tensor |

The kernel's conversion helper was **64% of its Python time**. Two distinct
pieces of waste:

**1. The kernel re-converted its own workspace on every launch.** Eleven of
those twenty conversions were scratch buffers that `_ws` already caches by
shape — their pointers never change. `_ws` now returns already-converted CuTe
tensors. (−99 µs)

**2. Every operand was converted twice.** The engine built torch tensors from
the `OperandBuffer` views purely so the kernel could convert them again to CuTe.
`OperandBuffer` implements DLPack, so the engine now converts straight to CuTe
and calls a new `run_cute()` entry point. The stream is passed explicitly as
well, rather than pushing a `torch.cuda.stream` context for the kernel to read
back out of thread-local state. (−112 µs)

`run()` is unchanged as the torch-tensor entry point, so standalone and test
callers are unaffected; `run_cute()` is purely an added fast path.

### Where the remaining time goes

| T=2048, H=12, N=1 | wall | GPU | host | engine cost over raw kernel |
|---|---|---|---|---|
| raw kernel | 306.7 µs | 97.4 µs | 209.3 µs | — |
| via cuDNN (before) | 551.0 µs | 96.7 µs | 454.3 µs | **+251 µs** |
| via cuDNN (after) | **343.5 µs** | 97.3 µs | 246.2 µs | **+37 µs** |

At T=4096 the engine costs **+6 µs** over calling the kernel directly (385.7 vs
379.2). cuDNN's dispatch is effectively free now.

The residual ~209 µs is the **CuTeDSL runtime's own launch path**, paid
identically by a standalone caller: nine genuinely unavoidable per-call operand
conversions plus the DSL's JIT argument marshalling. Reducing it further means
changing `nvidia_cutlass_dsl`, not this repo.

That is also what bounds the per-shape results. Where the GPU phase is long the
op wins outright — 8192/16/1 at **1.86×**, 8192/24/1 at 1.58×, 4096/12/1 at
1.40×. Where it is short, ~209 µs of DSL launch cost cannot be amortised:
1024/4/1 at 0.53×, 16384/16/8 at 0.79×, 2048/12/1 at 0.94×.

### Tests

`pytest -m L0 linear_attention/test_la.py` on H100 SXM, unchanged from before
the optimisation:

| selection | result |
|---|---|
| all backends | **420 passed, 3554 skipped, 1 xfailed, 0 failed** |
| `-k hopper` | **56 passed, 1020 skipped, 0 failed** |
| `-k "hopper and kda"` | **47 passed, 216 skipped, 0 failed** |

`pre-commit run` is green.

### Correcting myself

In the previous comment I suggested this overhead was largely a pre-existing
property of the linear-attention op layer, because `kda_cutile` showed a
comparable host cost. That was true of cuTile's numbers but wrong as an
explanation for this engine — the double conversion and the workspace
re-conversion were specific to this path, and removing them cut 211 µs without
touching any shared code.

Anerudhan:
@cudnn-ci-bot run frost

jhjpark (approved):

PR #1033, opened 2026-09-12
fix(kda): assign backward tail work by CTA ID

### LINKED ISSUE #1032: [Bug]: FROST KDA backward skips tail work under concurrent kernels on B300
### Description

FROST KDA backward can leave gradients unwritten when an independent kernel runs concurrently on another CUDA stream.

The failure reproduces 3/3 times on upstream commit `73758aa1b1352c44d6537dd487947c9ba11b0f86`, with finite random BF16 inputs, `batch_invariant=True`, `safe_gate=True`, and `gate_lower_bound=-5`.

### Suspected cause

In [kda_bprop_f16.py](https://github.com/NVIDIA/cudnn-frontend/blob/73758aa1b1352c44d6537dd487947c9ba11b0f86/python/cudnn/linear_attention/frost/kernel/kda_bprop_f16.py#L1427), the persistent scheduler assigns the final wave using:

```python
tail_row = tail_base + cute.arch.smid()
```

A grid with as many CTAs as SMs does not guarantee one CTA per SM. With other kernels occupying some SMs, multiple CTAs can execute on the same remaining SMs, so some reserved tail work items are never processed.

A local diagnostic change to use the unique CTA index resolved the standalone reproducer and produced bitwise-identical gradients to the no-overlap reference:

```diff
-    tail_row = tail_base + cute.arch.smid()
+    tail_row = tail_base + bidx
```

### Reproduction

Save the script below as `repro.py`, then run on B300:

```bash
CUDA_VISIBLE_DEVICES=0 python3 repro.py
```

Requires an existing cuDNN Frontend/PyTorch installation and CUDA 13 `nvcc` with a C++ compiler for the small occupancy kernel. Do not set `CUDA_LAUNCH_BLOCKING=1`, since this would serialize the streams.

The script uses only random inputs and a small CUDA kernel that temporarily occupies 32 SMs. It fills newly allocated floating-point buffers with NaNs to expose missing writes; all operator inputs and the upstream gradient are finite. The no-overlap reference is finite.

Expected: the six computed gradients remain finite and bitwise equal to the no-overlap reference.

Observed: repeated runs leave NaNs in `dq/dk/dv/dg/da_log/ddt_bias`. For example, `dq/dk/dv/dg` each contained 822,016 unwritten elements (247 tokens × 26 heads × 128 channels). The missing sequence/head regions match the SM-ID-based tail assignment.

<details>
<summary>Standalone reproducer</summary>

```python
"""Direct cuDNN KDA backward under concurrent SM occupancy (B300, no FLA)."""

import argparse
import os
import time
from unittest.mock import patch

os.environ.setdefault("MAX_JOBS", "2")
os.environ.setdefault("CUBLAS_WORKSPACE_CONFIG", ":4096:8")

import torch
from torch.utils.cpp_extension import load_inline

CUDA_SOURCE = r"""
#include <torch/extension.h>
#include <c10/cuda/CUDAGuard.h>
#include <c10/cuda/CUDAStream.h>
#include <c10/cuda/CUDAException.h>

__global__ void occupy_sms(int* smids, unsigned long long cycles) {
    extern __shared__ volatile unsigned char scratch[];
    scratch[threadIdx.x] = static_cast<unsigned char>(threadIdx.x);
    __syncthreads();
    unsigned int smid;
    asm volatile("mov.u32 %0, %%smid;" : "=r"(smid));
    if (threadIdx.x == 0) smids[blockIdx.x] = smid;
    unsigned long long start = clock64();
    while (clock64() - start < cycles) {
        asm volatile("nanosleep.u32 1000;");
    }
    scratch[threadIdx.x] = scratch[threadIdx.x];
}

void block_sms(torch::Tensor smids, int64_t cycles) {
    c10::cuda::CUDAGuard guard(smids.device());
    constexpr int shared_bytes = 160 * 1024;
    C10_CUDA_CHECK(cudaFuncSetAttribute(occupy_sms,
        cudaFuncAttributeMaxDynamicSharedMemorySize, shared_bytes));
    occupy_sms<<<smids.numel(), 32, shared_bytes,
        c10::cuda::getCurrentCUDAStream()>>>(smids.data_ptr<int>(), cycles);
    C10_CUDA_KERNEL_LAUNCH_CHECK();
}
"""


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--expect-fixed", action="store_true")
    parser.add_argument("--block-sm", type=int, default=32)
    parser.add_argument("--repeats", type=int, default=3)
    args = parser.parse_args()
    torch.set_num_threads(4)
    torch.cuda.set_device(0)
    from cudnn.linear_attention import kimi_delta_attention

    extension = load_inline(
        name="kda_diag_sm_blocker",
        cpp_sources="void block_sms(torch::Tensor smids, int64_t cycles);",
        cuda_sources=CUDA_SOURCE,
        functions=["block_sms"],
        extra_cuda_cflags=["-O2", "-gencode=arch=compute_103,code=sm_103"],
        verbose=True,
    )
    torch.manual_seed(42)
    shape = (8192, 64, 128)
    xs = {name: (torch.randn(shape, device="cuda", dtype=torch.bfloat16) * 0.1).requires_grad_()
          for name in ("q", "k", "v", "g")}
    xs.update(
        beta=torch.ones(shape[:2], device="cuda", dtype=torch.bfloat16),
        a_log=torch.zeros(64, device="cuda", requires_grad=True),
        dt_bias=torch.full((64, 128), -4.0, device="cuda", requires_grad=True),
        cu_seqlens=torch.tensor([i * 248 for i in range(32)] + [7935, 8182, 8192],
                                device="cuda", dtype=torch.int32),
    )
    options = dict(scale=128 ** -0.5, batch_invariant=True, safe_gate=True,
                   gate_lower_bound=-5.0, plan_name="kda_frost")
    dy = torch.randn(shape, device="cuda", dtype=torch.bfloat16) * 0.01
    targets = [value for value in xs.values() if value.requires_grad]
    names = [name for name, value in xs.items() if value.requires_grad]
    out = kimi_delta_attention(**options, **xs)[0]
    reference = tuple(g.detach().clone() for g in torch.autograd.grad(out, targets, dy, retain_graph=True))
    torch.cuda.synchronize()
    assert all(bool(g.isfinite().all()) for g in reference)

    stream = torch.cuda.Stream()
    smids = torch.empty(args.block_sm, device="cuda", dtype=torch.int32)
    # Compile/load the occupancy kernel before the measured overlap experiment.
    with torch.cuda.stream(stream):
        extension.block_sms(smids, 1000)
    torch.cuda.synchronize()
    original_empty, original_empty_like = torch.empty, torch.empty_like

    def poison_empty(*args, **kwargs):
        result = original_empty(*args, **kwargs)
        if result.is_cuda and result.is_floating_point():
            result.fill_(float("nan"))
        return result

    def poison_empty_like(*args, **kwargs):
        result = original_empty_like(*args, **kwargs)
        if result.is_cuda and result.is_floating_point():
            result.fill_(float("nan"))
        return result

    failures = 0
    for iteration in range(args.repeats):
        with torch.cuda.stream(stream):
            extension.block_sms(smids, 500_000_000)
        time.sleep(0.02)
        with patch("torch.empty", poison_empty), patch("torch.empty_like", poison_empty_like):
            gradients = torch.autograd.grad(out, targets, dy, retain_graph=True)
        torch.cuda.synchronize()
        print("OCCUPIED_SM", smids.cpu().tolist(), flush=True)
        for name, got, expected in zip(names, gradients, reference, strict=True):
            equal = torch.equal(got.contiguous().view(torch.uint8), expected.contiguous().view(torch.uint8))
            failures += int(not equal)
            nan = ~got.isfinite()
            print("RESULT", args.expect_fixed, iteration, name, "bit_equal", equal,
                  "nonfinite", int(nan.sum()), flush=True)
            if got.ndim == 3 and bool(nan.any()):
                indices = nan.any(dim=2).nonzero()
                print("MISSING", name, "tokens", int(indices[:, 0].min()), int(indices[:, 0].max()),
                      "heads", indices[:, 1].unique().tolist(), flush=True)
    if args.expect_fixed:
        assert failures == 0, failures
    else:
        assert failures > 0, "This run did not reproduce the overlap failure."
    print("EXPECTED_RESULT_CONFIRMED", "fixed" if args.expect_fixed else "original failure", flush=True)


if __name__ == "__main__":
    main()
```

</details>

The default script asserts that the failure was reproduced. After applying the one-line diagnostic change, `python3 repro.py --expect-fixed` instead asserts that all three repeats match the no-overlap reference; that flag does not modify cuDNN.

### Environment

- NVIDIA B300 (SM103, 148 SMs), driver 580.105.08
- Python 3.12.3, PyTorch 2.9.0, CUDA toolkit 13.0.88
- cuDNN Frontend 1.29.0; cuDNN backend 9.24.0
- CuTe DSL runtime 4.7.0
- KDA inputs: BF16, T=8192, H=64, D=128; int32 sequence offsets

Proposed fix: https://github.com/NVIDIA/cudnn-frontend/pull/1033

mdy666:
## Before submitting

- [x] I ran the complete pinned pre-commit hooks successfully (see the local limitation below).
- [x] I reviewed the applicable Hard Rules in `AGENTS.md`.
- [x] Required labels: `cat-bugfix`, `area:linear_attention`, `orig-external`.

Maintainer help is needed to apply the listed labels; the labels API returned HTTP 403 for this account.

## Affected area

FE OSS kernels / FROST KDA backward.

## Summary

Assign reserved backward tail work using the unique CTA index (`bidx`) instead of `cute.arch.smid()`. Add a standalone concurrency regression for default and batch-invariant KDA backward.

## Why

Concurrent kernels can occupy some SMs while multiple CTAs execute on the remaining SMs. SM IDs therefore cannot identify unique owners of reserved tail tasks. Some tasks are skipped, leaving gradient outputs unwritten. CTA IDs provide unique ownership without changing the launch grid or numerical computation.

## Related issues

Fixes #1032.

No public API, runtime dependency, dtype, or arithmetic changes. Only the tail scheduler in `kda_bprop_f16.py` changes. The regression uses finite random BF16 inputs with safe gate enabled and lower bound -5.

## Testing

B300 (148 SMs), PyTorch 2.9.0, CUDA 13.0, CuTe DSL 4.7.0; based on `ef47eabd`.

From `test/python`:

```bash
MAX_JOBS=2 CUBLAS_WORKSPACE_CONFIG=:4096:8 python3 -m pytest \
  linear_attention/test_kda_tail_scheduler.py -m L1 -s -q
```

- Before the fix: **2 failed** (default and BI), both with unwritten gradients.
- After the fix: **2 passed**; each case repeats backward three times under concurrent SM occupancy, requiring finite and byte-identical gradients relative to its no-overlap reference.

```bash
CUBLAS_WORKSPACE_CONFIG=:4096:8 python3 -m pytest \
  linear_attention/test_la.py -m L0 -s -q \
  -k 'frost and kda and not fp16 and (test_bwd_parity or test_bwd_varlen or test_bwd_partial_final_chunk or test_bwd_initial_state or test_determinism_bwd or test_determinism_multi_tile_bwd or test_batch_invariance_bwd)'
```

- **15 passed, 2 skipped** (existing parameter skips).
- Separate no-overlap before/after runs at T=8192, H=64, D=128, BF16/BI: output and all six tested gradients had identical byte hashes, for both a single sequence and 34 packed sequences.
- Backward timing smoke check: CUDA events around `torch.autograd.grad`, 5 warmups + median of 20 repeats, same GPU.

| Sequence layout | Before | After |
| --- | ---: | ---: |
| Single sequence | 2.765 ms | 2.763 ms |
| 34 packed sequences | 1.639 ms | 1.652 ms |

Formatting: Black 24.8 (line length 160), the SPDX pre-commit hook, and `git diff --check` passed. The complete pinned hook run was attempted, but the local Python 3.9 interpreter cannot install Black 26.3.1; the repository's hook configuration is unchanged.

### DIFF
--- a/python/cudnn/linear_attention/frost/kernel/kda_bprop_f16.py
+++ b/python/cudnn/linear_attention/frost/kernel/kda_bprop_f16.py
@@ -1424,7 +1424,8 @@ def tmaldg_warp(
     scheduler_state = PipelineState.start(phase=1)
     tail_count = ((total_tiles - cutlass.Int32(1)) % num_ctas) + cutlass.Int32(1)
     tail_base = (total_tiles - tail_count) if tail_count * 2 >= num_ctas else total_tiles
-    tail_row = tail_base + cute.arch.smid()
+    # CTA IDs remain unique when concurrent kernels prevent one CTA per SM.
+    tail_row = tail_base + bidx
     tail_row = tail_row if tail_row < total_tiles else cutlass.Int32(1 << 28)
 
     elect_one = nvvm.elect_sync()
--- a/test/python/linear_attention/test_kda_tail_scheduler.py
+++ b/test/python/linear_attention/test_kda_tail_scheduler.py
@@ -1,0 +1,124 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""KDA tail work must not depend on which SM executes a persistent CTA (#1032)."""
+
+import os
+import time
+from unittest.mock import patch
+
+import pytest
+import torch
+
+pytestmark = [pytest.mark.L1, pytest.mark.gpu_exclusive, pytest.mark.xdist_group(name="gpu_exclusive")]
+
+_CUDA_SOURCE = r"""
+#include <torch/extension.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <c10/cuda/CUDAStream.h>
+#include <c10/cuda/CUDAException.h>
+
+__global__ void occupy_sms(int* smids, unsigned long long cycles) {
+    extern __shared__ volatile unsigned char scratch[];
+    scratch[threadIdx.x] = static_cast<unsigned char>(threadIdx.x);
+    __syncthreads();
+    unsigned int smid;
+    asm volatile("mov.u32 %0, %%smid;" : "=r"(smid));
+    if (threadIdx.x == 0) smids[blockIdx.x] = smid;
+    unsigned long long start = clock64();
+    while (clock64() - start < cycles) {
+        asm volatile("nanosleep.u32 1000;");
+    }
+    scratch[threadIdx.x] = scratch[threadIdx.x];
+}
+
+void block_sms(torch::Tensor smids, int64_t cycles) {
+    c10::cuda::CUDAGuard guard(smids.device());
+    constexpr int shared_bytes = 160 * 1024;
+    C10_CUDA_CHECK(cudaFuncSetAttribute(occupy_sms,
+        cudaFuncAttributeMaxDynamicSharedMemorySize, shared_bytes));
+    occupy_sms<<<smids.numel(), 32, shared_bytes,
+        c10::cuda::getCurrentCUDAStream()>>>(smids.data_ptr<int>(), cycles);
+    C10_CUDA_KERNEL_LAUNCH_CHECK();
+}
+"""
+
+
+@pytest.fixture(scope="module")
+def sm_blocker():
+    if not torch.cuda.is_available() or torch.cuda.get_device_capability()[0] != 10:
+        pytest.skip("requires Blackwell FROST KDA")
+    if os.environ.get("CUDA_LAUNCH_BLOCKING") == "1":
+        pytest.skip("requires concurrent CUDA streams")
+    from cudnn.frost.buffers import cutedsl_state, cutedsl_too_old
+
+    available, version = cutedsl_state()
+    if not available or cutedsl_too_old(version):
+        pytest.skip("requires a FROST-compatible CuTe DSL")
+    from torch.utils.cpp_extension import CUDA_HOME, load_inline
+
+    if CUDA_HOME is None:
+        pytest.skip("requires nvcc for the occupancy kernel")
+    major, minor = torch.cuda.get_device_capability()
+    return load_inline(
+        name="kda_tail_scheduler_test",
+        cpp_sources="void block_sms(torch::Tensor smids, int64_t cycles);",
+        cuda_sources=_CUDA_SOURCE,
+        functions=["block_sms"],
+        extra_cuda_cflags=["-O2", f"-gencode=arch=compute_{major}{minor},code=sm_{major}{minor}"],
+    )
+
+
+@pytest.mark.parametrize("batch_invariant", [False, True], ids=["default", "bi"])
+def test_kda_backward_concurrent_tail(sm_blocker, batch_invariant):
+    from cudnn.linear_attention import kimi_delta_attention
+
+    # This work count reserves a partial tail wave on the 148-SM B300.
+    num_sms = torch.cuda.get_device_properties(torch.cuda.current_device()).multi_processor_count
+    tail_count = (34 * 64 - 1) % num_sms + 1
+    if tail_count * 2 < num_sms:
+        pytest.skip("this SM count does not exercise the reserved tail wave")
+    torch.manual_seed(42)
+    shape = (8192, 64, 128)
+    xs = {name: (torch.randn(shape, device="cuda", dtype=torch.bfloat16) * 0.1).requires_grad_() for name in ("q", "k", "v", "g")}
+    xs.update(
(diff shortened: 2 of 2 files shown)


jhjpark (approved):

jhjpark:
@cudnn-ci-bot run frost

jhjpark:
Thanks for helping to fix this!

PR #1011, opened 2026-09-11
perf(dsa): specialize SM100 H96 backward as H64 + H32

icavan:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and the changes comply.
- [ ] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`. External contributors cannot add labels to this repository; suggested labels are `cat-enhancements`, `area:sparse_attention`, `op: DSA`, and `orig-external`.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

This change adds an SM100 DSA backward path for BF16/FP16 H96, Dqk=576, Dv=512. It composes the existing generic H64 M64 body with the existing single-Q H32 M64 specialization, while preserving one shared FP32 dKV accumulation workspace and one full-H96 auxiliary workspace.

The optimized sequence uses six CuTe kernel launches:

1. Clear the shared FP32 dKV accumulator.
2. Compute `sum(O * dO)` and scaled LSE for all 96 heads.
3. Run the generic H64 backward body for heads 0:64.
4. Compute dSink for all 96 heads.
5. Run the specialized H32 backward body for heads 64:96.
6. Convert the shared FP32 dKV accumulator to the public BF16/FP16 dKV output.

This removes the duplicated clear, auxiliary, dSink, and conversion work of a naive two-sequence composition. The upstream H96 path uses five launches; this specialization pays one additional main-kernel launch in exchange for avoiding the half-populated second H64 head block.

## Why

The generic H96 kernel decomposes the head dimension into two H64 blocks. Its second CTA has only 32 live heads but retains the H64 kernel's scheduling, register allocation, and tensor-memory geometry. The existing H32 specialization is a better match for this tail: it uses an H32 MMA layout and H32-specific TMEM offsets and register budgets.

The H96 path therefore keeps the well-tuned generic H64 body for the first 64 heads and reuses the established H32 source structure for the remaining 32 heads. No new attention formula or reduced-precision accumulation mode is introduced.

## Related issues

None.

## API and compatibility impact

[...]

PR #926, opened 2026-09-07
Fix DSA backward boundaries and relax SM100 BSA FP8 shape limits

jiayus-nvidia:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes. (`pre-commit` is unavailable in the test environment; Black, `py_compile`, and `git diff --check` results are listed below.)
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and the changes comply.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`.

## Affected area

FE OSS kernels or CuTeDSL

## Summary

- Extend robust sparse-row boundary handling to the dedicated SM100 H16 and H32 DSA backward kernels.
- Extend the SM90 kernel to reject negative and positive out-of-range indices in both compact and non-compact layouts, and clamp `topk_length` to the physical top-k width.
- Carry packed per-row validity masks from the KV loader into probability recomputation so invalid rows contribute zero probability and zero gradient.
- Use predicated `cp.async` in the SM90 loader to suppress invalid global reads while zero-filling their shared-memory rows.
- Recheck sparse indices before every dKV scatter and define empty-row / saturating-sink limits without NaNs.
- Add regression coverage for active invalid indices, mixed invalid holes across tiles, oversized lengths, empty rows, and saturating sinks.
- Expand SM100/SM103 Sage FP8 BSA from the conservative `B=1`, `H in {4, 8}` support boundary to any positive batch and head counts; retain D=128, MHA, and 64-token sequence alignment requirements.
- Update the package-wide CuTe DSL floor documentation and the DSA/CSA acknowledgements.

## Why

#877 fixed invalid sparse-row handling in the generic SM100 kernel, but the dedicated H16 and H32 implementations still trusted active indices and retained the unsafe sink-limit behavior. #785 fixed the original SM90 empty-row, padded-tail, and sink failures, but active negative indices in compact mode, positive out-of-range indices in either mode, and `topk_length` values wider than the physical index row could still reach KV gathers, probability recomputation, or dKV scatter.

[...]

PR #817, opened 2026-08-31
Add SM100 two-CTA DSA backward specialization for BF16 H128/D512 — 1.10-1.16x faster

thynics:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I added GitHub labels: `cat-enhancements`, `mod-cutedsl`, `orig-nv-eng`.
- [ ] Milestone and Projects are left to maintainers, per the template's note on external contributions.

## Affected area

FE OSS kernels / CuTe DSL — DeepSeek Sparse Attention backward on SM100 and SM103.

## Summary

Add a shape-gated two-CTA specialization for DSA backward, selected only when all of the following hold:

- device capability is SM100 `(10, 0)` or SM103 `(10, 3)`;
- input dtype is BF16;
- `H = 128`;
- `D_qk = D_v = 512`;
- `topk_max` is one of `128`, `512`, `1024`, `1152`, or `2048`.

The public API, argument order, output tensors, stream semantics, and the existing generic/H16 implementations are unchanged. Every configuration outside this envelope retains the existing backend.

Main implementation changes:

- Map one query token to a `(2, 1, 1)` CTA cluster and use `tcgen05.CtaGroup.TWO` for score, dP, dQ, dV, and dK.
- Partition the 128 query heads across the two CTA ranks and eliminate the duplicate score-K gather performed by the two independent H64 CTAs.
- For top-k 512 and larger, reuse each rank's N32 score-K image when constructing the K-to-dQ operand and gather only the peer-owned half. Including this supplemental fetch, total sparse-K traffic decreases from 128 KiB to 96 KiB per token-tile. For top-k 128, it remains 128 KiB per token-tile.
- Stream transposed Q/dO operands through a fixed 32 KiB TMA ring: short rows use K64 with two stages; long rows use K32 with four stages. Unlike the baseline's SMEM recast reuse, the new path refetches 128 KiB per CTA per tile from L2, or 256 KiB per token-tile.
- Compute folded LSE and O·dO in the reducer-warp prologue, overlapping them with the prologue TMA/gather work and score/dP/softmax preparation. The first dS and dKV computation remains dependent on O·dO.
- Rebalance the fixed CTA register budget by warp role and hand retired reducer registers to the dQ epilogue.
[...]

PR #1048, opened 2026-09-14
Add SM100 2-CTA DSA backward for BF16 H128/D576

thynics:
## Before submitting

- [ ] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL (plus the Python API surface of `cudnn.DSA.SparseAttentionBackward`, its docs, and the DSA benchmark README).

## Summary

Adds a `(2,1,1)`-cluster CuTe DSL backward kernel for BF16 DeepSeek Sparse Attention with `H=128`, `D=576`, `D_v=512` on SM100-class devices (compute capability 10.0 and 10.3), and routes the public backward entry point to it from tensor metadata.

- New backend `dsa_bwd_sm100_h128_d576_2cta.py` with a thin compile/execute layer in `_interface_sm100_d576.py`, plugged into the dispatch and workspace structure introduced by #817: `_select_sm100_backend` returns `h128_d576_2cta_m64` after the `deterministic=True` precedence check, `_get_sm100_kernel_class` and `flash_attn_bwd_sm100` know the backend (direct callers are delegated to the same plan-time-compiled kernel), and the scratch size is the shared SM100 formula (`flash_attn_bwd_sm100_workspace_size`). Score, dP, dQ, dV and dK run on `cta_group::2` tensor cores with the two CTAs of a cluster splitting the 128 heads; the D64 QK/dQ tails are separate CG2 rounds and the dK tail is a per-CTA warp MMA with FP32 atomics. P and dS are exchanged between the CTAs with `shared::cluster` bulk copies. Slot validity is applied before every score and gradient operation, including the K rows retained for dQ, so ignored top-k slots (beyond `topk_length`, negative or out-of-range indices) cannot contaminate gradients. The sink-folded LSE is guarded on the log2(e)-rescaled sink and LSE values (as #926 did for H16/H32), so a large finite sink whose rescale overflows to +inf (for example 2.4e38) zeroes the row like an infinite sink instead of producing `inf - inf`.
- Short query batches spread each query over several clusters (one dQ owner plus dV/dK workers); long sequences use one cluster per query.
- Dispatch (`_select_sm100_backend`) requires compute capability 10.0 or 10.3 (`_BLACKWELL_CAPABILITIES`, validated on B200 and B300), `deterministic=False`, contiguous BF16 inputs, `H=128`, `D=576`, `D_v=512` and `topk_max ∈ {128, 512, 1024, 1152, 2048}`. FP16, other head counts and top-k sizes, noncontiguous inputs, `deterministic=True`, SM107 and SM90 keep their existing paths.
- The plan compiles from tensor metadata (no priming launch) and executes without allocation. The wrapper allocates `dq`/`dkv`/`d_sink` and a reusable `uint8` workspace sized by the existing `SparseAttentionBackward.scratch_workspace_bytes()` (the same bytes the generic SM100 pipeline needs for this shape); zero-initialization of the FP32 dKV/dSink accumulators happens in-kernel on the launch stream. The wrapper cache key now includes device and strides.
- Separate commit `fix(dsa): guard the H128/D512 two-CTA sink fold and dSink reduction`: the D512 two-CTA kernel from #817 had the same gap plus two more (`lse == +inf` was its only guard). On B200 it returned NaN `dq`/`dkv`/`d_sink` for `attn_sink = +inf` and for the 2.4e38 overflow, and NaN `d_sink` for an empty row with the sink disabled. It now tests the rescaled values in the fold and gives its dSink reduction the generic kernel's +inf / -inf sink handling. Top-k index validity (`-1` padding, active negative, active out-of-range, entries past `topk_length`) was already aligned with the generic kernel on all three routes (checked with a boundary matrix on B200).
- Docs: `docs/fe-oss-apis/dsa.md` (dispatch rule, `topk_length` semantics, workspace contract); `benchmark/dsa/README.md` (H128/D576 invocation).
- Tests: dispatch/fallback, numerics against the PyTorch reference for `topk 128..2048` in both cluster regimes, NaN-poisoned ignored slots with workspace reuse, sink normalization, empty and saturating sink limits (2e38, 2.4e38, +inf, -inf with empty rows) for both H128 two-CTA routes at supported top-k widths across the single-query, split-cluster and single-cluster schedules, active out-of-range indices on both H128 two-CTA routes, zero/out-of-range `topk_length`, non-default-stream ordering.

## Why

The production MLA shape of DeepSeek sparse attention (128 heads, 576 QK dims, 512 value dims) was served by the generic one-CTA M64 backward pipeline, which gathers every K row once per CTA and drives the tensor cores with a 64-head tile. Pairing two SMs per query lets the cluster issue 128-head `cta_group::2` MMAs, gather each sparse K row once per cluster half and reuse it as the dQ operand, and halve the per-tile dKV atomic traffic. The change is confined to a new backend selected purely from metadata, so every other shape, dtype and architecture is untouched, and the 2-CTA plan's allocation-free `execute()` fits stream-ordered and graph-captured training loops.

## Related issues

Builds on #817 (two-CTA DSA backward for the D512 shape, merged): this kernel derives from the same design, adds the D64 QK/dQ/dK tails, and reuses #817's dispatch, deterministic-precedence and caller-workspace contracts. Related to #676 (items 1–3: invalid top-k slots, empty rows, saturating sinks) and #926: both H128 two-CTA routes now match the generic kernel on those boundaries, with the H16/H32 rescaled-value sink guard carried over. Item 4 of #676 (recomputing the delta from Q·K·dO) is intentionally not addressed here, matching the generic kernel.

- **Public API additions** (backward compatible, appended parameters): `SparseAttentionBackward.execute(..., d_sink=None)` and `_select_sm100_backend(..., is_contiguous=True)`. `workspace=` and `scratch_workspace_bytes()` exist since #817 and are reused unchanged. The two-CTA D576 plan compiles in `compile()` and its `execute()` requires caller-provided `dq`, `dkv`, `d_sink` and workspace (it never allocates or compiles); the other backends reject a caller `d_sink`.
- **Behavior change**: on SM100-class devices (CC 10.0 and 10.3), contiguous BF16 `H=128/D=576/D_v=512` inputs with a supported `topk_max` now dispatch to the two-CTA kernel. Outputs agree with the generic path within BF16 rounding (relative L2 on the benchmark shapes: `dq ≤ 1.4e-3`, `dkv ≤ 1.6e-4`, `d_sink ≤ 2.4e-3`). Direct callers of `_interface_sm100.flash_attn_bwd_sm100` with this shape are delegated to the same kernel. The D512 two-CTA route (unchanged shapes) now returns zero gradients instead of NaN for infinite or overflowing sinks and for empty rows with the sink disabled. All other inputs are unaffected; `topk_length` clamping to `[0, topk_max]` is a two-CTA-route guarantee, the pre-existing backends still expect `topk_length <= topk_max` (unchanged). Tensor validation moved earlier: `check_support()` now rejects non-CUDA inputs before querying the device, and `execute()` resolves the architecture from `q.device` and launches under that device context instead of the ambient current device (review comment on this PR).
- **Supported versions**: SM100 and SM103 (CC 10.0 and 10.3, both validated on hardware); SM107 keeps the generic pipeline until it has been validated. Works with the declared `nvidia-cutlass-dsl >= 4.6.2` floor (validated on 4.6.2 on hardware) and also traces and compiles on 4.5.2 and 4.7.1. No C++, build or packaging changes.
- **Performance**: with the repository benchmark at default clocks against upstream develop @ e0317d1f, the two-CTA route is faster than the one-CTA pipeline by a geometric mean of 1.122x on B200 (1.116x without `topk_length`; `topk = 128` 1.04–1.10x, `topk >= 512` 1.10–1.17x) and 1.145x on B300 (with and without `topk_length`; `topk = 128` 1.05–1.10x, `topk >= 512` 1.14–1.20x); see Testing.

### Correctness

Hardware: NVIDIA B300 SXM6 AC (CC 10.3) and NVIDIA B200 (CC 10.0, 148 SMs); `nvcr.io/nvidia/pytorch:26.08-py3` (Python 3.12.3, CUDA 13.4, PyTorch 2.14.0a0+4fdf77b940.nv26.08, nvidia-cutlass-dsl 4.6.2, cuda-python 13.3.1, cuDNN backend 9.25.0), driver 610.57.04; editable install of this branch.

```bash
cd test/python
pytest fe_api/dsa/test_DSA_sparse_attention_backward.py -m "L0 or L1"
# B200 (SM100), branch on develop @ 367855f2 incl. the sink fixes: 158 collected, 144 passed, 14 skipped (SM90-only cases), 0 failed  (254 s)
# B300 (SM103), branch on develop @ e0317d1f, before the sink fixes:  124 collected,  95 passed, 29 skipped, 0 failed  (162 s)
#   B300 skips: 15 cases that upstream gates to CC (10, 0) (D512 two-CTA), 14 SM90-only cases
```

Every two-CTA D576 case ran on both GPUs (dispatch and fallback, 14 numerics configurations, NaN-poisoned ignored slots with workspace reuse through the wrapper and the plan, sink normalization, sink limits, length boundaries, non-default-stream ordering). The suite also covers the existing SM100 backends (generic M64, deterministic M64, H16 M128, H32 M64) alongside the new route, so the dispatch change is regression-tested for the shapes that keep their old path.

Static checks: `pre-commit run` (black 26.3.1 at line length 160, SPDX headers) passes on every changed file; `pyflakes` clean; the kernel was traced and compiled through `cute.compile` for eight `(topk_max, topk_length, single_query, split_regime)` configurations on nvidia-cutlass-dsl 4.5.2 and 4.7.1 (`sm_100a`) and three on `sm_103a`.

Notes for reviewers reproducing the run:

- One pre-existing defect of the generic M64 path was observed while it still served H128/D576 on SM103: with `topk_length > topk_max` that kernel does not clamp the length (`dsa_bwd_sm100.py` derives the tile count from the raw value) and corrupts 43 of 663552 `dq` elements of the affected query; develop @ 606e16f reproduces it on the same GPU. The H128 length-boundary case is therefore gated to devices that take the two-CTA route, the docs state the clamp guarantee for that route only, and the generic/H16/H32 clamp is left to a separate change.
- The sink-limit fixes were validated on B200 with a boundary matrix (generic 1-CTA, D512 two-CTA, D576 two-CTA; `-1` padding, active negative and out-of-range indices, entries past `topk_length`, empty rows with disabled and finite sinks, single-key rows with sinks 2e38 / 2.4e38 / +inf at `s_q` 1 and 4): after the fixes both two-CTA routes are finite and zero where the sink owns all mass; the generic kernel still returns NaN for the 2.4e38 overflow (pre-existing, #926 covered only H16/H32), which is left to a separate change.
- `test_DSA_sparse_attention_backward_sm100_deterministic_boundaries[H64-q1-kv1]` (upstream, unrelated to this route) needs a writable `TRITON_CACHE_DIR`: its PyTorch reference dispatches a `bmm` through Triton in this container, and it fails with `ENOSPC` when the home directory is full, on develop and on this branch alike.

### Performance

Environment: the container above, default clocks (no clock locking), SM clock and power sampled every 500 ms across the whole measurement window. Baseline: upstream develop (the generic one-CTA M64 pipeline that served this shape), measured from its own checkout with the same benchmark script. Candidate: this branch (two-CTA route through the public wrapper).

Matrix: `seqlen_q = seqlen_kv ∈ {4096, 8192}`, BF16, `H = 128`, `D = 576`, `d_v = 512`, `topk ∈ {128, 512, 1024, 1152, 2048}`. Attention sink enabled; `topk_length` present with every row set to the configured full top-k (benchmark default).

Protocol: `benchmark/dsa/benchmark_dsa_sparse_attention_backward.py` as shipped (per configuration 10 warmup calls, then CUDA-event timing of 50 calls of `sparse_attention_backward_wrapper`; JIT compilation and the reference forward excluded). The two checkouts were run alternately, three full sweeps each; every cell is the mean of its three sweeps. Outputs of the two routes agree to relative L2 `dq ≤ 1.4e-3`, `dkv ≤ 1.6e-4`, `d_sink ≤ 2.4e-3`.

```bash
python benchmark/dsa/benchmark_dsa_sparse_attention_backward.py --nheads 128 --head-dim 576 --topks 128,512,1024,1152,2048
```

The tables were measured with develop @ e0317d1f as the baseline; the branch was since rebased onto 367855f2 (#1011, H96 composite), whose changes to the generic kernel are `const_expr` flags at their default values, so the H128/D576 baseline path is unchanged and the matrix was not re-run. The two-CTA timings are unchanged by the rebase: the B300 below measured the pre-rebase branch at a geometric mean of 5.008 ms with `topk_length` and 5.022 ms without (5.009 / 5.022 ms now). The generic baseline moved with the upstream changes to `dsa_bwd_sm100.py` between 606e16f and e0317d1f (with `topk_length` 5.528 → 5.734 ms, without 6.216 → 5.748 ms), which is why the pre-rebase ratios (B300 1.104x / 1.238x, B200 1.080x / 1.193x) differ from the tables below.

#### SM100 / B200, baseline upstream develop @ e0317d1f (after the rebase)

NVIDIA B200 (CC 10.0, 148 SMs; umbriel 8-GPU node, one GPU allocated), boost up to 1965 MHz, 1000 W power limit; per-cell sweep-to-sweep spread ≤ 0.033 ms.

| Sq = Skv | K | Native one-CTA (ms) | Integrated two-CTA (ms) | Speedup |
|---:|---:|---:|---:|---:|
| 4096 | 128 | 1.192 | 1.116 | 1.068x |
| 4096 | 512 | 3.078 | 2.655 | 1.160x |
| 4096 | 1024 | 5.628 | 4.792 | 1.174x |
| 4096 | 1152 | 6.263 | 5.414 | 1.157x |
| 4096 | 2048 | 10.727 | 9.256 | 1.159x |
| 8192 | 128 | 2.311 | 2.224 | 1.039x |
| 8192 | 512 | 6.081 | 5.438 | 1.118x |
| 8192 | 1024 | 11.134 | 9.962 | 1.118x |
| 8192 | 1152 | 12.399 | 11.182 | 1.109x |
| 8192 | 2048 | 21.278 | 18.905 | 1.126x |
| Geometric mean | | 5.967 | 5.319 | **1.122x** |

With `--no-topk-length`: 1.059–1.166x, geometric mean 1.116x (one-CTA 5.971 ms, two-CTA 5.352 ms). SM clock sampled every 500 ms over the whole window: the one-CTA pipeline averaged 1945 MHz with 1.9% of its busy samples below 1900 MHz, the two-CTA kernel 1940 MHz with 10.7% below 1900 MHz; both drew up to the 1000 W power limit (997 W / 1002 W). The two-CTA kernel spends more of its time at the limit, so the ratio widens on GPUs with more power headroom (B300 below). Before the rebase, another umbriel B200 (umb-b200-260) measured this branch's two-CTA route at a geometric mean of 5.370 ms with `topk_length`, matching this node's 5.319 ms.

#### SM103 / B300, baseline upstream develop @ e0317d1f (after the rebase)

NVIDIA B300 SXM6 AC (CC 10.3), boost up to 2032 MHz, 1100 W power limit; per-cell sweep-to-sweep spread ≤ 0.023 ms.

| Sq = Skv | K | Native one-CTA (ms) | Integrated two-CTA (ms) | Speedup |
|---:|---:|---:|---:|---:|
| 4096 | 128 | 1.135 | 1.072 | 1.059x |
| 4096 | 512 | 2.964 | 2.537 | 1.168x |
| 4096 | 1024 | 5.412 | 4.516 | 1.198x |
| 4096 | 1152 | 6.023 | 5.108 | 1.179x |
| 4096 | 2048 | 10.311 | 8.645 | 1.193x |
| 8192 | 128 | 2.232 | 2.130 | 1.048x |
| 8192 | 512 | 5.846 | 5.121 | 1.142x |
| 8192 | 1024 | 10.705 | 9.243 | 1.158x |
| 8192 | 1152 | 11.918 | 10.345 | 1.152x |
| 8192 | 2048 | 20.422 | 17.561 | 1.163x |
| Geometric mean | | 5.734 | 5.009 | **1.145x** |

With `--no-topk-length`: 1.078–1.197x, geometric mean 1.145x (one-CTA 5.748 ms, two-CTA 5.022 ms). SM clock over the window: mean 2005 MHz, maximum 2032 MHz (546 samples, including idle samples between sweeps); power up to 1091 W under the 1100 W limit.

Not tested: SM107 devices, FP16 (not dispatched to this kernel), and mixed-compute-capability processes.


jiayus-nvidia:
The new H128/D576 two-CTA kernel still misses the **finite-sink rescale-overflow guard** addressed by #926 for H16/H32 (related to #676).

Reproduced on this PR's current head, `70656b8c22379361cac0559618e50a6d5145cff5`, on B200 / SM100, driver 595.58.03, CUDA 13.2, PyTorch 2.12.0a0+5aff3928d8.nv26.05, cuDNN backend 9.22.0, and nvidia-cutlass-dsl 4.6.2.

**Minimal reproduction** — run with this PR's Python package selected; the import-path print and backend assertion prevent accidentally testing another checkout or the generic fallback. `topk=128` is intentional because `topk=64` does not select this backend.

```python
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0
import torch
import cudnn
from cudnn import DSA
from cudnn.deepseek_sparse_attention.sparse_attention_backward._interface_sm100 import _select_sm100_backend

assert torch.cuda.get_device_capability() in ((10, 0), (10, 3))
torch.manual_seed(786)
q = torch.randn(1, 128, 576, device="cuda", dtype=torch.bfloat16)
kv = torch.randn(1, 576, device="cuda", dtype=torch.bfloat16)
out = torch.zeros(1, 128, 512, device="cuda", dtype=torch.bfloat16)
dout = torch.randn_like(out)
topk_idxs = torch.full((1, 128), -1, device="cuda", dtype=torch.int32)
topk_idxs[:, 0] = 0
scale = 192**-0.5
lse = ((q.float() * kv[0].float()).sum(dim=-1) * scale).contiguous()

print("cudnn:", cudnn.__file__)
backend = _select_sm100_backend(
    128, 576, head_dim_v=512, dtype=q.dtype, max_topk=128,
    device_capability=torch.cuda.get_device_capability(),
)
print("backend:", backend)
assert backend[0] == "h128_d576_2cta_m64"

failed = []
for sink_value in (2e38, float("inf"), 2.4e38):
    attn_sink = torch.full((128,), sink_value, device="cuda", dtype=torch.float32)
    result = DSA.sparse_attention_backward_wrapper(
        q, kv, out, dout, lse, attn_sink, topk_idxs, softmax_scale=scale,
    )
    torch.cuda.synchronize()
    nonfinite = {name: int((~torch.isfinite(value)).sum()) for name, value in result.items()}
    print(f"sink={sink_value}: nonfinite={nonfinite}")
    # For these sinks all probability mass belongs to the sink: O and all
    # three gradients should be zero at the supported numerical precision.
    if not all(torch.equal(value, torch.zeros_like(value)) for value in result.values()):
        failed.append(sink_value)
assert not failed, f"Expected zero gradients; failing sink values: {failed}"
```

Observed output:

```text
backend: ('h128_d576_2cta_m64', 64)
sink=2e+38: nonfinite={'dq': 0, 'dkv': 0, 'd_sink': 0}
sink=inf: nonfinite={'dq': 0, 'dkv': 0, 'd_sink': 0}
sink=2.4e+38: nonfinite={'dq': 73728, 'dkv': 576, 'd_sink': 0}
AssertionError: Expected zero gradients; failing sink values: [2.4e+38]
```

Expected: all three gradients are zero for every case above, since the sink owns all probability mass at the supported numerical precision. Instead, the finite `2.4e38` case makes **every dq and dkv element NaN**.

The cause is in [`_compute_folded_lse`, lines 1194–1206](https://github.com/NVIDIA/cudnn-frontend/blob/70656b8c22379361cac0559618e50a6d5145cff5/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h128_d576_2cta.py#L1194-L1206): `sink_value` is finite, but `sink_value * log2(e)` overflows to `+inf`. The subsequent `sink_log2 - maximum` evaluates `inf - inf`. The guard checks the original `sink_value`/`lse_value`, so it misses this overflow. #926 checks the **rescaled** values instead; see [the H16 guard](https://github.com/NVIDIA/cudnn-frontend/blob/70656b8c22379361cac0559618e50a6d5145cff5/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h16.py#L746-L760).

As a diagnostic, changing only those infinity comparisons to `sink_log2`/`lse_log2` made all 32 sink-boundary checks pass, versus 24 passes / 8 failures before. That matrix covered `(Sq, topk) = (1,128), (9,512), (129,128), (129,512)`, with and without `topk_length`, and sink values `2e38`, `2.4e38`, `+inf`, plus empty rows with `-inf`. The temporary change was reverted after verification.

Please carry over the rescaled-value guard and add H128/D576 coverage with a supported top-k width. The existing specialized sink-limit test covers H16/H32/H96 and uses topk=64, so the current full backward suite's `122 passed, 14 skipped` does not exercise this boundary on the new route.

For scope: the generic H128/D576 baseline at `e0317d1f` also fails the finite-overflow case. This is a missing fix in the new specialization, rather than a newly introduced regression relative to that baseline.

jiayus-nvidia:
@cudnn-ci-bot run oss

thynics:
Thanks for the report — fixed on the current head (b074b911), plus a couple of neighbouring gaps found while aligning both two-CTA routes with the generic kernel on the #676 boundaries:

- **D576 two-CTA (this PR):** `_compute_folded_lse` now guards on the log2(e)-rescaled LSE/sink values, as H16/H32 do since #926, so a large finite sink (2.4e38) whose rescale overflows zeroes the row like `+inf` instead of producing `inf - inf`. Your script passes for 2e38 / +inf / 2.4e38.
- **D512 two-CTA (#817):** its fold only special-cased `lse == +inf`, so `attn_sink = +inf` and the 2.4e38 overflow gave NaN `dq`/`dkv`/`d_sink`, and an empty row with the sink disabled gave NaN `d_sink` through the dSink reduction. Same rescaled-value guard, plus the generic kernel's `+inf → p = 1` / `-inf → p = 0` handling in the reduction.
- **D512 two-CTA, padded slots under a very negative LSE:** zero-filled K/V for invalid slots is not a softmax mask, so with the sink disabled and every valid logit far below zero `exp2(0 - LSE)` overflowed and `inf * 0` turned the whole `dq` row into NaN. Fixed by clamping the exp2 argument (NaN-preserving), which leaves valid slots unchanged and keeps the zero rows contributing exact zeros. Performance-neutral on B200; a per-slot P mask was measured at 5–8% slower and rejected.
- Index validity (`-1` padding, active negative / out-of-range indices, entries past `topk_length`) and empty rows were already consistent across generic, D512 and D576 (checked with a boundary matrix on B200).

Tests: sink limits (2e38, 2.4e38, +inf, -inf with empty rows) and padded slots under a very negative LSE for both H128 two-CTA routes at supported top-k widths across the single-query, split-cluster and single-cluster schedules; the active-out-of-range test parametrized over D512/D576; wrapper/plan tests now assert the two-CTA launch actually ran. Full `-m "L0 or L1"` file at head: B200 146 passed / 14 skipped / 0 failed, B300 124 passed / 36 skipped / 0 failed.

Out of scope: the generic 1-CTA H128/D576 path still returns NaN for the 2.4e38 case (pre-existing; #926 covered H16/H32 only)

YangXu1990uiuc:

The recorded OSS run covered `fcc7d7f`, before the latest D512 probability-path and route-assertion changes. I have approved the pending fork Style/Docs checks and am requesting OSS CI for `b074b911cb67`. The following standalone `@cudnn-ci-bot run oss` command is also a Codex bot action. Use the same command after further updates; a maintainer may need to approve fork workflows.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc (approved):

Reviewed `b074b911cb67` against base `c3b5b143da79`.

The H128/D576 two-CTA design makes sense: the contiguous dispatch contract, plan-time compilation, caller-owned storage and split-query schedule are consistent, and both measured D576 workloads improve. The previously reported finite-sink rescale issue is addressed. One P2 remains in the neighboring D512 padded-slot fix; it is an incomplete repair of a pre-existing problem, not a new regression in the D576 kernel.

**[P2] Bound the downstream dS product, not only exp2** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/b074b911cb679948d472bac5b01b30f73c590a58/python/cudnn/deepseek_sparse_attention/sparse_attention_backward/dsa_bwd_sm100_h128_2cta.py#L2704))

The new upper bound of 64 keeps P finite but still allows an invalid slot to carry P = 2^64. Its subsequent `(dP - O.dot(dO)) * P` can overflow before multiplication by zero-filled K, producing NaN dQ. I reproduced this on the current head with BF16 Q=(4,128,512) filled with 16, KV=(1,512) and O=(4,128,512) filled with -1, dO filled with 2^60, FP32 LSE=-16*sqrt(512), disabled sink (-inf), and topk=128 with only index 0 valid. The correct dQ and dSink are exactly zero and dKV is finite (512*2^60), but all 262144 dQ elements are NaN. The same D512 case with dO=1 passes now; the D576 route passes both scales. Consider a tighter bound derived from the probability range, accounting for the downstream product/conversion, or explicitly zero invalid P/dS as D576 does. Please extend the padded-slot regression beyond unit-scale dO. This does not require accepting the previously measured cost of an unconditional per-slot mask if a sound cheap bound suffices.

Evidence: Independent SM100 public-wrapper probe at b074b911: dO=1 gives finite zero dQ; dO=2^60 gives 262144 NaNs in dQ, with finite dKV/dSink. Baseline 367855f2 fails both D512 scales; baseline generic D576 and head two-CTA D576 pass both. Analytic one-key reference gives dQ=0, dSink=0 and finite dKV. Attribution: Incomplete fix in this PR for a pre-existing D512 padded-slot overflow; ordinary-scale behavior is improved, but the claimed finite dS guarantee is still false.

Validation: Independent SM100 run of the full DSA backward file with explicit -m 'L0 or L1': 145 passed, 1 failed, 14 skipped, 7 deselected. Verified the reviewed Python import path and exercised both H128/D576 schedules, supported widths, lengths/sentinel padding, ignored NaN rows, workspace reuse, sink limits and nondefault streams. The one failure is the existing H16/D576 wrapper dSink case. The baseline control reproduces the identical mismatch (1/16 elements, max absolute error 0.0716968775), so it is not attributed to this PR. ABBA performance spot check against 367855f2 on the same SM100 device, BF16 H128, KV=4096, full-length tensors, scale=1/sqrt(D), steady-state SparseAttentionBackward.execute with reusable scratch/dQ/dKV (and the new explicit dSink output on D576). D576 Sq=256/topk=2048: GPU 2528 us -> 2065 us, about 18.3% less; CPU median per arm 100-122 us -> 48-50 us. D576 Sq=9/topk=512: GPU 102.2 us -> 57.5 us, about 43.7% less; CPU 82-84 us -> 38-39 us. Actual new backend and split counts 1/3 were recorded. D512 control at Sq=256/topk=2048: GPU 1671 us -> 1674 us (about 0.2%); CPU 78-82 us -> 83 us, no material regression established. Every timing arm passed reference-gradient checks, poisoned-output/workspace replay and a changed-dO negative control; sync-debug mode reported no execute-time synchronization. Reviewed the split ownership, staged-index bounds, shared-memory reuse and completion barriers, D64 QK/dQ/dK tails, empty-row paths, dispatch fallbacks and new tests. The compared SM100 backward kernels/interfaces and reference are unchanged between the comparison baseline and the reviewed develop base; develop additionally contains SM90 changes/tests, outside this independent GPU run. Started current-head OSS CI and verified its acknowledgment; it is still running. Approved the pending read-only fork Docs/Style workflows after checking their contents; both completed successfully on b074b911. See the adjacent CI assistance comment for the supported trigger procedure.

Limitations: Independent GPU evidence is SM100 with CuTe DSL 4.8; no independent SM103 run, DSL 4.6.2 run or sanitizer result is claimed. Current-head OSS CI is pending. The timings cover the listed prepared API calls, not allocation-inclusive wrapper or full-model performance. The D512 numerical finding uses deliberately extreme finite gradients with a disabled sink.

jiayus-nvidia (approved):

jiayus-nvidia:
@cudnn-ci-bot run oss

PR #1049, opened 2026-09-14
fix(dsa): synchronize SM90 indexer dK staging

Denny991:
## Before submitting

- [x] I ran the repository pre-commit hooks and committed formatting changes.
- [x] I reviewed the Hard Rules in the root and `python/cudnn/AGENTS.md`.
- [x] Required labels are present: `cat-bugfix`, `op: DSA`, `area:sparse_attention`, and `orig-external` (added by a maintainer).

## Affected area

FE OSS kernels / CuTeDSL: SM90 DSA Indexer backward.

## Summary

Add two synchronization points per top-k block to protect the dK shared-memory staging buffer. Each compute warpgroup uses its own 128-thread named barrier because the two warpgroups own disjoint halves of the buffer.

## Why

Multiple threads write each staging row, while an individual thread issues the bulk reduction for that row. The per-thread bulk wait does not prevent other threads from overwriting staging before every prior asynchronous reader finishes. The async-proxy fence alone also does not make each reduction launcher wait for all staging producers.

The patch adds a warpgroup barrier after the bulk read wait and another after the staging stores and proxy fence. Existing GEMM scheduling barriers and the final bulk wait remain intact.

This synchronization gap is consistent with intermittent nonfinite dK values observed in historical H200 training. It has not been confirmed with a standalone reproducer.

## Related issues

No duplicate SM90 Indexer staging issue identified.

No public API, tensor layout, or dependency changes. The shared SM90 sparse/dense Indexer compute path gains two warpgroup barriers per block. SM100 code is unchanged. Performance on this revision has not been measured.

## Testing

Candidate commit `ca64645` is based on upstream develop `367855f256136d8a0e6926924689f8ff1a3f9ff9`. Local validation of this revision covers static checks only:

- `pre-commit run --files python/cudnn/deepseek_sparse_attention/indexer_backward/indexer_backward_sm90.py`: passed (Black, Black Jupyter, SPDX; clang-format not applicable).
- `git diff 367855f..HEAD --check`: passed.
- `python3 -m py_compile python/cudnn/deepseek_sparse_attention/indexer_backward/indexer_backward_sm90.py`: passed.
- A maintainer has started OSS CI for `ca64645`; results are pending.

The corrected model-free script was run separately on an idle H200 worker (script SHA256 `8bf75e2dcde8ecfb89d67caed2f9ef27f740e64c0ffacb603139a9e55f697793`): stock completed 20/20 calls with one stream and 400/400 calls with four streams; the historical barrier overlay completed 400/400 calls with four streams. All returned finite, nonzero outputs. This is a schedule-stress smoke test, not a deterministic reproducer and not a complete training-equivalent workload.

Fresh integration validation on 2026-09-14 used Frontend 1.26.0 and an isolated overlay with the same barrier placement, rather than the develop-based candidate:

- 8×H200, PyTorch 2.9.1+cu128, EP4/DeepEP joint training, packed sequence length 16384, top-k 2048.
- Stock Frontend 1.26.0 produced nonfinite training signals: one run reached `reduced_indexer_loss: nan` at step 12, and an instrumented run reached `grad_norm: nan` at step 9.
- The same 35-step recipe with the barrier overlay completed all 280 rank-step records with no `nan` in indexer loss, local loss, or grad norm.
- The fresh training instrumentation did not emit a dK tensor dump, so these runs establish nonfinite training behavior rather than a new direct dK capture. The direct dK-first observation was recorded in the historical investigation.

Historical observations for the same barrier placement on cuDNN Frontend 1.26.0:

- 8 H200 GPUs; BF16; sequence length 16384; head dimension 128; top-k 2048; PyTorch 2.12.1+cu132. Unpatched training intermittently produced nonfinite dK while dQ/dWeights and forward losses remained finite.
- Patched training completed three 35-step runs and one 200-step run (305 steps total) without observed nonfinite gradients or optimizer skips.
- A separate PyTorch 2.9.1+cu128 comparison observed unpatched NaNs and a completed patched 35-step run without them.
- Fixed-input standalone replay did not reproduce the failure. A deterministic standalone reproducer is not yet available.


jiayus-nvidia:
@cudnn-ci-bot run oss

jiayus-nvidia (approved):

PR #1056, opened 2026-09-14
la(kda): swap in the campaign's final Hopper sm90 KDA kernel (1.09x -> 1.19x FlashKDA)

Anerudhan:
## What

Replaces the sm90 KDA prefill kernel landed in #1017 (`b134720b`,
`kda_sm90_segscan_leader_checkpoint`) with the one the same Kernel Factory
campaign finished on: `4e60c30a`, `kda_sm90_runtime_partial_tail_v64`.

The campaign kept running after the #1017 branch was cut. This is only the
better kernel from it — **one file changes**, and `run_cute` keeps the same
15-argument signature, so `kda_engine.py` is untouched and no dispatch,
engine-manifest or support-envelope behaviour changes.

## Why it is faster

It transposes the state so every `wgmma` is full-M (128 rows) instead of the
16-row shape the untransposed recurrence gives, and feeds both `St` and `Rt`
to WGMMA as register A-operands (`OperandSource.RMEM`), so the `[128,128]`
state never round-trips through shared memory. That removes ~96 KB of SMEM
traffic and two CTA barriers from every step of the serial chain.

## Numbers

H100 80GB HBM3 **SXM**, production gate (`gate_lower_bound = -5`), non-zero
`initial_state`, geomean over ten shapes, FlashKDA as an **in-run control**
(measured in the same process, same harness, not quoted from elsewhere):

| metric | this kernel | kernel it replaces | FlashKDA | vs FlashKDA |
|---|---|---|---|---|
| per-call | **368.5 us** | 402 us | 438.7 us | **1.19x** |
| pipelined | **257.5 us** | — | 398.2 us | **1.54x** |

against 1.09x per-call for the kernel it replaces.

"per-call" synchronizes each iteration, so it includes host dispatch;
"pipelined" queues iterations back to back and measures steady-state GPU
throughput. Both are reported because KDA is launched in both regimes.

## Correctness

Unchanged envelope, and the kernel is scored by the campaign's own fp64
oracle at the production gate with a non-zero `initial_state` — the gate that
matters, since the v1 campaign winner was correct at `gate_lower_bound = -1`
and returned 100% NaN at `-5`.

## Test

`test/python/linear_attention` on H100: **494 passed, 3577 skipped, 1 xfailed**.

## Summary by CodeRabbit

- **Performance**
  - Improved Hopper GPU performance for linear attention prefill workloads.
  - Reduced launch overhead and improved efficiency across full and partial sequence chunks.
  - Optimized processing for vector-valued outputs and larger workloads.

- **Compatibility**
  - Improved handling of variable-length sequences and tail segments.
  - Enhanced workspace management for repeated workloads and transposed intermediate data.


YangXu1990uiuc:
Codex bot review · model `gpt-6-astra`

I am requesting the Python test CI for head `4d5f820f45a0`; the following standalone trigger comment is also from this bot. This target includes the Hopper Python tests. For future updates, request it with a separate comment containing only `@cudnn-ci-bot run python_tests`; contributors without trigger access can ask a maintainer. I am also checking this kernel independently on H100.

YangXu1990uiuc:

YangXu1990uiuc (approved):

Reviewed `4d5f820f45a0` against base `a3f82cdf9a7b`.

The transposed recurrence and segmented scan make sense, and the longer sampled workloads improve. I found one correctness regression in the unchanged layout contract and a small-workload performance regression; no P0.

**[P1] Preserve declared strides in the new input and state views** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/4d5f820f45a0077900c50aeafdfddf88946be8ea/python/cudnn/linear_attention/hopper/kernel/kda_prefill_sm90.py#L205))

The new PREP computes Q/K/V/g addresses as contiguous THD and beta as contiguous TH. The new state slab views likewise hardcode row stride 128 (lines 465, 660 and 731). However, KdaHopperEngine.check_support() still accepts noncontiguous tensors; its THD check only checks rank. Valid row-padded inputs now silently read the wrong elements. For example, allocate Q backing storage as [33, 2, 256] and pass [..., :128], giving strides (512, 256, 1). Independently, padding only initial_state to row stride 256 also breaks both O and final_state. Preserve each operand's declared strides when hoisting the views, or explicitly decline the unsupported layouts at support checking. Add Hopper layout coverage: the current noncontiguous tests select FROST only.

Evidence: Fresh-process public kimi_delta_attention(..., plan_name="kda_hopper") on H100 NVL, lengths [17,0,16], H=2, BF16 QKV, FP32 gate/beta/state, nonzero initial_state and production gate. Q-only padding: O relative RMS 0.00310 on baseline versus 7.045 on head. Initial-state-only padding: O/final-state relative RMS 0.00310/0.000540 on baseline versus 15.493/10.392 on head. Same values, oracle and environment; both baseline probes passed. Attribution: Introduced by this PR: baseline 85aa39bdfdfb indexes the declared tensor layouts. These are not failures caused by warming the shape cache with a different layout.

**[P2] Keep a low-launch-count path for small packed batches** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/4d5f820f45a0077900c50aeafdfddf88946be8ea/python/cudnn/linear_attention/hopper/kernel/kda_prefill_sm90.py#L855))

The combined PREP path is restricted to nseq == 1, so even a tiny packed batch always launches separate full-chunk and tail PREP kernels. For lengths [17,0,16], H=2, the complete GPU call regresses from about 9.8 us to 13.2 us (+35%). Please include small packed/continuation batches in the performance set and consider selecting combined PREP for them too; PREP_MODE=0 already contains the varlen mapping. This can coexist with the gains from the split path on larger workloads.

Evidence: H100 NVL, baseline/head/head/baseline in separate processes, warmed public API captured as eight calls per CUDA Graph. Median GPU us per call: 9.825 / 13.214 / 13.216 / 9.792. All four arms passed the fp64 oracle, changed-Q replay and poisoned-output/workspace checks before timing. Each arm uses seven timing groups. Attribution: Measured regression from baseline 85aa39bdfdfb to this head. The extra PREP launch is a source-supported explanation; this probe does not independently isolate its entire cost.

Validation: H100 NVL targeted Hopper KDA tests with explicit -m "L0 or L1": 16 passed, 6 skipped (22 collected); source and kernel route verified. Correctness-gated ABBA GPU spot checks also improved for lengths [1024], H=4 (baseline 63.6–63.9 us, head 35.7–36.2 us) and [1025,0,31], H=2 (62.7–62.9 us to 50.6–50.7 us). Warmed public-API CPU enqueue medians for the same three shapes were baseline/head: 161–165/166–170 us, 176–177/177–184 us, and 177–178/174–176 us. These short samples do not establish a broad CPU regression. Current-head Python test CI was missing; requested it with the signed explanation above and verified its running acknowledgment. Style passed.

Limitations: Timings cover three H100 NVL workloads and exclude compilation; they do not reproduce the PR's H100 SXM/FlashKDA geomean. Python CI is still running. The unused single-segment workspace allocation already noted by CodeRabbit also exists in the baseline; it is a valid cleanup, not a newly introduced allocation regression.

Anerudhan:
Folded into #1061 so a single PR carries both sm90 engines: the kernel swap commit (`4e60c30a`) is cherry-picked there unchanged, and #1061's comparison table now measures this kernel as its own row against current `develop`.

Closing to avoid two PRs touching the same Hopper path.

YangXu1990uiuc:
**Codex bot review follow-up · model: `gpt-6-astra`**

I see this patch was folded into #1061. Please make sure the [P1 layout finding](https://github.com/NVIDIA/cudnn-frontend/pull/1056#pullrequestreview-5203171317) is addressed there before merging: on the reviewed `4d5f820f45a0`, accepted row-padded inputs or initial state silently produce incorrect output/state because the hoisted views assume contiguous strides. This is a high-risk correctness issue. Preserve the declared strides or reject unsupported layouts, with Hopper regressions for input padding and state padding separately. I have not yet reviewed #1061's current head.

Issue #737, opened 2026-08-25
frost(sdpa): THD ignores bound ragged offsets — TE-style padded THD is silently mis-addressed, and cannot be declined at plan time


vedaanta:
## Summary

The THD lowering reconstructs each sequence's base row as `prefix(lens) x token_stride` and **never reads the ragged-offset tensors the caller bound**. `fwd/engines.py` states this outright:

> THD lowerings assume FULLY-PACKED storage: the packed addressing is re-derived as `prefix(lens) x token stride`, and the graph's bound ragged-offset values are never read. TE-style padded THD (offsets from `cu_seqlens_padded != cu_seqlens`, gaps between sequences) is NOT served — **and being runtime data, cannot be declined at plan time.**

That last clause is the problem. A padded-THD caller is not rejected; the graph is claimed and **silently mis-addressed**.

## Concrete failure

TE's `DotProductAttention` documents this layout directly. A batch of 3 sequences with identifier tokens inserted between them:

```
[a a a 1 b b 2 2 c c c c 3]
cu_seqlens        = [0, 3, 5,  9]
cu_seqlens_padded = [0, 4, 8, 13]
```

Sequence 1 starts at row **4** (`b`). We would derive `prefix(lens)[1] = 3` and start it at row **3** — the identifier token `1`. Every sequence after the first is shifted, output is wrong, and nothing reports an error.

Memory safety is not at risk: the per-sequence offset enters as a TMA *coordinate* (`tma_k(0, head, kv_row_base + ... + kv_seq_off, tma_batch)`), which the hardware bounds-checks against `GLOBAL_DIM`, and #706 bounds capacity by the view's element span. So this is silent wrong results, not a fault.

## Proposal: consume the bound offsets on device

The offsets are already device tensors in the variant pack, and the setup kernel already runs on device and writes the `[seq_kv | cu_q | cu_k]` metadata. It could read the caller's ragged-offset tensors instead of reconstructing offsets from lengths:

- **packed callers unaffected** — bound offsets equal the derived ones by definition;
- **padded callers become correctly served** rather than silently mis-served;
- **no host read**, so #552's zero-host-read contract and CUDA-graph capture are untouched.

This converts "cannot be declined at plan time" into "does not need to be declined" — strictly better than any validation we could add, since validating a device-resident prefix sum would require exactly the D2H read #552 removed.

## What needs checking before calling it simple

[...]

PR #1058, opened 2026-09-14
frost(sdpa_bwd_sm80): read the bound ragged offsets on device (issue #737)

egilliam-nv:
## Summary

`sdpa_bwd_sm80` now **reads the bound ragged offsets on device** and places every port's sequences at their own token origins, so a padded THD layout — TE's `cu_seqlens_padded`, gaps between sequences, per side or per port — is read and written where the caller put it. First row to land issue #737; the metadata extension and the analyzer/lowering plumbing are shared, for the other rows to adopt. Follows the design note on #737.

## Why

Every FROST THD lowering derived each sequence's base as `prefix(lengths) × token stride` and never read the offset values, so a padded graph was accepted and silently mis-addressed (#737, and the Codex finding on #950). TE and Megatron are the padded-THD producers (`PackedSeqParams` carries `cu_seqlens_*_padded`), and the pad-zeroing work on the backend side (NVBug 4718457) needs the same knowledge of where each sequence's rows start.

## What changed

**Setup launch** (`thd_helpers.build_thd_sm80_bwd_meta_kernel`): after the shared `[seq_kv | cu_q | cu_k | …]` metadata it writes one **token origin per port and sequence**, `org_p[b] = ro_p[b] × M_p / ts_p` — `M` the port's `ragged_offset_multiplier`, `ts` its token stride, both plan-time constexprs; `ro` the bound device tensor. A port bound without an offset copies its side's `cu`, so the direct adapter and the standalone wrapper are unchanged. Append-only extension of `tile_dsl.thd` (`THD_ORG_OFF`, `THD_ORG_META_WORDS`, `THD_ORG_DEAD`).

**Kernels**: two kinds of row origin. The **internal** packed buffers (fp32 dQ accumulator, `do_dot`, GQA dK/dV partials) stay at `prefix(lengths)`, so their sizing and the mask / loop bounds do not change. The **caller's** ports go through `ORG`: Q/dO loads and the Stats read (Q side), K/V loads and the direct-bound dK/dV stores (KV side), and the bounded dQ cast, dK/dV fold and O·dO dot, which iterate `(sequence, row)` over the envelope `S_max` and touch `org_p[b] + row` for `row < len[b]` only. Gap rows are never read or written.

**Contract: offsets are whole tokens.** The setup kernel checks the remainder; a violation cannot be declined at plan time (device data), so the sequence is marked **dead** on every port: length 0 for the kernels, its rows untouched (there is no correct address to write to), and the metadata flag word set for a test to assert on. Declared `max_total_seq_len_q/kv` must cover the padded span, which is what a padded-THD producer declares.

**Plumbing**: `SdpaBinding` carries the nine ragged-offset tensors; the lowering passes each port's `(offset dtype, multiplier)` to the adapter at plan time and the bound tensors at execute, only to an adapter whose `execute()` takes them. `Capabilities.thd_ragged_offsets` names the rows that read the offsets and gates the one rule that assumed `prefix(lengths)` addressing (the head-major Stats head-stride bound).

## Tests

`test/python/sdpa/frost/test_sdpa_bwd_thd_sm80_offsets.py` (new, 9 cases): TE-style padding per side in element and token units; every port padded on its own schedule; both Stats packings; GQA + causal; a declared total past the pads; the dead-sequence contract with the flag word; a probe check that offsets do not change eligibility. Inputs are re-homed in NaN-filled padded records (a gap read poisons the result), gradients are NaN-filled and their gap rows asserted untouched.

Local:

| suite | device | result |
|---|---|---|
| `test_sdpa_bwd_thd_sm80_offsets.py` | A100 | 9 passed |
| SM80 THD / dense / wrapper / integration / analyzer / FlashInfer-shaped / knob vocabulary / execute-is-async | A100 | 214 passed, 21 skipped |
| `test_sdpa_bwd_thd_sm100.py` + `test_sdpa_graph_analyzer.py` (shared probe + lowering) | SM100 | 141 passed |
| `test_mhas_v2.py::test_sdpa_random_bwd_ragged_L0` with `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1` | A100 | 383 passed, 1 skipped (harness skip), 0 failed — every draw served by `sdpa_bwd_sm80` |

## Docs

Tracker footnote ᵏ (Rule S2), `Capabilities.thd` (fwd) and `thd_ragged_offsets` (bwd) comments, `_sm80_spec` docstring, kernel header, and a design-doc section *Per-port token origins from the bound ragged offsets*.

## Open for review (see the #737 note)

Whether the origins block belongs in the shared `tile_dsl.thd` layout as done here, and whether a bad remainder should leave the sequence untouched (as here) or zero it — the latter has no correct address to write to, which is why this PR leaves it untouched and sets the flag.

## Labels

`cat-feature` · `area:frost` · `area:global_attention` · `orig-nv-eng`


egilliam-nv:
Implements the design note on issue #737 for the SM80 backward; opened as a draft pending Vedaanta's feedback on the note (origins block in the shared THD layout, dead-sequence handling).

egilliam-nv:

egilliam-nv:

egilliam-nv:

PR #1002, opened 2026-09-10
Add state_indices to FROST LA to support paged states

Anerudhan:
## What

Adds an optional `state_indices` argument to `cudnn.linear_attention.gated_delta_net`, so the GDN recurrent state can live in a **paged pool** owned by the caller instead of a compact buffer laid out in sequence order.

With `state_indices=[N] int32`, `initial_state` is a pool `[N_pool, HO, V, K]` and sequence `i` reads row `state_indices[i]`. Under `output_final_state` that same row is overwritten in place, so the pool is both source and destination and the returned `final_state` **is** the pool.

## Why

vLLM and SGLang both keep linear-attention recurrent state in a paged pool and hand kernels one row id per sequence. FlashInfer's `chunk_gated_delta_rule` already models this (`state_indices`), and its docstring gives the reason directly: it lets callers "avoid gathering the active rows into a packed buffer and scattering the result back."

cuDNN had no equivalent — `ops/gdn.py` asserted `initial_state.shape[0] == N` — so a caller holding a pool had to gather in and scatter out around every call, which costs more than the kernel saves. This closes that gap and is a prerequisite for any serving stack routing GDN prefill through cuDNN.

The naming and shape contract follow the existing in-repo precedent from `causal_conv1d_update`'s `conv_state_indices`.

## How

**Kernel** (`frost/kernel/gdn_prefill_f16.py`) — thread an optional `mStateIdx`. Absent it, the state row is the sequence, exactly as today. Present, the seed read uses `mStateIdx[batch_idx]` and the final-state write maps the work item's existing `WORK_ITEM_FINAL_DST` through the same table, so both ends of the recurrence land on the caller's row.

Two properties made this cheap and are worth calling out for review:
- state moves by **plain global access, not TMA**, so this is a slot substitution rather than descriptor work;
- `host()` already reshapes the state to put batch in the last mode *and preserves `stride[0]`*, so a **non-compact pool needs no extra handling**.

Presence of the table is a `const_expr` branch and therefore a distinct compilation, so it joins the compile cache key.

**Op layer** — because the pooled path writes through an input, it is a **separate custom op** (`cudnn::gated_delta_net_fwd_pooled`) declaring `mutates_args=("initial_state",)`, rather than a flag on the existing op whose `mutates_args=()` would then be a lie under `torch.compile`. It returns only `o`; the updated state is the caller's pool, not a returned alias. The shared body moved to `_gdn_fprop_impl` and both ops delegate to it.

## Scope, and what is refused

[...]

PR #1061, opened 2026-09-14
la(kda): sm90 KDA — fused-CUDA engine (fwd+bwd), CuTeDSL kernel swap, and default-engine heuristics

Anerudhan:
## What

Two sm90 KDA changes that belong together, plus the routing change that makes
them reachable by default:

1. **A new engine, `kda_hopper_cuda`** — CUDA C++ compiled with NVRTC and
   launched through the driver API, serving **both directions**. Its backward is
   the **first sm90 KDA backward** in the tree: `kda_engine.py` has a forward
   kernel only, so a Hopper KDA backward currently falls to `kda_cutile`.
2. **The Hopper CuTeDSL kernel swap** (`b134720b` → `4e60c30a`), which the
   Kernel Factory campaign finished after #1017's branch was cut. One file; the
   `run_cute` signature is unchanged. (This was #1056, folded in here so one PR
   covers the Hopper path.)
3. **A KDA heuristics hook**, so the fastest engine on a device is the default
   instead of the lowest slot number.

## Numbers

H100 80GB HBM3 **SXM**, production gate (`gate_lower_bound = -5`), non-zero
`initial_state`, geomean over ten shapes. **Every row below was measured in one
job, with fla and FlashKDA re-measured inside each run as controls** — no row is
quoted from another process or another day.

### Forward

| backend | geomean | vs FlashKDA | vs fla |
|---|---|---|---|
| `kda_cutile` — *today's default on sm90* | 734.3 us | 0.58x | 1.16x |
| `kda_hopper` — CuTeDSL, kernel on `develop` today | 397.4 us | 1.08x | 2.15x |
| `kda_hopper` — CuTeDSL, **with the swap in this PR** | 364.7 us | 1.17x | 2.33x |
| **`kda_hopper_cuda` — new, the new default** | **209.2 us** | **2.04x** | **4.00x** |
| FlashKDA (control) | ~427 us | 1.00x | — |
| fla Triton (control) | ~849 us | — | 1.00x |

So against what an sm90 user gets today, this PR is **3.5x**. The one loss is
the smallest shape (1024/4/1, 0.91x vs FlashKDA), where the kernel is short
enough that host dispatch dominates — see below.

### Backward — six shapes, through autograd

| | per-call | pipelined | vs cuTile |
|---|---|---|---|
| **`kda_hopper_cuda`** | **632.5 us** | **486.0 us** | **2.80x** |
| `kda_cutile` | 1771.5 us | 1361.7 us | 1.00x |

6/6 shapes faster, 6/6 correct.

### Host dispatch, and CUDA graphs

These engine-level figures sit well above the kernels standalone (forward
53.9 us, backward 316.2 us) because the path is **CPU-bound**: at 2048/12/1 one
call takes 148 us of host time to issue against a ~60 us kernel. Decomposed by
stubbing stages out:

| stage | us | share |
|---|---|---|
| cuDNN FE op + graph + variant-pack layer | 84.1 | 57% |
| `variant_pack.operands()` | 12.5 | 8% |
| address dict (9x `data_ptr`) | 5.5 | 4% |
| driver launch, bare | 11.7 | — |

The dominant term is shared FE machinery that **every** engine pays — on the
same shape `kda_cutile` pays 417 us of host time against this engine's 148 us.

**CUDA graph capture removes it**, and this path captures:

| | eager wall | captured wall | eager host | captured host |
|---|---|---|---|---|
| `kda_hopper_cuda` | 176.1 us | **27.2 us (6.5x)** | 151.4 us | **3.0 us** |
| `kda_cutile` | 497.0 us | 206.7 us (2.4x) | 416.7 us | 2.9 us |

Captured, the new engine is **7.6x** cuTile at that shape. Replay reproduces
eager numerics in both cases. `test_cuda_graph_capture_replays` guards this,
because capture breaks silently the moment anything on the path allocates or
synchronises per call.

Not attempted: caching the ctypes parameter block would save ~4 us, but it is
rebuilt per call precisely so two threads executing one graph cannot hand each
other the other's pointers.

### FlashInfer path

The numbers above are the cuDNN op. These are the same engine measured through
**FlashInfer's own entry point**, `flashinfer.cudnn.linear_attention.cudnn_recurrent_kda`,
on FlashInfer's **default** KDA contract — in-kernel q/k L2 norm, in-kernel safe
gate from `A_log`/`dt_bias`/`lower_bound`, beta as logits, bf16 state, int64
`cu_seqlens`. That is the contract vLLM and SGLang use, and before this PR **no
sm90 engine served it at all**.

The baseline needs stating precisely. Because no other sm90 engine accepts that
contract (cuTile declines a bf16 state and an int64 chunk table), "the same call
on the other engine" does not exist to measure. The `before` column is therefore
what a Hopper caller would have had to do instead: convert the state to fp32 and
`cu_seqlens` to int32 so cuTile can serve it, paying those conversions per call.
Both columns compute the same function through the same entry point; only the
contract the caller has to meet differs. H100 SXM, geomean over five shapes:

| | after (this PR) | before (adapt + cuTile) | speedup |
|---|---|---|---|
| eager | 606.4 us | 1011.5 us | **1.67x** |
| CUDA-graph captured | **172.0 us** | 659.9 us | **3.84x** |

Per shape, captured: 3.42x / 4.04x / 3.83x / 3.84x / 4.09x — the win is flat
across 2048-16384 tokens, so it is the engine and not one favourable size.
Captured is the headline because the eager path is host-bound by the FE
op/graph layer at ~150 us, which sits above the kernel and compresses the ratio.

### What the in-kernel fusions cost

Captured, H100 SXM, against the same call with the transforms done outside:

| fusion | cost |
|---|---|
| q/k L2 norm | 1.26x |
| safe gate | 1.23x |
| beta sigmoid | 1.02x |
| all three | **1.47x** |

### No regression for callers who use none of this

The fusion flags are compile-time (`-DKDA_FUSED_FLAGS`, one cubin per
combination) and the equal-head case is specialised the same way, so the plain
contract compiles to the pre-fusion code path — 68440 bytes against 68792.
Three interleaved repeats on H100 SXM, captured geomean:

| | pre-fusion tree | this PR |
|---|---|---|
| rep 1 | 56.86 us | 56.71 us |
| rep 2 | 56.88 us | 56.83 us |
| rep 3 | 56.75 us | 56.92 us |

Within 0.3% either way. (As *runtime* branches the same measurement showed a
2.7% regression, which is what motivated the specialisation.)

## Routing: the new engine becomes the default on sm90

The KDA family declared no heuristics hook, so it fell back to **slot order —
which is identity, not preference**. A slot is fixed forever because an autotune
result is `(engine_id, knobs)`, and `kda_cutile` holds slot 1 because it shipped
first. That is why the slowest engine on Hopper was winning by default.

`kda_heuristics.recommend` is **ordering only**: on sm90 it promotes
`kda_hopper_cuda`, then `kda_hopper`, and leaves every other engine exactly
where slot order had it; off sm90 it returns slot order unchanged, so **no
non-Hopper graph changes engine**. `plan_name` still overrides.

### One shared-code change, and why it is required

`heuristics.rank()` built `offered` from every candidate engine, unfiltered,
while the hookless path (`_unranked`) has always filtered by `check_support` —
and `_build_plan_at()` builds a ranked entry **without re-affirming
`check_support`**, so whatever reaches the list is taken as eligible.

Ordering alone was therefore not safe: ranking the 128-only sm90 engines ahead
of a head-dim-64 graph launched the kernel on 64-dim tensors and raised an
**illegal memory access**. `rank()` now filters `offered` by `check_support`,
which is what the name always meant and what the hookless path already did.
Families with a capability table (SDPA, gemm) can only see the same set or a
smaller one, so this can remove a plan that would have failed at build but never
adds one. It does mean every family now pays a `check_support` probe per
candidate at plan time.

**This is the one part of the PR outside linear attention and deserves an FE
reviewer.** `fe_api` is green (370 passed), which covers the SDPA and gemm
heuristics paths.

## Correctness

Scored against the campaign's fp64 oracle at the production gate with a non-zero
`initial_state` — the gate that matters, since the v1 campaign winner was correct
at `-1` and returned 100% NaN at `-5`.

* **forward**: 6/6 workloads, both outputs, 100% element match, worst scaled
  error 0.34 against a tolerance boundary of 1.0.
* **backward**: 6/6 shapes, **all six gradients**, scored by the definition's
  three-term checker (per-element, hard cap, RMS — built to reject a K-major
  `d_initial_state`, a zeroed output, and a uniform 5% bias a per-element check
  is blind to). Worst scaled error 4.6e-03 to 7.1e-03, matching the standalone
  campaign figures (4.57e-03 to 7.14e-03), which is the evidence the integration
  is faithful.

## Implementation notes worth a reviewer's time

**Templated kernels need their own compile path.** The backward is four
`__global__`s and seven of the nine reachable entry points are template
instantiations. A templated kernel has no stable C name, and NVRTC only emits and
names one if it is registered with `nvrtcAddNameExpression` before the compile
and looked up with `nvrtcGetLoweredName` after it (a `__global__` cannot be
called from another `__global__`, so an `extern "C"` wrapper is not an option).
Hence `cuda_bwd_host.py` has its own compile, caching the cubin and the
lowered-name map **together** — a cubin without its names is useless here.

**Constants are restated in Python behind a compile-time tripwire.** NVRTC
internalises a `__device__` variable no kernel references — it lands in the cubin
as a LOCAL symbol `cuModuleGetGlobal` cannot resolve — so the launcher cannot
read `sizeof(FusedSmem)` or the backward's tiling constants out of the module.
Python owns them and passes them as `-D`; the bodies `static_assert` against
those, so a future change to the kernel's tiling is a compile error naming the
constant rather than a wrong grid. Verified RED-then-green in both bodies.

**Vendoring is mechanical and re-runnable.** `vendor_bwd_kernel.py` concatenates
the campaign's multi-file nvcc project into one NVRTC translation unit; no device
code is edited, so a re-vendor is a clean diff.

`compile_cubin` / `KernelLibrary` grow a **defaulted trailing** `extra_options`
so the `-D` threads through without touching the Blackwell call sites; it is part
of the cubin cache key, so option sets cannot collide.

## Test

`hopper_cuda` joins `test_la.py`'s shared `backend` fixture — `Backend.plan()`
already composes `<variant>_<name>`, so every existing KDA case runs against it
and configurations it declines waive themselves, exactly as `cutile` and `hopper`
do. `test_kda_sm90_cuda.py` adds what that fixture cannot reach: default
selection, `plan_name` override, the head-dim-64 fall-through (the RED-then-green
record of the IMA above), and CUDA graph capture.

On H100: **`linear_attention` 576 passed, `fe_api` 370 passed, 0 failed.**

## Summary by CodeRabbit

- **New Features**
  - Added a Hopper GPU backend for Kimi Delta Attention (KDA) on SM90-compatible hardware.
  - Added fused forward execution and backward gradient support, including variable-length sequences and optional state outputs.
  - Added runtime compilation options with improved kernel-cache handling.
  - Added specialized handling for full and partial sequence chunks.
  - Added automatic engine ordering and filtering based on graph and device support.
- **Tests**
  - Expanded linear-attention coverage for the Hopper CUDA backend, including supported backward scenarios and CUDA graph capture.


Anerudhan:
@cudnn-ci-bot run frost

YangXu1990uiuc:

I am requesting `python_tests` and `oss` test CI for the current head (`81e12b2931b8`) with the next standalone command. The existing FROST run is for `6417267`; its linear-attention lane is SM100, so it does not exercise this new Hopper backend. The Python-test target includes Hopper linear-attention/routing tests, and OSS covers the shared frontend heuristics change. For future Hopper updates, include `@cudnn-ci-bot run python_tests` and verify the run acknowledges the new head.

YangXu1990uiuc:
@cudnn-ci-bot run python_tests oss

YangXu1990uiuc (changes requested):

Reviewed `81e12b2931b8` against base `0ed2710bc6db`.

The separate CUDA engine and shared eligibility filter make sense, and the new default is substantially faster in the three measured forward cases. Two reproduced silent-correctness hazards need fixing before approval: unsupported layouts and optional zero-state initialization. The earlier CuTe layout/performance comments and the linked CodeRabbit comments remain unresolved; the new CUDA layout finding extends beyond the earlier CuTe reports. The folded CuTe source is unchanged from [the #1056 review](https://github.com/NVIDIA/cudnn-frontend/pull/1056#pullrequestreview-5203171317); its small-packed PREP-launch performance follow-up is carried forward there, without claiming a fresh pinned-CuTe timing run.

**[P1 · blocks approval] Validate layouts before launching the new raw-pointer engines** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/81e12b2931b8aca74cf70bcdb1e86e02ebfc513b/python/cudnn/linear_attention/hopper/cuda_engine.py#L120))

The new forward and backward engines discard operand strides and pass only addresses to kernels that index compact THD/state arrays. Neither check_support nor execute rejects padded outer strides. This now silently miscomputes ordinary fused-projection slices, padded recurrent states, and a row-padded incoming dO on the default Hopper route. The existing stride tests in test_la.py explicitly restrict backend to frost, so adding hopper_cuda to the shared fixture does not exercise them. Honor the layouts or reject unsupported operand/output layouts before launch (and make the public Torch adapter/plan selection agree). Check every port, including dO and optional states. This is a high-risk P1 because accepted calls silently return incorrect outputs and gradients.

Evidence: H100 NVL, BF16, D=128, H=2, lengths [17,0,16], nonzero FP32 initial state: compact CUDA O/state relative RMS errors were 0.00310/0.000663; changing only Q to stride (512,256,1) made O error 7.045; changing only state to stride (65536,32768,256,1) made O/state errors 15.49/10.39. Separately, a dO with the same logical values and stride (512,256,1) changed all five gradients, with relative errors 1.32–2.47 versus contiguous dO. All backing allocations were valid; failures were numerical, not intentional invalid-memory probes. Attribution: New CUDA engine/default-route defect. The folded CuTe swap also retains the already-reported #1056 layout defect: its source is unchanged from that reviewed head, and a fresh baseline 85aa39bdfdfb CuTe control passes both Q/state layouts. See https://github.com/NVIDIA/cudnn-frontend/pull/1056#pullrequestreview-5203171317 and https://github.com/NVIDIA/cudnn-frontend/pull/1061#discussion_r4011876740 for that existing CuTe finding.

**[P1 · blocks approval] Order optional zero-state initialization on the execution stream** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/81e12b2931b8aca74cf70bcdb1e86e02ebfc513b/python/cudnn/linear_attention/hopper/cuda_engine.py#L144))

With initial_state omitted, torch.zeros queues initialization on the ambient Torch stream, while the kernel launches on ctx.stream. There is no dependency between them. A fresh plan can therefore read old allocation contents when the caller uses a handle on another stream. Caching the tensor also treats capture-time allocation as completed initialization: if the first execute is captured, an eager execute before that graph replays reads the uninitialized cached seed. Backward _zero() uses the same pattern for absent initial_state/d_final_state. Please make initialization and ownership safe across execution streams and capture; a null/zero-state kernel path or per-execution, stream-ordered scratch avoids publishing a not-yet-ready cached constant. Warm capture tests with explicit initial_state do not cover these cases. This is a high-risk P1 because valid handle/capture usage silently corrupts results.

Evidence: Two independent current-head H100 public graph-API reproductions, with kernel loading warmed first: (1) delay the ambient stream, reuse an allocation previously filled with ones, and execute a fresh no-state plan on a separate handle stream; (2) first capture the no-state plan in a shared graph pool containing a freed ones allocation, then call it eagerly before first replay. Both verified pointer reuse and produced O/state relative errors 34.09/81.11. After the zero fill completes or the captured graph replays, the same plan returns 0.00299/0.00277. The initial un-warmed stream probe passed because first-use kernel loading synchronized; it is not evidence against the warmed reproduction. Attribution: Introduced by the new CUDA plans. Forward reproduced on hardware; the equivalent backward initialization hazard is source-supported, not independently GPU-reproduced.

- **[P2] Still open:** [Pass the forward shared-memory define to NVRTC](https://github.com/NVIDIA/cudnn-frontend/pull/1061#discussion_r4010532299).

- **[P2] Still open:** [Assert the requested cuTile route positively](https://github.com/NVIDIA/cudnn-frontend/pull/1061#discussion_r4011926017).

Validation: Reviewed all 13 changed files (4,811 additions/226 deletions), callers, eligibility/dispatch, NVRTC cache options, both CUDA kernels, and the folded CuTe delta; clean merge check against develop 0ed2710bc6db. Matching native binding reused for these Python-only changes. Current head on H100 NVL, Torch 2.13.0+cu130, cuDNN 9.25.1, CuTe DSL 4.8: 7 focused forward/backward numerical tests passed (including empty/partial sequences, initial-state and final-state gradients); all 5 new routing/capture tests passed. Explicit L0-or-L1 selection prevented silent deselection. Shared dispatch and SDPA heuristics checks: 96 passed, 2 skipped. An initial harness-induced failure disappeared after removing its global FROST opt-in flag. Fresh guarded layout probes and public graph stream/capture probes reproduced the two P1s; compact-layout and initialized-state controls passed. Padded dO alone also reproduced the backward layout defect. Correctness-gated ABBA between baseline 85aa39bdfdfb and head 81e12b2931b8, same H100 NVL and dependencies, default route verified by profiler (cuTile to kda_fused). For lengths [17,0,16]/H2, [1024]/H4, [1025,0,31]/H2, median CPU enqueue was 313.7→100.4, 316.1→103.3, 318.4→102.6 us; captured GPU time per call was 37.12→13.50, 79.26→20.95, 77.88→20.53 us. Compilation/autotuning excluded; changed-input, poisoned-output replay checks passed. Requested python_tests and oss CI, verified the mirrored full head equals 81e12b2931b8. Hopper jobs are running/pending; the Rubin OSS runner failed during setup for lack of disk space, before tests.

Limitations: Targeted H100 validation, not a full multi-architecture regression or backward performance comparison. The three forward timing cases do not validate the broader performance claims in the source docstring. The equivalent backward optional-zero lifecycle hazard is source-supported; only forward stream/capture cases were reproduced. Current-head CI is not yet complete.

YangXu1990uiuc:

CI follow-up for head `81e12b2931b8`: the requested Python-test lanes passed on SM80/SM90/SM100 (dev and release), as did OSS on SM80/SM90 and both SM100 shards. The Hopper dev Python lane reported 5,175 passed and 2,641 skipped.

The two remaining failures are outside the PR's code delta: the Rubin runner ran out of disk space before tests, and the text scan flags `Conv->GEMM` in a convolution comment from develop. These green lanes do not resolve the independently reproduced layout and optional-zero-state stream/capture P1s in the review; please address both before merging.

jhjpark (approved):

Anerudhan:
@cudnn-ci-bot run oss, python_tests

Anerudhan:
Thanks — both P1s were real, and both are fixed in `b3b56a19`. Your repros were precise enough to reproduce and regression-test directly, which I appreciated.

**P1 · padded layouts.** Confirmed and fixed in `hopper/layout.py`, with two guards because a layout arrives from two places:

- `declared_layout_reason()` in `check_support` reads the strides the **graph** declares. That is the routing gate — such a graph is declined and another KDA engine serves it. Since the op layer declares no strides, this is the path a public-API graph hits, which is how FlashInfer drives cuDNN.
- `packed_addresses()` per `execute` reads what the **caller actually passed**, because `_normalize` takes each operand's geometry from the producer's DLPack vtable and never checks it against the declaration.

One deliberate asymmetry: a padded **input** is repacked into a contiguous copy rather than rejected. `ops/kda.py` has always accepted strided q/k/v slices — `test_la.py::test_fwd_innermost_contiguous_inputs` covers exactly that — so rejecting would have been a capability regression *caused by* making this engine the sm90 default. A padded **output** still raises, since it is written in place and copying back would be a second silent behaviour. The CuTe engine gets the declared gate too, for the `stride=(128, 1)` slab issue you carried forward from #1056.

**P1 · optional zero state.** Confirmed and fixed. The allocation is still cached, but the fill is now `buffers.memset_zero_async` on **the execution stream on every call**. That fixes both halves of what you found: the missing dependency on `ctx.stream`, and the capture case, where the allocation happens once at capture time while every replay is what has to see zeros. Backward's `_zero()` had the identical pattern for `initial_state` and `d_final_state` and got the same treatment — you flagged that as source-supported rather than reproduced, and it was.

**P2 · forward shared-memory define.** You were right, and it was worse than a nit: `_SMEM_DEFINE` was constructed but never passed to `KernelLibrary`, so `KDA_FUSED_SMEM_BYTES` was undefined and the `#ifdef`-guarded `static_assert` compiled to nothing. The tripwire my earlier commit message claimed was armed **had never been armed**. Now threaded through, and verified three ways: correct value compiles, off-by-one fails with the intended message, and no define compiles — which is what made it inert.

**P2 · cuTile assertion.** Now asserts cuTile's own kernels ran, not merely that `kda_fused` did not.

**CodeRabbit `PNT`.** Valid — `k_prep` uses `__launch_bounds__(PNT, 1)` and the launch hardcoded 512. Mirrored with its own `static_assert`, verified failing on an off-by-one.

Both new tests were confirmed **RED** with the guard neutralised, so they have teeth. On H100: `linear_attention` 578 passed, `fe_api` 370 passed, 0 failed.

Still open, and I'd rather be explicit than quietly defer them:

- The two CuTe kernel nits on the folded swap (unused `bf16_c_to_a`, `range_constexpr(1, nseq)` unrolling with `N`) are untouched. They are in vendored campaign-generated code and neither is a correctness issue; I did not want to hand-edit that body in the same change as the correctness fixes.
- Your point that three forward timing cases do not validate the broader docstring claims is fair. The docstring figures are a ten-shape geomean on H100 SXM with FlashKDA as an in-run control, and the backward is six shapes against cuTile; the per-shape tables are in the PR body rather than the source.

YangXu1990uiuc (commented):

Reviewed `7be230558846` against base `94939f3af7f0`.

The execution-stream zero fills fix both of the earlier warmed stream/capture reproductions; the NVRTC shared-memory define and positive cuTile route assertion are also fixed. However, the serving-dtype commit reintroduces the CUDA layout bug after the intermediate fix, and its shared conversion buffers introduce a concurrent-execution race. The previously reported CuTe runtime-state layout issue also remains. An existing approval is present, so this is a comment review; please address these high-risk P1s before merging.

**[P1 · high risk] Preserve the packed input address through dtype marshalling** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/7be230558846424ec2017a5adcb534b1cb0fe8f6/python/cudnn/linear_attention/hopper/marshal.py#L68))

After packed_addresses() repacks an input, execute() calls as_input() with the original view and overwrites that address. For an already-FP32 state/gate (or already-int32 sequence table), this fast path returns the original padded pointer. On H100, changing only initial_state to stride (65536,32768,256,1) again gives O/state relative RMS errors 15.49/10.39; a padded FP32 gate also fails. The same state probe passes on b3b56a19 and fails on 7be23055, so this is a regression introduced by the dtype-marshalling commit. Preserve the validated address through the conversion step, or reject the unsupported runtime layout. Add same-dtype padded state/gate cases; the new 15-case suite passes despite this. High-risk P1: supported calls silently return incorrect values.

Evidence: Exact-head public Torch API probes, H100 NVL, BF16 Q/K/V, D128, H2, lengths [17,0,16], identical logical inputs and valid padded backing storage. Compact/Q-padded controls pass; FP32-state padding fails. Intermediate b3b56a198db1c9a4d38509c7260459512842121f passes the identical state probe. Attribution: The original raw-pointer layout defect was reported in review 5210171376; b3b56a19 fixes the tested CUDA case, and the newly reviewed 7be23055 commit reintroduces it through a new address overwrite.

**[P1 · high risk] Give simultaneous executions separate conversion scratch** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/7be230558846424ec2017a5adcb534b1cb0fe8f6/python/cudnn/linear_attention/hopper/marshal.py#L53))

The compiled plan owns one _convert buffer per port, shared by every execute(). Two threads can execute the same cached graph with different operands and streams. If A finishes converting its BF16 initial_state and is descheduled before launch, B can overwrite that same FP32 buffer and complete; A then reads B's state. I reproduced this with controlled host scheduling: A's output is bitwise equal to B's serial reference instead of A's, while isolating the staging buffers makes both calls bitwise match their own serial results. Use caller-owned/per-execution scratch with appropriate stream lifetime rather than mutable data buffers shared by a compiled plan. High-risk P1: valid concurrent calls can silently consume another call's data.

Evidence: Exact-head H100 public Torch API, two host threads and CUDA streams, T128/H2/N2, BF16 initial states of opposite sign. A scheduling barrier is inserted between conversion and the unmodified CUDA launch; no kernel or operand values are patched. Shared-cache A error is 0.01953125 and B error is zero; isolated-cache control errors are both zero. Attribution: Introduced by 7be23055 serving-dtype support and the new per-plan _convert cache. Earlier code declined BF16 initial states on this engine.

- **[P1 · high risk] Still open:** [Guard the actual CuTe state layout as well as the declaration](https://github.com/NVIDIA/cudnn-frontend/pull/1061#discussion_r4011876740).

**[P2] Reduce the new overhead for compact, kernel-native operands** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/7be230558846424ec2017a5adcb534b1cb0fe8f6/python/cudnn/linear_attention/hopper/cuda_engine.py#L131))

Even compact BF16 Q/K/V with FP32 gate/state and int32 cu_seqlens now pays the extra layout/dtype marshalling and temporary Python containers on every execute. In a correctness-gated H100 ABBA comparison against 81e12b29, median CPU enqueue rises from 103.5 to 122.7 us for lengths [17,0,16]/H2, and by about 19-20 us in the other two sampled cases. The small case still launches only kda_fused and its captured GPU time is essentially unchanged (13.49 to 13.53 us), localizing this to host work. Please keep the runtime safety checks while preparing/reusing the invariant marshalling work or simplifying the native-dtype fast path. This is nonblocking; the new engine remains much faster than the previous default in the earlier comparison.

Evidence: Same H100 NVL/dependencies, 81e12b29 versus 7be23055, ABBA, three workloads, compilation/setup excluded, profiler route check plus changed-input/poisoned-output capture validation. Larger-workload GPU timings are noisy and support no GPU-regression claim. Attribution: Introduced by the current layout/marshalling revision; original baseline-to-new-engine performance evidence is separate.

Validation: Reviewed the entire eight-file incremental diff from 81e12b29 through b3b56a19 to 7be23055, including both CUDA directions, the CuTe declaration gate, and the new tests. Reused a matching native binding for these Python/kernel changes. Exact head on H100 NVL, Torch 2.13.0+cu130, cuDNN 9.25.1, CuTe DSL 4.8: all 15 routing/capture/serving-dtype tests passed with explicit L0-or-L1 selection. The independent padded-Q and padded-dO checks now pass; padded FP32 state/gate and the CuTe state check still fail as described. Both warmed public graph-API zero-state probes pass: a separate execution handle stream and eager use before the first captured replay. Allocation reuse was confirmed in both, so these exercise the original failure modes. The same padded CUDA state probe passes at intermediate b3b56a19 and fails at final 7be23055. The new shared-conversion-cache concurrency reproduction fails only with shared staging and passes with isolated staging. Correctness-gated H100 ABBA covers three compact forward workloads with actual kda_fused routing, modified inputs and poisoned captured outputs. CPU overhead is reported above; GPU variability limits the larger-case timing conclusions. Current-head GitHub pre-commit and merge requirements checks passed. The available OSS/Python CI acknowledgment is still for 81e12b29, so its results do not validate 7be23055. The owner already knows and uses the CI trigger; no duplicate request was posted for the old running pipeline.

Limitations: Targeted H100 revision review, not a full multi-architecture regression or backward performance sweep. Concurrency is reproduced using an explicit host-scheduling barrier; ordinary stress testing is not claimed. The previously discussed folded-CuTe small-packed performance concern is unchanged and has not received a new pinned-CuTe performance sweep.

Please make sure the high-risk P1 findings above are addressed before merging. Existing approvals are left unchanged.

YangXu1990uiuc (commented):

Reviewed `bcbad8a22eae` against base `07c939a6cce4`.

The new input fusions, grouped value-head mapping and revised L2 normalization pass the focused H100 checks, with large gains for the measured fused workloads. The earlier three high-risk correctness findings remain open; please address them before merging. Safe-gate support also exposes the existing address-overwrite bug to padded FP32 dt_bias. One new nonblocking P2 below covers the slower no-fusion path. This is COMMENT because the PR already has an approval. I will request python_tests for this head with the following standalone bot command; that target includes Hopper coverage.

- **[P1 · high risk] Still open:** [Preserve the packed input address through dtype marshalling](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5214875584).

- **[P1 · high risk] Still open:** [Give simultaneous executions separate conversion scratch](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5214875584).

- **[P1 · high risk] Still open:** [Guard the actual CuTe state layout as well as the declaration](https://github.com/NVIDIA/cudnn-frontend/pull/1061#discussion_r4011876740).

- **[P2] Still open:** [Reduce the new overhead for compact, kernel-native operands](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5214875584).

**[P2] Preserve a specialization for calls without input fusion** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/bcbad8a22eae1f0213aeff393663030ecf1d7e62/python/cudnn/linear_attention/hopper/cuda_kernels/kda_fused_sm90.cu#L416))

The new fusions remain runtime branches even though their flags are fixed per graph node. In H100 ABBA measurements against 7be23055, the existing no-fusion path regresses from 13.50 to 13.86 us, 20.79 to 22.19 us, and 20.54 to 22.17 us for lengths [17,0,16]/H2, [1024]/H4 and [1025,0,31]/H2 respectively (2.7%, 6.7%, 7.9%). Correctness and the actual kda_fused route match. Consider specializing the node-constant flags, potentially also the equal-head case, with those specialization choices included in the compilation cache key. Uniform branches do not guarantee unchanged code generation. This is nonblocking: the separately measured fusion-enabled workloads improve substantially.

Evidence: Exact 7be23055 versus bcbad8a2, same H100 NVL, Torch 2.13.0+cu130, cuDNN 9.25.1 and DSL 4.8. Eight ABBA processes across plain/fused modes, three cases each, seven timing repetitions per case and eight calls per captured graph. All correctness, changed-input/poisoned-output replay and profiler route checks passed; compilation/setup excluded. Attribution: New since the last published 7be23055 review. An unpublished 98c96144 comparison already showed a smaller 2.7-6.5% no-fusion slowdown; the current head retains it after the grouped-head and L2-layout changes.

Validation: Reviewed the full incremental delta since 7be23055, including the 98c96144 fusion revision and the two subsequent grouped-head/L2 commits. Checked launch ABI, Q/K head remapping versus gate/state/output heads, normalization epsilon and zero rows, gate warmup/tail handling, scale/backward eligibility, layout and dtype marshalling. The base-only #1074 changes do not change these successful KDA paths. Exact bcbad8a2 H100 checks: 18 routing/serving/capture tests and 13 selected shared numerical tests passed; 17 unsupported configurations explicitly skipped, with L0 and L1 enabled. Compatible native binding reused. Twelve additional public-op cases passed: HQ/HV=1/4,2/6,3/3; lengths [17,0,16] and [1025,0,31]; fusions disabled/enabled. Included zero and sub-epsilon Q/K rows, distinct per-head gate parameters, nondefault scale, exact empty-sequence state preservation, changed-Q/poisoned-output CUDA Graph replay, and positive verification of default grouped routing to kda_fused. Three correctness-gated forward ABBA workloads with all three fusions enabled: default routing changes from cuTile to kda_fused. GPU medians improve 44.52->16.85, 87.92->16.88 and 85.79->19.36 us; CPU enqueue medians improve 390.6->130.8, 382.1->133.4 and 415.9->134.3 us. This is the full published-head-to-current-head gain, not an isolated measure of the last L2 commit. Without fusions, GPU medians regress as described in the P2. CPU medians are 120.9->128.8, 122.8->126.5 and 122.2->130.6 us, with more variation than GPU measurements. The previous 81e12b29-to-7be23055 host-overhead finding remains separate. Existing defect evidence carried forward by source identity: 98c96144 padded-state and concurrent-conversion reproductions, its padded dt_bias O/state relative errors 12.2%/19.5%, and the preceding CuTe runtime-state-layout reproduction. The latest commits extend features and performance; they do not change the defective marshalling or CuTe code. Existing completed Python/OSS CI acknowledgments are for 81e12b29, not this head. Current-head python_tests will be requested after publication and remains pending.

Limitations: Focused H100 correctness and three forward performance workloads, not a complete all-shape or multi-architecture regression. Grouped-head correctness is new-head evidence; the reported timings use equal Q/K/V head counts. The three high-risk correctness findings and Python marshalling overhead are linked existing comments. This review adds one measured no-fusion performance concern and wider evidence for the existing pointer bug; no additional high-risk finding.

Please make sure the high-risk P1 findings above are addressed before merging. Existing approvals are left unchanged.

YangXu1990uiuc:

YangXu1990uiuc:

CI follow-up for `bcbad8a22eae`: **all six requested Python GPU lanes passed** (SM80/SM90/SM100, dev and release). The SM90 dev run reports 5,229 passed and 2,600 skipped.

The remaining text-scan failure points to the local `dkg` identifier in [kda_fused_sm90.cu](https://github.com/NVIDIA/cudnn-frontend/blob/bcbad8a22eae1f0213aeff393663030ecf1d7e62/python/cudnn/linear_attention/hopper/cuda_kernels/kda_fused_sm90.cu#L934), including its four uses. Please rename that descriptor variable. The independently reproduced layout/concurrent-staging findings in [the review](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5215865541) remain open despite the passing CI suites; please address them before merging.

Anerudhan:
Fixed in `c36eab54`. All three P1s and the no-fusion P2. Thank you for the reproductions — the address-overwrite one in particular I would not have found from the diff.

**P1 · repacked address overwritten.** Correct, and it was mine: `packed_addresses()` recorded the repacked address, then the dtype pass called `as_input()` with the *original* view, whose fast path returns `view.data_ptr()` when the dtype already matches — overwriting the good address with the padded one. An already-fp32 padded state or gate was then read as if compact.

The two passes are now **one**. `resolve_inputs()` decides each port once, and a single `copy_` fixes layout and dtype together — which is all a mismatch ever needed. Nothing downstream re-derives an address, so the failure mode is structurally gone rather than patched.

**P1 · shared conversion scratch.** Also correct. Scratch is now **per execution**, held alive by the caller until after the launch. This one stings: an earlier commit in this same PR deliberately did *not* cache the ctypes parameter block, citing the exact hazard `VariantPack`'s docstring warns about — and then I cached data buffers, which is the same mistake with worse consequences. Nothing is allocated at all on the common path, where every operand is already packed and already the right dtype.

**P1 · CuTe actual state layout.** Added. `check_support` gates the declared layout (that is the routing gate); the engine now also verifies the operands it was actually handed, since `_normalize` takes geometry from the producer's DLPack vtable and never checks it against the declaration.

**P2 · no-fusion path.** You were right that uniform control flow does not guarantee unchanged code generation, and right about the size. The flags are fixed per graph node, so they are now **compile-time** (`-DKDA_FUSED_FLAGS`, one cubin per combination, keyed by the existing option-based cubin cache), and I took your suggestion to specialise the **equal-head** case too (`-DKDA_FUSED_GQA`) so the q/k address folds back to the shared one.

That closes it. `flags=0 gqa=0` compiles to **68440 bytes** against the pre-fusion kernel's **68792**, and a five-repeat interleaved A/B against the pre-fusion tree gives **115.23 us vs 116.19 us** captured — overlapping, no regression. With runtime branches the same measurement showed the 2.7% you reported, so this reproduces your finding and then removes it. Fusion cost is unchanged at 1.37x with all three on.

**On the padded fp32 `dt_bias` you noted under safe gate:** that is the same overwrite bug and is covered by the single-pass fix — `dt_bias` and `a_log` are ports in the same resolution map.

Tests on H100: `linear_attention` 641 passed, `fe_api` 370 passed, 0 failed; forward and backward 6/6 at the production gate; all four flag/GQA specialisations compile.

One thing I have **not** done and want to be explicit about: I have no new adversarial test for the concurrent-execution race. Your repro used controlled host scheduling with a barrier between conversion and launch, which I cannot reproduce from a plain pytest without the same instrumentation. The fix is structural — there is no longer a buffer shared between executions to race on — but it is argued, not pinned by a test, and I would rather say so than imply coverage I do not have.

YangXu1990uiuc (commented):

Reviewed `1d32c921af25` against base `473d35b877fa`.

The single-pass operand resolution and per-execution staging fix the three high-risk correctness findings. I reran the original adverse host schedule: the old head still makes call A consume call B's state, while this head matches both serial references exactly and uses separate staging addresses. The compile-time flags/grouping specialization also removes the measured no-fusion GPU regression. One existing nonblocking host-overhead P2 remains. This is COMMENT because another reviewer's approval already exists. The earlier text-scan issue also still needs its dkg descriptor-variable rename.

- **[P2] Still open:** [Reduce the new overhead for compact, kernel-native operands](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5214875584).

Validation: Exact-head H100 NVL, Torch 2.13.0+cu130, CuTe DSL 4.8 and cuDNN 9.25.1. All 18 routing/serving/capture tests passed with L0 and L1 enabled. Twelve additional grouped/equal-head cases passed reference and changed-input/poisoned-output CUDA Graph checks, including empty sequences, zero/sub-epsilon normalization inputs, per-head gate parameters and nondefault scale. The concurrency control uses one compiled graph with BF16 state inputs, two host threads and two CUDA streams. After warming serial references, monkeypatch only cuda_host.launch: pause A after its conversion has completed, run B through stream synchronization, then resume A. On bcbad8a2, A equals B's output and differs from its own reference by 0.01953125; an isolated-scratch control passes. On 1d32c921, both calls match their own serial O/final-state references bit-for-bit, with distinct input and output staging addresses. This schedule is suitable for a focused pytest regression without replacing production marshalling. Padded Q/gate/beta/FP32 state now match the contiguous CUDA results and reference bounds. Padded dO produces exactly the same five gradients as contiguous dO; padded FP32 dt_bias produces exactly the same O/final state. The CuTe contiguous control still passes, while an actually padded state with a packed declaration is rejected before launch with the offending port and strides. All 16 combinations of the three fusion flags and equal/grouped value heads passed numerical checks in one process and produced distinct library-cache entries. The in-memory key and persistent cubin options both include the specialization; the host launch ABI matches the removed runtime flags argument. Three no-fusion forward ABBA cases against 81e12b29: GPU 13.475->13.480, 20.904->20.853, 20.537->20.547 us, all within 0.25%. Against the previous bcbad8a2 head with all fusions enabled, the same three cases improve 16.813->16.727, 17.066->16.285, 19.352->18.862 us. This closes the prior no-fusion GPU P2; the remaining CPU P2 is quantified above. Eight separate processes, fixed CPU affinity, seven timing repetitions and eight calls per captured graph; every timed arm passed correctness/replay checks and selected kda_fused. The requested Python GPU lanes on bcbad8a2 completed successfully; that is previous-head evidence. No current-head run was found. I will request python_tests with the following standalone bot command: @cudnn-ci-bot run python_tests. The standalone command is a Codex bot action associated with this signed review; current-head CI remains pending.

Limitations: Focused H100 validation and three forward workloads, not a complete suite or multi-architecture benchmark. Compatible native binding reused for the Python/NVRTC-source delta. The surviving CPU overhead is nonblocking; no P0 or high-risk P1 remains from this review. The dkg identifier from the prior text-scan failure is unchanged at current lines955/960/961/967/968; this is source-verified carry-forward, not a new-head CI result.

YangXu1990uiuc:

YangXu1990uiuc:
**Codex bot CI follow-up (model: gpt-6-astra)**

For `1d32c921`, all six requested Python GPU lanes passed: SM80, SM90 and SM100, each with the development and release backend. The text scan still flags the `dkg` descriptor identifier in `kda_fused_sm90.cu`; please rename it as noted earlier. The API-index job also fails because it installs JAX 0.6.2 while the JAX BSA exports require >=0.9.1. The focused correctness and performance results are in [the review](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5217080277).

Anerudhan:
Re-verified against `develop` now that #1002 has merged (`fd409b43`), since both touch `hopper/kda_engine.py` and #1002 is what makes this PR's `state_indices` guards live for the first time.

* **Merges cleanly** — `manifest.py`, `hopper/kda_engine.py` and `test_la.py` all auto-merge, no conflicts.
* **The pooled-state guard now actually fires.** It was written with `getattr` defaults so it was inert until #1002 landed; on the merged tree a `state_indices` graph is declined both unpinned and pinned to this engine, and an ordinary call still reaches `kda_fused`. That mattered because this engine is the sm90 default, so a pooled graph is the one it would have reached first — and ignoring `state_indices` is silent, not loud: it reads row *i* instead of row `state_indices[i]`.
* **Suites on the merged tree, H100:** `linear_attention` 552 passed / 0 failed (the count differs from the 641 I quoted earlier because #1002 restructured `test_la.py`), `test_kda_sm90_cuda` 18 passed, forward and backward 6/6 at the production gate.

Worth noting for anyone looking at pooled state on Hopper: no sm90 engine serves it yet — cuTile declines it too — so a `state_indices` graph has no Hopper path at all right now. That is a gap, not a regression, and it is separate from this PR.

YangXu1990uiuc (approved):

Reviewed `1d32c921af25` against base `fa7a433ee844`.

Correction to my previous follow-up: the three approval-blocking correctness findings are resolved, so I should have submitted APPROVE to clear my earlier REQUEST_CHANGES. Another reviewer's approval and my subsequent COMMENT did not clear that block. The fixes for single-pass operand resolution, per-execution staging and compile-time specialization remain verified on this head; no approval-blocking finding remains. The existing CPU-overhead P2 stays with the owner.

- **[P2] Still open:** [Reduce the new overhead for compact, kernel-native operands](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5214875584).

Validation: Reused the completed exact-head review: 18 H100 routing/serving/capture tests, 12 additional grouped/equal-head cases, all 16 fusion/grouping specializations, padded-operand checks and the deterministic two-thread/two-stream concurrency regression (old head fails, current head matches both serial references exactly). All six requested Python GPU CI lanes passed on this head in pipeline 68073611. Fresh integration against current develop fa7a433ee844 merged cleanly; H100 checks confirmed all four unsupported serving-option cases reject before launch and two ordinary automatic/pinned-engine cases produce identical finite outputs.

Limitations: The focused integration probe is not a full rerun. The previously noted text-scan identifier and API-index dependency issues remain separate CI cleanup items.

This approval clears my earlier [bot request for changes](https://github.com/NVIDIA/cudnn-frontend/pull/1061#pullrequestreview-5210171376); the approval-blocking findings are resolved.

PR #1067, opened 2026-09-15
benchmark/dsa: config-driven DeepSeek-V4 sparse attention suite

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes (black, line-length 160).
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and my changes comply (benchmark-only; no library code touched).
- [ ] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`.

## Affected area

Benchmarks or performance

## Summary

- Arrange `benchmark/dsa` like `attention_training` / `attention_inference`: `configs/` presets, `runner.py` (one subprocess per case, CSV + chart), `charts.py`, `run_all.sh`, `results/<config>/<gpu>/`.
- New `deepseek_v4` config: V4-Flash (H64, d=512, K=640 = 512 indexer-selected + 128 window, `indexer_topk=512`) and V4-Pro (H128, d=512, K=1152 = 1024 + 128, `indexer_topk=1024`); `s_q == s_kv` in {2k, 4k, 8k, 16k, 32k}; fwd + bwd; bf16; sink + `topk_length`.
- `benchmark_single_dsa.py` times `SparseAttentionForward.execute` / `SparseAttentionBackward.execute` with preallocated outputs (CUDA events, 256 MiB L2 flush, median). Kernel `check_support` rejections land as `skipped` rows. TFLOPS count gathered rows only; a dense-MMA peak (`peak_mma_tflops`, via optional pynvml) is the SOL denominator.
- B200 results committed under `results/deepseek_v4/b200/`. The existing single scripts are untouched and documented under "Single Scripts" in the README.

## Why

The DSA directory had two ad-hoc scripts with no shared config, CSV schema or chart; there was no way to hand someone a "run this on arch X" sweep comparable to the dense suites. This mirrors the structure the other attention suites already use so DSv4 sparse-attention numbers can be produced per architecture the same way.

## Related issues

None

## API and compatibility impact

None (benchmark-only). Forward requires an SM100-family GPU; backward runs on SM90/SM100 — other archs record skipped rows. The DSA sparse-attention kernels are fp16/bf16 only, so the `data_types` axis is bf16 for now.

## Testing

```
[...]

PR #1073, opened 2026-09-15
Reduce CPU launch overhead in DSA and BSA

jiayus-nvidia:
## Before submitting

- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and my changes comply, or I explain the exception below.

## Affected area

FE OSS kernels or CuTeDSL — DSA/BSA Python launch paths.

## Summary

Reduce repeated CPU preparation in warmed DSA/BSA calls:

- Reuse the current CUDA stream context and cache immutable device capabilities by device index. DSA forward uses one outer launch context for normalization, allocation, degenerate-output fills, and kernel execution, while retaining stream/device validation and allocator lifetime tracking.
- Cache BSA layout metadata without retaining tensor storage; move BSA forward kernel construction and compactify DLPack conversion into compile-cache misses.
- Validate immutable indexer backward plan layouts once. Continue checking runtime dtype, shape, and stride before any in-place score update.
- Reuse the local-to-global index kernel for compressed top-k, preserving caller-owned outputs and negative sentinels. Order compactify's noncontiguous-input copy on its explicit launch stream.
- Add a CUDA Graph replay correctness test that changes KV and poisons outputs before replay.

## Why

Repeated stream wrapping, metadata queries, and tensor conversion can dominate CPU submission time even when the compute kernel is fast. Compressed top-k also launched several small torch kernels for index conversion.

Recorded B200 measurements, B=2, Sq=2048, Skv=4096 (microseconds):

| Path | CPU submission before → after | Eager event time before → after |
|---|---:|---:|
| BSA64 backward | 230.2 → 115.0 | 397.8 → 287.4 |
| Sparse indexer score | 71.7 → 36.0 | 119.4 → 84.1 |
| Compressed indexer forward + top-k | 171.9 → 97.3 | 215.1 → 171.3 |
| Indexer backward | 97.8 → 56.4 | 323.2 → 281.8 |
| Compactify | 34.6 → 21.4 | 45.5 → 32.1 |

BSA uses H=16, D=128 and eight selected 64-token blocks. Indexer score / compressed forward / backward use H=16/32/64 respectively, D=128 and topk=512; compressed forward uses ratio=1 and global indices.

Measurements use a warmed full-wrapper harness, 40 alternating eager/Graph samples and medians, with the queue drained and backward input reset outside each timed call. CPU submission is measured separately from CUDA-event elapsed time. These measurements were collected before rebasing onto current `develop`; conflict resolution was validated separately below.

Nsight Systems confirms compressed top-k goes from **9 to 3 kernels per call**. Its Graph time drops from 139.3 to 121.0 us at Sq=2048 (13.2%); pure host changes leave Graph time approximately unchanged. The original attention-forward optimization is excluded from this table because the audit baseline already included it.

## Related issues

Related to #1067.

No public signature, default, output shape/dtype, or minimum dependency-version changes. Preallocated outputs, explicit streams, and negative index sentinels retain their behavior. `max_seqlen_q/k` remain optional: THD calls that omit them still infer exact maxima from current device data, including the existing synchronization. Existing wrapper allocation and normalization contracts are preserved.

## Testing

NVIDIA B200 (148 SM), PyTorch `2.12.0a0+5aff3928d8.nv26.05`, CUDA 13.2, CuTeDSL 4.6.2, cuDNN backend 92200. Imports verified against this checkout.

From `test/python`, with `PYTHONPATH` pointing at this checkout's `python` directory:

```bash
pytest -q -m 'L0 or L1 or L2' fe_api/dsa/test_DSA_sparse_attention_forward.py
# 53 passed, including the retained Graph replay test.

pytest -q fe_api/dsa/test_DSA_sparse_attention_backward.py
# 118 passed, 16 skipped after integrating the upstream DSA backward changes.

pytest -q fe_api/bsa/test_BSA_attention_forward.py fe_api/bsa/test_BSA_attention_backward.py
# 22 passed, 23 skipped after integrating the upstream BSA SM120 changes.
```

- Initial DSA/BSA family sweep: 160 passed, 23 skipped; included indexer forward/compressed/top-k/backward, dense score/backward, sparse score/attention backward, and BSA forward/backward.
- Local performance harness: 18 paths × two query lengths, all **36 configurations** passed eager/Graph output checks.
- Explicit-stream probes verified that forward normalization `full`/`cat` and degenerate `zero_`/`fill_` run on the selected launch stream, including explicit default and nondefault handles.
- `pre-commit run --files <changed files>` and `git diff --check` passed.
- SM90/SM103/SM120 and multi-GPU execution were not tested on hardware.

## Summary by CodeRabbit

* **Performance**
  * Improved compilation and metadata caching to reduce repeated processing and accelerate repeated attention operations.
  * Optimized CUDA device and kernel capability lookups.

* **Reliability**
  * Improved CUDA stream handling across sparse attention, indexing, and backward operations.
  * Added stronger validation for device, stream, layout, and empty-input scenarios.
  * Improved support for CUDA graph replay.

* **Behavior**
  * Added support for preserving negative index sentinels during local-to-global index conversion.
  * Improved handling of variable-length inputs and padded top-k indices.

* **Tests**
  * Added coverage for sparse attention execution during CUDA graph capture and replay.


jiayus-nvidia:
@cudnn-ci-bot run oss

YangXu1990uiuc:

The previous OSS CI request failed before a pipeline started. I am retrying it for the current head (`453bd555db04`) with the next standalone command.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc (approved):

Reviewed `453bd555db04` against base `0ed2710bc6db`.

The overall approach makes sense: cache immutable device/layout facts, keep stream selection per call, and move construction/DLPack preparation off cache hits. Independent same-GPU measurements show substantial CPU savings without a material GPU regression in the sampled workloads. I found two localized P2 follow-ups, including a public-API cache-order regression; no P0 or high-risk P1.

**[P2] Include zero-stride specialization in the conversion cache contract** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/453bd555db0426ffa3d871b75bd1624c27652998/python/cudnn/deepseek_sparse_attention/indexer_top_k/local_to_global_dsl.py#L194))

The newly shared conversion cache keys only dtype, varlen and preserve_negative, but mark_layout_dynamic retains literal zero strides. A legal caller output shaped (1, 1, 64) with stride (0, 0, 1) can therefore initialize the global-conversion specialization with both outer strides fixed at zero. A later ordinary batched compressed-top-k call hits that entry and raises `Mismatched mLocal.strides[0] ... expected to be 0`. Please canonicalize unobservable singleton strides using metadata-only views, or include the relevant zero-stride masks in the cache key, and cover both call orders. The failure is a visible exception with a normal-stride workaround, so this is nonblocking.

Evidence: Reproduced through DSA.indexer_forward_top_k_wrapper on SM100: warm the unchanged stage-2 path with global=False; execute B=Sq=1, Sk=128, H=32, D=128, top_k=64, global=True into the singleton output above; then call B=2, Sq=128, Sk=256 with ordinary outputs and global=True. Base 0ed2710bc passes, including exact per-row index-set comparison; head 453bd555d raises in _launch_local_to_global. A direct conversion probe also passes regular->padded->singleton order but fails singleton->regular on head. Attribution: Newly exposed on the compressed-top-k API by replacing its Torch in-place conversion with the shared DSL cache. The standalone conversion cache was already narrow before this PR.

**[P2] Run the new stream-replay guard in the default test selection** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/453bd555db0426ffa3d871b75bd1624c27652998/test/python/fe_api/dsa/test_DSA_sparse_attention_forward.py#L886))

This small replay test is marked L1, while pytest.ini defaults to -m L0 and the regular OSS runner does not override that marker. Consequently the new guard is excluded from the usual CI run. Please promote it to L0, or explicitly include it in the regular CI selection, so the stream optimization stays protected after this PR.

Evidence: On this exact head, collection with -k graph_replay_uses_capture_stream and default options reports no tests collected (53 deselected); adding -m L1 collects the test. The test itself passed in the independently run L0/L1/L2 forward suite. Attribution: New test introduced by this PR; the existing default marker is unchanged.

Validation: Reviewed all 14 changed files, affected stream/cache/layout callers, and the applicable repository rules. The PR is based directly on current develop 0ed2710bc; git diff --check passed. Independent SM100 (148 SMs), PyTorch 2.13.0+cu130, cuDNN 9.25.1, CuTe DSL 4.8: DSA forward 53 passed; compressed top-k 19 passed; BSA forward/backward 22 passed and 23 skipped; indexer runtime-contract checks 6 passed. L0/L1/L2 were selected explicitly and imports resolved to the reviewed checkout. Additional probes: 96 layout-metadata permutations/singleton/empty-shape comparisons passed; padded-output conversion preserved -1/-7 sentinels; THD conversion with an empty batch and two row strides passed, including changed-prefix/poisoned-output graph replay; explicit noncurrent-stream compactify with noncontiguous input passed. The separate public cache-order probe reproduces the first finding against a passing base. Base/head ABBA performance on the same SM100, after correctness and warm-up, with CPU submission separated from GPU time and 20 calls per captured graph: eight configurations at Sq=128/1024, Sk=2*Sq. BSA blk128 forward and blk64 backward used B=1,H=4,D=128; compressed top-k used B=2,H=32,D=128,K=64, deterministic global indices; compactify used (2*Sq,64). CPU submission fell 28-60%. Compressed GPU time fell 17.02->10.36 us and 49.01->42.21 us; other sampled GPU times stayed within 1%. Every arm passed changed-input/poisoned-output replay checks. Retried the author's failed-before-launch OSS CI request and verified the full mirrored head SHA. The pipeline started; Rubin failed before checkout/tests because the runner ran out of disk space.

Limitations: Independent GPU evidence is limited to the local SM100 and the stated configurations/dependencies; it does not establish SM90/SM103/SM120, multi-GPU, or the complete OSS CI matrix. Remaining CI jobs are pending evidence.

YangXu1990uiuc:

CI follow-up for head `453bd555db04`: OSS passed on SM80, SM90 and both SM100 shards. The SM100 shards together reported 3,261 passed, 1,963 skipped and 2 xfailed. The Rubin runner ran out of disk space before checkout/tests, so there is no Rubin test result; the other failure is a text-scan false positive on an unchanged `Conv->GEMM` convolution comment.

The default OSS selection still excludes the new L1 stream-replay test, as noted in the review. That test passed in the separate expanded SM100 run, but its regular CI coverage and the singleton zero-stride cache-order case remain follow-ups for the owner.

PR #1074, opened 2026-09-15
fix-output-alignment-check

yanqinz2:
## Before submitting

## Summary

Fix output alignment check and some test brokage

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graph validation so detailed node-level errors remain available when higher-level validation fails.
  * Relaxed alignment requirements for byte- and halfword-sized output data.
  * Improved memory budgeting and descriptor alignment for block-scaled MoE operations.

* **New Features**
  * Added support for M-major MoE output layouts and correct descriptor resizing.
  * Improved segmented row-scale capacity and alignment handling across supported data types.

* **Tests**
  * Expanded coverage for validation, output layouts, alignment, scaling, caching, and runtime correctness.


yanqinz2:
@cudnn-ci-bot run frost

YangXu1990uiuc (approved):

Reviewed `9449acabbeac` against base `0ed2710bc6db`.

The natural byte/halfword store fix and aligned-ring SMEM accounting make sense. The new M-major MoE TMA selection has a supported-DSL compilation regression and a separate endpoint-eligibility gap; both have a working STG alternative. No P0 or high-risk P1 found. Please address the findings before relying on the new TMA path.

**[P1] Keep a working path when the DSL rejects ordinal-zero replacement** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/9449acabbeacc121df5a2c53df8819a91254f8f1/python/cudnn/gemm/frost/sm100/kernel_templates/_tile_helpers.py#L78))

The new helper cannot compile with CuTe DSL 4.7.1 or the tested 4.8 build: their tensormap_replace wrapper/verifier admits ordinals [1,5] and raises on ord=0. Consequently an alignment-annotated M-major MoE graph that previously ran through STG now fails during JIT; plain and block-scaled MoE are affected. Dimension zero is the correct dimension to update, so changing this to ord=1 would be wrong. Use a compatible dimension-zero implementation or retain STG until the installed DSL supports it. This is a bounded, visible compilation failure with the default alignment hint/STG as a workaround, rather than an approval-blocking correctness hazard.

Evidence: SM100: 13 plain/mixed-output tests and all 12 block-scale M-major tests fail with 'tensormap.replace: ord must be in [1, 5], got 0'. One focused case reproduces on DSL 4.7.1. Current-head GEMM CI reports the same 25 failures (5,771 passed, 3,075 skipped). A public-API control with S=512, N=256, K=128 and alignment_value=8 passes exactly on base 0ed2710bc. Attribution: Introduced by enabling the new M-major TMA route against a dependency whose ordinal-zero support is not yet available.

**[P2] Include the implicit last-group endpoint in TMA eligibility** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/9449acabbeacc121df5a2c53df8819a91254f8f1/python/cudnn/gemm/frost/sm100/compiler.py#L3568))

alignment_value promises the values stored in first_token_offset; it does not promise alignment of S, the implicit last endpoint. For BF16 S=510, offsets [0,104,304], alignment_value=8, and an M-major output with padded column stride 512, every explicit offset satisfies the promise. Base executes this layout correctly through STG. This branch selects TMA, although its runtime alignment gate then rejects the 4B-aligned contiguous extent against the 16B requirement. Even after fixing ordinal-zero compilation, this accepted layout would regress. Include the declared endpoint/output extent in the store-mode gate and retain STG when it cannot meet TMA's contract.

Evidence: Base public API passes exact integer-input numerics and padding guards. Head selects ('tma',); the head CompiledMoeGemm runtime validation, isolated from JIT with a never-called dummy launch target, rejects 'output[0]: alignment 4B < required 16B'. Removing the alignment hint selects STG and the unmodified head passes. Attribution: Newly introduced by relaxing MoE M-major eligibility based solely on the explicit offset-value promise.

**[P2] Include the validation regressions in the default test selection** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/9449acabbeacc121df5a2c53df8819a91254f8f1/test/python/gemm/frost/test_gemm_python_validate.py#L107))

This file has no level marker, so the default -m L0 selection excludes the new error-preservation tests along with the existing controls. The regular FROST runner does not override that marker. Add a module-level L0 marker (these checks are small) and verify default collection so the shared pygraph.validate change stays protected.

Evidence: Exact-head default collection: no tests collected, 13 deselected, exit 5. With the marker filter cleared, all 13 pass. Attribution: The file was already unmarked; this PR adds and changes regressions that inherit the coverage gap.

Validation: Independent SM100, 148 SMs, Torch 2.13.0+cu130, cuDNN 9.25.1, DSL 4.8.0a0: selected suites reported 46 passed and 25 failed; all failures are the ordinal-zero issue above. The passing subset includes 13 graph-validation checks, 21 SMEM-budget/numeric checks and the two-process compiled-cache regression. Newly accepted BF16 and FP8 outputs with N=255 and a one-element pointer offset pass exact numerical and leading/trailing guard checks on both DSL 4.7.1 and 4.8. Base rejects them at the old 4-byte floor. Reviewed all 15 changed files and affected store, descriptor, runtime-alignment and validation consumers. git diff --check passed. Reused the author's exact-head FROST CI; its GEMM lane confirms the 25 failures above.

Limitations: No independent Rubin or SM120 execution. The SM100 numeric checks do not establish Rubin's larger-SMEM boundary behavior. No before/after performance claim: the new TMA route fails correctness setup, and the newly accepted narrow stores were rejected by base. Timing these as comparable successful paths would be misleading. Other requested CI lanes are still running.

yanqinz2:
@cudnn-ci-bot run frost

YangXu1990uiuc (commented):

Reviewed `fe8eb1c310b7` against base `94939f3af7f0`.

Verified the fixes for all three findings in my [previous review](https://github.com/NVIDIA/cudnn-frontend/pull/1074#pullrequestreview-5213210265). The ordinal-zero replacement now compiles and runs on both tested DSL versions, endpoint eligibility includes S for output stores and global descriptors, and the validation tests enter the default L0 selection. No new actionable findings; the existing approval is preserved.

Validation: Exact fe8eb1c3, SM100 (68 SMs), cuDNN 9.25.1: 82 selected tests passed. DSL 4.8: 13 validation, 40 plain/mixed-output MoE and 24 block-scale MoE tests; DSL 4.7.1: five IR/TMA regressions. These include S=510 STG fallback, S=504 TMA with per-group descriptor updates, empty groups, both CTA-group sizes, mixed FP32/BF16/FP8 outputs, and padding guards. Reviewed the complete six-file delta. The inline PTX retains side effects and a memory clobber before the existing warp synchronization, descriptor copy and release fence. The SM120 global-descriptor gate uses the same explicit-offset plus implicit-endpoint contract. Performance spot check against original base 0ed2710b on the same SM100: three BF16 MoE shapes, baseline/head/head/baseline order, seven timing replicates per arm, ten calls per captured graph. Actual compiled STG/TMA modes, exact output checks, poisoned-output replay, changed-input replay and padding guards passed. GPU median changes were within 0.1%: S/N/K=512/256/128 and 8192/512/512 switch from STG to TMA; 510/256/128 retains STG. Between-process CPU samples varied. A same-process interleaved control for S/N/K=512/256/128 measured 44.85 us for STG and 45.09 us for TMA, with no material route-overhead regression in that case. The author already requested FROST CI for fe8eb1c3; I verified the full mirrored GitHub SHA and reused that run. It remains pending.

Limitations: No independent Rubin or SM120 execution. Local results cover the specified SM100 configurations, not the full GEMM suite or model performance. The preceding head's Rubin CI timed out while waiting for resources before pytest; that supplies no target-GPU correctness evidence.

YangXu1990uiuc:

CI follow-up for `fe8eb1c310b7`: **SM100 GEMM and linear attention passed**, as did the SM80/SM100/SM103/SM120 SDPA lanes. In particular, the current-head GEMM lane is green after the ordinal-zero TMA descriptor fix.

The two remaining failures are outside this PR's code change: Rubin timed out waiting for a GPU allocation before pytest started, and the text scan flags the unchanged `Conv->GEMM` comment from develop (subsequently fixed by #1076). This adds CI evidence to [the exact-head local review](https://github.com/NVIDIA/cudnn-frontend/pull/1074#pullrequestreview-5214546689); it does not establish Rubin validation.

PR #796, opened 2026-08-28
grouped gemm: accept canonical (natural row-major) layouts and flat SF buffers

hwanseoc:
## Description

Accept canonical (natural row-major) layouts in the contiguous grouped SwiGLU/dSwiGLU APIs while preserving every existing pre-permuted form. The intended consumer is the contiguous MXFP8 TE/JAX fusion: this PR makes its buffers representable as ordinary JAX arrays; the stacked JAX bridge #1079 supplies the custom calls. It does not establish completed TE integration or full-model training validation.

Current TE/PyTorch uses the unified GLU/dGLU wrappers, which are outside this PR. Canonical support for those APIs is a separate follow-up. Dense and discrete-weight JAX APIs already exist and do not depend on this change.

| Operand | Existing form | Added form |
|---|---|---|
| A | `(m,k,1)` | row-major `(m,k)` |
| B | strided `(n,k,experts)` | C-contiguous `(experts,n,k)` |
| backward C | `(m,2n,1)` | row-major `(m,2n)` |
| SFA/SFB | MMA-tiled 6-D view | dense physical 6-D or flat buffer, with the same packed bytes and element count |
| prob | `(m,1,1)` fp32 | `(m,)` fp32 or bf16 |
| alpha | explicit per-expert tensor | unchanged: required |

**Flat SF means already packed MMA-tiled bytes in physical storage order. It does not mean ordinary row-major logical scale values.**

Canonical A selects natural 2-D C/D outputs, 1-D backward dprob, and physical C-contiguous SFD `(1, ceil(mn/128), rest, 32,4,4)`. A supplied backward dprob buffer retains its declared rank.

Canonical operands compile at their own rank. Samples normalize once for support checks; trace-time helpers rebuild the kernel-facing views. Scale factors use their base pointers because the kernels rebuild their layouts. `execute()` adds no host views or copies. The only device-body change is converting loaded bf16 prob to fp32. Compile keys distinguish operand forms, prob dtype/rank, SF dtype/contiguity/rank, and effective dprob output rank.

## Correctness

Alpha remains explicit, as on develop. The proposed optional-alpha global device cache has been removed, eliminating the capture-created constant lifetime/initialization issue. None is rejected before cache dispatch, including during graph capture.

[...]

PR #1079, opened 2026-09-15
grouped gemm: dual Torch/JAX SwiGLU wrappers and namespace APIs

hwanseoc:
## Description

The existing contiguous grouped SwiGLU/dSwiGLU wrappers now accept both Torch tensors and canonical MXFP8 JAX arrays/tracers. JAX calls use the same implementation as the explicit namespace APIs, eagerly or under `jax.jit`:

| Operation | JAX | Torch alias |
|---|---|---|
| Forward | `cudnn.jax.grouped_gemm_swiglu` | `cudnn.torch.grouped_gemm_swiglu` |
| Explicit backward | `cudnn.jax.grouped_gemm_dswiglu` | `cudnn.torch.grouped_gemm_dswiglu` |

Existing `cudnn.grouped_gemm_{swiglu,dswiglu}_wrapper_sm100` names, signatures, defaults, and Torch execution remain supported. Each `cudnn.torch` alias is the same function object as its existing wrapper and inherits dispatch. JAX inputs dispatch lazily through the `cudnn.jax` API and `cudnn.jax.call`, without importing Torch. XLA owns buffers and stream ordering. No device kernels or Torch execution paths changed.

Rebased onto `develop` at `adb963c01`, including merged canonical-layout PR #796 and shared-bridge errata #1081. This PR contains the JAX bridge, wrapper dispatch, and namespace additions. Kernel bodies are unchanged.

```python
import jax
from cudnn.jax import grouped_gemm_swiglu

result = jax.jit(grouped_gemm_swiglu)(**jax_inputs)
```

```python
import torch
from cudnn.torch import grouped_gemm_swiglu

result = grouped_gemm_swiglu(**torch_inputs, d_dtype=torch.float8_e4m3fn, sf_vec_size=32)
```

The shared wrapper can also be jitted:

```python
from functools import partial
import cudnn
import ml_dtypes

result = jax.jit(partial(
    cudnn.grouped_gemm_swiglu_wrapper_sm100,
    sf_vec_size=32,
    d_dtype=ml_dtypes.float8_e4m3fn,
))(**jax_inputs)
```

Both return `TupleDict` with the existing field order, integer/key access, and unpacking behavior. JAX preserves it as a pytree across JIT. API classes accept metadata-only descriptors for shared support validation.

## Contract and limits

- Canonical (natural row-major) A `(m,k)`, B `(experts,n,k)`, backward saved C `(m,2n)`, and prob `(m,)` fp32/bf16. SF inputs contain already-packed E8M0 bytes in physical or flat form; uint8 bit patterns are accepted. Ordinary logical scales are not packed by this API.
- JAX requires explicit alpha/prob/norm_const arrays and backward beta. Missing operands and mixed-framework inputs are rejected. Outputs have natural shapes and C-contiguous physical SF storage.
- Direct `cudnn.jax` entry points fix scale-vector size to 32 and default D to FP8 e4m3. Shared wrappers preserve their existing defaults, requiring explicit `sf_vec_size=32` and FP8 `d_dtype` for JAX inputs. FP8 e4m3/e5m2 inputs are supported; forward D supports both, backward D supports e4m3 only because the existing packed quantizer cannot emit e5m2.
- XLA owns streams and output buffers. Direct JAX APIs omit Torch-specific parameters. Shared wrappers reject unsupported JAX options with `ValueError`: caller streams/output buffers, non-FP32 accumulation, non-`n` layout, vectorized-f32 mode, non-default alignment, discrete-column SF, and non-identity backward epilogues. Existing Torch defaults are unchanged.
- `m` is a positive multiple of 256. Device offsets must be nondecreasing, 256-aligned, and within `[0,m]`; validation does not read their values to the host.
- Output initialization defines untouched padding; SF initialization writes raw zero bytes. Backward dprob is freshly zero-initialized for atomic accumulation.
- Cached callables retain static metadata, never runtime arrays or pointers. This avoids repeated eager tracing while retaining fresh initialization on every call. A warm-call regression failed before caching and passes with `jax.no_tracing(True)` afterward.
- CuTeDSL >=4.7 is gated before importing the JAX implementation. SM100 tested. Packed FP4, BF16 JAX outputs, bias, unified GLU/dGLU, automatic `jax.grad`, TE integration, and full-model training remain outside this PR.

## Validation

SM100, Python 3.14, Torch 2.13+cu130, JAX 0.11.1, CuTeDSL 4.7.1, cuDNN 9.20:

**320 passed, 200 skipped** in 358.63 seconds across canonical Torch/JAX tests, both legacy SwiGLU/dSwiGLU suites, JAX rejection/default-contract tests, and the shared-call regression suite.

- Both the shared wrapper and explicit JAX namespace run eager/JIT byte parity with Torch (dprob rtol/atol 1e-4), physical/flat SF, bf16 prob, one/four experts, changed alpha/runtime offsets, empty experts, padding, and forward-to-backward command-buffer replay.
- Torch namespace aliases execute the existing kernels and are identical to the original functions. Fresh-process tests verify both JAX entry paths with Torch blocked, lazy JAX exports, and Torch alias imports with JAX blocked. The shared-wrapper parity regression failed on the previous head before dispatch was added.
- All sixteen namespace/wrapper parity configurations verify warm eager calls with tracing disabled, including freshly initialized backward accumulators.
- No exports changed in this dispatch update. Previous namespace validation: API index includes all four new exports. Eight scanner fixture tests passed. Full runtime index comparison has 15 missing native exports and one extra SDPA member in this local installation; a control using current develop sources and the same native extension reproduces the identical differences. All four new exports resolve. Full wheel-based API-index qualification remains for CI.
- Changed-file pre-commit and `git diff --check` passed. No benchmark scripts are included. No new performance measurements or current-head CI claims.

@coderabbitai ignore


Anerudhan (approved):

YangXu1990uiuc:
I am requesting test CI for `9cd71adb1d42` with the following standalone command: `@cudnn-ci-bot run oss`. For future updates, request the appropriate test target before review; a maintainer may need to trigger it.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc (commented):

Reviewed `9cd71adb1d42` against base `55fdf97ca929`.

Reviewed 9cd71adb1d42; no actionable correctness or approval-blocking findings. The shared metadata validation, explicit JAX output initialization, and Torch namespace aliases are consistent with the documented contracts. Leaving a comment because an effective approval already exists.

Validation: Independent SM100 run against this head and a freshly built matching FE extension: all 37 canonical grouped-GEMM JAX tests passed, with zero skips (JAX 0.11.0, CuTe DSL 4.8). This covers eager/JIT/warm execution, Torch parity, runtime offsets with empty experts and padding, forward/backward chaining, and the JAX-only import path. Additional forward and backward probes with M=768, N=512, K=128 and three experts matched Torch in eager, JIT, and warm no-tracing execution; forward also exercised float32 C output. Started OSS CI pipeline 68381776 and verified its mirror commit identifies this exact GitHub head. The pipeline is still running.

Limitations: Independent GPU coverage is SM100 only; full CI is pending. No CPU/GPU timing comparison was made for the new JAX entry points, which have no pre-change JAX baseline; existing Torch aliases retain the same implementation and kernel bodies are unchanged.

YangXu1990uiuc:
Starting OSS CI for the wrapper-dispatch update at `e44690f9e79b`; pipeline 68381776 covers the previous `9cd71adb1d42` revision. The existing review evidence still applies to the unchanged bridge and kernels; the added dispatch is being checked separately.

YangXu1990uiuc:
@cudnn-ci-bot run oss

YangXu1990uiuc (commented):

Reviewed `e44690f9e79b` against base `55fdf97ca929`.

Re-reviewed the wrapper-dispatch update from 9cd71adb1d42. The new JAX branch forwards all supported parameters and explicitly rejects the unsupported Torch options, while retaining the existing wrapper signatures/defaults and Torch execution path. No new actionable finding.

Validation: Exact-head SM100 run: 71 related tests passed, zero skipped, covering shared-wrapper/namespace eager and JIT parity, warmed no-tracing calls, explicit option/default rejection, legacy-layout rejection, and fresh-process execution with Torch unavailable. Used the matching native extension from the previous head after verifying native-source identity. Measured direct JAX API versus shared wrapper on forward/backward with 1/4 experts (M=256/1024, N=K=256), using nine interleaved batches per route. Warm eager host-call medians increased by 9-15 us, approximately 0.6-0.9% over the 1.63-1.65 ms direct-call medians. Under JIT both routes were about 60-63 us, with median differences between -0.4 and +1.1 us; no material dispatch regression was observed in these cases. All four pairs produced identical StableHLO after normalizing the module name. git diff --check and current-head Style/Docs checks passed. Started OSS CI pipeline 68394900 and verified its mirror commit carries the full reviewed GitHub SHA; it remains pending evidence.

Limitations: The performance spot check measures host-call cost, not standalone GPU execution time. The device kernels and lowered JIT programs are unchanged, but GPU timing was not newly measured. Full CI is still running; independent coverage is SM100 only.

PR #1082, opened 2026-09-16
Fix SDPA Stats log-base compatibility and cover ragged decode LSE

YangXu1990uiuc:
`graph.sdpa_fp8(stats_use_log2=True)` raised `TypeError` because the dedicated FP8 wrapper omitted the argument. Append the default-false parameter and forward it to `sdpa_internal`. Default/false preserve natural-log Stats; true converts the entire LSE to base 2. Also keep SM120 FP16/FP8 split-KV partial LSE in natural-log units at the kernel entry point; the final combine converts once.

The CI dev headers also lack the pending native Stats log-base attribute even though their version is newer than the original guard. Detect the actual enum member at compile time and return not-supported when it is absent; retain the runtime version check and avoid hard-coding an unreleased enum value. Remove a CI guardword from the Stats documentation.

Add a deterministic `test_mhas_v2.py::test_sdpa_ragged_decode_stats` regression. The random single-token inference sweep did not request Stats, leaving the native ragged GQA decode store untested even when O was correct. The new L0 test initializes Stats to NaN, pins a backend plan, checks every head against an fp64 reference, and checks O independently. It covers FP16/BF16, int32 token offsets with multipliers / int64 element offsets, unequal KV lengths, and padded-Stats, MHA, and prefill controls. Record the detector and split-LSE invariant in the relevant agent guides.

Validation:

- New decode regression on unfixed cuDNN 9.26: **4 failed, 12 passed** with `--runxfail`; affected cases leave **18/24 Stats entries NaN**, while O passes. With the local backend fix: **16 passed** on SM100.
- Known affected backends below 9.28 have a strict xfail registered only after graph execution and all O checks pass. A backport produces XPASS; 9.28+ must pass normally. This is a regression gate, not a claim that the fix has shipped.
- After rebuilding the extension, public FP16/BF16 and FP8 LSE-base tests plus the new regression, cuDNN 9.25.1 with FROST enabled: **26 passed, 4 xfailed**. The decode test still explicitly selects a backend plan.
- FP8 graph regression was red before the binding fix (2 failures), then passed after rebuilding. Prior focused SM100 suite: **22 passed, 6 SM120 skips**; checks default/false/true, masked/unmasked Stats, and unchanged O.
- Pinned pre-commit hooks passed. SM120 runtime remains for target CI.

The previous CI run also exposed two failures in the existing FP8 intermediate-reconstruction self-test. Both reproduced locally with the CI TF32 override (max error 0.0071678). Compute that independent reconstruction in fp64, keeping the tolerance unchanged: **44 self-tests passed** with TF32 enabled.

Header validation: the old source fails on the missing enum with the matching dev headers; the corrected umbrella header compiles against release 9.25, dev 9.28 without the enum, and companion headers with it, using GCC/Clang and the cuDNN exception macro on/off. The exact detection helper also passes strict C++17 compilation under both compilers for all three header sets. Local Clang 14 umbrella checks suppress two pre-existing missing-override warnings; all other warnings remain errors.

## Summary by CodeRabbit

* **New Features**
  * Added the optional `stats_use_log2` setting to FP8 scaled dot-product attention, returning statistics in base-2 format while preserving attention outputs.
  * The setting defaults to natural-log statistics for compatibility.

* **Bug Fixes**
  * Corrected split-KV attention statistics so partial results remain compatible during combination.
  * Improved handling when base-2 statistics are unavailable by reporting a not-supported status.

* **Documentation**
  * Clarified the format and availability requirements for base-2 attention statistics.

* **Tests**
  * Expanded coverage for FP8 statistics, ragged decode scenarios, split-KV execution, and multiple attention layouts.


YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc:

YangXu1990uiuc:
@cudnn-ci-bot run frost

vedaanta (approved):

YangXu1990uiuc (commented):

Reviewed `5cbaad94385e` against base `fa7a433ee844`.

The fixes make sense: append/default/forward the FP8 Python flag, detect the actual native enum rather than infer its existence from a development version, and keep split partials in the combiner's natural-log units. I found no new P0 or high-risk P1 in those changes. The poisoned, backend-pinned ragged Stats test is a useful detector and correctly keeps O checks outside the expected-failure region. Current CI is not yet green: besides the two cleanup items below, its cuDNN9.28 SM80/SM120 runs still leave 18/24 ragged-decode Stats entries NaN after O passes. The native fix therefore has not reached those runners. Rubin also reports native plan-finalization/NVRTC failures in several new int64 decode/control cases; that is a separate backend follow-up, not a demonstrated regression in the FE flag forwarding. Please coordinate those backend fixes/results before describing the integration as validated. One useful coverage follow-up is a direct SM120 template test with stats_log2=True and split_kv>1: the existing adapter test passes False to the partial kernel, so it cannot detect regression of these two new kernel-entry guards.

**[P2] Update the three contract checks left stale by the log2 addition** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/5cbaad94385e4e8d6f6614be6daab608ad47e84c/test/python/sdpa/frost/test_split_kv_heuristic.py#L473))

The arity detector counts stats_log2 in the Python _host signature as a runtime argument, but it is cutlass.Constexpr[bool] and is consumed by compilation. The eight-argument compiled calls are correct; changing production calls to nine arguments would be the wrong fix. Check the compiled runtime contract or exclude compile-time parameters in this detector. The two constructor-prefix tests likewise still demand that the last two arguments be sample_amax_o,pv_bf16, although stats_log2 was safely appended; preserve the legacy-prefix assertion while allowing the new trailing parameter.

Evidence: Exact-head CI pipeline68115480 reports these same three failures in SM80 and SM120 FROST jobs: test_new_constructor_parameters_follow_the_legacy_positional_prefix, test_thd_stats_padded_is_appended_to_the_public_signature, and test_every_combine_call_site_matches_the_compiled_arity. Independently exercised the actual compiled combiner for four split/log-base combinations on SM100; all four passed with the existing runtime call ABI. Attribution: Inherited from merged#931, not introduced by this fix; an actionable leftover in the same log2 integration and its CI cleanup.

**[P3] Use a neutral phrase in the new expected-failure comment** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/5cbaad94385e4e8d6f6614be6daab608ad47e84c/test/python/test_mhas_v2.py#L378))

The current text-scan job flags the new phrase 'Known broken on older backends'. Changing this to 'Known issue on older backends' preserves the explanation and avoids replacing the removed documentation violation with another one.

Evidence: Exact-head text-scan job441883382 reports one violation at test_mhas_v2.py:378, rule COMMENT:broken; the earlier graph_properties.h violation is gone. Attribution: Introduced by the new regression-test comment in this PR.

Validation: Built the exact-head Python extension independently. On SM100/cuDNN9.25.1/DSL4.8, the new FP8 API plus ragged Stats selection gave14passed4xfailed. The two FP8 tests fail with TypeError against the prior#931 extension and pass with this extension. A separate A100 run of the sixteen ragged cases gave12passed4xfailed; all O/control checks pass, and only the known four decode-GQA ragged-Stats cases xfail. With the CI TF32 override enabled, all44 FP8 reconstruction/flip-budget selftests and four actual split-combine LSE tests passed (48 total). No tolerance was relaxed. Real release9.25, development9.28 without the enum, and companion headers with the enum: all12 GCC/Clang C++20 umbrella-header checks passed, including exception-macro on/off; all6 strict C++17 checks of the exact new detection helper passed. Full strict C++17 umbrella checks encounter the unchanged empty-variadic-macro warning in cudnn_frontend_shim.h, so those are not claimed as passes. Existing Clang missing-override warnings were suppressed in the umbrella checks, as in the reported author validation. Verified pipeline68115480's mirror commit carries the exact full GitHub head. Both development Linux builds, release builds and Clang analysis now pass. SM120 FROST completed3230passed/3398skipped with the seven failures described above; SM80 completed2691passed/3937skipped with the same seven. The user already requested python_tests and frost on this head; no duplicate CI trigger was sent.

Limitations: Independent GPU runs here are SM100 and A100. SM120/Rubin observations above are current-head CI evidence. The direct SM120 template-entry guard is source-reviewed, not independently executed; existing split tests use the adapter. Some other pipeline lanes are still running. No new hot-path performance claim: enum detection and Python argument forwarding affect graph construction, while the split-store condition is compile-time and existing adapter specializations already disable partial-log2 conversion. No full-model benchmark was warranted for this delta.

YangXu1990uiuc:
CI follow-up for `5cbaad94385e`: pipeline **68115480** has finished.

The header compatibility fix is confirmed by both development builds, release builds, and Clang checks. All six regular Python GPU lanes and all three C++ sanitizer lanes passed; FROST GEMM and linear also passed (linear: 1,626 passed, 169 skipped, 8 xfailed).

The remaining SDPA failures are consistent across SM80/SM100/SM103/SM120: the three stale contract assertions described in my review, plus four new native ragged GQA decode Stats cases. SM100 finished with **4,810 passed / 7 failed**. Those native Stats failures still occur with the CI cuDNN reporting `92800`; output `O` passes, but Stats retains NaNs. Rubin additionally encounters native plan/NVRTC compilation failures in the new int64-offset cases (its lane reports 12 failed / 3,118 passed).

Please align the backend rollout and test expectations with the actual fixed backend build: `>= 92800` alone does not establish that the ragged Stats fix is present. Keep the `O` checks and the failing Stats coverage. The comment guardword and the unrelated existing API-index dependency failure remain as noted in the review. I found no additional FE runtime regression in the final lanes.

YangXu1990uiuc:

YangXu1990uiuc:
@cudnn-ci-bot run frost

YangXu1990uiuc (commented):

Reviewed `c0108c35f1b4` against base `fa7a433ee844`.

Follow-up: both findings from my previous review are addressed. The constructor checks retain the legacy positional prefix and verify the appended default-false option; the arity detector correctly excludes Constexpr parameters without changing the compiled call sites; the guardword is gone. The new direct SM120 template tests also cover the adapter bypass discussed previously, checking partial LSE units and final combination for FP16/FP8, split/unsplit, and ln/log2. The C++ declaration now supplies the default requested by CodeRabbit. No new approval-blocking issue found in this delta. The native backend failures from the previous CI run still require the separate backend rollout/result follow-up already documented; this revision does not change that backend behavior.

Validation: Independently ran the three formerly failing contract tests from this exact source tree: 3 passed. All eight new SM120 cases were collected and correctly skipped on the local SM100 device; this is not SM120 runtime validation. Reviewed all changes since 5cbaad94385e. The new direct-template launch arguments match the FP16/FP8 compiled entry points; nonzero per-chunk logits make partial log-base errors observable before combination. Verified that owner-started pipeline68127598 mirrors the full current GitHub head. Current development/release builds, Clang checks, sanitizer build, and guardword scan pass. GPU lanes are still running; no duplicate CI was requested.

Limitations: The eight new direct-template cases require SM120; their target CI execution and seen-RED check remain unverified by me. Existing current-head SM120 CI is running. Reused the independently built previous-head extension for the Python contract tests: the only native delta is a declaration default, with no runtime ABI/implementation change. Current-head native builds are independently confirmed by CI. Previous-head native ragged Stats/NVRTC failures are not claimed fixed by this FE test cleanup. No new timing comparison was needed for this tests/comments/default-argument delta.

PR #1109, opened 2026-09-16
frost(sdpa): d256 decode tile for decode-shaped f16/bf16 graphs (swap-AB, SM100)

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes. (Formatting was applied with the pinned versions directly, `black==26.3.1 --line-length 160` on every changed `.py` file, rather than through the pre-commit hook; no C++/CUDA files are touched, so `clang-format` does not apply. `test/python/test_mhas_v2.py` is under `# fmt: off` and was hand-aligned like its neighbours.)
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches (see [root AGENTS.md § Reviewing a PR](https://github.com/NVIDIA/cudnn-frontend/blob/develop/AGENTS.md#reviewing-a-pr-human-or-agent)) and my changes comply, or I explain the exception below.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

A second CuTe-DSL template for the SM100 f16/bf16 d256 SDPA forward flavor, `python/cudnn/sdpa/fwd/kernels/sm100/decode_d256_f16.py`, selected by the existing `sdpa_fwd_prefill_sm100` FROST engine row when the graph is decode-shaped: `S_q x G <= 16` packed Q rows, with `G` the PackGQA group (16:1 at `S_q = 1`, 8:1 at `S_q <= 2`, 4:1 at `S_q <= 4`, MHA at `S_q <= 16`; the decode tile packs the WHOLE group -- 96/8 puts its 12 heads in the 16-column tile -- it has no partial-PackGQA form, so `G` here is the full ratio, not #1104's `gcd(G, 128)`). Same graph contract and decline discipline as `prefill_d256_f16.py`: dense padded / dense unpadded / paged K/V caches (page 16..128, HND / NHD), padding, top-left and bottom-right causal, SWA, right band, dense padded-Q trim, sink, Stats natural or base-2, KV-split partials for the shared `split_combine_sm100`, `O := 0` / `LSE := -inf` on dead rows, nothing read on the host.

[...]

PR #1099, opened 2026-09-16
FROST SDPA fwd: explicit pointer/int host entries and plan-time launch state (execute host 149 → 51 µs, #1086)

YangXu1990uiuc:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`.

## Affected area

- FE OSS kernels or CuTeDSL
- Python API or bindings (engine contract, internal)

## Summary

FROST SDPA-forward host path (#1086), two commits on top of develop (`ae80d63a`, i.e. after #1102 / #1104 / #1110):

- **Part A — lowering layer, tensor entries unchanged:** `lower_dsl_prefill` resolves the plan's launch state once (operand ids, IR layouts, feature-operand sources incl. the epilogue gate, quantized-operand ids, a `CUstream` cache) and `_FrostSdpaFwdPlan.execute` hands its resolved `{id(ir_tensor): buffer}` map straight to `execute_resolved`. Reviewable and measurable on its own (below).
- **Part B — explicit pointer/int host entries:** the compiled entry of every SM100 / SM107 forward template takes **device pointers, runtime extents and the caller's (batch, seq, head) strides** instead of `cute.Tensor`s, through one shared prologue; the adapter builds the argument list itself. Public API unchanged; no toggles; no generated code.

[...]

Issue #1098, opened 2026-09-16
[Bug]: FROST SDPA fwd engine ignores execute-time shape overrides and rejects the graph by its declared shape


YangXu1990uiuc:
### Summary
With `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1`, a graph built with `is_override_shape_enabled=True` and executed with `override_uids/override_shapes/override_strides` (FlashInfer #5245: declared cache batch 4096, real batch 64) is served by the FROST SDPA forward engine, which validates the caller's buffers against the **declared** shape and raises:

```
ValueError: cu_seq_len_q must have B + 1 = 4097 elements (prefix sums); got 65
```

`pygraph.execute()` sets `overriding = True` and skips `_normalize`, so a python plan (`takes_variant_pack=False`, uid-dict path) never sees the overrides; `_FrostSdpaFwdPlan.execute` → `engines._execute` → `SdpaFwdDslSm100._thd_pack` then treats the declared B as the runtime B. The backend path applies the overrides correctly (same graph, FROST disabled: passes).

### Expected
Either the FROST engines consume `override_shapes` (the THD kernels already take dynamic token extents, so the override is exactly the `t_q`/`t_kv`/batch they want), or the engine declines at `check_support` when the graph is override-enabled so the heuristic falls through to a backend plan instead of failing at execute.

### Environment
cudnn-frontend develop (b2712de2d, 1.30), CuTe DSL 4.7.1, cuDNN 9.27, SM100; graph = FlashInfer ragged SDPA prefill with cu_seq_lens, ragged offsets and Stats.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
[...]

PR #1096, opened 2026-09-16
frost(sdpa): paged KV on the d192x128 f16/bf16 flavor; gate paged head dims on the selected flavor

vedaanta:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes. (Formatting was applied directly with the pinned tool versions — `black` 26.3.1 `--line-length 160` on every changed `.py` file — rather than through the pre-commit hook; `--check` reports all changed files unchanged. No C++ files are touched.)
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches (see [root AGENTS.md § Reviewing a PR](https://github.com/NVIDIA/cudnn-frontend/blob/develop/AGENTS.md#reviewing-a-pr-human-or-agent)) and my changes comply, or I explain the exception below.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

## Affected area

FE OSS kernels or CuTeDSL

## Summary

Paged KV cache on the SM100 f16/bf16 **d192×d128** FROST flavor (`sm100/prefill_d192_d128_f16.py`) and a head-dim gate for paged graphs that tests the flavor the lowering *selects* instead of the raw dims.

Based directly on `develop` (`7ac535419`, after #1094 and #1095); no other branch is carried. #1107 was closed by policy and nothing from it is in this diff. Three commits: the port + selected-flavor gate, the THD + left-window predecode fold-off pin, and `test_paged_graph_d192x128_sink` (the sink fold from #1095 and this PR's paged loader compiled together on the d192×d128 kernel for the first time; matches the fp32 reference at `s_q` 1 / 4, NHD / HND).

[...]

PR #1119, opened 2026-09-17
FROST SDPA f16/bf16 SM100/SM107: explicit pointer host entry, layout-only compile key, prepared THD launch bound from the VariantPack (#1086, #1098)

YangXu1990uiuc:
## Before submitting

## Affected area

- FE OSS kernels or CuTeDSL (the 8 f16/bf16 SM100 / SM107 prefill templates' host entry and compile key; the only kernel-body change is the four SM107 templates' dense LSE store, a slice-then-store replaced by the direct three-index store `lse_arr[b, h, row]` the SM100 templates already use)
- Python API or bindings (FROST SDPA adapter / engine, internal; `VariantPack` native metadata)

## Summary

**Scope: the f16 / bf16 forward on SM100 and SM107 (`sm100/prefill_*_f16.py`, `sm107/prefill_*_f16.py`: dense, paged, split-KV and THD), on develop.** Re-scoped from the earlier stacked version (see the conversation above): the 16 FP8 / MXFP8 templates, the SM100 d128 decode tile, the SM80 / SM120 templates, `split_combine.py` and the quantized adapter arms are byte-identical to develop — they follow in later PRs on the same pattern. What does change for everyone is the shared plumbing: the lowering-layer resolution (commit 1), `VariantPack` / `_pygraph` normalization metadata and the FROST plan class (commit 3), and the shared `_common_blackwell.py` gains two helpers (existing functions untouched); the out-of-scope gate below covers that. #1099 is left open as the wide-rollout reference and is not a prerequisite.

Four commits on develop `d34a6909c` (`b49d83e1a` → `51eb30cbe` → `ba9fc6701` → `e8a0be3c4`; the last records `SdpaFwdDslSm100.template_params` in the API index — a develop gap from #1102 that fails the pipeline's api_index job):

1. **`engines.py` / `engine.py` hot path** (`65b5cadd8`, the first commit of #1099): the plan resolves its operand roles once at lowering (`_execute.execute_resolved`), the per-execute adapter glue goes.
2. **Explicit pointer/int host entry for the 8 f16 templates + layout-only compile key + prepared THD launch.** `_host(q_ptr, k_ptr, …, problem_size, q_strides: (batch, seq, head), …, lse_kind: Constexpr, stream)`: every extent and stride is a runtime argument, so `compile(d_qk, d_v, has_lse, lse_kind, paged_hnd)` produces **one artifact per layout kind** instead of one per (b, qh, kh, sq, skv, strides…) key; strides are the caller's, never derived (S=1 BHSD storage, the #1099 review P1). The tensor building is one shared prologue in `_common_blackwell.py` (`sdpa_operand_tensors`, `sdpa_gate_tensor`). Kernel bodies are unchanged except the SM107 LSE store noted above. The adapter (`api_dsl.py`) gains an `EXPLICIT_ABI` arm next to develop's tensor arm: `_explicit_compile_kwargs`, `_execute_dense_explicit` (dense / paged / split, pointers + strides through the DSL wrapper), and the THD arm binds through **`cudnn/sdpa/fwd/prepared.py`** — `build_thd_spec` (plan-time positional argument template, capacity decls, Stats layout kind, workspace regions, build-time zeroed dummies; torch-free), `resolve_thd_geometry` + `bind_thd` (one device / dtype / alignment / extent rule per role, capacities from observed spans, all-KV-zero clamp, paged pools / tables, padded-Stats seed; every check a cold branch), calling the artifact's **positional `tvm_ffi.Function`** (`compiled_cache.positional_entry`) with integer addresses and stride tuples.
3. **The graph plan binds the same launch.** `VariantPack` records the producer's `observed_bytes` / `observed_device` at `read_operand` / `set_operand` (untouched by `describe_from` / `override_operand`) and exposes `facts(indices)`, one native crossing per execute; `_FrostSdpaFwdPlan` takes the pack (`takes_variant_pack` per instance) and calls `PreparedThdLaunch.execute`; `lower_dsl_prefill` attaches it for f16 THD plans without bias / gate / synthesized KV padding. Graph and standalone execute build byte-identical frames (tested).

Not in this PR (follow-ups, in order): dense f16 through the positional entry (today: DSL keyword wrapper + a `make_ptr` cache), the native override loop in `_normalize`, a native binder, then FP8 / MXFP8 / decode / SM120 joining the explicit entry.

## Why

Host time of the public `graph.execute()` on the FlashInfer-shaped THD graph (b=64 × q4, kv 1024, GQA 8/2, d128 bf16, Stats, ragged declaration), SM100 idle, drain then one call, median of 500 (relative numbers from a shared dev box; absolute SM100 numbers to follow from the perf lab):

| head | fixed geometry | fused-slab K/V (runtime token stride 512) | bounded batch override (32 of 64) |
|---|---|---|---|
| develop `ae80d63a` | 132–137 µs | – | – (#1098) |
| stacked #1119 `f5a0ef263` | 29.6–31.6 | 28.7–32.6 | 74–102 |
| **`6f29345a8`** (two rounds; `6bd94d711` changes no hot-path code on this graph) | **29.6 / 30.5** | 29.0 / 38.0 | 73.1 / 75.1 |

**Device time is unchanged.** Same box (SM100), develop `5fb8d3b44` vs this head, FROST kernel time per call (torch.profiler, median of 40; the THD and paged legs count setup + main kernel), O / LSE digests compared bit-for-bit; paged cases are reference-checked by the test helper on every call:

| case | develop µs | head µs | ratio | O / LSE |
|---|---|---|---|---|
| THD b64 × q4, kv 1024, GQA 8/2, d128 bf16 | 273.8 | 271.6 | 0.992 | identical |
| THD same, fp16 | 273.8 | 271.7 | 0.993 | identical |
| THD b8 × q512, kv 512, MHA 8, d256 bf16 | 40.9 | 40.4 | 0.989 | identical |
| dense b4 h16 s2048 d128 bf16, causal + Stats | 105.8 | 105.1 | 0.993 | identical |
| dense b2 h8 s1024 d128 fp16 | 22.4 | 22.2 | 0.993 | identical |
| dense b2 h8 s1024 d192×128 bf16 | 22.1 | 22.1 | 1.004 | identical |
| dense b2 h8 s1024 d256 bf16 | 20.1 | 20.1 | 1.000 | identical |
| dense b1 h8 s512 d512 bf16 | 19.0 | 19.1 | 1.002 | identical |
| paged b5 GQA 8/2 d128 p16 NHD fp16, Stats | 11.5 | 11.4 | 0.994 | ref-checked |
| paged b3 GQA 8/2 d256 p32 HND fp16, Stats | 15.5 | 15.6 | 1.007 | ref-checked |
| paged b1 32k KV d128 p16 HND fp16 (heuristic split) | 44.7 | 44.6 | 0.998 | ref-checked |

**SM107 (Rubin host w2u1g-lc-0030, DSL 4.8.0.dev0, cuDNN 9.27; no CUPTI there, so device time is CUDA-graph replay of 20 launches per graph, median of 40 replays; shared box, load 14–17; two samples):**

| case | develop µs | head µs | ratio (sample 1 / 2) | O / LSE |
|---|---|---|---|---|
| THD b64 × q4, kv 1024, GQA 8/2, d128 bf16 | 607 / 583 | 611 / 555 | 1.007 / 0.951 | identical |
| THD same, fp16 | 583 / 507 | 532 / 497 | 0.913 / 0.981 | identical |
| THD b8 × q512, kv 512, MHA 8, d256 bf16 | 43.9 / 44.0 | 42.6 / 42.7 | 0.970 / 0.971 | identical |
| dense b4 h16 s2048 d128 bf16 | 63.6 / 63.7 | 64.0 / 64.1 | 1.006 / 1.005 | identical |
| dense b2 h8 s1024 d128 fp16 | 20.5 / 20.5 | 20.7 / 20.7 | 1.007 / 1.010 | identical |
| dense b2 h8 s1024 d192×128 bf16 | 23.1 / 23.2 | 23.2 / 23.4 | 1.004 / 1.006 | identical |
| dense b2 h8 s1024 d256 bf16 | 19.5 / 19.5 | 19.5 / 19.5 | 0.999 / 0.999 | identical |
| dense b1 h8 s512 d512 bf16 | 21.1 / 21.1 | 21.2 / 21.2 | 1.002 / 1.002 | identical |

Dense is within ±1 % on both samples; the THD develop numbers themselves move 5–15 % between samples on that box, and the head is never the slower side. Paged has no FROST plan on SM107 (as on develop). The same A/B doubled as a capture test and found the stream bug below.

Attribution at the stacked head (instrumented, fixed geometry): `_uid_to_data` 1.1 · `_normalize` 4.5 · `bind_thd` 10.1 · positional call 6.6 · plan / graph glue ≈ 6. Kernel and outputs unchanged: device time identical and O / LSE bit-identical between the tensor entry and the explicit entry (kernel-level A/B in #1099).

## Related issues

#1086 (execute host overhead); #1098 (a bounded execute-time override — batch ≤ declared, TMA-expressible strides — now runs through the public API on the prepared FROST plan); #1099 (wide rollout, kept open as reference).

None at the graph API. Internal: the 8 f16 templates' `compile()` / `_host` signatures change (module-level `EXPLICIT_ABI = True`; direct callers use `frost_test_utils.launch_f16`); `VariantPack` gains `observed_bytes(i)` / `observed_device(i)` / `native.facts(indices)`, `VariantPackNative.set_operand` takes the observed facts as optional trailing arguments; `CompiledPlan.takes_variant_pack` is per instance on the FROST SDPA plan. A THD operand passed as a bare address is rejected on the f16 path (a ragged operand needs a sized buffer). The FP8 / MXFP8 / decode-tile / SM120 / SM80 templates and adapter arms are unchanged; they share the changed lowering / graph glue / `VariantPack` plumbing.

## Rebase onto develop `d34a6909c`

develop moved ten commits under the PR (#1096 paged KV on the d192×128 flavor, #1095 attention sink at S_q == 1, #1109 d256 decode tile, #1127 `stats_use_log2`, …). The d192×128 template's host was rewritten by #1096, so develop's version was re-converted with the same two converter scripts used for the original eight (the result is our explicit host plus develop's paged additions); the sink changes in d128/d256/d512 are kernel-body hunks and merged clean; the new d256 decode tile keeps its tensor entry and rides develop's arm behind the `EXPLICIT_ABI` gate like the d128 one. Everything below was re-run on the rebased head.

## Found by the SM107 A/B: CUDA-graph capture without a handle stream

The A/B on the CUPTI-less Rubin host times by CUDA-graph replay, and the prepared THD head replayed in 0.1 µs with correct digests: the kernel had run *during* the capture and the graph was empty. `_FrostSdpaFwdPlan.execute` passed stream 0 (the legacy default) when the handle carried no stream, while the tensor path resolves None to the caller's current torch stream (`APIBase._get_default_stream`). Fixed in commit 3: the lowering hands the plan `default_stream = lambda: api._get_default_stream(None)` and the plan uses it for a None context stream (`engine.py` stays torch-free). `test_capture_without_a_handle_stream_records_the_launch`: the capture leaves the outputs untouched, replay reproduces the eager result, a second replay follows new inputs. Reproduced and verified on SM100 too (`capture_probe.py` in the backup dir).

## Review fixes folded into the commits

- **Zero-element producers** (Codex P1): `VariantPackNative.read_operand` records a zero observed span for any zero-extent shape (the affine span formula assumes a nonempty index domain and claimed capacity for a strided empty O). `test_empty_operands_never_launch_whatever_their_geometry` covers a strided empty O and an empty interleaved Q slice (the negative-span case): no launch, nothing written.
- **Padded Stats storage forms** (Codex P2): the declared per-batch strides apply over the caller's storage — the declared `(B, H, S)` view, the same allocation as contiguous `(B, S, H)` or flat storage all bind; only a gapped view that is neither the declaration nor storage is rejected (`test_padded_stats_accept_the_declared_view_or_contiguous_storage`).
- **SM107 f16 hosts** referenced an undefined `SPLIT_KV` (the SM107 f16 kernels carry no split; the SM100 templates import it from the split helpers) — every SM107 f16 graph build failed on the `frost-sdpa:cutlass-4.8:sm107` lane (519 `NameError`s on pipeline 68357760). Now `split_kv=1`. Not reachable from an SM100 box; the SM107 lane is the evidence.
- A bare-address Stats operand is rejected (was: capacity from `numel` of an unshaped fact); `_build_thd_spec` propagates `NotImplementedError` instead of leaving the plan without a launch (CodeRabbit).
- **Gate slots on SM107 d256** (Codex, third round; 125 of the 137 failures on the SM107 lane of pipeline 68419868): the d256 f16 host declares `gate_ptr` / `gate_strides` unconditionally, but both the dense launch and the prepared THD frame only filled them when a gate was present, so every ungated d256 call on SM107 failed. Both paths now bind the absent gate as `None` + `(0, 0, 0)`. Two plan-time guards make this class impossible to miss on an arch the developer cannot run: the dense adapter declines a host whose runtime slots are outside `_DENSE_EXPLICIT_SLOTS`, and `build_thd_spec` declines a slot that neither the build template nor `bind_thd` fills. `test_every_f16_host_slot_is_bound_by_both_launch_paths` loads all eight templates on both arches (host-side, no GPU) and checks both vocabularies.

CI notes on pipeline 68419868 (`6bd94d711`, before the gate-slot fix): `frost-sdpa:cutlass-rel:sm100` passed in 3076 s; sm80 / sm120 passed; sm107 = the 125 gate-slot failures above + the 12 NVRTC cases below; sm103 never got past `git fetch` within its 2 h job limit (runner, not the tree). Earlier, on pipeline 68357760 (`9c063d150`): analysis / api_index / guardwords / sm80 / sm103 / sm120 lanes pass; the `frost-sdpa:cutlass-rel:sm100` lane hit the 1 h job limit at 89% with no failures — develop's own runs of that lane take 3516–3547 s against the 3600 s cap (pipelines 68381101, 68306102) and the `6f29345a8` run passed in 3263 s, so this is the lane's margin, not a regression of this head. The remaining SM107 failures (12 × `CUDNN_STATUS_INTERNAL_ERROR_COMPILATION_FAILED`, NVRTC) reproduce on develop's own pipeline.

## Testing

Local SM100 (this box; py3.12 pybind built from this tree), rebased head `e8a0be3c4` (in-scope gate re-run after the stream fix: 1315 passed / 0 failed; the other rows are from `ffd49abce`, which differs only by that fix):

| gate | files | result |
|---|---|---|
| in scope + shared plumbing | `test_sdpa_prepared_thd`, `test_flashinfer_shaped_sdpa`, `_fwd_paged_sm100`, `_frontend_integration`, `_execute_is_async`, `_stream_ordering`, `_fwd_dsl_sm100`, `_fwd_split_kv_sm100`, `_compiled_cache_gpu`, `test_split_kv_heuristic`, `_fwd_decode_d128_sm100`, `_fwd_decode_d256_sm100`, `_fwd_heuristics`, `_fwd_api_contract`, `test_knob_vocabulary_frost`, `_python_validate`, `_graph_analyzer`, `test_variant_pack_normalization` | **1314 passed, 5 skipped, 0 failed** (41 deselected: `stats_use_log2` / `strided_stats` ids, local pybind reason) |
| out of scope, must match develop | `_fwd_fp8_sm100`, `_fwd_mxfp8_sm100`, `_mxfp8_sf_bind`, `_fwd_dsl_sm107`, `test_sdpa_fp8_sm107` | **467 passed, 58 skipped, 0 failed** |
| FlashInfer (`yanxu/lse-base` worktree, FROST engines on) | `tests/attention/test_cudnn_prefill_token_indptr.py`, `tests/attention/test_lse_base.py` | **135 passed; 14 passed, 2 xfailed** (the xfails are the pre-existing paged `return_lse` wrapper bug tracked in FlashInfer #5257) |

`test_sdpa_prepared_thd.py` (13 tests + the 8-template slot check): plan prepared and bound from the pack; capacity from the observed span under FlashInfer's ragged declaration; independent frames per call; shrinking lengths and zero capacity (no launch); zero-KV clamp binds the build-time stub; fused-slab K/V at their runtime stride; sub-head-dim head stride rejected; bare address / CPU K / lengths / Stats rejected before launch; no allocation and no synchronization on the warmed path; bounded override through public `graph.execute()` with adapter / lowering / compiler tripwired, then out-of-domain rejected; graph and standalone frames byte-identical. The SM107-named files ran on SM100 here (host-side contract tests; GPU cases skipped) — SM107 GPU coverage, including the LSE store change, is the frost CI pipeline on this head (`@cudnn-ci-bot run frost python_tests`).

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer · worktree /home/scratch.yanxu_gpu/fe-wt-stage1 (local worktree /data/yanxu/worktrees/scratch-gpu/fe-wt-rebase, branch yanxu/frost-f16-rebase → PR head yanxu/frost-prepared-launch; backup /home/scratch.yanxu_gpu/fe-sites/pr1119_backup)</sub>


YangXu1990uiuc:
Test result for `8256d0d7` (the Testing section said "in progress"): SM100, `test_flashinfer_shaped_sdpa`, `test_sdpa_frontend_integration`, `test_sdpa_execute_is_async`, `test_sdpa_stream_ordering`, `test_sdpa_fwd_dsl_sm100` — **610 passed, 5 skipped, 0 failed** (the `stats_use_log2` graph tests deselected for the local pybind reason).

Second commit `f0d039e8`, per Codex's architecture feedback ("one owner per category — buffer normalization owns observation"): the producer's **observed span and DLPack device are recorded once in the native pack** (`read_operand`, and `set_operand` for the Python-described fallback producers; `-1` = unknown, a bare address), apart from the effective geometry that `describe_from` / `override_operand` rewrite. `VariantPack.observed_span(i)` / `observed_device(i)` expose them; the Python `_observed_span` loop is gone; the prepared launch derives every THD capacity from the observed span and checks each operand's device against the plan's. `_normalize` on the FlashInfer graph: 15.0 → 8.8 µs (`pygraph.execute` 49.6). Built and run against a fresh py3.12 pybind from this tree.

Direction for the next commits on this PR, agreed with the review: **consolidate, do not add a second family.** The THD rules (capacities, checks, Stats forms, zero-KV clamp, dummies, declared ops) move into one prepared representation built by the adapter; the graph plan and the standalone `execute()` both bind through it; the adapter's `_thd_pack` / `_thd_operand` / `_thd_lse_operand` / `_launch_thd_explicit` / `_seed_padded_lse` arms for this domain are deleted once equivalence is shown. Then the fixed-CFG `supports()` domain + bounded override binding, then re-measure.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
I am starting the OSS and FROST test lanes for `f0d039e893c6`, including the native VariantPack changes. The current-head checks show no existing test pipeline, and the isolated matching-binding review is underway. Results are pending.

YangXu1990uiuc:
@cudnn-ci-bot run oss frost

YangXu1990uiuc:
Consolidation, per the architecture feedback — two more commits (`0bc3b229`, then the dummies commit):

**One THD f16 launch core, two entries.** `cudnn/sdpa/fwd/prepared.py` now has one owner per category:
- `BufferFacts` — what a caller buffer is (address, dtype, DLPack device, the span the producer guarantees, shape/strides), read once: `facts_of_pack(pack, i)` from the normalized `VariantPack` through the native scalar accessors (no `OperandBuffer` objects), `facts_of_tensor(t)` for the standalone adapter's torch tensors. Nothing downstream touches a buffer object.
- `ThdLaunchSpec` / `build_thd_spec(api)` — the adapter's plan-time facts as one record: positional template with the constants filled, slot indices, per-operand dtype rules, capacity `decl`s, Stats form, workspace regions, the declared per-call operation, and the read-only dummies it owns (`cuMemAlloc` + D32 memset on the launch stream; no torch in `prepared.py`).
- `bind_thd(spec, facts, workspace, stream)` — the runtime rules once: dtype/device/alignment/extent checks, token capacities from the observed spans, the all-KV-zero clamp, paged pools and page tables, the independent frame; None when no Q token is addressable.

The graph plan (`PreparedThdLaunch` = spec + this graph's uids) and `SdpaFwdDslSm100.execute()`'s THD arm both call `bind_thd` and the artifact's positional entry. Deleted from the adapter: `_launch_thd_explicit`, `_thd_kernel`, and the f16 THD use of `_thd_pack` (paged THD included). Quantized THD still binds through `_thd_pack` until its roles join the spec — that is the next family, and it adds roles to the spec, not a second binder.

**Acceptance tests** (`test_sdpa_prepared_thd.py`, 7 passed on SM100): plan is prepared and binds the pack; capacity from the observed span under FlashInfer's ragged declaration (t_q = T; poisoned O/LSE fully overwritten); independent frames per call; shrinking ragged lengths and zero capacity (no launch); bare address / CPU buffer rejected before launch; no allocation and no synchronization in execute (sync-debug mode); graph and standalone execute bind byte-identical frames (all 27 slots). FlashInfer with FROST through this path: `test_cudnn_prefill_token_indptr` 135 passed, `test_lse_base` 14 / 2 xfailed. The wider SM100 THD / paged / integration / stream / async / dsl suite run is in progress; result follows as a comment.

Codex's acceptance questions, as of this head: each buffer fact captured once (native pack) — yes; each THD f16 rule implemented once — yes; execute depends only on prepared state and this call's inputs — yes (`spec` + facts + workspace + stream); a valid override updating geometry without re-entering the adapter — not yet (next: `supports()` domain + bounded override on the spec); can the old THD f16 executor be removed — removed.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
`978f6369` closes the three boundaries from the follow-up review:

1. **Resource lifecycle.** `build_thd_spec` allocates and zero-fills (`cuMemsetD32` + sync, on the plan's device) every read-only dummy a frame can need — the sinks buffer when the plan has no sink, the all-KV-zero V stub when unpaged — so execute never allocates or initializes, on any stream, including the first zero-KV call. The standalone `workspace=None` scratch is per-invocation again (GPU-written metadata is never shared through the spec); the graph path keeps the caller's workspace.
2. **Shared acceptance + geometry.** The binder now checks a packed THD operand's effective (token, head, elem) strides — rank-4 graph description or rank-3 caller buffer — against the plan's declared ones and rejects a mismatch before launch, instead of running on the declared strides. That is the explicit "reject unsupported geometry changes" interim; binding admitted runtime strides from a `supports(effective_geometry)` domain is the next commit. The presence/device rules that lived in the adapter preamble are in `bind_thd` for the migrated roles (lengths, Stats, sinks, page tables).
3. **Observation units.** The native pack records `observed_bytes` in the producer's own element width (`read_operand` from the DLTensor; `set_operand` from the Python-described fallback through the producer's `element_size`); `facts_of_pack` converts to elements of the effective dtype. A uint8 buffer accepted for a bf16 declaration is no longer read as twice its capacity. Unknown stays `-1`, distinct from zero.

Tests added: zero-KV clamp binds the build-time stub with no execute-time allocation; a K sliced out of a wider slab (different head stride) is rejected with no launch; the allocation test is now described as the warmed-path check it is. `test_sdpa_prepared_thd.py` 8 passed; FlashInfer graph: graph-path and standalone frames identical, outputs unchanged vs fa2. The wider SM100 suite is running in a fresh process (the previous run died when I replaced the pybind module under it — my doing, not a code failure); result follows.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
Correction to my previous comment, with the fix: in `978f6369` the device rule ran only for Q/K/V/O; the length buffers, Stats, sinks and page tables were not checked (Codex's probe: a CPU Q-length and a device-1 Stats reached the frame). `618103a72` applies one `on_plan_device` rule to every role `bind_thd` binds — a known producer device must be the plan's CUDA device; an unknown one (bare address) is not a mismatch — and the page table must cover its declared `(B, max_pages)` extent. Test: CPU lengths and CPU Stats are rejected before any launch. `test_sdpa_prepared_thd.py` 8 passed.

Scoping, as requested: "no execute-time allocation or initialization" is the graph / caller-workspace path's contract; the standalone `execute(workspace=None)` convenience deliberately allocates its scratch per invocation (as the old `_thd_pack` did) and shares nothing writable through the spec.

Next commit, on this representation: one geometry resolver that validates the effective (override) geometry against the plan's fixed CFG and returns the strides/dimensions the binder actually writes (no separate boolean `supports()` with its own rules); the acceptance case is a real shape/stride override through public `graph.execute()` on the same prepared artifact, then a return to the original geometry, with adapter re-entry / heuristics / compile tripwired and an out-of-domain override rejected before any seed or launch.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
`cb5a59aac` — the bounded override slice, on this representation:

**One resolver.** `resolve_thd_geometry(spec, facts)` validates the call's effective geometry against the plan's fixed specialization and returns the values `bind_thd` writes (strides per packed role, the batch, the row spans the capacities use). There is no separate boolean `supports()`; admission and binding are the same computation. Fixed by the artifact: head counts, head dims, layout kind, page size. Runtime per call: the batch (from the length buffers, ≤ the declared one — workspace and unit envelope are sized for it; the host takes B as an argument, over-launched units are dead by contract) and each packed operand's (token, head, elem) strides, admitted when TMA-expressible. Rejected before any seed or launch: more sequences than prepared, a padded-Stats buffer with a changed batch, other head counts / dims, a layout TMA cannot express.

**The decisive acceptance case** (`test_bounded_override_through_graph_execute`): a real shape override through public `graph.execute(..., override_uids/shapes/strides)` on the same prepared artifact — 3 sequences on a plan declared for 8, every per-batch operand re-described — with the adapter's `execute` / `compile`, `lower_dsl_prefill` and `compile_cached` tripwired; the frame carries the override (`problem_size` = (3, …, 12, 192)), O / LSE match an independent reference; then the declared geometry again on the same plan; then an out-of-domain override (10 sequences) rejected with no output seed and no launch. Plus: K/V sliced from a fused (T, 2·H_kv, D) slab bind their runtime token stride and produce the reference output; a sub-head-dim head stride is rejected. `test_sdpa_prepared_thd.py` 9 passed; FlashInfer graph outputs unchanged.

Next: measure the integrated public path again and batch the Python/native crossings the profile points at (`_normalize` and the per-role facts reads), without changing this contract. The wider suite result is still pending on the fresh run.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
`79ba942b5` closes the two domain gaps from the bounded-override review: the Stats branch now rejects an effective layout that is not the compiled kind (token-major packed / head-major / declared padded strides — the host builds Stats from the kind, so another layout must not be admitted), and the paged branch validates **both** tables against the runtime `(B, max_pages)` with one shared stride pair plus the pools' `(n_pages, KH, page_size, D)` relations. Tests: a head-major-shaped Stats override on a token-major plan is rejected before seed/launch; paged THD graph tests unchanged. `test_sdpa_prepared_thd.py` 10 passed + the paged THD graph tests.

Next, as agreed: uninstrumented public `graph.execute()` on the FlashInfer THD graph as the baseline (fixed geometry / bounded batch override / fused-slab K/V reported separately, same artifact and cache state, drain-then-call submission), component attribution alongside, then the native batching of the per-role facts reads with one production binder (Python stays the oracle). Suite result for the final head follows.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
**Suite result (final head before the last domain-check commit, `21c87655`):** SM100 `test_sdpa_prepared_thd`, `test_flashinfer_shaped_sdpa`, `test_sdpa_fwd_paged_sm100`, `test_sdpa_frontend_integration`, `test_sdpa_execute_is_async`, `test_sdpa_stream_ordering`, `test_sdpa_fwd_dsl_sm100` — **666 passed, 5 skipped, 0 failed** (`stats_use_log2` graph tests deselected for the local pybind reason). The one failure of the previous run was real and is what `21c87655` fixed: a CPU padded-Stats buffer was recorded as an *unknown* device and reached the `-inf` seed; a CPU tensor is now a known kDLCPU device and is rejected as before. On `79ba942b` (the two domain checks): `test_sdpa_prepared_thd` 10 passed + the paged THD graph tests; FlashInfer with FROST through the prepared path: `test_cudnn_prefill_token_indptr` 135 passed, `test_lse_base` 14 passed / 2 xfailed.

**Integrated baseline, uninstrumented public `graph.execute()`** (FlashInfer-shaped THD graph b=64 × q4, kv 1024, GQA 8/2, d128 bf16, Stats, ragged declaration; SM100 idle; drain, then one call; median of 500; two rounds):

| geometry | `graph.execute()` |
|---|---|
| fixed | 47.7 / 47.9 µs |
| bounded batch override (32 of 64 sequences, 10 operands re-described) | 91.5 / 92.5 µs |
| fused-slab K/V (runtime token stride 512) | 47.1 / 47.1 µs |

Attribution (instrumented, fixed geometry): `_uid_to_data` 1.1 · `_normalize` 4.7 · `facts_of_pack` 1.8 × 7 roles = 12.6 · `bind_thd` 19.5 · positional call 6.5 · plan / graph glue ≈ 6. The override case's extra ~44 µs is the graph layer's Python override loop in `_normalize` (`storage_geometry` + `_in_axis_order_of` + `override_operand` per operand), not the binder.

Next commit, per the agreed order: one native read of all roles' facts per call (`VariantPackNative.facts(indices)`), replacing the 42 scalar accessor crossings behind `facts_of_pack`; the binder stays the single production implementation.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
`7f18a164c`: one native crossing for a launch's operand facts (`VariantPackNative.facts(indices)`, consumed by `facts_of_roles`), replacing 6 accessor calls × 7 roles. Same graph, same method as the baseline above:

| geometry | before | after |
|---|---|---|
| fixed | 47.7 µs | **38.9** |
| fused-slab K/V | 47.1 | **38.2** |
| bounded batch override | 91.5 | 84.4 |

Attribution now: `_uid_to_data` 1.1 · `_normalize` 4.5 · `bind_thd` 19.2 · positional call 6.3 · glue ≈ 6. `bind_thd` is the remaining Python block; when it moves native it stays the single production implementation (Python `bind_thd` becomes the oracle in the tests), and the override case's extra ~45 µs is the graph layer's Python override loop in `_normalize`, a separate item. Acceptance file 9 passed on the rebuilt module.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc:
**Final suite result for this head, `f5a0ef263`:** SM100, `test_sdpa_prepared_thd`, `test_flashinfer_shaped_sdpa`, `test_sdpa_fwd_paged_sm100`, `test_sdpa_frontend_integration`, `test_sdpa_execute_is_async`, `test_sdpa_stream_ordering`, `test_sdpa_fwd_dsl_sm100` — **666 passed, 5 skipped, 34 deselected, 0 failed** (deselected: the `stats_use_log2` graph tests and the two `strided_stats` ids that pass that kwarg; this box's prebuilt pybind predates the binding).

The run on `2eafa311` had 13 failures, all real: the Stats layout-kind check I added in `79ba942b` rejected flat (T·H) and (H, ext) head-major Stats buffers as "unsupported rank". `f5a0ef263` accepts compact rank ≤ 2 storage for the token-major and head-major kinds (it carries no layout that could contradict the kind; the tensor path bound it the same way) and keeps the rank-3 / rank-4 description check. The public-entry numbers are unchanged on this head (fixed / fused-slab / batch override re-measured: fixed 29.6 / 30.2 / 31.6 µs, fused-slab K/V 28.7 / 29.1 / 32.6, bounded batch override 74–102 (three runs; the override loop is Python-heavy and the box had other GPU processes at the time)).

The PR description's validation table records every head with what ran on it. Handing over to overall review at this checkpoint; the two follow-ups (generic native override normalization in `VariantPack`, then the native THD binder decided from the remaining profile, one production implementation) are separate PRs.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

YangXu1990uiuc (commented):

Reviewed `f5a0ef263501` against base `5fb8d3b44f49`.

The shared binder and batched observation are a sound direction. Current-head probes confirm the CPU-Stats guard and the compact/head-major Stats fixes. Two high-risk P1s still need addressing before merge, plus a narrower remaining Stats compatibility issue. Other inherited #1099 findings remain tracked at https://github.com/NVIDIA/cudnn-frontend/pull/1099#pullrequestreview-5231001557.

**[P1 · blocks approval] Return zero observed bytes for every zero-element producer** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f5a0ef263501fefaf05b9d905481323f86f7f64c/python/pygraph/variant_pack.cpp#L553))

The affine span formula assumes a nonempty index domain. Empty O with shape (2,8,0,128) and legal strides (8192,128,1024,1) records observed_bytes=16384 despite numel=0. The binder derives T_q=8 and reaches the raw call with o_ptr=0. Empty interleaved Q/KV instead produce negative bytes and are rejected as bare addresses. Short-circuit any zero extent to span 0 before applying the formula, keeping unknown capacity distinct. Fabricating writable capacity for a null output makes this high-risk P1; please cover empty axes in different positions.

Evidence: Reproduced on f5a0ef263 with matching native sources: bulk/scalar observations agree; empty O reaches a launch interceptor with problem_size=(2,8,4,8,15,0), o_ptr=0. Empty interleaved Q/KV report -2048/-1024 bytes. The null-pointer launch was intercepted before CUDA executed it. Attribution: Introduced by this PR's move from Python capacity observation, which checks numel==0, to native read_operand.

**[P1 · blocks approval] Integrate develop's decode template with the explicit ABI** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f5a0ef263501fefaf05b9d905481323f86f7f64c/python/cudnn/sdpa/fwd/api_dsl.py#L2238))

The stack unconditionally compiles selected templates with _explicit_compile_kwargs, including lse_kind. Current develop adds decode_d128_f16.py with the earlier tensor-based compile/host interface, so ordinary decode graph builds fail with unexpected keyword argument 'lse_kind'. The template is absent from this branch, which hides this in branch-only testing. Port its complete compile/launch interface or retain a compatible adapter before landing; dropping only that keyword leaves the host ABI mismatch. This is high-risk P1 because a normally selected core decode path fails at plan construction.

Evidence: Pipeline 68327331 integrated superseded f0d039e8 with newer develop. SM103 job 443593238 fails test_graph_decode_selects_the_decode_tile[NHD/HND] and other decode cases with this TypeError. Inspected the exact tested tree, then verified the incompatible adapter is unchanged on f5a0ef263 and the legacy decode template remains in current base 5fb8d3b4. This does not attribute every failure in that lane. Attribution: Inherited from #1099 and exposed by the decode kernel now on develop; not introduced by the prepared THD binder.

**[P2] Preserve padded Stats storage views for nondefault declared strides** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/f5a0ef263501fefaf05b9d905481323f86f7f64c/python/cudnn/sdpa/fwd/prepared.py#L317))

The low-rank fix covers ordinary B/H/S storage, but the padded branch still confuses producer storage geometry with a layout override. Declare padded Stats with strides (64,1,4) over B=2,H=4,S=16 storage: passing its declared (B,H,S) view succeeds, while passing the same allocation as contiguous (B,S,H) or flat storage raises ValueError. Both forms satisfy the existing _checked_padded_lse contract, which applies declared strides over caller storage. Normalize storage-only inputs using the declaration, while retaining rejection of explicit incompatible overrides. This is nonblocking: supplying the declared view works around it.

Evidence: Same-device standalone probe: both storage forms pass on parent 8fd1831b with results identical to the declared-view control; both are rejected on f5a0ef263. Separate current-head probes confirm ordinary flat padded Stats and head-major (H,T) storage now pass. Attribution: Remaining compatibility regression from this PR's Stats layout-kind check; f5a0ef263 fixes the broader head-major/compact failures but retains the nondefault padded-stride restriction.

Validation: Current-head SM100 probes use the extension built from 7f18a164 after verifying native sources are identical, with DSL 4.8 and cuDNN 9.25.1. Normal O/LSE execution passes; CPU Stats is rejected before fill; empty-buffer and remaining padded-storage findings reproduce. Reviewed both incremental Python commits; no PR source was changed. Reused scoped evidence from preceding revisions: 79ba942b boundary suite had 68 passed, 6 failed, 1 skipped, 1 deselected. The six failures concerned the subsequently changed Stats checks; this is not a current-head suite result. On 7f18a164, analytic correctness, changed-address binding, explicit-stream capture/replay with changed lengths and poisoned buffers passed. Compared with parent 8fd1831b, two BF16 D128 THD cases reduced median CPU enqueue 56.87 to 44.34 us and 60.40 to 50.29 us; GPU medians stayed within 0.1%. These timings precede the cold-error-branch optimization and are not f5a0ef263 measurements. CI pipeline 68327331 is for superseded f0d039e8, not this head; its SM100 lanes are still pending. Owner reports 666 passed/5 skipped/34 deselected on f5a0ef263; that broader run was not independently repeated.

Limitations: Focused SM100 validation, not full parent-stack or Rubin qualification. Performance evidence covers two workloads on a 68-SM GPU and is not an application-level claim. The null-pointer call was safely intercepted; no GPU crash is claimed.

YangXu1990uiuc:

YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `9c063d150b10` against base `5fb8d3b44f49`.

The narrower f16/bf16 scope resolves the previous decode ABI blocker: the EXPLICIT_ABI gate keeps develop's decode template on its tensor interface, and both NHD/HND graph decode regressions now pass. The empty-buffer high-risk P1 and the narrower padded-Stats P2 from the previous review remain reproducible on this head. Please address the empty-buffer issue before merging.

- **[P1 · blocks approval] Still open:** [Return zero observed bytes for every zero-element producer](https://github.com/NVIDIA/cudnn-frontend/pull/1119#pullrequestreview-5233537185).

- **[P2] Still open:** [Preserve padded Stats storage views for nondefault declared strides](https://github.com/NVIDIA/cudnn-frontend/pull/1119#pullrequestreview-5233537185).

Validation: Exact-head SM100 run with matching native bindings, DSL 4.8 and cuDNN 9.25.1: 13 passed, zero skipped. This includes all 9 prepared-THD tests, both NHD/HND graph decode cases, prefill selection and THD/decode exclusion. Additional exact-head probes: ordinary graph O/LSE succeeds, CPU Stats is rejected before a CUDA fill, and the two retained findings reproduce. Compared the rescope against both develop and the previously reviewed revision; all eight f16 templates retain the prior implementation, apart from two comment markers in SM107 d256. Pipeline 68357760 is verified against head 9c063d150. The API-index and other analysis/build checks passed at inspection; GPU lanes were running or pending. The owner already requested this run, so no duplicate CI was started.

Limitations: Focused follow-up, not a new full-suite or Rubin qualification. CPU/GPU timings from the previous review belong to earlier revisions and were not remeasured for this rescope. No GPU fault is claimed: the unsafe launch was intercepted.

YangXu1990uiuc:
CI follow-up on `9c063d150`: the narrower scope still carries the Rubin blockers from [the earlier #1099 review](https://github.com/NVIDIA/cudnn-frontend/pull/1099#pullrequestreview-5231001557), in addition to the empty-buffer finding above.

Pipeline 68357760 / SM107 job 443837599 has finished with **531 failed, 2720 passed, 1685 skipped**. Its trace includes `NameError: name 'SPLIT_KV' is not defined` in the new f16 `_host` prologue. All four current SM107 f16 templates still reference that undefined global; representative sites are d128:2091, d192:2095 and d256:2133. This is a retained PR regression, not an infrastructure failure. I have not attributed all 531 failures to it.

The other inherited issue also remains visible in the current d256 f16 interface: `_host` requires `gate_ptr` / `gate_strides`, but `_execute_dense_explicit` adds them only when a gate is present. After fixing the compile error, keep the ordinary **ungated** case in the regression selection; the previous same-hardware diagnostic run isolated its missing-argument failure. That diagnostic belongs to the earlier revision, while the current evidence for this second issue is the unchanged call/signature contract.

These remain high-risk P1s to address before merge. My preceding follow-up confirmed the decode fix and reran the prepared-THD findings; it did **not** clear the inherited Rubin findings. SM80 FROST and SM90 Python lanes are green; the remaining GPU lanes were still running or pending at inspection. No duplicate CI was started.

YangXu1990uiuc:
Head `6bd94d711` (rebased; fixes folded into their commits):

- **Codex P1 — zero-element producers**: fixed in `read_operand` (zero span for any zero-extent shape, so the affine formula never claims capacity for an empty operand); new test covers a strided empty O and an empty interleaved Q slice, both bind no launch.
- **Codex P2 — padded Stats storage forms**: the declared per-batch strides apply over the caller's storage; the declared view, contiguous `(B, S, H)` and flat storage all bind, only a gapped non-declared view is rejected. CPU test added.
- **SM107 lane**: the four SM107 f16 hosts referenced an undefined `SPLIT_KV` (519 `NameError`s on 68357760) — `split_kv=1`, these kernels carry no split. The 12 NVRTC `INTERNAL_ERROR_COMPILATION_FAILED` cases in that lane reproduce on develop's own pipeline.
- **SM100 lane 1 h timeout**: develop's own runs take 3516–3547 s against the 3600 s cap; the `6f29345a8` run passed in 3263 s. Lane margin, not this head.
- CodeRabbit: `_build_thd_spec` now propagates `NotImplementedError`; the `launch_f16(host=)` finding does not hold — the compiled f16 entry exposes the host's parameter names (every direct-launch f16 test in `test_sdpa_fwd_split_kv_sm100` / `_paged_sm100` passes without `host=`); the fp8-template threads refer to files that are byte-identical to develop on this head.

Local SM100 gate on `6bd94d711`: in-scope 1187 passed / 0 failed; the PR body has the details.

YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `6bd94d711156` against base `7ac5354191cb`.

The empty-buffer P1 and padded-Stats P2 are resolved, and the four SM107 hosts now define the unsplit geometry correctly. One previously reported Rubin ABI issue still blocks approval: ordinary ungated d256 calls need valid gate argument slots in both the dense wrapper and prepared THD frame. Current hardware checks and exact-head CI confirm it. Please address this before merging; leaving COMMENT, with no additional CI trigger.

- **[P1 · blocks approval] Still open:** [Bind the disabled gate slots in both d256 launch paths](https://github.com/NVIDIA/cudnn-frontend/pull/1119#issuecomment-5712612199).

Validation: Reviewed the full eight-file incremental diff against 9c063d150. Built matching native bindings from this head. On SM100 with cuDNN 9.25.1 and DSL 4.8, 16 focused prepared-THD/decode tests passed without skips. Independent boundary probes confirm empty interleaved Q/KV and strided empty O all record zero observed bytes. Empty Q/O do not launch; zero KV uses the intended one-row dummy path. CPU Stats is rejected before a CUDA fill. Declared, flat and contiguous BSH padded-Stats storage all produce identical correct O/LSE, resolving both prior findings. Independent current-head SM107 run: four gated FP16/BF16 d256 cases passed, and the ungated half of the LSE-independence test failed with the missing gate_ptr/gate_strides TypeError. The remaining fix must also initialize prepared.py's THD gate_strides slot to a tuple: its current None value causes the separate positional-ABI error. Owner-started pipeline 68419868, mirror metadata verified against full head 6bd94d71115640bead3a44975f42232dbaa6ba0c: SM107 job 444326834 finished 137 failed/3104 passed/1695 skipped. The full failure summary groups exactly into 46 dense missing-gate-argument failures, 79 THD gate_strides tuple failures, and 12 native ragged-decode Stats NVRTC failures. The 125 gate-slot failures are actionable PR regressions, separate from the native NVRTC cases. SM100/SM103 lanes were still running at inspection; no duplicate pipeline was started. Incremental head/previous/previous/head spot checks on SM100 for BF16 THD D128, Hq8/Hkv2, (B,Sq,Skv)=(4,4,128) and(64,4,1024): analytic O/LSE, changed buffer addresses, changed-length captured replay and sync-debug checks passed in every arm. GPU medians stayed within 0.3% between versions (about 26.7 us and 810 us); CPU enqueue varied 33-46 us across runs with no stable directional signal, so no small CPU-performance claim is made.

Limitations: Focused SM100/SM107 validation and current CI failure attribution do not establish a green full current-head suite. Performance checks compare this fix against the previous reviewed head, not an application-level baseline. No PR source was modified.

RomanAnders90 (approved):

YangXu1990uiuc:
Head `ffd49abce`: rebased onto develop `d34a6909c` (the d192×128 template re-converted on top of #1096's paged rewrite; #1095 / #1109 ride develop's arms), plus the third-round fix:

- **Gate slots (Codex P1)**: the dense launch and the prepared THD frame now bind the absent gate as `None` + `(0, 0, 0)` on every gate-capable host. Two plan-time guards make this class of miss impossible on an arch the developer cannot run: the dense adapter declines a host with runtime slots outside `_DENSE_EXPLICIT_SLOTS`, `build_thd_spec` declines a slot neither the build template nor `bind_thd` fills, and `test_every_f16_host_slot_is_bound_by_both_launch_paths` loads all eight templates on both arches host-side and checks both vocabularies.
- The 12 remaining SM107 `INTERNAL_ERROR_COMPILATION_FAILED` cases reproduce on develop's own pipeline; on pipeline 68419868 the SM100 lane passed in 3076 s and the SM103 lane never got past `git fetch` in its 2 h limit.

Local SM100 on this head: in-scope 1314 passed / 0 failed (incl. develop's new d256 decode tests), out-of-scope fp8 / mxfp8 / SM107 files 467 passed / 0 failed, FlashInfer 135 + 14 (2 known xfail). Device-time A/B table in the body is from the previous base; kernel bodies changed only by develop's sink hunks since.

YangXu1990uiuc:

YangXu1990uiuc:
@RomanAnders90 one more change after your approval, worth a look: head `e8a0be3c4` (= `ffd49abce` + one fix in commit 3).

**CUDA-graph capture without a handle stream.** The SM107 device-time A/B (Rubin hosts have no CUPTI, so it times by CUDA-graph replay) replayed the prepared THD head in 0.1 µs with correct digests: the kernel had run *during* the capture and the graph was empty. `_FrostSdpaFwdPlan.execute` passed stream 0 (legacy default) when the handle carried no stream, while the tensor path resolves None to the caller's current torch stream (`APIBase._get_default_stream`). The plan now gets `default_stream` from the lowering and uses it for a None context stream; `test_capture_without_a_handle_stream_records_the_launch` pins it (outputs untouched by the capture, replay == eager, a second replay follows new inputs). Reproduced on SM100 as well.

Also new in the body: the SM107 device-time table (dense ±1 %, THD within that box's noise and never slower, all O/LSE digests identical to develop). CI on `ffd49abce`: the SM107 lane is down to develop's 12 NVRTC cases; the `py_test` lanes' 4 `linear_attention/test_la.py::test_fwd_bwd_correlated_keys` failures are in code this PR does not touch (last changed by #1124).

YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `e8a0be3c40d3` against base `d34a6909c9d1`.

The Rubin gate-slot high-risk P1 is resolved, and the additional no-handle-stream capture fix is verified. I reproduced the empty-graph failure on ffd49ab and confirmed the new head records and replays the launch correctly. My earlier capture checks explicitly supplied a handle stream, so they did not cover this branch. No remaining approval-blocking finding from this follow-up. This is a COMMENT on my own account's PR; the existing reviewer approval is preserved.

Validation: Reviewed the complete rebase/range-diff through ffd49ab, then the three-file e8a0be3 increment. Both dense and prepared THD paths bind the disabled Rubin gate; the implicit-stream fix resolves the caller's current stream at execution time while retaining explicit handle-stream behavior. Native sources are identical to the exact ffd49ab binding built for this review. SM100, CuTe DSL 4.8, cuDNN 9.25.1.1: the new capture test fails on ffd49ab because capture already modifies the outputs and the graph is empty (4.93 s); on e8a0be3 it passes (4.20 s), including unchanged outputs during capture and changed-input replay. The entire current prepared-THD suite passed 20/20 with zero skips in 4.32 s. Before the stream-only increment, 26 prepared/d192-paged/split/sink cases passed on ffd49ab; unaffected evidence is reused. Verified full-head provenance for both CI pipelines. On ffd49ab, pipeline 68495007's Rubin job 445018151 completed with 3229 passed, 1963 skipped and 12 failed. All 12 failures are native ragged-decode Stats NVRTC errors; the previous 46 dense missing-argument and 79 THD tuple failures are gone. This is CI evidence, separate from local SM100 execution. The current e8a0be3 pipeline 68499224 is already running; no duplicate CI was triggered. Interleaved ffd49ab/6bd94d7/6bd94d7/ffd49ab SM100 checks passed analytic O/LSE, changed-address binding, changed-length capture with poisoned buffers and sync-debug checks. For BF16 THD D128, Hq8/Hkv2, (B,Sq,Skv)=(4,4,128) and (64,4,1024), GPU medians stayed within 0.2% (about 26.7 us and 810 us); CPU enqueue medians overlapped at 32.6-34.2 us and 35.3-37.6 us. These runs explicitly supplied a handle stream, whose execution branch is unchanged in e8a0be3; they do not measure the new implicit-stream lookup.

Limitations: CI is not all green. Current-head API-index passed and GPU lanes were still running at inspection. The guardwords failure concerns SUPPORT_MATRIX_TRACKER.md:311 inherited unchanged from develop d34a6909. The preceding pipeline's three inspected Python lanes each fail four cuTile correlated-key tests added on develop; the tests and linear-attention implementation are unchanged by this PR. No fresh same-environment develop GPU control was run for those numerical failures, and neither they nor the native NVRTC failures are claimed fixed or waived. Independent GPU execution here covers SM100. Rubin evidence is the verified preceding-head CI run; the current stream-only increment is locally tested and awaits its own CI completion. Performance coverage is limited to the two explicit-stream THD cases above.

YangXu1990uiuc:
Current-head CI follow-up for `e8a0be3c`: pipeline **68499224** has one additional Rubin failure that should be triaged separately from the known NVRTC cases.

SM107 job **445061994** reports **3228 passed, 1963 skipped, 14 failed and 1 error**. Twelve failures are the native ragged-decode Stats compilation cases. The remaining failure/error entries describe one `test_sdpa_random_bwd_L0[test420]` launch failure and its crashed worker: BF16, B=10, Hq/Hkv=6/1, Sq/Skv=8192/5951, Dqk/Dv=192/128, dense strided storage.

The error first surfaces in `compute_and_compare_reference` while reading output statistics. The harness synchronizes after both FE forward and backward execution, so the traceback does not establish that a FROST launch failed. This case also uses an explicit handle and dense storage, outside the new implicit-stream prepared-THD branch. I have not reproduced or attributed the fault; please isolate this case with launch blocking on the matching cuDNN 9.28 stack before treating the Rubin lane as only the known 12 failures.

SM80/SM120 FROST passed; SM100/SM103 FROST were still running. Both inspected SM100 Python lanes fail only the four cuTile correlated-key cases already noted in the review. This updates the pending-CI limitation in my review, without changing its verified gate-slot and capture fixes.

YangXu1990uiuc:
CI on `e8a0be3c4` (internal pipeline 68499224) is complete. FROST SDPA lanes: **sm80, sm100 (5752 passed, 51.6 min), sm103 (5752 passed), sm120 all green**. Every remaining red is present on develop's own pipelines and outside this PR's diff:

- `frost-sdpa:cutlass-4.8:sm107`: 14 failed / 3228 passed / 1 error — the 12 `test_sdpa_ragged_decode_stats` NVRTC `INTERNAL_ERROR_COMPILATION_FAILED` cases plus the `test_sdpa_random_bwd_L0` "unspecified launch failure" worker crash; develop's own sm107 runs (pipelines 68381101, 68306102) show the identical 14 / 3228 / 1 signature. The two regressions this lane found earlier in the PR (`SPLIT_KV`, ungated gate slots) are gone.
- `py_test:*` (six lanes): 4 × `linear_attention/test_la.py::test_fwd_bwd_correlated_keys` (non-finite `o`); linear attention is untouched here and last changed by #1124.
- `analysis:guardwords_scan`: `SUPPORT_MATRIX_TRACKER.md:311` ("trtllm-gen"), a develop file from #1109, byte-identical here.

PR #1122, opened 2026-09-17
[DSv4.1] Fuse native-layout vision RoPE backward and QKV gradient packing

YangXu1990uiuc:
The published DSv4.1 vision attention path produces BF16 Q/V gradients in HTD storage and K gradients in HDT storage. This adds `VisionRoPEBackward` and its allocation wrapper with `backend="frost"`: one SM100 kernel reads those layouts directly, applies split-half rotary backward, and writes contiguous packed dQKV. It removes the separate K-layout copy and fuses Q/K rotation with V packing.

The API supports dynamic token counts without recompilation, caller-owned output, explicit streams, and CUDA Graph capture. It validates dtype, layout, alignment, device, and output overlap. Trigonometric tables use FP32; multiply and add/subtract round separately before BF16 output. The implementation targets 16 heads and head dimension 64.

The mathematical contract follows [DeepSeek-V4.1-Flash vision.py](https://huggingface.co/deepseek-ai/DeepSeek-V4.1-Flash/blob/dba1be0a40aa45a94ad051997016db3960a90277/inference/vision.py). The GPU addressing, shared-memory transpose, and fused packing are authored for this implementation.

Validation:

- 29 focused GPU tests pass on SM100, covering six model-derived image geometries, tail tiles, fresh buffers, dynamic T, invalid layouts/dtypes/alignment, overlap, streams, and Graph replay.
- B200 validation loads the actual checkpoint's first vision block. All 24 ordinary/checkpointed forward-and-backward comparisons pass across six geometries and two input banks; intermediate operands and the public FE kernel route are verified.
- A separate B200 run validates the complete 32-layer vision tower and aligner with actual checkpoint weights, six model-derived geometries, and two input banks per geometry. Original 3D math attention and RoPE forward remain unchanged; non-reentrant checkpointing is enabled per block. Each run checks the output, all 32 hidden states, input gradient, and all 263 parameter gradients against the source implementation. All 297 tensors match exactly numerically; outputs and hidden states also match bitwise. Source-repeat and Torch-formula adapter controls pass, and profiling confirms one native FROST backward call per layer (384 calls across 12 cases). The compiled PTX and cubin are identical to the operator benchmark.
- Two separate B200 runs qualify the kernel. The public API run compares original autograd, Transformer Engine, FlashInfer, four compiled Torch variants, and a fused FROST control that first copies dK. All required layout conversions and packed-output writes are timed. Each cache regime uses eight paired blocks with five samples per provider and block.

The following **backward operator** speedups use the fastest same-run control, which was the copy-dK-plus-FROST implementation:

| Vision tokens | Graph GPU speedup (hot / evicted) |
| ---: | ---: |
| 1521 | 1.36x / 1.37x |
| 1610 | 1.36x / 1.34x |
| 4070 | 1.39x / 1.38x |
| 5476 | 1.40x / 1.40x |
| 8418 | 1.41x / 1.42x |
| 8649 | 1.39x / 1.39x |

Across these shapes, public-API Graph GPU speedup is 1.34–1.42x and Graph wall-time speedup is 1.29–1.42x. Every paired block exceeds 1.05x for both Graph clocks and cache regimes. Relative to the fastest external provider, Graph GPU speedup is 1.76–1.91x. Eager wall-time medians improve 1.03–1.13x, with smaller and noisier gains.

A separate B200 timing run covers forward plus checkpointed backward of the full 32-layer vision tower and aligner, using the same actual checkpoint, six image geometries, two input banks, and all nine providers. Each geometry uses eight paired blocks with three samples per provider; compilation is excluded and outputs/gradients are rechecked after timing. Relative to the fastest same-run control, median whole-phase wall-time ratios are **0.999–1.003x**, and CUDA-event span ratios are **0.999–1.003x**; against only the numerically exact controls, wall-time ratios are 1.001–1.006x. None of the six geometries shows a stable >5% whole-phase gain. These results do not demonstrate an overall vision-training speedup.

The performance benefit demonstrated here is confined to the RoPE backward-and-pack operator. The vision tests retain the source 3D math attention path and exclude optimizer updates and the language backbone. Only the kernel, public API, user documentation, and focused tests are included.

## Summary by CodeRabbit

* **New Features**
  * Added experimental Vision RoPE backward support for compatible Blackwell systems.
  * Added APIs for compiling and executing Vision RoPE backward operations, including a convenience wrapper.
  * Supports packed BF16 QKV gradients, optional CUDA streams, dynamic token sizes, and CUDA Graph execution.

* **Documentation**
  * Documented the Vision RoPE backward API, requirements, tensor layouts, validation rules, and usage.


vedaanta (approved):

Reviewed `959440e2762f` against base `3c79b970410e`.

Reviewed the native HTD/HDT addressing, shared-memory transpose, split-half gradient formula, packing, runtime validation, dynamic compile identity, and stream/device handling. No actionable findings. The evidence supports this backward operator, not an overall vision-training speedup.

Validation: Inspected the entire diff and relevant APIBase, dependency-gating, package discovery and lazy-export paths. The fixed 16-head/64-channel contract makes the overlap range calculation valid for these dense storage layouts. Checked the gradient signs against the pinned upstream apply_rotary: https://huggingface.co/deepseek-ai/DeepSeek-V4.1-Flash/blob/dba1be0a40aa45a94ad051997016db3960a90277/inference/vision.py. git diff --check passed. Fixed-head SM100 validation with cuDNN 9.26.0, PyTorch 2.13 and CuTe DSL 4.7: all 29 shipped tests passed in 6.12 s. Covers numerical equality, fresh buffers, dynamic T without recompilation, metadata-only construction, invalid contracts, overlap, explicit streams and graph replay. Additional independent autograd VJP comparisons passed exactly for T=1,2,17,31,32,33,63,64,65,1521. Verified bit-preserving V packing across all 65,536 BF16 bit patterns. A delayed side-stream wrapper call stayed asynchronous and its freed output was not recycled by ambient-stream allocation churn; the copied result remained exact. Limited timing, five interleaved repeats after warm-up, compilation excluded: at T=1521, graph GPU medians were 46.73 us for the cached FROST class API versus 214.19 us for an unfused Torch-formula reference; at T=8649, 223.29 versus 1152.40 us. Separate eager CPU enqueue medians were 21.45/123.83 us and 22.96/129.47 us respectively. Graph and eager outputs matched exactly. This is not a comparison with the fastest optimized external provider. Current-head GitHub Style and Docs checks passed. Merge-requirements failures request a Project board and Milestone; no Actions workflow awaits approval. No existing test-pipeline request was found. CI assistance is deferred until a current target is confirmed to collect this new suite on SM100 with DSL >=4.7; no broad run was launched merely to produce skipped tests.

Limitations: Did not reproduce the author's full checkpoint/model qualification or nine-provider timing campaign. Local timing is a small unlocked-clock operator spot check against an unfused formula, with allocations included in that reference's eager host path. Tests loaded the fixed-head Python source with a compatible existing FE extension; the new operation executes through CuTe DSL, not modified native bindings. No wheel-build or older-PyTorch campaign was run. Current CI lane configuration was not accessible for independent target verification. Before requesting an OSS run, confirm that its SM100 lane selects fe_api/rope/test_vision_rope_backward.py and uses CuTe DSL >=4.7.

YangXu1990uiuc:
@cudnn-ci-bot run oss python_tests

YangXu1990uiuc:
Merged develop after Engram and both RoPE QDQ PRs landed. Vision API and kernel sources are byte-identical to the reviewed head. The shared family now lazily resolves Vision through its API module and QDQ through its QDQ module, preserving each dependency boundary. Exact-checkout export/missing-dependency checks and targeted pre-commit hooks passed. Native binding inputs match the validated reference.

The existing GPU pipeline remains the regression evidence; no full rerun was requested for this export integration. The canceled SM107 OSS lane is outside the explicit SM100-only contract; all applicable GPU/build/analysis jobs passed. No SM107 coverage is claimed.

PR #1128, opened 2026-09-17
Norm torch custom ops

rmhaskarnvidia:
## Before submitting

## Affected area
- Python API or bindings

## Summary
Adds experimental PyTorch custom operators for LayerNorm and RMSNorm, including forward, backward, fake-tensor, autograd, and `torch.compile` support.
The implementations live under `cudnn.ops.norm` and are lazily exposed through:
```python
from cudnn.experimental.ops import layer_norm, rms_norm
```
The implementation includes thread-local, device-pinned cuDNN handles; bounded thread-safe graph caches; input validation; and non-differentiable saved-stat outputs.

## Why

This provides PyTorch-native access to cuDNN graph-backed LayerNorm and RMSNorm without requiring users to construct and execute cuDNN graphs manually.
The operators integrate with PyTorch dispatch, autograd, fake tensors, and compilation while preserving the framework-free `import cudnn` boundary. Keeping the public entry points under `cudnn.experimental.ops` allows the API to mature without adding stable top-level exports.

## Related issues
None 

Adds two experimental public APIs:
- `layer_norm(input, normalized_shape, weight=None, bias=None, eps=1e-5)`
- `rms_norm(input, weight, bias=None, eps=1e-5)`
Current constraints:
- CUDA tensors only.
- FP16, BF16, and FP32 activations.
- Normalization over one final dimension.
- Inputs must contain at least one normalization row.
- RMSNorm weight and bias must match the input dtype.
- LayerNorm supports FP32 parameters with FP16/BF16 input.
- The RMSNorm signature intentionally supports optional bias and is not a drop-in replacement for `torch.nn.functional.rms_norm`.
No existing stable APIs are changed.
Validated on SM100 with PyTorch 2.11.0+cu128 and cuDNN 9.26.0. `CUDNN_FRONTEND_CUDART_LIB_NAME=libcudart.so.13` was set to make cuDNN Frontend runtime selection explicit.
The benchmark completed successfully, although these custom operators were slower than PyTorch for the tested shapes, primarily due to per-call overhead. CUDA 13 PyTorch should be used before treating those measurements as final apples-to-apples performance results.

## Testing

Environment:
- SM100 GPU, compute capability 10.0
- PyTorch 2.11.0+cu128
- cuDNN 9.28.0 (`cudnn.backend_version() == 92800`)
- Local editable installation from this checkout
- `CUDNN_FRONTEND_CUDART_LIB_NAME=libcudart.so.13`
Commands and results:
```bash
pre-commit run --from-ref "$BASE" --to-ref HEAD
```
Passed. `clang-format` skipped because the PR contains no C/C++/CUDA changes; Black, Black-Jupyter, and SPDX checks passed.
```bash
cd test/python
pytest -q test_torch_norm_ops.py -m L0
```
Result: `52 passed`
```bash
pytest -q test_torch_norm_ops.py -m L1
```
Result: `2 passed, 2 skipped`
The skipped tests require two GPUs; the test node has one GPU.
```bash
pytest -q test_import_boundaries.py
```
Result: `15 passed`
```bash
python benchmark/norms/benchmark_torch_norm_ops.py \
    --op all --warmup 10 --repeat 100
```
Completed successfully.

## Summary by CodeRabbit

- **New Features**
  - Added experimental CUDA-accelerated `layer_norm` and `rms_norm` operations for PyTorch.
  - Supports FP16, BF16, and FP32 inputs, autograd, and `torch.compile`; RMSNorm also supports optional bias.
  - Exposed normalization operations through the public operations packages with lazy loading and multi-device support.

- **Documentation**
  - Added usage, compatibility, and input requirements for the new normalization APIs.
  - Added an experimental benchmark comparing PyTorch and cuDNN normalization performance across supported scenarios.


YangXu1990uiuc (approved):

Reviewed `74f1c3a30071` against base `d34a6909c9d1`.

The lazy exports, per-call workspace, graph caches and autograd integration are reasonable. Please address the device guard and benchmark comments below. I am treating the multi-device issue as nonblocking because it produces an explicit launch error in a bounded configuration and callers can work around it with an input-device context. The independent checks also support the description's caution about eager overhead; RMSNorm's GPU benefit becomes visible under capture.

**[P1] Keep the input device current through graph planning and execution** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/74f1c3a300715f4f97fbbf63ec54d3fd4cdffbf6/python/cudnn/ops/norm/_common.py#L29))

The device context covers handle creation only. After get_handle returns, both CUDA implementations build/execute the graph with the caller's ambient device current. With x/weight on SM100 and the current device set to an A100, both public ops fail with CUDA error 209 (no kernel image is available). The PR's own two create_handles_on_input_device tests also fail on this pair. Please guard the entire forward and backward CUDA implementation with torch.cuda.device(x.device), including graph-cache misses and execution, and compare values/gradients in the two-device test. A fresh-process control that keeps the input device current passes forward and backward on both GPUs, including repeated cache hits. This is a visible multi-device failure with a direct caller workaround, rather than demonstrated silent corruption, hence nonblocking P1.

Evidence: Exact-head SM100/A100 run, PyTorch 2.13.0+cu130 and cuDNN 9.25.1: both existing L1 device tests failed (2 failed, 54 deselected). Independent analytic FP32 probes reproduced the same error for both ops; input-device controls passed 8 forward/gradient comparisons across the two devices and repeated calls. Attribution: Introduced by these new CUDA implementations; creating a device-pinned handle does not guard later native operations.

**[P2] Use the same epsilon before comparing RMSNorm timings** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/74f1c3a300715f4f97fbbf63ec54d3fd4cdffbf6/benchmark/norms/benchmark_torch_norm_ops.py#L46))

The cuDNN call defaults to eps=1e-5, while the PyTorch call leaves eps=None and uses its own version/dtype-dependent default. They therefore benchmark different functions. On the tested PyTorch 2.13 build, BF16 input filled with 1e-3 and unit weights produces 0.30078125 from this wrapper versus 0.9453125 from the benchmark's PyTorch expression; passing eps=1e-5 to PyTorch produces 0.30078125. Pass one explicit epsilon to both providers and add a correctness check outside the timed region.

Evidence: Independent numerical probe using the benchmark's two call signatures; the explicit-epsilon control matched. Performance measurements below used matched eps=1e-5. Attribution: Introduced in the new benchmark's provider comparison.

Validation: Built the Python extension from this exact head. On SM100, the complete norm-op L0/L1 file passed 54 tests and skipped the 2 multi-device tests; running those 2 on SM100+A100 reproduced the finding above. The 4 new framework-free import tests passed. Both ops passed independent fullgraph=True, dynamic=True compiled forward and backward checks for row counts 8, 17 and 8 again. CUDA Graph timing controls also checked poisoned outputs and changed-input replay. Matched-epsilon cuDNN/PyTorch/PyTorch/cuDNN spot check, BF16, hidden=4096, rows=32/1024, 68-SM SM100, cuDNN 9.25.1: RMSNorm CPU enqueue was 65-66 us versus 6.9 us; captured GPU time was 1.54/6.14 us versus 4.06/9.76-9.85 us. LayerNorm CPU enqueue was about 108-109 us versus 7.5-7.6 us; GPU time was 12.6/28.2 us versus 8.3/19.7-19.9 us. GPU timing captured 16 calls per graph to avoid host enqueue gaps. These are comparisons of the new optional APIs, not regressions of an existing FE route.

Limitations: No broad CI pipeline started while the focused two-device tests remain red. The usual OSS-only lane does not collect test/python/test_torch_norm_ops.py; follow-up CI should include the regular Python tests after the guard is addressed. Performance scope is two forward shapes on one SM100 GPU. The local backend is 9.25.1, not the author's 9.28 environment.

YangXu1990uiuc (commented):

Reviewed `93099d27e5ea` against base `ebe3bba998b8`.

Both findings from my previous review are resolved. The input-device guard now encloses the complete forward/backward execution body, including cache misses and graph execution; pinning only handle creation was insufficient. The benchmark now passes the same explicit epsilon to both providers and checks correctness before timing. No remaining findings from this follow-up; preserving the existing approval.

Validation: Exact-head SM100+A100 run with PyTorch 2.13/cuDNN 9.25.1: all 56 norm tests passed, zero skipped, including both formerly failing multi-device tests and their new backward/cache-hit checks. An independent analytic probe also passed eight alternating-device forward/gradient comparisons with the other device ambient. Matched prior/head/prior checks at BF16 hidden4096, rows32/1024: captured GPU times stayed within about1%; numerical, poisoned-output and changed-input replay checks passed. Public CPU timings were noisy. A same-process ABBA comparison of the identical execution body with/without the new guard measured median CPU increments of1.94us for LayerNorm and2.54us for RMSNorm. This small correctness cost is nonblocking; it is not an end-to-end model latency measurement. Both revised benchmark provider pairs passed their new correctness gate. Current-head Style, Docs and merge-metadata checks passed. No exact-head GPU pipeline existed at the final CI lookup; the completed focused review supports starting the regular Python-test target, which collects test_torch_norm_ops.py.

Limitations: Independent validation used cuDNN9.25.1 rather than the author's9.28 environment; broader architecture/backend CI is still needed. Native-extension reuse was checked against identical native sources.

YangXu1990uiuc:
I am requesting test CI for `93099d27e5ea` with the following standalone command: `@cudnn-ci-bot run python_tests`. For future updates, request the appropriate test target before review; a maintainer may need to trigger it.

YangXu1990uiuc:

PR #1131, opened 2026-09-18
Add optional max-logit output to Flex Attention

jiayus-nvidia:
## Before submitting

- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and my changes comply.

## Affected area

FE OSS kernels or CuTeDSL; Python API.

## Summary

Add `return_max_logit=True` to `flex_attn_func`, returning the maximum scaled, mask-visible attention score per query head across all batches and query/key positions. The FP32 `[Hq]` result is non-differentiable and can be returned alongside LSE. Support covers SM90 and the SM100/SM103 Q-stage 1, Q-stage 2, and D=256 forward paths.

Expose the same output through `FlexAttentionFwd(..., sample_max_logit=...)` and `execute(..., max_logit_tensor=...)`, with descriptor validation, compile-cache specialization, and output reset on the launch stream before reduction.

Also fix Q-stage 2 empty tiles reading stale TMEM: multiplying a previous NaN by zero left all-masked output rows as NaN. The empty-tile epilogue now writes literal zeros. Add numerical, autograd, stream/capture, output-reuse, and regression coverage, plus API documentation.

## Why

Callers can obtain maximum-logit statistics from the attention kernel's row maxima without recomputing the attention scores. Empty rows and reused output buffers need defined results, including during CUDA graph replay.

## Related issues

None.

- The new option defaults to `False`; existing return values remain unchanged. Enabling it returns `(out, max_logit)` or `(out, lse, max_logit)` when LSE is also requested.
- The statistic requires a non-negative `softmax_scale`. Fully masked heads return `-inf`; visible heads return zero at zero scale.
- Existing fixed-length, variable-length, GQA, dtype, and architecture restrictions remain in place. No new dependencies.
- Explicit execution uses caller-provided output storage. Initialization and atomic reduction are enabled only when the statistic is requested. Tracking the exact maximum disables the SM100 softmax rescaling threshold for that specialization, so enabling the option can affect performance.

## Testing

With imports verified to resolve to the feature worktree, on B200 with CUDA 13.2, cuDNN 9.22, and CuTe DSL 4.6.2:

```bash
cd test/python
python -m pytest -q fe_api/flex_attention/ -m 'L0 or L1 or L2' --maxfail=1 --disable-warnings
```

**56 passed, 4 skipped** (unsupported packed-GQA D=256 combinations). This includes FP64 maximum-logit references, output/backward comparisons, preallocated-output validation, non-default-stream execution, CUDA graph replay, and allocation/compile/synchronization guards.

The empty-TMEM regression test was also run against the original code and failed with NaNs in all 140,288 output elements before the fix.

SM90 compile-only validation passed 16 combinations: fixed/variable length, packed/unpacked GQA, D=64/256, and LSE enabled/disabled. H100 and SM103 execution were not available locally.

Pre-commit formatting and SPDX checks passed for all changed files. Subsequent amendments only remove documentation text and update the new helper's copyright/license header; the tested kernel bodies are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Flex Attention can now optionally return per-head maximum attention logits alongside outputs and log-sum-exp statistics.
  * Supports fixed- and variable-length inputs, grouped-query attention, masking, and preallocated CUDA buffers.
  * Fully masked heads return `-inf`; maximum-logit results are non-differentiable.
* **Documentation**
  * Added API guidance covering output combinations, scaling requirements, buffer usage, and edge cases.
* **Bug Fixes**
  * Prevented stale values from affecting results in empty output tiles.


jiayus-nvidia:
@cudnn-ci-bot run oss

YangXu1990uiuc (approved):

Reviewed `d22bb282aa7f` against base `d469317de89f`.

The optional-output API, per-head reduction and autograd/stream integration make sense. Independent SM90 and SM100 execution passed the submitted coverage and additional capture/GQA checks. Please address the two localized comments below: zero-scale max-logit handling in Q-stage 1, and a reproducible performance regression in the internally selected Q-stage 2 variant. Neither is an approval blocker under this review's scope; the default Q-stage 1 route retained its GPU performance.

**[P2] Preserve the visible row maximum before clearing invalid normalization state** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/d22bb282aa7fa8f49c7bf74b3ae441a84d20fd26/python/cudnn/flex_attention/kernels/sm100/fwd/forward_qstage1.py#L357))

The zero-scale exception here is too late: row_max0/1 were already replaced with -inf when their row sums were NaN. With BF16 Q/K/V filled with ones, B=1, Sq=65, Sk=257, Hq=4/Hkv=2, pack_gqa=False, endpoints=1 and softmax_scale=0, every head has one visible zero logit, but max_logit is [-inf, -inf, -inf, -inf]. This reproduces for qstage1_1cta/D64 and qstage1_2cta/D192; Q-stage 2 and D256 return the expected zeros. Preserve the raw masked maxima for this statistic independently of normalization validity, and add a one-visible-key zero-scale case. The existing tests can still obtain zero from another valid tile, hiding the lost contribution. The underlying zero-scale O/LSE failure also occurs on the baseline; the new issue is the advertised statistic inheriting it.

Evidence: Exact-head SM100 execution; an independent assertion expecting four zeros fails with four infinite mismatches. The baseline and head have the same pre-existing O/LSE behavior for this case, while only the new head exposes max_logit. Attribution: Introduced in the new Q-stage 1 statistic by consuming the normalization-sanitized row_max_combined.

**[P2] Retain the exact-zero fix without slowing nonempty Q-stage 2 tiles** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/d22bb282aa7fa8f49c7bf74b3ae441a84d20fd26/python/cudnn/flex_attention/kernels/sm100/fwd/forward.py#L1521))

This empty-tile change also affects feature-off, full-mask execution. On a 68-SM SM100 with DSL 4.8, BF16 D128, B=2, Hq=8/Hkv=2, pack_gqa=False and _fwd_variant='qstage2_1cta', captured GPU time increases from 16.41 to 18.87 us at Sq=Sk=512, 26.18 to 31.07 us at 1024, and 292.63 to 368.54 us at 4096 (+15%, +19%, +26%). Reversing run order reproduces the differences. A diagnostic copy that only restores the old TMEM load/multiply body returns to 16.41/292.64 us, isolating this lowering change; that is a performance control, not a safe fix for the NaN bug. Please investigate the generated code or separate the zero-fill epilogue so the normal path keeps its performance. This is nonblocking because the affected variant requires an internal override; the default qstage1_2cta path stays flat at the two checked lengths.

Evidence: Matched baseline fa7a433e and head d22bb282, identical native binding/dependencies, correctness checked before timing, 100 executions captured per graph. Baseline/head/head/baseline comparisons and isolated epilogue ablation reproduce the effect. Default D128 GPU time is 16.60 us at length 512 and about 339.4 us at 4096 on both revisions. Attribution: Introduced by the incidental empty-epilogue correctness fix, independently of enabling max_logit. Compiler/resource-level cause is not yet established.

Validation: Built the native extension from the exact head. With PyTorch 2.13.0+cu130, cuDNN 9.25.1 and DSL 4.8, the full Flex Attention L0/L1/L2 selection passed 56 tests with 4 unsupported-combination skips on SM100, and 46 tests with 14 architecture/variant skips on H100. No suite failures. 22 additional SM100/H100 checks passed with Hq=16/Hkv=1, B=17, negative masked maxima, empty heads, tail tiles and multiple scheduler waves. Both omitted and explicit streams left poisoned outputs unchanged during capture, then produced the expected O/max_logit after changed-input replay. Feature-off performance comparisons covered SM90 D64/D128/D256 and SM100 Q-stage 1, Q-stage 2 and D256. GPU differences outside the Q-stage 2 finding were below 0.4% in these checks. A separate warmed default-path CPU control, with post-compilation garbage collected before timing, measured execute medians of 65.4-66.0 us on baseline versus 66.5-66.7 us on head; allocating-wrapper medians overlapped. These are bounded spot checks, not an exhaustive performance claim. Verified full GitHub-head provenance for the owner's pipeline 68519146. Its SM90 and both SM100 OSS jobs passed; Style and Docs checks also passed on this head. Reused that run and did not start duplicate CI.

Limitations: The pipeline is not entirely green: the Rubin OSS job has 11 gated-attention device-declaration errors (cuda versus cuda:0), and guardwords rejects an inherited SUPPORT_MATRIX_TRACKER.md line. Those files are unchanged by this PR; no independent same-environment Rubin baseline control was run. Flex Attention does not advertise SM107 support. No independent SM103 execution. Performance measurements use one 68-SM SM100 and one H100 with DSL 4.8; they do not establish B200 or DSL 4.6.2 performance. Zero-scale O/LSE remains a separate pre-existing issue, as confirmed by the baseline control.

PR #1135, opened 2026-09-18
fix(dsa): synchronize SM100 dense indexer dK staging

C-TC:
The SM100 dense-indexer backward reducer lacks synchronization between the threads writing its dK shared-memory staging buffer and the thread issuing asynchronous bulk reductions. Add two warpgroup rendezvous to order buffer reuse and publication.

## Before submitting

- [x] I agree to license this contribution under the terms of LICENSE.txt.
- [x] I ran the repository pre-commit hooks and committed formatting changes.
- [x] I reviewed the applicable AGENTS.md Hard Rules.
- [ ] Required labels (maintainer permission needed): `cat-bugfix`, `op: DSA`, `area:sparse_attention`, `orig-external`.

## Affected area

FE OSS kernels / CuTeDSL: SM100 dense-indexer backward.

## Summary

- Add one dedicated 128-thread barrier in `dense_indexer_backward_sm100.py`: after the preceding bulk-read wait and after the staging stores/proxy fences. A per-thread wait or fence does not wait for other staging writers.
- Preserve single-thread DMA issuance. A full-CTA barrier would include threads that do not participate in the reducer.
- Extend the existing dense backward test module with ten synthetic cases: D64/D128 crossed with short, single-tile and multi-tile sequences. Alternate positive/negative Q; strictly negative QK requires exactly zero dQ/dK/dW.

## Related issues

Related to #1049, which fixed analogous SM90 staging. This change targets the separate SM100 dense reducer.

Six added runtime lines; GEMM arithmetic, dtypes, scaling, public interfaces and launch geometry are unchanged.

## Testing

GB200, CUDA 13.2, PyTorch 2.11.0a0, CuTe DSL 4.6.2. Compared against develop `ebe3bba998b8`, changing only this kernel for the numerical control:

- Unpatched staging regression: **10 passed** ordinarily; **6 failed / 4 passed** under Compute Sanitizer instrumentation. Failures occur at Q/K=(32,128), (33,129), (32,513), each at D64/D128.
- Patched full dense-backward test module, including reference and CUDA graph tests: **12 passed**.
- Patched staging regression under the same instrumentation: **10 passed**.
- Changed-file Black/SPDX hooks passed.

The unpatched instrumented failures are **numerical assertions**, not sanitizer-issued race reports; racecheck displays zero hazards in both arms. A single-KV-tile case fails too, so buffer publication matters independently of multi-tile reuse.

Reproduce from `test/python`:

```bash
CUDNN_TEST_NO_ISOLATION=1 compute-sanitizer \
  --tool racecheck --racecheck-report analysis --error-exitcode 99 \
  python -m pytest -o addopts='' -n 0 -v \
  fe_api/dsa/test_DSA_dense_indexer_backward.py -k staging_sync
python -m pytest -o addopts='' -n 0 -v \
  fe_api/dsa/test_DSA_dense_indexer_backward.py
```

CUDA-graph GEMM-only microbenchmark, H64/D128 and Q/K=(32,128), (512,2048), (2048,8192): no measured slowdown in matched A/B/A medians; differences are below 1%. Each arm uses warmup plus seven batches of 20 replays, including dK reset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization during dense indexer backward processing to ensure reliable gradient reductions.

* **Tests**
  * Added coverage for varied query lengths, key lengths, and head dimensions.
  * Verified finite gradients and correct zero-gradient behavior for negative queries.


vedaanta:
@cudnn-ci-bot run oss

YangXu1990uiuc (approved):

Reviewed `604f373ffa1a` against base `ebe3bba998b8`.

The two reducer-only rendezvous correctly order staging reuse and publication while preserving single-thread DMA issuance. The 128 participants follow the same loop bounds, and the new named barrier is separate from the existing compute/allocation barriers. No approval-blocking findings; retaining the positive-output test improvement already noted below.

- **[P2] Still open:** [Check the positive dK values, not just finiteness](https://github.com/NVIDIA/cudnn-frontend/pull/1135#discussion_r4048105772).

Validation: Independent SM100/CuTeDSL 4.7 validation: the complete dense-backward module passed all 12 cases, including reference and CUDA Graph tests. Under racecheck instrumentation the new staging tests passed 10/10; the same tests against exact base ebe3bba998b8 failed 6/10. Both runs reported zero sanitizer hazards: the red evidence is numerical assertion failure under instrumentation, not a racecheck-issued hazard report. Additional analytic checks covered all 10 new shapes with repeated positive/negative inputs. dQ/dK/dW passed, negative gradients were exactly zero, and positive dK matched an oracle accounting for BF16 operands and dS exactly. Please preserve a positive-output assertion in the committed test as well. Matched base/head/base timing, H64/D128 BF16, GEMM-only plus dK reset: CUDA Graph device time was about 14.98 -> 14.77 us at Q32/K128 and 380.05 -> 379.95 us at Q512/K2048. Warm CPU calls remained around 15 us and 19-21 us respectively. Correctness and graph replay were checked before timing; no material regression in these samples. Verified pipeline 68627204 belongs to this exact head. SM80, SM90 and both SM100 OSS shards passed; the remaining failures are the unchanged documentation-scan path and 11 Rubin gated-attention cuda-versus-cuda:0 declaration errors, outside this diff. Current Style and merge-metadata checks passed. Reused the existing CI run.

Limitations: Independent probes used one SM100 device rather than GB200 or Rubin. The timing sample covers two shapes and does not establish performance for every supported configuration.

Issue #186, opened 2025-12-03
The populate_cuda_graph creates an unnecessary CUDA graph


zcbenz:
In `populate_cuda_graph`, the cuDNN graph is first populated to a intermediate CUDA graph first, and then the intermediate CUDA graph is added as a child graph to the user's CUDA graph:

https://github.com/NVIDIA/cudnn-frontend/blob/be6c079be8aaffa0fc079fcf039887e637c289c7/include/cudnn_frontend/graph_interface.h#L588-L604

which I think is just adding overhead and complexity.

When using the `populate_cuda_graph` APIs, users are always using an intermediate CUDA graph to store the results from cuDNN, i.e.:

```c++
CudaGraph cuda_graph(device);
graph.populate_cuda_graph(..., cuda_graph);
cudaGraphNode_t node;
cudaGraphAddChildGraphNode(&node, root_cuda_graph, NULL, 0, cuda_graph));
```

I guess the current behavior of `populate_cuda_graph` assumed use case like below so it handles the work of creating an intermediate CUDA graph:

```c++
graph.populate_cuda_graph(..., root_cuda_graph);
```

However when updating the CUDA graph the assumption would break because you can't do things like below:

```c++
graph.update_cuda_graph(..., root_cuda_graph);
```

So to use the native CUDA graph APIs the users always need to create an intermediate CUDA graph themselves, and the final graph would look like this:

Root graph -> Intermediate child graph by user -> Intermediate child graph by cuDNN -> Actual kernel

and the intermediate child graph by cuDNN would be redundant.

Refs https://github.com/ml-explore/mlx/pull/2857.

egilliam-nv:
Thanks for reaching out!

As currently designed, cuDNN's native CUDA Graph API expects to populate an empty CUDA graph which the caller will typically add as a child node of a larger graph. In fact, since you're using the cuDNN frontend, there will be two levels deep of child graphs:
- the backend (BE) is given an empty graph, and populates it...
- which the frontend (FE) will copy into a child node of its own graph (this is the code snippet you pointed out above), and sometimes surround it with additional kernels/memsets/etc. of its own...
- which the user (framework) will typically copy into a child node of *its* own graph.

[...]

PR #1140, opened 2026-09-18
sdpa: CUDA graph topology evidence and an opt-in direct populate/update path

0z5a:
**Incremental review:** [Files changed](https://github.com/NVIDIA/cudnn-frontend/pull/1140/files) against `develop`.

## Problem

`populate_cuda_graph` wraps the backend's graph in an extra frontend child node, and a caller that embeds the result in its own parent graph adds another level. The backend's native CUDA graph API does not need either: it accepts an empty caller graph and populates it in place, provided update targets the original graph or its clone with the original execution plan.

## What this measures first

On an L20 the candidate plan reports `SUPPORTS_CUDA_GRAPH_NATIVE_API` (`prepare_native_plan built=true native_api=true`), so the question was measurable on Ada rather than only on Hopper. One `Graph` object is used for every path, so every path is the same plan (`eng7_k24=39`, workspace 0):

| path | top-level nodes | recursive | kernel nodes | child_graph nodes | depth |
|---|---|---|---|---|---|
| P1 `populate_cuda_graph` | 1 | 2 | 1 | 1 | 1 |
| P1 inner backend graph | 1 | 1 | 1 | 0 | 0 |
| P2 caller parent wrapping P1 | 1 | 3 | 1 | 2 | 2 |
| P3 direct populate | 1 | 1 | 1 | 0 | 0 |

The kernel-node count never changes; only the wrapping does. P3 is item-for-item identical to P1's inner backend graph, which is also the evidence that the two paths are the same plan rather than two different ones. Nothing here hardcodes how many roots or kernels a backend graph "should" have.

## Eligibility

A plan qualifies when `cached_workspace_modifications` contains no entry with `operation_type` in {0,1}; populate and update share the rule. Measured, and this is the point: SDPA backward has 264320 B of workspace and **zero** frontend nodes, so it is eligible, while SDPA forward with an alibi tensor has 16 B of workspace and one memcpy node, so it is not. `workspace_size != 0` does not imply a frontend node.

## Dispatch

I could not find a safe way to infer the path from topology -- a backend graph may itself have a single root, or a root that is a child graph node -- so this PR does **not** ship an implicit dispatch and does **not** add a global `cudaGraph_t -> mode` map. The direct path is an explicit pair, `populate_cuda_graph_direct()` / `update_cuda_graph_direct()`, which refuse an ineligible plan with `INVALID_VALUE` (measured code 14). The default wrapped path is unchanged, so existing callers are unaffected.

## Cost, host and GPU separated

Median of medians over three independent processes; host timings from `steady_clock`, GPU timings from CUDA events. The columns are never merged.

| phase | P0 | P1 | P2 | P3 | P3 vs P1 |
|---|---|---|---|---|---|
| host populate cycle (create+populate+destroy) | - | 4.526 us | - | 3.171 us | -29.9% |
| host instantiate | - | 5.817 us | 6.452 us | 5.062 us | -13.0% |
| host update | - | 2.894 us | 2.897 us | 2.247 us | -22.4% |
| host exec update | - | 0.228 us (ChildGraphNodeSetParams) | 0.280 us | 0.220 us (cudaGraphExecUpdate) | -3.5% |
| host exec update via re-instantiate | - | 5.681 us | - | 4.886 us | -14.0% |
| host CPU launch submission | 6.731 us (execute) | 3.654 us | 3.676 us | 3.656 us | ~0 |
| GPU replay, single | 175.616 us | 172.032 us | 172.032 us | 172.032 us | 0% |
| GPU replay, batched | - | 169.656 us | 169.606 us | 169.779 us | noise |

P0 versus P1 is the existing wrapper's cost: per-call host submission 6.731 -> 3.654 us (-45.7%) and GPU 175.616 -> 172.032 us (-2.0%), against a one-time 10.343 us setup, so the crossover is about 3.4 replays. P2 versus P1: instantiate +10.9%, exec update +22.8%, replay unchanged.

The honest reading is that the direct path removes host build and update overhead only. **GPU replay is 172.032 us on every path**, so I am not claiming any throughput or tokens/s improvement, and a caller that builds once and replays many times will see nothing.

## Correctness and lifecycle

G01-G08 and G10 pass with data checks rather than return codes alone:

- G02 asserts the new output received the new data while the old buffer still held its sentinel.
- G04 clone-then-update on both the wrapped and the direct path.
- G05 propagation through `cudaGraphExecChildGraphNodeSetParams` to an instantiated parent.
- G06 the auxiliary memcpy precedes the backend node and its destination equals its host source after replay.
- G07 empty and foreign target graphs plus a bad variant pack return explicit errors, leak no handle or graph, and leave the caller's graph usable.
- G08 at 10/50/100/200 create-update-instantiate-destroy loops: zero checksum failures, zero host-RSS delta, zero device-free delta, and per-loop time flat (73.7 -> 66.1 us).
- G09 runs for real under `CUDA_VISIBLE_DEVICES=0,1`; the single-device run skips it with "needs two visible CUDA devices; only 1 visible".
- G11 the legacy Slice alias (cuDNN < 9.22, or a frontend compiled against it) reaches all four CUDA graph paths: a variant pack that names only the slice SOURCE and the graph output populates, updates and replays correctly, and a variant pack that names neither end of the alias still fails closed. On the pre-fix tree the same case fails with `Uid 22 does not exist in variant pack.`

Results: 14 passed / 1 skipped with 7229 assertions on one device, and the multi-device case passes on two. The upstream `[cudagraph]` suite still passes 3/3 (8245 assertions) on both the base and the patched tree.

The sample also builds with CUDA 13: the six dependency queries go through version-guarded shims (the 4-argument form with `cudaGraphEdgeData *` under `CUDART_VERSION >= 13000`, `nullptr` edge data below), so the default `samples` target compiles with `nvcc` 13.2.51 instead of failing at six calls, as the review found on the unmodified head.

## Not in scope

No general DAG flattening, no CUDA graph scheduler, no cross-device graph merging, no backend kernel changes, and no change to the default path. Multi-threaded sharing of one `Graph` object and the full `deserialize()` + direct chain are untested.

Related to #186.

## Summary by CodeRabbit

- **New Features**
  - Added direct CUDA graph population and update support for eligible graphs.
  - Added eligibility checks for graphs without cached memory-copy or memory-set nodes.
  - Added cached variant replacement support across CUDA graph population and update paths.
  - Added validation for missing replacement sources during graph updates.
  - Added a sample demonstrating CUDA graph topology usage.
  - Added CUDA-version compatibility for retrieving graph node dependencies.
  - Preserved pointer rebinding and pass-by-value support.
  - Ineligible graphs return an error directing callers to wrapped APIs.


YangXu1990uiuc (commented):

Reviewed `c85e494b7738` against base `ebe3bba998b8`.

Thank you for your contribution! The explicit populate/update pair and auxiliary-node eligibility rule make sense, and the measured benefit is appropriately framed as host-side graph construction/update savings. I am withholding approval for the CUDA 13 sample-build regression below. I also reproduced the existing slice-binding comment on cuDNN 9.21.1; that narrower, visibly reported error is nonblocking. The runtime results below required a probe-only compatibility adapter, so they do not clear the build issue.

**[P1 · blocks approval] Use CUDA-version-compatible graph dependency queries** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/c85e494b7738ce89e7be3717830d84f257f0c1f3/samples/cpp/misc/cudagraph_topology.cpp#L288))

CUDA 13 removed the three-argument forms of cudaGraphNodeGetDependentNodes and cudaGraphNodeGetDependencies: their signatures now require an edgeData argument before the count. The six new calls at lines 288, 291, 304, 307, 1717 and 1720 fail to compile with CUDA 13.2 (size_t* cannot convert to cudaGraphEdgeData*). This file is added unconditionally to samples, and CUDNN_FRONTEND_BUILD_SAMPLES defaults to ON, so this blocks the normal C++ build on supported CUDA 13 toolkits. Please use the existing detail::cuda_graph_node_get_dependent_nodes compatibility shim and an equivalent version guard/helper for GetDependencies, supplying nullptr edge data on CUDA 13+. Keep the older-toolkit signature in the other branch. This is approval-blocking because it breaks the default build, independently of which tests are selected.

Evidence: Exact-head sample compilation with GCC, C++17, CUDA 13.2 and cuDNN 9.25.1 fails at all six calls. The existing fp16 forward/backward and cudagraphs translation units compile with the same flags. A probe-only adapter adding the missing nullptr arguments makes the new translation unit compile; the reviewed source was not edited. Attribution: Introduced by this PR's new sample and unconditional CMake source entry.

- **[P2] Still open:** [Apply slice pointer replacements before gathering graph variant-pack UIDs](https://github.com/NVIDIA/cudnn-frontend/pull/1140#discussion_r4049994519).

Validation: With only a probe-side CUDA 13 signature adapter: all 14 added cudagraph_186 cases, the existing matmul CUDA graph case, and two independent slice/serialization probes passed (17 cases, 19544 assertions, no skips). Main execution used SM100; the two-device case also used H100. cuDNN 9.25.1, CUDA 13.2 headers; runtime reported 13.4. Independent serialization probe verified eligible matmul deserialize -> direct populate -> replay, and preserved ALiBi ineligibility after deserialize. The separate legacy slice probe passed 4102 assertions after explicitly demonstrating the expected missing-UID failures and the binding workaround. Same-process, same-plan FP16 batched matmul spot check (B=8,M=N=K=1024): wrapped/direct populate+create+destroy medians 3.68/2.78 us; update 2.49/2.28 us; ABBA launch submission 1.54/1.55 us; batched GPU replay 51.212/51.201 us. This supports host setup savings with no observed replay regression in this workload. Timings used the optimized build, before the separate diagnostic legacy rebuild. Exact head/base and current reviews rechecked. Style workflow is awaiting fork authorization, not a completed test failure. Automatic CI assistance is deferred until the build blocker is fixed.

Limitations: The unmodified new sample does not compile with CUDA 13; runtime success with the isolated adapter is not an unmodified-head build pass. No full CI, Windows, CUDA 12, or shared-Graph multithreading validation was performed. The main workspace-address test selects a zero-workspace matmul plan on this machine; the separate eligibility case observes nonzero SDPA workspace but does not establish direct-update correctness for that case. Performance evidence is a bounded wrapped-versus-direct comparison, not a full workload sweep.

YangXu1990uiuc (approved):

Reviewed `908adb0ff8c4` against base `667fe4ce8ce4`.

Thank you for your contribution! The CUDA 13 build blocker and the legacy slice-binding issue are resolved. The current source compiles without the previous probe-only adapter, and an independent cuDNN 9.21 test verifies fresh source-only maps rebind both wrapped and direct graphs to new input/output buffers. I am comfortable approving this follow-up. Two nonblocking items remain: the cross-major dynamic-loader diagnostic issue already noted by CodeRabbit, and a legacy test that skips before exercising the fix in the environment below.

- **[P2] Still open:** [Select the dependency-query signature from the loaded CUDA runtime](https://github.com/NVIDIA/cudnn-frontend/pull/1140#discussion_r4051871409).

**[P2] Use a legacy-supported graph for the G11 binding regression** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/908adb0ff8c41e173d8daf4ea8718bf99e156725/samples/cpp/misc/cudagraph_topology.cpp#L2714))

On cuDNN 9.21.1 with SM100, the rank-2 slice-plus-pointwise graph has no supported engine, so G11 skips before reaching any replacement binding assertion. On modern cuDNN it intentionally skips because Slice is native. Please use a graph known to build in the intended legacy test lane, such as the rank-3 slice-to-batched-matmul case, and verify that the selected regression actually executes there. A fresh-map update with changed input/output pointers also guards against retaining an alias created during populate. This is nonblocking because the independent working graph verifies the implementation fix.

Evidence: Current-head G11 on cuDNN 9.21.1 reports 'slice plan could not be built' with NOT_SUPPORTED_LAYOUT. In the same process/environment, an independent rank-3 Slice -> matmul graph runs both wrapped and direct populate/update successfully: initial O is exactly 8, rebinding a fresh source-only map produces exactly 16 in the new output, and the old output remains the -123 sentinel. Attribution: Coverage gap in the regression added by this follow-up; functional binding fix independently passes.

Validation: Unmodified current-head sample/probe plus existing forward, backward and cudagraph translation units compile and link with GCC C++17, CUDA 13.2 headers, cuDNN 9.25.1 headers and warnings as errors. No CUDA-signature compatibility adapter was used. The prior reviewed head failed this build at all six raw query calls. Before publication, develop advanced to 667fe4ce8ce4 by merging the Python-only #1146; the intervening diff changes no C++ sources or build files, so this evidence remains applicable. cuDNN 9.25.1 on SM100, with H100 also visible for the multi-device case: CUDA-graph lifecycle and independent rebinding selection reports 16 passed, 1 intentionally skipped, 40,017 assertions passed. The skip is G11's native-Slice guard. cuDNN 9.21.1 on SM100: the independent fresh-map slice-rebinding case passes both wrapped/direct paths (24,584 assertions); G11 skips at plan construction as described above. Earlier base-side missing-UID failures and the explicit-alias workaround were independently recorded in the first review. Current same-plan SM100 component timings: wrapped/direct graph population cycles 3.700/2.850 us, update 2.510/2.270 us; CPU launch submission 1.550/1.550 us and batched GPU replay 51.187/51.201 us. These retain the host setup/update benefit, with no observed replay regression in this workload. The previous reviewed direct-path medians were 2.78 us populate, 2.28 us update and 51.201 us replay; cross-run values are a stability check, not a controlled speedup claim.

Limitations: The build used CUDA 13.2 headers; an older-toolkit full build was not repeated. The separate CUDA 12.4 dynamic-runtime probe establishes only the stated query behavior. Validation is focused local C++/GPU testing, not a completed full current-head CI run. Style workflow 35415999998 is awaiting maintainer approval.

YangXu1990uiuc:
The enabled [Style run for `908adb0ff8c4`](https://github.com/NVIDIA/cudnn-frontend/actions/runs/35415999998) failed only because clang-format modifies `include/cudnn_frontend_shim.h` and `samples/cpp/misc/cudagraph_topology.cpp`. Please run `pre-commit run clang-format --files include/cudnn_frontend_shim.h samples/cpp/misc/cudagraph_topology.cpp` and commit the resulting formatting changes. The functional review approval remains in place.

0z5a:
Review follow-up on both findings, on head `908adb0f`. Nothing else in the PR changed.

## [P1 · blocks approval] Use CUDA-version-compatible graph dependency queries — fixed (`7ad62aab`)

The six calls now go through the frontend's own shims:

- `detail::cuda_graph_node_get_dependent_nodes(...)` (already present), and
- `detail::cuda_graph_node_get_dependencies(...)`, which `cudnn_frontend_shim.h` now defines for both worlds: the four-argument form taking `cudaGraphEdgeData *` under `#if CUDART_VERSION >= 13000`, and the three-argument form for older toolkits, passing `nullptr` edge data.

Evidence from the project's own build on the L20 host (GCC 11.4, C++17, `nvcc` 13.2.51, cuDNN 9.19 headers/runtime, driver 595.91.07), `-DCUDNN_FRONTEND_BUILD_SAMPLES=ON`:

| tree | `samples` target |
|---|---|
| pre-fix head `c85e494b` | 8 compile errors in `samples/cpp/misc/cudagraph_topology.cpp`, all six calls (`size_t *` cannot convert to `cudaGraphEdgeData *`) |
| this head `908adb0f` | `BUILD_EXIT=0`, `[100%] Built target samples` |

Runtime on that build: `bin/samples "[cudagraph_186]"` → 14 passed, 1 skipped, 7227 assertions (the skip is the two-device case; it passes under `CUDA_VISIBLE_DEVICES=6,7`), `bin/samples "[cudagraph]"` → 3/3 (8245 assertions), `bin/samples "[graph]"` → 68 passed / 42 skipped (9273 assertions).

## [P2] Apply slice pointer replacements before gathering graph variant-pack UIDs — fixed (`908adb0f`)

`make_variant_pack_replacements()` now runs after the workspace binding in all four CUDA graph entry points (wrapped populate, wrapped update, direct populate, direct update), which is what the ordinary execute path already does through its variant-pack template. The path is live on this build: the samples link the cuDNN 9.19 headers, so `SliceNode::collect_variant_pack_replacements_node` fills the map rather than short-circuiting.

New case `186 G11` drives a `Slice -> pointwise` graph through all four paths with a variant pack that names only the slice SOURCE and the graph output, never the derived destination:

| tree | wrapped populate |
|---|---|
| pre-fix (`graph_interface.h` reverted, patch in `evidence/186/prefix-replacement-patch.diff`) | `Uid 22 does not exist in variant pack.` — the case fails |
| this head | wrapped populate OK, wrapped update OK, direct populate OK, direct update OK |

Each replay is checked against `2 * X[1, :]` after poisoning the output buffer, so the alias is verified numerically rather than by a return code. A variant pack that names neither end of the alias still fails closed, and the case asserts that error too. A variant pack that spells the alias out by hand stays accepted.

The default wrapped path, the eligibility rule for the direct pair, and the GPU replay time (172.032 us on every path) are unchanged; this is not a throughput change.

0z5a:
Both non-blocking findings and the style check are addressed on this head (`dcc958f2`).

## Style: clang-format

`clang-format` 21.1.6 (the version the pre-commit hook pins) reformatted `include/cudnn_frontend_shim.h` and `samples/cpp/misc/cudagraph_topology.cpp`; `clang-format --dry-run -Werror` is clean on both now. Formatting only — the samples target still builds and `[cudagraph_186]` still passes on both toolchains (14 passed / 1 skipped / 7229 assertions on CUDA 13.2 and on CUDA 12.8, upstream `[cudagraph]` 3/3).

## [P2] Use a legacy-supported graph for the G11 binding regression

G11 now drives a rank-3 `Slice -> batched matmul` graph (`X[b, s, d] -> Y[b, 1, d] @ W[b, d, n]`) instead of the rank-2 `Slice -> pointwise` one, so the legacy lane builds it rather than skipping at plan construction. It also does what you asked for the retention question: after populate it re-binds the **same** graph to **fresh source and output buffers** and checks the new result, on the wrapped path and on the direct path, plus that the first output buffer still holds its own value — an alias remembered from populate could not pass that.

Verified against the pre-fix header: the same probe fails with `Uid 22 does not exist in variant pack.` before the fix and passes after (wrapped populate/update, direct populate/update, and the negative case that names neither end of the alias, which still fails closed).

## [P2] Select the dependency-query signature from the loaded CUDA runtime

Fixed in `dcc958f2`. You are right that this is reachable: `load_cudart_so()` prefers `libcudart.so.12` when both runtimes are present, so a CUDA 13-header build can load the CUDA 12 runtime, where the base symbol still has three parameters and the edge-data form is the `_v2` symbol. The four-parameter shim was calling the three-parameter function, so the count pointer came from the slot holding the edge-data argument and the three-parameter wrappers passed a null count pointer.

Dispatch is now on the loaded runtime: base symbol for a CUDA 13 runtime, `_v2` from 12.3 on, and for the three-parameter entry points the three-parameter base symbol directly on any CUDA 12 runtime. `get_cuda_runtime_version()` resolves `cudaRuntimeGetVersion` (same ABI on both) and caches it. The static build's code path is unchanged. I applied the same dispatch to the `cudaGraphNodeGetDependentNodes` twins, which had the identical latent issue in upstream code — the probe below shows they fail the same way without it.

Probe, built with the CUDA 13.2 headers and linked against the CUDA 12.8 runtime (`PROBE_RUNTIME cudart=12080 compiled=13020`), graph built through the linked runtime and queried through the frontend shims:

| query | before | after |
|---|---|---|
| `cuda_graph_node_get_dependencies` (count, then node) | `cudaErrorInvalidValue`, count 0 | count 1, node = the one the graph was built with |
| `cuda_graph_node_get_dependent_nodes` (count, then node) | `cudaErrorInvalidValue`, count 0 | count 1, node correct |
| `..._dependencies_v2` / `..._dependent_nodes_v2` (edge-data form) | `cudaErrorInvalidValue`, count 0 | count 1, node correct |

All seven queries fail before (`failures=7`) and pass after (`failures=0`); the probe source and both binaries are in `evidence/186/abi-probe/`.

YangXu1990uiuc (commented):

Reviewed `dcc958f2793d` against base `667fe4ce8ce4`.

Thank you for your contribution! Both remaining items from my previous review are resolved: the dynamic-loader queries now use the loaded runtime's ABI, and G11 actually exercises legacy Slice bindings, including fresh-buffer updates. I found no new actionable regression in this incremental review. My existing approval remains in place.

Validation: CUDA 13.2-header dynamic-loader probe: CUDA 12.4 and CUDA 13.2 runtimes each pass 12 checks covering dependency/dependent counts, returned node identity, edge-data entry points, and repeated calls. The prior reviewed head independently failed the CUDA 12.4 count queries with cudaErrorInvalidValue. cuDNN 9.21.1 on SM100: the revised G11 and the independent fresh-map rebinding probe both pass, with 24,598 assertions and no skips. G11 verifies wrapped/direct outputs change from exactly 64 to 128 after fresh source/output bindings, preserves the old output, and rejects a missing slice source. cuDNN 9.25.1 on SM100 with H100 visible for the two-device case: focused CUDA-graph lifecycle selection passes 16 cases and 40,017 assertions; G11 alone intentionally skips because Slice is native. Exact-head C++17 translation units compile and link with CUDA 13.2 and the existing harness warning settings. clang-format 21.1.6 --dry-run -Werror passes both changed files. Same-plan current-head SM100 spot check retains the expected scope: wrapped/direct CPU launch submission 1.54/1.55 us and batched GPU replay 51.196/51.206 us. Direct populate/update medians are 2.94/2.32 us versus wrapped 3.84/2.54 us. These are component measurements, consistent with the previous review; they do not establish an end-to-end throughput gain.

Limitations: Focused local validation only; a full current-head CI run and a full build with older CUDA headers were not repeated. Style workflow 35423823386 is awaiting maintainer authorization, rather than reporting a formatting failure.

YangXu1990uiuc:
The [Style run for `dcc958f2793d`](https://github.com/NVIDIA/cudnn-frontend/actions/runs/35423823386) is now green after maintainer authorization. This also clears the formatting failure from the previous revision. The verified fixes and retained approval are recorded in my latest review.

vedaanta:
@egilliam-nv please help review/approve/merge this. 🙇

include/cudnn_frontend_shim.h
```diff
@@ -177,6 +177,24 @@ get_cuda_symbol(CudaLibrary library, const char *function_name) {
     return symbol;
 }
 
+// Version of the CUDA runtime this process actually loaded, as major * 1000 +
```
egilliam-nv: Can this function be moved further down, above cuda_graph_create, so it can take advantage of the NV_FE_CALL_TO_CUDA macro?

include/cudnn_frontend_shim.h
```diff
@@ -303,12 +321,28 @@ cuda_graph_node_get_dependent_nodes_v2(cudaGraphNode_t node,
...
+    // three parameters and the edge-data form is the _v2 one, so resolving by
+    // name alone would pass edgeData where the count is expected.
+    using get_dependent_nodes_t = cudaError_t (*)(cudaGraphNode_t, cudaGraphNode_t *, cudaGraphEdgeData *, size_t *);
+    if (get_cuda_runtime_version() >= 13000) {
+        return reinterpret_cast<get_dependent_nodes_t>(
+            get_cuda_symbol(CudaLibrary::CUDART, "cudaGraphNodeGetDependentNodes"))(
+            node, pDependentNodes, edgeData, pNumDependentNodes);
+    }
+    return reinterpret_cast<get_dependent_nodes_t>(get_cuda_symbol(
+        CudaLibrary::CUDART, "cudaGraphNodeGetDependentNodes_v2"))(node, pDependentNodes, edgeData, pNumDependentNodes);
```
egilliam-nv: Likewise here and everywhere you do a `return reinterpret_cast<...>`, can we use NV_FE_CALL_TO_CUDA?

0z5a:
Both points are addressed in `ba510802`.

`get_cuda_runtime_version` now sits with the other CUDA shims, just above `cuda_graph_create`, and resolves the version through a `cuda_runtime_get_version` shim that uses `NV_FE_CALL_TO_CUDA`; the caching accessor keeps its previous behaviour. The four version-dispatched dependency queries call through the same macro as well, so the local function-pointer typedefs and the raw `reinterpret_cast` calls are gone.

Verified on an RTX 5090 host (nvcc 13.0, cuDNN 9.24, GCC 11.4): the `samples` and `tests` targets build with the project's `-Wall -Wextra -Wpedantic -Werror`, `[cudagraph_186]` passes (13 passed, 2 skipped, 7215 assertions), `[cudagraph]` passes (8245 assertions), the full cpp suite passes (35 passed, 3 skipped, 632 assertions), and a small probe that builds a two-node graph and queries it through all four shims passes both with and without `NV_CUDNN_FRONTEND_USE_DYNAMIC_LOADING`. `clang-format` is clean on the file.

`get_backend_version` has the same hand-rolled pattern but predates this PR, so I left it alone to keep the diff scoped - happy to convert it here or in a follow-up, whichever you prefer.

YangXu1990uiuc (commented):

Reviewed `ba510802f9c4` against base `4d4e428bd747`.

Thank you for your contribution! The requested shim-macro refactor preserves the CUDA 12/13 query ABI in independent checks. I found one integration issue with newly merged #1195: the direct graph path bypasses the execution-plan retention now provided by the wrapped path. Please address this high-risk P1 before merging. Existing approvals are preserved; this is a comment, not a request for changes.

**[P1 · high risk] Retain the execution plan on direct-populated CUDA graphs** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/ba510802f9c432a27229e44b636d8500924d599c/include/cudnn_frontend/graph_interface.h#L1016))

Please carry #1195's plan retention into the two new direct entry points. That merged fix attaches the selected plan to the caller's CUDA graph because cuDNN can release runtime-compiled kernel code with the plan. A clean merge of this head onto current develop retains the fix in populate_cuda_graph/update_cuda_graph, but the direct methods still return immediately after the backend call. Consequently an application that instantiates/clones a direct graph and releases its FE Graph can leave a live graphExec without its plan/code owner. Call retain_on_cuda_graph(cudnn_cuda_graph) on the selected plan after successful direct population/update, as the wrapped methods now do, and extend the outliving-Graph regression to the direct path. This is high risk because normal graph-owner teardown can invalidate code that a later replay needs.

Evidence: Independent SM100/cuDNN9.26 probe on the unmodified clean merge preview: populate, clone, instantiate, destroy both source CUDA graphs, then destroy FE Graph while graphExec remains alive. A weak reference to the backend plan descriptor remains live for wrapped population, expires for direct population, and remains live for a direct-plus-explicit-retention control. All three execute correctly before FE teardown. No kernel was launched after the descriptor expired, so this demonstrates the ownership gap without claiming a reproduced stale-code launch failure. Attribution: Integration gap in the PR's new direct path against newly merged #1195; not caused by this follow-up's shim-macro substitutions.

Validation: Reviewed the complete incremental delta from dcc958f2: only cudnn_frontend_shim.h changes, replacing manual casts with the existing macro and moving the cached runtime-version accessor alongside the CUDA shims. Exact-head dynamic-loader probe built with CUDA13.2 headers: CUDA12.9 and CUDA13.2 runtimes each pass all12 dependency/dependent-node checks, including counts, node identity, edge-data variants and repeated calls. Created a separate clean merge preview against develop4d4e428b without changing the PR. The statically linked ownership probe compiles against that preview and distinguishes wrapped, direct and explicit-retention control as described in the finding.

Limitations: No unsafe replay after plan destruction was attempted. The finding is based on observed descriptor ownership plus the runtime-code lifetime contract fixed by #1195, rather than a claimed new numerical corruption or CUDA fault. No new broad CI was requested while the lifetime issue remains. Current-head Style is awaiting maintainer authorization; that is not a formatting failure. Previous lifecycle/performance measurements remain historical evidence for the unchanged direct implementation.

Please make sure the high-risk P1 findings above are addressed before merging. Existing approvals are left unchanged.

YangXu1990uiuc (commented):

Reviewed `283a7025eb32` against base `79ed0b2cae8b`.

Thank you for your contribution! This head has exactly the same complete Git tree as ba510802. The previously reported high-risk P1 remains after a clean integration with current develop: the direct populate/update methods still bypass execution-plan retention. Please address it before merging. Existing approvals are preserved.

- **[P1 · high risk] Still open:** [Retain the execution plan on direct-populated CUDA graphs](https://github.com/NVIDIA/cudnn-frontend/pull/1140#pullrequestreview-5286422989).

Validation: Verified complete tree identity with ba510802: 5cbae4944fa7f979ad06e6c98d5595a912f5a7d7. Constructed an isolated, unpublished clean merge with develop 79ed0b2c. The graph interface, execution-plan and CUDA shim headers are byte-identical to the previously probed integration; the wrapped/direct ownership difference remains unchanged. Reused the earlier SM100/cuDNN 9.26 ownership evidence rather than repeating identical GPU work.

Limitations: No fresh GPU launch or performance run. The existing evidence demonstrates descriptor ownership, not an unsafe replay after releasing kernel code. CI remains deferred while this high-risk finding is unresolved.

Please make sure the high-risk P1 findings above are addressed before merging. Existing approvals are left unchanged.

Issue #601, opened 2026-08-14
frost(sdpa): collapse the mask Capabilities onto the band vocabulary (causal/swa/bottom_right flags are uniformly True)


vedaanta:
Follow-up to the diagonal-band mask model (#485) and BR+SWA (#584): the mask-related `Capabilities` flags in `sdpa/fwd/engines.py` (and bwd) predate the band model and several no longer discriminate between rows.

**Current state** (post-#584, 6 fwd + 2 bwd rows):

| flag | rows differing | verdict |
|---|---|---|
| `causal` | True on all 8 | dead — delete |
| `swa` | True on all 8 | dead — delete |
| `bottom_right` | True on all 8 | dead — delete |
| `right_band_widening` | 7/8 (SM120-fp8 lacks it) | keep, rename into the band vocabulary |
| `bottom_right_with_swa` | 5/6 (MXFP8 stopgap only, see #600) | delete once #600 lands |
| `bottom_right_padded_seq_q` | SM120 only | keep — real kernel gap |

**Proposed end-state**: capabilities mirror the band facts instead of enumerating mask spellings —

- `window_left: bool` — serves a left bound (re-added only when some row actually differs; today it wouldn't exist at all);
- `window_right: "none" | "causal" | "any"` (or two booleans) — replaces `causal` + `right_band_widening`;
- `bottom_right: bool` — anchoring;
- combination flags exist **only** while a kernel has a genuine interaction bug (the `thd_bottom_right` / `thd_stats` precedent: both were deleted the moment all rows agreed).

A capability that is `True` on every row is dead weight: its `mismatch()` gate never fires and it costs every future row a line of boilerplate. Deleting the three uniformly-True flags and folding the rest onto the band vocabulary makes "can this engine serve this band?" answerable by comparing two records with the same field names (facts already speak `window_left` / `right_bound` / `bottom_right`).

Blocked on #600 only for the `bottom_right_with_swa` deletion; the rest is independent.

0z5a:
Census, canonical model and a behaviour-preserving migration, with an explicit compatibility gate. Base `ebe3bba9`.

[...]

Issue #600, opened 2026-08-14
frost(sdpa): MXFP8 executor SF-layout mismatch keeps the MXFP8 row at ~0% mhas routing


vedaanta:
The SM100 MXFP8 executor's scale-factor view disagrees with the cuDNN-convention F8_128x4 layout for some mhas-shaped graphs, which keeps the MXFP8 row at effectively **zero routing** on test_mhas_v2 (322/323 mxfp8-fwd graphs native in the latest census) and is the reason `Capabilities.bottom_right_with_swa` stays `False` on that row (PR #584): the one graph BR+SWA admits there fails at execute.

**Symptom** (test_mhas_v2 `test_sdpa_mxfp8_bwd_L0[test118]`'s fwd graph, when routed):
```
ValueError: MXFP8 SF size mismatch: got 6656 bytes, expected 1664*1*13*512=...
```
`api_dsl._reshape_sf` validates the SF tensor byte count against `b*h*n_tiles*sf_smem_size` and the incoming tensor's geometry disagrees wholesale (its leading dim isn't batch-like), i.e. the producer's F8_128x4 packing and the executor's expected per-tile `[B, H, n_tiles, sf_smem_size]` view use different padding/rounding rules for this shape family. The same mismatch was seen earlier from `benchmark/sdpa_benchmark_training`'s `create_scale_factor_tensor_mxfp8`.

**Work**:
- Root-cause the layout disagreement (executor `_reshape_sf` / `_execute_mxfp8` vs the cuDNN SF-tensor convention — dims rowwise `(B, H, S_pad, ceil(ceil(D/32)/4)*4)`, columnwise `(B, H, S/32-pad, D)`, 512-byte 128×4 atoms) for non-tile-aligned S.
- Fix the executor view (or decline precisely on the true trigger).
- Then flip `bottom_right_with_swa=True` on the MXFP8 row — and delete the flag entirely (see the mask-capability cleanup issue), since every row will agree.

Found while landing BR+SWA (#584); prior sighting in the OSS benchmark work.

vedaanta:
Closing — all three work items are done on develop:

[...]

PR #1138, opened 2026-09-18
frost(sdpa): SM89 (Ada / L20) d64 dense forward row

0z5a:
## What

`sdpa_fwd_prefill_sm89` -- the FROST d64 dense SDPA forward row for Ada (L20, cc 8.9), proposed as a narrow opt-in row. Discussion in #1137.

## Why a new row instead of widening the SM80 gate

The SM80 row is pinned to cc 8.0 exactly and that pin is load-bearing: it has to carry the d=256 flavor, whose pinned point needs 128 KiB of opt-in SMEM (sQ_buf 64 KiB + sK_buf 64 KiB), which fits A100's 164 KiB but not Ada's 99 KiB (101376 B on an L20, queried from the device). The d=64 gptoss point allocates 32 KiB, so the same `sm80/prefill_f16.py` template runs on Ada unchanged -- only the device gate and the advertised box differ.

Files:

- `python/cudnn/sdpa/fwd/engines.py` -- the row: `sm_lo=sm_hi=89`, `d_shapes={(64, 64)}`, `d_pad_multiple=1`, FP16/BF16, dense `bshd`/`dense_flex`.
- `python/cudnn/sdpa/fwd/api_dsl.py` -- the SM80-family adapter gains a device-family split plus a plan-time SMEM gate checked against the device's real per-block opt-in limit, so an unvalidated geometry is declined with a diagnosable error instead of failing at launch.
- `python/cudnn/sdpa/fwd/config_sm80.py` -- the SM89 flavor view and the SMEM arithmetic helper.
- `python/cudnn/engines/manifest.py` -- slot 17, appended, never reordered.
- `python/cudnn/sdpa/frost/SUPPORT_MATRIX_TRACKER.md` -- the SM89 table, per Rule S2 in `python/cudnn/sdpa/AGENTS.md`.
- `test/python/sdpa/frost/test_sm89_d64.py` -- 39 cases.

The SM80 row keeps `sm_lo=sm_hi=80` and its full feature set, and a regression test pins that.

## Mask claim, measured rather than assumed

[...]

PR #1141, opened 2026-09-18
frost(sdpa): canonical band-mask capability model

0z5a:
**Incremental review:** [Files changed](https://github.com/NVIDIA/cudnn-frontend/pull/1141/files) against `develop`.

## Problem

The four mask capability flags (`causal`, `bottom_right`, `swa`, `right_band_widening`) can only *widen* a row: each defaults to False and a True is a yes. A restricted claim such as "this kernel serves unmasked graphs only" is not expressible at all, even though `sdpa_bwd_sm100_mxfp8` already is one, and an incoming SM89 d64 row needs to be another.

## Change

`cudnn/sdpa/band.py` adds two objects and keeps them deliberately distinct:

- `BandFacts` is the band a **graph** asks for: left bound, right mode (unbounded / causal / finite-right), the diagonal anchor, plus the one real combination rule (a bottom-right anchor needs a diagonal to anchor to). `BandFacts.covers_kv_tail` is the band x geometry rule that was previously re-derived inline at each call site.
- `BandSupport` is the band set an **engine** serves. Every axis refuses an empty set, so the default `BandSupport()` is the NARROWEST claim (unbounded / no left bound / top-left, i.e. `unmasked_only()`); the convenience constructors name the shapes rows actually take. A row may declare `band` ALONE: only the legacy flags that are actually SET constrain it, each of them must be inside the declared band, and an all-default legacy block is the spelling of "nothing declared". A causal-only set -- which the flags cannot spell, because `from_legacy_flags` always serves the unbounded case -- is therefore expressible.

`Capabilities` (forward and backward) gains an appended `band` field, resolved once in `__post_init__` by the single normalization layer `BandSupport.from_legacy_flags`. The probes decide with the model, not the flags, and they ask every axis in BOTH directions: the right-mode and left-mode checks are membership tests on the mode the graph requests (unbounded and no-left-bound included), and the anchor check is symmetric across both anchors. Absence of a graph flag therefore cannot bypass a restricted claim, which the legacy one-way checks could not prevent.

## Compatibility

The field is appended, so the positional prefix contract holds, and the append-only test was extended rather than deleted. Declaring both spellings is accepted only while they state the same claim; a contradiction raises at construction rather than one spelling silently winning by field order. The flags stay the declaration spelling of every pre-existing row, so no existing row had to change.

## Behaviour preservation

Base versus patch, two separate processes against two checkouts, an independent comparer, and nothing filtered:

| check | result |
|---|---|
| engines / facts compared | 13 (9 forward + 4 backward, enumerated dynamically) / 754 |
| support-boolean mismatches | 0 |
| reject-category mismatches | 0 |
| reject-text changes (allowlist empty) | 0 |
| non-mask capability domain changes | 0 |
| capability diffs | 13 added `band` fields (the representation change) |

The census builds, per row, a legal facts sample that satisfies every non-mask constraint, so an arch or dtype rejection cannot mask a mask-behaviour change; all 13 rows report `baseline_ok=true`. Harness credibility was checked as well: an A/A run on the unmodified tree produces 0 diffs, and an injected-diff negative control makes the comparer exit 1.

One finding worth stating: no shipped row declines causal today, and the only genuinely restricted row is `sdpa_bwd_sm100_mxfp8` -- which is exactly why the model has to be able to express "unmasked only" for rows that do not exist yet.

## Tests

- new `test/python/sdpa/frost/test_band_capabilities.py`: 25 collected / 25 passed (including band-only construction and the per-axis symmetry cases)
- existing probe, append-only and heuristics: 184 collected (182 passed, 2 deselected as non-L0)
- consumer suites (import boundaries, split_kv heuristics, knob vocabulary): 149 collected (147 passed, 2 skipped as "needs a pre-Rubin SM100-line GPU")
- combined run: 348 executed, 346 passed, 0 failed, 2 skipped

The matrix covers all three sequence geometries, both anchors, default and extended right bounds, left bound presence and length, padding and length boundaries, positive and negative layout cases, and the four construction spellings including conflicting declarations.

## Cost, reported as a cost

Capability-table import and build: 79.3 / 82.3 ms base versus 78.6 / 80.1 ms patched, no stable difference. `mismatch()` per call over the 754-fact sweep: 8.438 / 9.568 us base versus 9.657 / 9.891 us patched (ABBA, re-measured on the review-follow-up head), so the band model adds on the order of **+1 us per call** rather than a fixed percentage, stable in both arms. Decomposed: the old four boolean reads cost 0.16 us, the new path costs ~1.0 us to normalize the facts and ~1.5 us to build and query the model. Across 13 rows that is roughly 22 us per graph, negligible beside lowering and compilation. This is a representation cost, not a speedup; the value is that a restricted claim is expressible and the KV-tail rule has one definition.

## Not in scope

Kernel mask algorithms, new kernel capabilities, shape and layout limits, user-facing graph-API mask semantics, and any cross-architecture numerical claim -- this is plan-time behaviour only, audited on an L20. It does not redo the `bottom_right_with_swa` cleanup from #600, which is already closed.

Related to #601.

## Summary by CodeRabbit

- **New Features**
  - Added canonical SDPA mask-band support for causal, sliding-window, right-bounded, unmasked, and diagonal-anchor configurations.
  - Added consistent band capability reporting across forward and backward engine support checks.

- **Bug Fixes**
  - Legacy mask settings are now normalized consistently with explicit capability declarations.
  - Improved validation for bounds, KV-tail coverage, layouts, and bottom-right anchors.
  - Unsupported configurations now receive standardized rejection reasons.

- **Documentation**
  - Expanded support documentation with canonical band models, restricted-row behavior, and capability tracking details.


YangXu1990uiuc (commented):

Reviewed `2e02ce18dba5` against base `5567e2733408`.

Thank you for your contribution! Centralizing the band geometry makes sense, and an independent base/head comparison preserved the decisions and rejection messages of all 13 existing rows over 2496 facts. However, this diff also adds an Ada engine, despite the description saying there are no new kernel capabilities. That addition is incomplete: its manifest slot and adapter/configuration implementation are missing. On an actual SM89 GPU the added suites report 27 failures and 30 passes. Please complete that integration or split the Ada addition into a follow-up; I am withholding approval until it is usable. The explicit-band declaration issue below is nonblocking.

**[P1 · blocks approval] Register the new SM89 engine in the manifest** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/2e02ce18dba5a782ed7c8e8bc5c1983308187ff5/python/cudnn/sdpa/fwd/engines.py#L2194))

Appending _sm89_spec() here does not expose it to dispatch. FrostSdpaFwdEngines only instantiates specs whose names occur in the manifest-provided ids, and engines/manifest.py has no sdpa_fwd_prefill_sm89 slot. Consequently, even with FROST enabled on cc 8.9, every new numerical/selected-plan test fails before launching the Ada kernel because no such plan is offered. Append a new stable, opt-in manifest slot without reusing retired IDs, and retain the explicit route assertions. This blocks approval because the entire newly advertised engine is unreachable and its added test suite is red, rather than an isolated unsupported edge case.

Evidence: On L40S, test_sm89_row_is_registered_and_opt_in fails and all 21 added numerical/selected-plan cases fail with 'no plan for engine sdpa_fwd_prefill_sm89'. Source tracing confirms the engine factory silently filters the unregistered spec. Attribution: Introduced by this PR's new SM89 engine row and tests; the manifest was not updated.

- **[P1 · blocks approval] Still open:** [Complete the Ada adapter and configuration implementation](https://github.com/NVIDIA/cudnn-frontend/pull/1141#discussion_r4050027274).

**[P2] Allow an explicit band declaration without duplicating legacy flags** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/2e02ce18dba5a782ed7c8e8bc5c1983308187ff5/python/cudnn/sdpa/fwd/engines.py#L427))

Both forward and backward __post_init__ require every explicit BandSupport to equal the legacy-derived set, even when all legacy flags were left at their defaults. For example, band=BandSupport.causal_and_unmasked() raises with no legacy flags supplied, and a causal-only set cannot be expressed at all because from_legacy_flags always includes unbounded. This contradicts the advertised independent/restricted-set spelling; the error's advice to clear flags does not help. Distinguish an omitted legacy declaration from a conflicting one and add band-only construction tests. When enabling genuinely restricted sets, also make the matcher check the unbounded/no-left-bound/top-left cases, so absence of a graph flag cannot bypass a restricted claim.

Evidence: Independent construction probes against this head reproduce the ValueError for causal_and_unmasked and causal-only declarations in both Capabilities classes. Existing engine rows retain their legacy behavior in the separate parity probe. Attribution: Introduced by the new explicit-band normalization contract in both capability classes.

Validation: Exact-head test_band_capabilities.py plus test_sm89_d64.py on L40S (SM89), cuDNN 9.25.1 and CuTe DSL 4.7, with -m 'L0 or L1': 30 passed, 27 failed, no skips. All 23 band-model tests passed; the failures are in Ada registration, adapter/configuration and route selection, before kernel execution. Existing graph-analyzer suite: 137 passed. Independent comparison against base ebe3bba998b8: 13 existing forward/backward rows, 2496 fact combinations, zero differences in acceptance or rejection text; each row's non-mask baseline was separately required to be accepted. Repeated CPU matcher microprobe measured base medians 5.75 and 5.94 us/call versus head 6.95 and 6.96 us/call. This is about 1.0-1.2 us extra in support matching, not an execute-path or GPU timing result; I am not treating that bounded overhead as an approval blocker. Publication recheck: develop advanced to 5567e2733408 through #1132 while the PR head remained unchanged. Its prepared dense-launch changes do not supply the missing Ada registration/configuration or alter the checked capability declarations. Automatic CI is deferred because the added engine still requires substantive integration fixes.

Limitations: No Ada numerical or GPU-performance claim is possible because the new engine never reaches execution. Existing-engine GPU kernels were not benchmarked; their source is unchanged and the checked change is in capability matching. The independent matcher comparison covers the stated finite set, not every possible graph or future restricted BandSupport declaration.

YangXu1990uiuc (approved):

Reviewed `1d2b445691a3` against base `b03c3f3321f5`.

Thank you for your contribution! Removing the incomplete Ada integration resolves both previous approval-blocking findings and brings the diff back to the stated band-model scope. The current-head focused suite passes, and all 13 existing engine rows preserve the independently checked acceptance decisions and rejection text. The previously reported P2 remains: an explicit band-only declaration still has to duplicate the legacy flags, and a causal-only claim cannot be constructed. Please address that construction contract before relying on it for new restricted rows; it does not block this behavior-preserving refactor.

- **[P2] Still open:** [Allow an explicit band declaration without duplicating legacy flags](https://github.com/NVIDIA/cudnn-frontend/pull/1141#pullrequestreview-5252466883).

Validation: Exact-head test_band_capabilities.py and test_sdpa_graph_analyzer.py: 160 passed, no skips or deselections, using PyTorch 2.13 / cuDNN 9.25.1 on SM89. These are capability/analyzer tests, not Ada kernel numerical validation. Reran the independent 13-row, 2496-fact probe against the current head and compared every acceptance/rejection string with the retained base and prior-head results: zero differences; every non-mask row baseline remains accepted. The separate explicit-construction negative probes still reproduce the P2. Current matcher median is 6.960 us/call versus 6.952/6.957 us for the prior reviewed head. The previously measured original-base medians were 5.754/5.936 us, so the roughly 1 us plan-time representation cost remains; this revision adds no observable additional matcher cost.

Limitations: This incremental revision changes no GPU kernels or execution paths. Numerical and GPU performance claims are not inferred from the capability tests. The current-head Style workflow is awaiting maintainer approval and no current-head FROST pipeline is recorded yet; neither is counted as passing.

YangXu1990uiuc (commented):

Reviewed `a80e1c68fc0f` against base `667fe4ce8ce4`.

Thank you for your contribution! The remaining explicit-band declaration issue is resolved: band-only causal claims now construct successfully, conflicting set legacy flags are rejected, and the matcher checks the unbounded/no-left-bound/top-left cases against restricted claims. No remaining correctness findings from my review. The existing approval stays in place.

Validation: Current-head band-capability and graph-analyzer suites: 162 passed, no skips or deselections, on SM89 with PyTorch 2.13 / cuDNN 9.25.1. Independent existing-row parity probe: 13 rows and 2496 facts retain exactly the prior acceptance/rejection strings. The four forward/backward explicit-construction probes that previously raised now succeed. An additional independent oracle checked all 1512 combinations of nonempty right/left/anchor claim sets and graph requests across forward/backward matchers. Expected results were derived directly from set membership and diagonal coherence, without calling the model's decision helpers; all passed. Matcher CPU median was 6.864 us/call versus 6.960 us in the previous reviewed head, with no observed additional plan-time overhead. GPU kernels and execution paths are unchanged, so this is not a GPU performance claim.

Limitations: These are focused plan-time behavior checks, not a full GPU numerical sweep. The latest-head Style workflow is awaiting maintainer approval.

YangXu1990uiuc:
The pending fork [Style workflow passed on `a80e1c68fc0f`](https://github.com/NVIDIA/cudnn-frontend/actions/runs/35416445934). For future updates, `pre-commit run --all-files` runs the same checks locally; a maintainer can enable a fork workflow when GitHub requests approval.

0z5a:
Review follow-up on all findings, on head `a80e1c68`.

## [P1 · blocks approval] Register the new SM89 engine in the manifest / [P1 · blocks approval] Complete the Ada adapter — the contaminated head is gone

This branch no longer contains an Ada engine, a manifest slot, adapter changes or SM89 tests. It is the band model alone: `python/cudnn/sdpa/band.py`, the two `Capabilities` classes, the two probes, the analyzer view, the tracker/README text and the phase-C test file — 8 files, no new kernel capability, exactly as the description says. Every file this branch touches is byte-identical to the content you reviewed minus the Ada additions.

The Ada row now lives in its own PR (#1138), stacked on this one, where its manifest slot and its adapter/configuration completion sit next to the row they belong to. That PR is reviewable as one commit on top of these three.

## [P2] Allow an explicit band declaration without duplicating legacy flags — fixed

`__post_init__` (forward and backward) now constrains an explicit `band` only by the legacy flags that are actually SET: each set flag must name an axis the declared band serves. The all-default legacy block is the spelling of "nothing declared", so it leaves an explicit claim alone.

- `Capabilities(..., band=BandSupport.causal_and_unmasked())` constructs with no flags (it used to raise).
- A causal-only set is expressible — `BandSupport(right=frozenset({RIGHT_CAUSAL}))` — which the flags cannot spell at all, because `from_legacy_flags` always serves the unbounded case.
- A set flag OUTSIDE the declared band still raises, naming the flag and the claim; the message no longer advises "clear the flags", since clearing them is now a legitimate state.
- `test_declaring_flags_and_band_together_is_explicit` covers the band-only construction, the causal-only claim and the still-refused conflict; `test_keyword_construction_and_dataclasses_replace` now distinguishes CLEARING a flag (accepted, the claim is kept) from WIDENING past the claim (refused).

## [P2] Make the matcher check the unbounded / no-left-bound / top-left cases — fixed

Both probes now ask every band axis in BOTH directions: the right-mode check is `serves_right_mode(band_facts.right_mode)` (unbounded included) and the left check is `serves_left_mode(band_facts.left_mode)`, with the anchor check symmetric across both anchors. Before this, a row that serves only the causal diagonal still served unmasked graphs by omission — the escape your finding describes.

`test_a_band_only_claim_is_enforced_on_every_axis` pins all three axes as (claim, band it serves, band it must decline, exact reason) and cross-checks `BandSupport.decline()` on the same facts.

## Inline comments

- **`band.py` "empty claim" wording**: the `BandSupport` docstring now states that no axis accepts an empty set, so the default is the NARROWEST claim (unbounded / no left bound / top-left, i.e. `unmasked_only()`), not "no band".
- **`engines.py` rationale vs docstring**: the mirror labels now exist for the axes that previously had no decline wording (`LABEL_RIGHT_UNBOUNDED`, `LABEL_LEFT_NONE`, `LABEL_ANCHOR_TOP_LEFT`), and `anchor_reason(noun, anchor=...)` words both anchors, so nothing has to be re-derived per probe.
- **Test/matcher vocabulary**: the stale "unmasked only, which a future SM89 row needs" sentence in `SUPPORT_MATRIX_TRACKER.md` is replaced by the two restricted claims that actually exist (the MXFP8 backward's causal-only claim, the SM89 forward's top-left-anchor-only claim), and the tracker now says the matcher asks every axis both ways.

## Behaviour preservation, re-run on this head

Two processes, two checkouts, the independent comparer, nothing filtered:

| check | result |
|---|---|
| engines / facts compared | 13 / 754 |
| support-boolean mismatches | 0 |
| reject-category mismatches | 0 |
| reject-text changes (allowlist empty) | 0 |
| `pass` | true |

The symmetric checks cannot fire for any pre-existing row, because every existing row's claim contains unbounded / no-left-bound / top-left; the census confirms that rather than assuming it.

## Tests and cost on the L20

- `test_band_capabilities.py` 25 passed (23 before, +2 for band-only construction and the symmetric matcher); 162 passed together with the graph-analyzer suite; 147 passed / 2 skipped for the consumer suites named in the PR description.
- Host-side cost, re-measured ABBA with one fresh process per arm: the band block costs ~1.5 us against ~0.2 us for the four raw boolean reads it replaces, and the full `mismatch()` sweep is 8.4–9.6 us/call on the base versus 9.7–9.9 us/call here. That is the same representation cost as before, not a new one.

vedaanta:
Thanks @0z5a !
Please do help resolve conflicts.

Adding @RomanAnders90 for review and approval.

YangXu1990uiuc (commented):

Reviewed `2bbec155618c` against base `79ed0b2cae8b`.

Thank you for your contribution! This head has the same complete Git tree as the previously reviewed a29e6911. I additionally checked its integration with current develop because the forward capabilities have since gained paged-FP8/ragged-decode support: the merged contract suite and current-base decision comparison pass. No new findings; the existing approval stands. The linked nonblocking tracker wording correction remains.

- **[P3] Still open:** [Drop the stale statement that every FROST engine is opt-in](https://github.com/NVIDIA/cudnn-frontend/pull/1141#pullrequestreview-5289013090).

Validation: Complete tree identity with a29e6911: ccfc742c2976368e81c2bd6e62bf670e8fed56af. Constructed an isolated, unpublished integration tree of this head with develop 79ed0b2c. Automatic merge was clean. The band-capability and graph-analyzer suites passed all 175 cases, including the current paged-FP8 capability checks; no skips or deselections. Compared the independent mask-fact matrix on current develop and the integration tree: all 13 rows and 2,496 facts retain identical acceptance/rejection strings. Verified each loaded source path and native-binding compatibility.

Limitations: These new checks cover host-side eligibility and integration, not GPU numerical execution or performance. The previous planning-overhead measurements were not repeated.

YangXu1990uiuc:
Style run **35962415853** for reviewed head `2bbec155618c` has passed after maintainer authorization. No additional GPU pipeline was started for this history-only update; the current-develop integration checks are recorded in the review above.

Issue #752, opened 2026-08-26
SDPA: CP-accurate numerics mode (chunk_size param, bitwise match to TE CP merge)


vedaanta:
## Feature request

Add a "CP-accurate numerics" mode to cuDNN SDPA: a `chunk_size` parameter that makes single-GPU SDPA output bitwise-equivalent to Transformer Engine's Context Parallelism (CP) ring-attention merge, rather than only "within tolerance."

## Motivation

Customers train with TE CP (multi-GPU, ring attention with cross-rank partial-result merging) and run inference/generation without CP on a single GPU. The two paths currently diverge numerically because CP's merge order (per-chunk softmax LSE correction + output rescaling) differs from a plain single-shot SDPA. This drift between train-time and gen-time numerics is a problem for long-context post-training at scale.

The ask (from the customer, NVBugs 5958792, comment from Zhiyu Li): even on a single GPU, SDPA should be able to force CP-style partial correction on outputs using a user-provided chunk size, and match TE's CP merge bitwise — not just approximately.

## Proposed design

1. cuDNN SDPA gains a `chunk_size` parameter. When set, the kernel processes K in `chunk_size`-tile groups, stores per-tile `(O, LSE)` in fp32, and applies TE's exact merge pattern at the end:
   - Phase 1: incremental LSE correction — `max + log(1 + exp(min - max))`
   - Phase 2: batched output correction — `out = sum_i O_i * exp(LSE_i - LSE_final)`
2. TE's CP code paths use the same `chunk_size`, so the cross-rank merge matches the single-GPU emulation bitwise.

Reference merge implementation: TE `transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py` (`flash_attn_fwd_softmax_lse_correction`, `flash_attn_fwd_out_correction[_init]`, `flash_attn_fwd_second_half_*` for the load-balanced causal split).

## Scope / open questions

- Needs fp32 partials in both fwd and bwd (per customer, comment #1) — bwd path is not yet modeled.
- Causal load-balanced CP splits K into first/second-half chunks per rank (`[r, N-1-r]`); a faithful single-GPU emulation needs to reproduce that visit order, not just contiguous chunking.
- Verification should use bitwise (or near-bitwise, fp32-limited) comparison against an SDPA-isolated oracle, not a loose end-to-end tolerance test.
[...]

PR #1142, opened 2026-09-18
test(sdpa): CP numerics reference, TE merge oracle and CP1/CP2/CP4 replay

0z5a:
**Incremental review:** [Files changed](https://github.com/NVIDIA/cudnn-frontend/pull/1142/files) against `develop`.

## What this adds

A numerical reference for the CP partial/merge path that does not depend on the code under test, plus layered CP1/CP2/CP4 replays.

`test/python/sdpa/cp_numerics/`:

- `reference_math.py` -- independent FP64 oracle. The monolithic attention is the expected value and the chunked partials are the inputs, so the closed-form merge is only a cross-check. Empty-chunk and fully-masked-row semantics are explicit (`L = -inf`, output zero).
- `te_adapter.py` -- the merge oracle bound to a fixed TransformerEngine revision (`ae34b34f`, `context_parallel.py` sha256 `37b389ce...`), transcribing the five CP helpers and the ring schedule, with source-anchor verification and a separately labelled self-written emulator lane so "my emulator agrees with itself" is never presented as fidelity.
- `trace_schema.py` -- the trace record fields and seven invariants (coverage without duplication, unique partial slots, no order change from receive order, global token order restored, no padding read as valid KV).
- `make_fixtures.py` plus nine fixtures under `fixtures/`, each with a header carrying the full precision contract (storage dtype, accumulator dtype, O dtype, LSE base, scale, mask, layout, head mapping, TE revision, fuser mode) and an explicit P -> schedule mapping.
- `run_cp_reference.py` -- the torchrun entry, `--mode frozen-partials|actual-attention`, binding the device from `LOCAL_RANK`, validating GPU count against world size, doing a pre-collective consistency check on every rank, and always destroying the process group.
- `test_cp_numerics_reference.py` -- 49 passed / 0 failed (the one checkout-gated case runs when `TE_CP_REF_REPO` is set).

## Measured

CP1, CP2 and CP4 frozen-partial replay all really ran on 1, 2 and 4 L20s, with each rank recording its device UUID. Against the FP64 oracle: max abs 2.95e-08 / 6.07e-08 / 7.18e-08, max abs over max|ref| <= 4.28e-08, zero NaN and zero Inf on both sides. Comparing integer bit patterns of the same dtype -- never `torch.equal` -- the gather against the single-process replay gives **0 mismatches out of 32768**, self-repeat gives 0, and the global token order is restored in every case.

Bitwise against FP64 stays **not passed, by design**: the in-tree bound is precision-derived (`32*eps(dtype)*max|ref| = 6.39e-06`) and no atol was invented. The reported max ULP is 5488 (CP2) and 9584 (CP4), but the distribution is p50 = 1, p90 = 4-5, p99 = 33-44 -- the maximum is an artefact of measuring ULP at the spacing of a near-zero output -- and abs/rel/ULP are reported raw rather than folded into one number.

## Three findings worth carrying into the design

1. **The fuser mode changes the bits.** CP2 eager and `--compiled-helpers` produce an identical error envelope (max abs 6.074739997141698e-08 in both) but different digests, and a CPU-inductor A/B differs in 2530 of 32768 elements. Fidelity claims are only meaningful with the fuser mode pinned, which is why the runner refuses without it.
2. **TE's helper returns NaN for a fully empty row** (`max = min = -inf` leads to `log1p(exp(nan))`), while the FP64 reference's documented policy is `L = -inf, O = 0`. A single empty chunk is absorbed cleanly. It is asserted as a behavioural difference and not silently rewritten.
3. **TE's CP accumulator dtype is the output dtype, not FP32,** on the non-FP8 path; the `out in torch.float32` comment only holds on the FP8-dequantised branch. Recorded in the fixture header and the trace contract.

## R2 status: executed, with the backend verified

The `actual-attention` lane computes the per-step partials with the pinned TransformerEngine fused attention and merges them, and it records which backend TE's own dispatcher selected for every step. Each step asks for a mask type the fused backend supports (`causal` for the diagonal section, `no_mask` for the two triangles); `arbitrary` is not used, because at the pinned revision it disables FusedAttention and FlashAttention alike. `R2_EXECUTED` requires every step to have run fused with no unfused and no unrecorded step, and the process exits `EXIT_R2_UNVERIFIED` otherwise.

TransformerEngine is built from the pinned revision (`2.21.0.dev0+ae34b34f`, installed `context_parallel.py` sha256 `37b389ce...`, matching the recorded pin), so the pin check binds to the module that actually runs. On 2 and 4 L20s, `--te-compute-dtype float16`:

| lane | fixture | ranks | status | fused steps | fused sub-backend | merged O max abs | merged LSE max abs |
|---|---|---|---|---|---|---|---|
| actual-attention | cp_p2_f32_s128_causal | 2 | **R2_EXECUTED** | 2/2 | 1 | 4.82e-04 | 4.68e-07 |
| actual-attention | cp_p4_f32_s128_causal | 4 | **R2_EXECUTED** | 4/4 | 1 | 4.82e-04 | 4.08e-07 |
| actual-attention (bfloat16) | cp_p2_f32_s128_causal | 2 | **R2_EXECUTED** | 2/2 | 1 | 3.83e-03 | 4.68e-07 |
| frozen-partials | cp_p2_f32_s128_causal | 2 | OK | - | - | 2.95e-08 vs FP64 | - |
| frozen-partials | cp_p4_f32_s128_causal | 4 | OK | - | - | 2.95e-08 vs FP64 | - |

The merged O error is the storage precision of the fused lane: TE's fused attention does not accept fp32 (fp32 selects the unfused backend), so an fp32 partial comparison is not available on this route and is not claimed.

One boundary remains and is documented in the module: the public TE forward returns only O at this revision, so each step's LSE still comes from the fixture and the merge is labelled `te_o_with_fixture_lse`. The merged O is TE end to end; an LSE *bookkeeping* error would not show up here.

## Merge cost, same bytes

The ratio is robust across three independent passes; the absolute throughput is not quotable, because host load average was 600-860 with other tenants on GPUs 1-3 and a first sweep on a different physical card came out about 4x faster.

| measurement | config | TE-ordered merge | naive sequential merge | ratio |
|---|---|---|---|---|
| merge-only | P=1 S=512 | 0.0665 ms | 0.5294 ms | 7.8-8.1x |
| merge-only | P=2 S=8192 | 0.1915 ms | 0.5622 ms | 2.9x |
| merge-only | P=4 S=8192 | 0.3277-0.4076 ms | 0.5683-0.7894 ms | 1.3-2.4x |
| end-to-end runner | CP1 / CP2 / CP4 | 2.083 / 2.170 / 2.561 s | - | CP4 is not faster than CP2 |
| compiled helpers | CP2 | 8.66 ms median | - | ~35x the end-to-end cost |

## Not in scope

No native FP32 partial production path, no TE bitwise parity claim, no NCCL ring tuning, no training-framework integration, and no backward kernels. The scope stays at the reference and validation subtask.

Related to #752.

## Summary by CodeRabbit

* **New Features**
  * Added context-parallel attention numerical validation with FP64 reference calculations and Transformer Engine-compatible merge checks.
  * Added support for causal and noncausal attention, multiple precisions, parallelism levels, empty chunks, and ragged sequences.
  * Added machine-readable trace validation, invariant checks, output reassembly, distributed reports, and optional runtime execution checks.
  * Added compute-dtype selection and fused-backend verification.
* **Tests**
  * Added coverage for accuracy, masking, schedules, precision, determinism, fixture contracts, and fail-closed execution requirements.


YangXu1990uiuc (commented):

Reviewed `07b18c931cc1` against base `13d61c87bce0`.

Thank you for your contribution! The monolithic FP64 expectation and separate merge path are a useful foundation, and the five transcribed merge helper bodies match the pinned TE source. I am withholding approval for the R2 provenance gaps below: the current code can attribute results to a fused backend and pinned installation that it has not actually exercised. Please address these before merging. I also reproduced the existing optional-import failure and reduced-precision metadata mismatch.

**[P1 · blocks approval] Verify the backend before reporting R2 as fused attention** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/07b18c931cc11c60e6e278973c077fd9182c4307/test/python/sdpa/cp_numerics/actual_attention.py#L228))

Every step selects `attn_mask_type="arbitrary"`. At the pinned TE revision, [the backend selector](https://github.com/NVIDIA/TransformerEngine/blob/ae34b34f18d85cc62d90ca35cf6efe84c5a45a7e/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1399-L1409) explicitly disables both FusedAttention and FlashAttention for this mask. Thus this path runs UnfusedDotProductAttention (or fails if that fallback is disabled), while the result notes describe pinned fused-attention partials. `--compiled-helpers` changes the merge helpers, not this choice. Use supported mask/dtype combinations and assert/record the selected backend, or label this as an unfused lane and leave fused R2 unverified. For the stock trace masks, the global-position predicate already matches the per-section causal/no-mask rule; I checked all 53 trace entries. This blocks approval because a successful run silently supplies evidence for a kernel it did not execute.

Evidence: Followed compute_partials -> _PinnedAttention.run -> pinned DotProductAttention backend dispatch. Executing the exact upstream arbitrary-mask branch with both backends enabled disables both. Source/CPU route proof, not a GPU TE rerun. Attribution: Introduced by this PR's actual-attention lane.

- **[P1 · blocks approval] Still open:** [Bind the pin check to the imported TE installation](https://github.com/NVIDIA/cudnn-frontend/pull/1142#discussion_r4050019948).

- **[P2] Still open:** [Keep TE optional for frozen-partial and refusal paths](https://github.com/NVIDIA/cudnn-frontend/pull/1142#discussion_r4050150092).

- **[P2] Still open:** [Record the executed accumulator dtype in reduced-precision fixtures](https://github.com/NVIDIA/cudnn-frontend/pull/1142#discussion_r4050019920).

Validation: Current-head reference suite with the normal repository conftest: 44 passed, 1 skipped, 2 failed; the two failures are the optional-TE imports above. The isolated CPU run had the same outcome. The skipped test requires a separate TE checkout. All five transcribed helper bodies match the pinned upstream AST; the pinned context_parallel.py raw SHA256 matches the recorded digest. The mocked installed-package mismatch is nevertheless accepted by the current pin check. Checked the pinned TE mask-dispatch branch and all 53 committed TE trace masks. Independently confirmed the two reduced-precision accumulator metadata mismatches.

Limitations: No independent multi-GPU or actual TE kernel replay, and no performance claim from these CPU/source checks. This PR changes test/reference tooling only; no production CPU/GPU hot-path benchmark was needed. CI was not triggered while the R2 path still needs correction. After updating, please request a maintainer-triggered Python test run on the new head.

YangXu1990uiuc (approved):

Reviewed `597f39098dc9` against base `667fe4ce8ce4`.

Thank you for your contribution! The previous approval blockers are addressed: the pin check is bound to the imported CP source, the lane requests supported TE masks and records backend selection, and frozen/refusal paths no longer import optional TE eagerly. I am approving this follow-up with the localized verifier issues below remaining nonblocking. This approval does not certify an actual fused-TE CP2/CP4 run on this revision; please refresh those records after the recorder fixes.

- **[P2] Still open:** [Handle cached backend selection and version-valued FlashAttention metadata](https://github.com/NVIDIA/cudnn-frontend/pull/1142#discussion_r4052121458).

**[P2] Return the unverified exit code when backend verification fails** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/597f39098dc947c370740c08dd82d5e3c6ece1b8/test/python/sdpa/cp_numerics/run_cp_reference.py#L627))

The new backend guard can return result.status == R2_UNVERIFIED, but this path still returns EXIT_OK unconditionally. The JSON says unverified while torchrun/shell automation receives success. I passed an explicitly mocked unverified result through the real run_actual_attention function: it wrote R2_UNVERIFIED and returned 0. Return EXIT_R2_UNVERIFIED for that status, consistent with the pin-refusal branch, and add a behavioral check of both the report and exit code. This is nonblocking because the report itself retains the correct unverified status.

Evidence: Current-head runner control-flow probe produced exit=0/status=R2_UNVERIFIED; missing-checkout refusal correctly produced exit=2/status=R2_UNVERIFIED. Attribution: Newly exposed by this follow-up adding a backend-derived unverified result to a path that previously always reported execution.

- **[P2] Still open:** [The reduced-precision accumulator metadata issue remains](https://github.com/NVIDIA/cudnn-frontend/pull/1142#discussion_r4050019920).

Validation: Normal exact-head reference suite: 46 passed, 1 skip requiring a separate TE checkout. Both previously failing optional-TE import tests now pass. Independent provenance controls reject an unrelated imported CP file despite a matching --te-repo, accept the byte-identical imported pin, reject an unrecorded digest, and write the expected exit-2 unverified report for a missing checkout. The pinned TE selector/cache probe confirms the existing warm-cache finding. Additional recorder detail: the pinned selector can return a packaging.version.Version for the available FlashAttention backend even when use_flash_attention=False and use_fused_attention=True; this head's int(flash_backend) raises TypeError. Use a string for that field. All 177 tensor payloads in the nine regenerated fixtures match the previous reviewed revision. All nine source digests are now populated; unknown schema versions are rejected. The two reduced-precision metadata fields still claim float32 while replay accumulates in float16/bfloat16; the new notes do not resolve that prior P2. Independent one-GPU torchrun frozen CP1 replay without TE completed with status OK: 32768 output elements, max absolute error 2.95e-8, zero NaN/Inf, zero gather/replay and self-repeat bit mismatches, and restored global token order. This validates frozen-partial replay, not actual attention.

Limitations: No actual TE GPU execution or new multi-GPU CP2/CP4 proof. Backend and pin probes explicitly mock TE or execute pinned upstream control flow; the GPU run consumes frozen partials. Previous actual-attention numbers should be refreshed because this follow-up changes input compute dtype and backend selection. Only test/reference tooling changed, so no production CPU hot-path or GPU performance claim is made. The current Style workflow awaits maintainer authorization; broad CI remains unnecessary while these small verifier fixes are pending.

YangXu1990uiuc:
The pending [Style check](https://github.com/NVIDIA/cudnn-frontend/actions/runs/35419476326) has now been authorized and passed for `597f39098dc9`.

I am leaving broad GPU CI until the recorder follow-up. The focused suite and frozen CP1 replay are complete; actual fused-TE replay remains outside my validated scope, as noted in the review.

YangXu1990uiuc (commented):

Reviewed `4a3387d996e2` against base `667fe4ce8ce4`.

Thank you for your contribution! All three findings from the previous round are resolved: repeated backend selections are recorded, FlashAttention versions are serialized correctly, unverified results exit 2, and the reduced-precision fixture/trace metadata matches execution. The existing approval stays in place. One new nonblocking test-dependency issue remains below.

**[P2] Keep the new exit-code test runnable without optional TransformerEngine** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/4a3387d996e27b9911f8f75b01bae60f5e56c775/test/python/sdpa/test_cp_numerics_reference.py#L692))

This new L0 test imports actual_attention before installing its mocks. That module still imports DotProductAttention at module scope, so a valid FE environment without optional transformer_engine fails with ModuleNotFoundError before the exit-code assertion runs. The exact-head suite reports 47 passed, 1 skipped, 1 failed here; the previous revision had 46 passed and 1 skipped in the same environment. Please stub the actual_attention module/result before importing it, or defer its real TE import until attention construction. This test checks runner reporting and exit status, so it can exercise the behavior without TE. The runner fix itself passed an independent mocked-result probe.

Evidence: Normal repository pytest on 4a3387d: TestR2Preconditions.test_runner_returns_unverified_exit_for_an_unverified_backend fails at line692 -> actual_attention.py:56 -> ModuleNotFoundError: No module named 'transformer_engine'. The independent current-head run_actual_attention probe returns exit2 and writes R2_UNVERIFIED. Attribution: Introduced by the new regression test in this follow-up. The previously fixed frozen-runner optional-import behavior remains intact.

Validation: Exact-head normal reference suite: 47 passed, 1 explicit TE-checkout skip, 1 new optional-TE test failure described above. Repeated the previous independent pinned-TE control-flow probes: both cold and warm selections now record fused execution and restore the cache flag; an available FlashAttention Version is recorded as text. Five aggregation cases correctly reject empty, missing, mixed or unfused evidence and accept complete fused records. The same unverified-result control that returned exit0 on 597f390 now returns exit2, with the JSON retaining R2_UNVERIFIED. Imported-source positive/negative controls and missing-checkout refusal remain correct. All 177 frozen tensor payloads remain unchanged. Both FP16/BF16 fixture headers and every corresponding trace entry now match the observed TE-ordered merge accumulator dtype. Prior frozen CP1 replay evidence remains applicable to the unchanged frozen path.

Limitations: No new actual-TE GPU or multi-GPU execution claim: recorder/exit checks use explicitly mocked TE and pinned upstream control flow. This follow-up changes test/reference tooling only, so no production performance benchmark is needed. I am leaving new CI requests for the small test-dependency follow-up to avoid spending capacity on the known L0 failure.

0z5a:
All three findings are addressed on head `4a3387d9`, and the R2 records you asked for are refreshed below with a real fused-TE run.

## [P2] Handle cached backend selection and version-valued FlashAttention metadata

The recorder now forces `_attention_backends["backend_selection_requires_update"] = True` for the duration of each recorded call and restores the previous value on exit, so the pinned dispatcher cannot skip the selector on a repeated non-causal step. An empty record is no longer an exception: it is counted as `unrecorded_steps`, the flag reads use `.get()`, and the verdict requires

```
executed = steps > 0 and fused_steps == steps and unfused_steps == 0 and unrecorded_steps == 0
```

with `unrecorded_steps` in the `merged.backend` block and named in the verdict text, so a step whose kernel choice was never observed can never be presented as fused evidence.

`flash_attention_backend` is recorded as text. You are right that the pinned selector can hand back a `packaging.version.Version` there even when `use_flash_attention` is False and the fused backend was chosen — `int()` cannot represent it, and the new source test pins `str(flash_backend)` and the absence of `int(flash_backend)`.

## [P2] Return the unverified exit code when backend verification fails

`run_actual_attention` now returns `EXIT_R2_UNVERIFIED` when the backend guard (not just the pin guard) produces `R2_UNVERIFIED`, so the shell agrees with the report. `test_runner_returns_unverified_exit_for_an_unverified_backend` drives the real function with a mocked unverified result and asserts both the exit code and the written report, including `unrecorded_steps`.

## [P2] The reduced-precision accumulator metadata

Fixed by setting the parameter rather than re-wording the note: the FP16 and BF16 fixtures now record `accumulator_dtype` equal to their storage dtype, which is what the replay actually accumulates in on the non-FP8 path, and all nine fixtures are regenerated. `test_reduced_precision_fixtures_record_the_accumulator_they_ran` pins the relationship for every fixture so the metadata cannot drift again. The tensor payloads are unchanged — the per-fixture digests are byte-identical to the previous revision (177 tensors), only the header metadata moved.

## R2 records, refreshed with the pinned TransformerEngine

TransformerEngine was built from the pinned source and installed over the PyPI wheel:

```
Successfully installed transformer_engine-2.21.0.dev0+ae34b34f
installed context_parallel.py sha256 = 37b389cefd7e80060ccb0a856f4970daafe307e0705b79ddc2f3cbce30fc002a  (= the recorded pin)
```

so the pin check passes on the imported module, not on a `--te-repo` copy. `torchrun` lanes on 2 L20s (CP2) and 4 L20s (CP4), `--te-compute-dtype float16`, plus a CP2 `bfloat16` arm:

| lane | fixture | ranks | status | fused steps | unrecorded | unfused | fused sub-backend | merged O max abs | merged LSE max abs | exit |
|---|---|---|---|---|---|---|---|---|---|---|
| actual-attention | cp_p2_f32_s128_causal | 2 | **R2_EXECUTED** | 2/2 | 0 | 0 | 1 | 4.82e-04 | 4.68e-07 | 0 |
| actual-attention | cp_p4_f32_s128_causal | 4 | **R2_EXECUTED** | 4/4 | 0 | 0 | 1 | 4.82e-04 | 4.08e-07 | 0 |
| actual-attention (bf16) | cp_p2_f32_s128_causal | 2 | **R2_EXECUTED** | 2/2 | 0 | 0 | 1 | 3.83e-03 | 4.68e-07 | 0 |
| frozen-partials | cp_p2_f32_s128_causal | 2 | OK | - | - | - | - | 2.95e-08 vs FP64 | - | 0 |
| frozen-partials | cp_p4_f32_s128_causal | 4 | OK | - | - | - | - | 2.95e-08 vs FP64 | - | 0 |

The per-step mask types are now `causal` for the diagonal section and `no_mask` for the triangles, and every step reports the fused backend. The merged O error is the storage precision of the fused lane — TE's fused attention does not accept fp32 (fp32 selects the unfused backend), which is also why I am **not** repeating the earlier "per-step partials match the fixture to 1.19e-07" number: that measurement came from the fp32 compute path, i.e. exactly the unfused route your review flagged. The fused lane cannot produce an fp32 partial to compare; its honest numbers are the merged ones above.

Evidence on the L20 host: `test/python/sdpa/test_cp_numerics_reference.py` 49 passed (46 + 3 new behavioural checks), 0 failed, with the TE-checkout cross-check enabled; the four `torchrun` lanes above all exited 0.

YangXu1990uiuc (commented):

Reviewed `8827220e329c` against base `a5d61f45c08d`.

Thank you for your contribution! This head has exactly the same Git tree as the previously reviewed 4a3387d9, so there is no implementation delta to reassess. The existing approval remains in place. The nonblocking optional-TransformerEngine test failure is still reproducible; please address the linked item before merging.

- **[P2] Still open:** [Keep the exit-code test runnable without optional TransformerEngine](https://github.com/NVIDIA/cudnn-frontend/pull/1142#pullrequestreview-5254695528).

Validation: Compared the complete Git trees: both 4a3387d9 and 8827220e resolve to 8c6fda924464ffca71c6de1db61f71792b31dd4a. Reran the exact-head reference suite with the normal repository conftest and explicit L0 or L1 selection: 47 passed, 1 skipped, 1 failed in 28.69 seconds. The sole failure is the unchanged optional-TE import issue above. Verified the actual imported FE path.

Limitations: No new actual-TE or multi-GPU replay claim. Further CI is deferred until the known L0 test-dependency fix, to avoid a run that needs repeating.

PR #1145, opened 2026-09-18
engines: offer opt-in FROST engines without the flag when the cuDNN backend proposes no plan or declines every plan at build

YangXu1990uiuc:
## Before submitting

## Affected area

- Python API or bindings (planning / dispatch: `engines/manifest.py`, `_pygraph.create_execution_plans` / `build_plans`)

## Summary

`CUDNN_FRONTEND_ENABLE_FROST_ENGINES` withholds an **optimization**, never an **operation**. Today a graph the cuDNN backend proposes no plan for — or whose every proposed plan declines at build time — fails with `no engine — python or backend — proposed a plan` even when an opt-in FROST engine serves it, because the opt-in engines are not candidates without the flag. This PR admits them exactly then:

- `create_execution_plans`: when `backend_plan_entries()` is empty, the family's opt-in engines become candidates (`pygraph._admit_opt_in_engines`) and are ranked as usual.
- `build_plans`: when every entry the backend proposed declines at build time, the opt-in engines are admitted, their proposals appended **after** the backend's entries, and the walk continues on the new tail. An explicit `select_plan()` pin stays strict (no silent fallback).
- `manifest.EngineFamily.offered_ids(include_opt_in=)` / `engines_for(include_opt_in=)` carry the choice; `EngineSlot.opt_in`'s docstring, the README and `docs/utilities/python_graph_and_execution_backends.md` state the rule.

Graphs the backend serves are **unchanged**: the opt-in engine is absent from their plan list, not ranked behind the backend, and the FROST module is not imported for them. This is the per-graph form of the existing rule that a sole-implementation family (GDN / KDA) is never gated. Forward and backward alike, every family.

Second commit (self-review round, two real gaps): `check_support()` used to re-raise the backend's aggregate decline whenever no python entry was already listed, so `build()` / `execute()` never reached the build-time admission — it now tolerates the decline when the family has admissible opt-in engines (a `select_plan()` pin still raises), and `build_plans()` carries the backend's decline into its final error when the admitted engines decline too. And a recorded `(engine_id, knobs)` of a plan offered without the flag can now be replayed with `create_execution_plan()` on a fresh graph: an explicit pin is deliberate, so the id resolves through `manifest.engine_for_id(include_opt_in=True)`. Carve-out stated in the docs: a graph the installed backend rejects at `validate()` still raises there, as before.

Direction (Yang): keep FlashInfer and every other consumer agnostic of the flag; graduate FROST where the backend has nothing. The per-config graduation where FROST **wins** a served graph (kernel-time table below) stays a separate, evidence-driven step.

## Why

On the SM100 dev box (cuDNN 9.27 headers, cudnn-frontend `develop@5567e2733`), bf16 dense causal SDPA, flag **unset**:

| graph | before | after |
|---|---|---|
| backward d = 512 | planning: `no engine — python or backend — proposed a plan` (backend proposes nothing) | plans `['sdpa_bwd_sm100']`, builds and runs; numerics vs fp32 torch pass (`_run` in `test_sdpa_bwd_dsl_sm100.py`) |
| backward d = 128 / d = 256 | backend `eng5` / `eng7` plans | unchanged (no FROST entry in the list) |
| forward d = 128 / 256 / 512 | backend `eng10` / `eng8` plans | unchanged |

Why this PR stops at "backend has nothing" rather than flipping `opt_in` per arch: that is a performance judgement, and it belongs to a table, not to this rule. For the record, `benchmark/attention_inference` re-run today against this develop on computelab (device time, `cudnn_oss ÷ cudnn`, geomean over the config's cells): RTX PRO 6000 (SM120) llama3.1 context **0.63** (28/28 cells faster), generation 0.29; B200 llama3.1 context 0.76, generation 0.14; qwen35 on B200 context 0.75, generation 0.06 — FROST ahead on both arches for these models, a large move from the 2026-08-26 CSVs in the tree (B200 llama context 1.43). The per-arch graduation of *served* graphs is the next PR, on that table.

## Related issues

#1098 (execute-time overrides; the reason FlashInfer must not need to know about the flag), #1119, #1132. Replaces a FlashInfer-side guard (flashinfer#5331, closed).

None at the graph API. `offered_ids` / `engines_for` gain a keyword-only `include_opt_in`. Plan lists of backend-served graphs are byte-identical to before; only graphs that previously raised at planning or build now get a plan.

## Testing

- `test/python/test_dispatch.py` (CPU, 64 passed): five new tests — opt-in stub admitted and executed when the backend proposes no plan; stays absent when the backend serves; admitted and appended when every backend plan declines at build, with the `select_plan()` pin still raising; the backend's aggregate `check_support()` decline reaching the admission through `check_support(); build_plans()` and through `execute()`'s auto-build; a flag-less replay via `create_execution_plan()`. The existing only-plan decline test now states the rule with an opt-in engine that declines the graph itself.
- `test/python/sdpa/frost/test_sdpa_bwd_dsl_sm100.py::test_d512_backward_runs_on_frost_without_the_opt_in_flag` (SM100): flag deleted, plan list is exactly the FROST backward engine, numerics vs torch pass, and the recorded `(engine_id, knobs)` replays on a fresh graph without the flag.
- GPU regression set on SM100 with the flag unset (`test_sdpa_bwd_dsl_sm100`, `test_sdpa_fwd_heuristics`, `test_sdpa_frontend_integration`, `test_native_backend_lowering`, `test_knob_vocabulary`): 111 passed, 1 skipped on the first commit; 101 passed on the second (same set minus `test_sdpa_frontend_integration`).
- In-scope / out-of-scope frost gates (17 + 5 files, 16 workers, SM100, flag per test via the frost conftest): **1323 passed / 5 skipped and 467 passed / 58 skipped on the head `a02e8a976`** (identical counts on the first commit).

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer · worktree /data/yanxu/worktrees/scratch-gpu/fe-wt-rebase (branch yanxu/frost-default-when-backend-declines) · evidence /home/scratch.yanxu_gpu/fe-sites/pr1132_backup/q12</sub>


YangXu1990uiuc:

YangXu1990uiuc:
Second commit `a02e8a976` (self-review round with an adversarial reviewer set; 19 findings, 2 real, both fixed):

- `check_support()` re-raised the backend's aggregate decline whenever no python entry was already listed, so `build()` / `execute()` never reached the build-time admission in `build_plans()` — the documented rule was reachable only by calling `build_plans()` directly. It now tolerates the decline when the family has admissible opt-in engines (a `select_plan()` pin still raises), and `build_plans()` names the backend's decline in its final error when the admitted engines decline too.
- A recorded `(engine_id, knobs)` of a plan offered without the flag could not be replayed with `create_execution_plan()` on a fresh graph (`engine_for_id` went through the gated `offered_ids()`). An explicit pin is deliberate: it now resolves with `include_opt_in=True` and the engine joins the graph's candidates.
- Docs state the carve-out: a graph the installed backend rejects at `validate()` still raises there, as before.

Tests: `test_dispatch.py` 64 passed (five new cases + the only-plan decline test restated with an opt-in engine that declines the graph itself); the SM100 bwd d=512 test now records its plan and replays it on a fresh graph without the flag. First-commit gates: in-scope 1323 / 5 skipped, out-of-scope 467 / 58; re-running on this head.

YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `a02e8a9767c3` against base `b03c3f3321f5`.

The fallback direction makes sense, and I found no approval-blocking issue at a02e8a976. Late admission preserves explicit pins, engine/note/workspace exclusions, and the caller's build context. I independently reproduced CodeRabbit's remaining live-flag/cached-candidate corner case below; I classify it as nonblocking P2 because it requires changing the environment after candidates were cached and produces a visible planning failure. Setting the flag before graph planning avoids it. This is a comment-only review because the PR is authored by the account hosting this bot.

- **[P2] Still open:** [Reconcile cached candidates when the opt-in flag changes](https://github.com/NVIDIA/cudnn-frontend/pull/1145#discussion_r4051391230).

Validation: Independent exact-head dispatch suite: 64 passed. Seven additional probes passed for late-admission name/numerical/behavior/workspace filters, repeated failed builds without duplicate plans, fatal RuntimeError propagation, ALL/default selection, and caller-context forwarding. Focused SM100 run: 59 passed and 2 failed. The new flag-free d512 backward execution/replay regression passed; running that same new test against merge base 5567e273 reproduced the prior no-plan failure. Both neighboring heuristics failures reproduced unchanged on that exact base (plan-count expectation and lowercase knob-name lookup). Default backend-served SDPA control, flag unset, FP16 B=2 H=8 S=256 D=128 on SM100: base/head retained the same native plan. ABBA CPU enqueue medians were base 9.43/8.92 us versus head 9.43/9.03 us; GPU medians were 10.88-10.89 us on both. Analytic output/LSE checks and poisoned-output CUDA Graph replay passed. No material regression in this spot check. Built a matching native extension and verified source imports; local validation used CUDA 13.2, cuDNN 9.25.1 and CuTe DSL 4.8 development dependencies. The base comparison used the same native binding after checking that the PR changes no native sources. Reused owner-triggered current-head CI 68689657. Inspected SM80/90/100 Python-lane failures: the same four cuTile correlated-key GDN/KDA numerical cases seen in prior runs; those direct implementation paths are unchanged here. Rubin reports 12 native ragged-decode Stats NVRTC failures, also seen in preceding runs. SM103 exited 141 during device detection before tests. No duplicate CI requested.

Limitations: The performance comparison covers one backend-served workload, not every family or device. Independent numerical validation here is SM100; other-architecture evidence is from CI. The complete CI pipeline was still running on the last GitHub status read. Final GitLab metadata refresh became unavailable, so the inspected failures are classified by their logs/source and prior evidence rather than claimed as a complete green pipeline.

YangXu1990uiuc:
CI 68689657 has finished: FROST SDPA on SM80/SM100/SM120, FROST linear/GEMM, builds, and C++ sanitizer jobs passed. The SM103 job labeled **New Failure** exited 141 during device detection, before pytest started; its inspected log does not show a code/test regression. The inspected Python and Rubin failures are described in the review above. No duplicate pipeline requested.

YangXu1990uiuc:
Closing: #1152 (merged as 51265b474) makes the SM100/SM120 f16 FROST rows default candidates through workload placement, so the backend-declines fallback this PR added is no longer needed for them; the remaining opt-in rows (fp8/mxfp8, SM107, GEMM) can get a fresh, rebased PR if a fallback is still wanted. The branch stays on the fork.

Issue #853, opened 2026-09-02
[Bug]: invalid argument error when create tensor map in cutlass dsl 4.8.


Adnios:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.

### Affected area

FE OSS kernels or CuTeDSL

### Description

```bash
python -m pytest -vv -s --tb=long \
    'test/python/test_mhas_v2.py::test_sdpa_random_bwd_ragged_L0[test247]'
```
hits invalid argument error when create tensor map in cutlass dsl 4.8.

### Steps to reproduce

Install cutlass dsl 4.8

mini reproducer:

```python
#!/usr/bin/env python3

import torch
import cuda.bindings.driver as cuda_driver

import cutlass
import cutlass.cute as cute
from cutlass.experimental import cuda


@cute.kernel
def consume_descriptor(descriptor: cutlass.GridConstant[cuda.TensorMap]):
    pass


@cute.jit
def host(q: cute.Tensor, stream: cuda_driver.CUstream):
    descriptor = cuda.create_tensor_map_tiled_from_view(
        q,
        box_dims=(1, 1, 2, 64),
        stride_order=(3, 2, 1, 0),
        swizzle=cuda.TensorMapSwizzle.s128b,
    )
    consume_descriptor(descriptor).launch(
        grid=(1, 1, 1),
        block=(1, 1, 1),
        stream=stream,
    )


outer_stride = cute.sym_int(divisibility=1)
fake_q = cute.runtime.make_fake_tensor(
    cutlass.Float16,
    (1, 1, 8, 256),
    (outer_stride, 8192, 256, 1),
    assumed_align=16,
)
compiled = cute.compile(
    host,
    fake_q,
    stream=cute.runtime.make_fake_stream(use_tvm_ffi_env_stream=False),
    options="--enable-tvm-ffi",
)

# 145,227,776 FP16 elements = 290,455,552 bytes. The size-1 outer
# dimension keeps the actual allocation at about 4 KiB.
q = torch.empty_strided(
    (1, 1, 8, 256),
    (145_227_776, 8192, 256, 1),
    dtype=torch.float16,
    device="cuda",
)
stream = cuda_driver.CUstream(torch.cuda.current_stream().cuda_stream)
compiled(q, stream)
torch.cuda.synchronize()
print("PASS")
```

### Expected behavior

```bash
CUDA Dialect Assertion failed on CUDA error: cudaErrorInvalidValue - invalid argument
[...]

PR #1146, opened 2026-09-19
FROST SDPA f16/bf16 SM100/SM107: Int64 stride leaves through the pointer host ABI (2^27 TMA batch-stride overflow since #1119)

YangXu1990uiuc:
## Before submitting

## Affected area

- FE OSS kernels or CuTeDSL (FROST SDPA forward f16/bf16, SM100 + SM107 explicit pointer host entries; `_common_blackwell._bshd`)

## Summary

Since #1119 the eight f16/bf16 forward hosts (`sm100/` and `sm107/` `prefill_{d128,d192_d128,d256,d512}_f16.py`) take the per-operand `(batch, seq, head)` element strides as `Tuple[int, int, int]`. The DSL types a Python-int tuple element as a dynamic **Int32** whatever the annotation says, and `_bshd` hands those leaves to the layout every TMA descriptor is encoded from. This PR carries the stride leaves as **Int64** end to end:

- the `compile()` placeholders for `q/k/v/o/lse_strides` (and `gate_strides` on SM107 d256) become `cutlass.Int64(0)` triples — the artifact's signature width comes from the placeholder values, so this is the change that widens the ABI; the annotations say `Tuple[cutlass.Int64, cutlass.Int64, cutlass.Int64]` to match (the tvm-ffi spec reads the annotation, the MLIR signature reads the value; the two must agree);
- `_bshd` coerces its three leaves with `cutlass.Int64(...)`, so the layout is Int64 even for a caller that still binds Int32;
- the stale "`S * seq_stride` (Int64)" remark in the host docstrings becomes the actual rule.

Callers are unchanged: `prepared.py` frames and the adapter keep passing Python ints (tvm-ffi packs them as int64; the 64-bit slot has no bound check or truncation). `problem_size`, `lse_ext`, `n_pages`, `table_strides` stay Int32 (extents and page-table strides, all < 2^31).

## Why

`create_tensor_map_tiled_from_view` converts a stride to 16-byte TMA units as `stride * width_bits // 128` in the stride's own width (`cutlass/experimental/cuda/tensor_map.py::_stride_to_tma_units`, widened to Int64 only afterwards). For a 16-bit operand that is a ×16 before the divide, so a per-batch extent `S * H * D`:

| per-batch elements | Int32 product | effect |
|---|---|---|
| 2^27 (256 MiB) | 2^31 → negative | `cuTensorMapEncodeTiled` rejects it: `CUDA Dialect Assertion failed on CUDA error: cudaErrorInvalidValue`, exit 134 |
| 2^28 … 2^30 | wraps to exactly 0 | the encoder accepts a zero batch stride: at B ≥ 2 every batch reads batch 0's operand and O of every batch lands on batch 0 — **silently wrong** |
| ≥ 2^31 | — | tvm-ffi rejects the argument: `Out of bound value on k_strides[0] … expected to be in int32 range` |

Observed, same FE code and the FROST plan verified as the executed plan in both cases: with nvidia-cutlass-dsl **4.7.1** these graphs run correctly (only the 2^31 ceiling shows), with **4.8.0.dev0** the abort and the aliasing reproduce on SM100 and SM107. The Python-level `tensor_map.py` scaling and the traced IR are identical in the two DSL versions (a standalone probe of `leaf * 16 // 128` on an Int32 layout leaf wraps in both), so the difference sits in the DSL's lowering / encode of that IR and is not chased further here — the Int32 declaration is the FE bug either way. This is the same DSL behaviour as #853; #974 already widened the SM100 d256 prefill and decode descriptors by hand (`_create_tma_desc` with `cutlass.Int64(tensor.stride[i])`), which is why d256 was the one flavor that passed. fp8 / MXFP8 hosts take `cute.Tensor` arguments (64-bit dynamic strides by default) and were never affected; THD binds the token stride in the batch slot (extent 1) and is immune on the TMA side.

Measured (dense, H = 128, token-major operands, FROST plan pinned; `before` = develop `b03c3f332`, `after` = this branch). SM100 = the local Blackwell board (correctness only); SM107 = Rubin `w2u1g-lc-0186`, DSL 4.8.0.dev0:

| case | before, DSL 4.7.1 (SM100) | before, DSL 4.8 (SM100) | before, DSL 4.8 (SM107) | after (all three) |
|---|---|---|---|---|
| bf16 B=1 S=2048 d=512 (Q, K, V, O all 2^27; the `benchmark_single_sdpa` repro) | ok | **abort** | **abort** | ok (3e-4) |
| Q-only 2^27: f16 B=1 S_q=8192 S_kv=1024 d=128 | ok | **abort** | **abort** | ok (4e-5) |
| Q-only 2^27: f16 B=1 S_q=4096 S_kv=1024 d=256 | ok | ok (#974 widened SM100 d256) | **abort** (SM107 d256 was not) | ok (4e-5) |
| K/V-only 2^27: bf16 B=1 S_q=1024 S_kv=2048 d=512 | ok | **abort** | **abort** | ok (2e-4) |
| K/V 2^27: f16 B=2 S_q=128 S_kv=8192 d=128 | ok | ok | **abort** | ok (1e-5) |
| Q 2^28 at B=2: f16 S_q=16384 S_kv=1024 d=128 | ok | **wrong, both batches** (max err 1.2e-1 / 8.0e-2) | **wrong, both batches** (1.2e-1 / 7.5e-2) | ok (4e-5) |
| K/V 2^28 at B=2: f16 S_q=16 S_kv=8192 d=256 (the reported decode-shaped case) | ok | ok | **batch 1 = ref(q1, K0, V0) to 1.1e-5**, 3.9e-2 off its own | ok (1e-5) |
| K/V 2^28 at B=2: f16 S_q=128 S_kv=16384 d=128 | ok | ok | **batch 1 aliases batch 0** (8.6e-6 vs 2.5e-2) | ok (9e-6) |
| K/V 2^30 at B=2: f16 S_q=16 S_kv=32768 d=256 | ok | ok | **batch 1 aliases batch 0** (5.9e-6 vs 1.9e-2) | ok (5e-6) |
| K/V 2^31: bf16 B=1 S_q=16 S_kv=32768 d=512 | `Out of bound value on k_strides[0]` | same | same | ok (5e-5) |

SM100 under DSL 4.8 shows fewer red cells than SM107: #974 widened the SM100 d256 encodes, and the SM100 d128 K/V cells happen to pass (not chased further here); the SM107 flavors show the full pattern of the report, K/V aliasing included.

## Related issues

#853 (the DSL 4.8 tensor-map behaviour), #974 (d256-only widening), #1119 (introduced the Int32 leaves), #1132 (moved the same launch into the prepared dense entry). Internal report: "Int32 overflow of the TMA batch stride since #1119" (Rubin bisect, 2026-09-18).

None at the graph or engine API. The compiled artifacts change signature (i32 → i64 stride slots) and are keyed by source digest, so they recompile once; no cache-key format change. The 2^31-element per-operand stride ceiling is lifted with it.

## Testing

- New `test_sdpa_fwd_dsl_sm100.py::test_dsl_sm100_stride_leaves_are_int64` (runs on SM100 and SM107; four ids: Q / K-V side at exactly 2^27 with the abort mode, and at 2^28 with B = 2 for the aliasing mode; per-batch fp32 reference on a row slice; skips under 6 GiB free). Seen red first: on the unfixed develop kernels with DSL 4.8 on SM107 the run aborts (`pytest` exit 134, the 2^27 id); on this branch all four ids pass on SM107 (Rubin, DSL 4.8.0.dev0) and on SM100 (DSL 4.7.1 in the gates below, and DSL 4.8.0.dev0 in the repro matrix). Unrelated on that Rubin stack: `test_dsl_sm100_singleton_seq_bhsd_storage` fails identically before and after ("no plan for engine 'sdpa_fwd_prefill_sm107'", S = 1 graphs), i.e. pre-existing there.
- Repro matrix above: 12 cases × {DSL 4.7.1, DSL 4.8.0.dev0} on SM100, all pass after the fix (before: 4 aborts + 1 silent corruption under 4.8, 1 `Out of bound` under both).
- SM100 regression gates on this branch (16 workers, DSL 4.7.1, same 17 + 5 file sets as #1132 / #1145): in-scope frost gate **1327 passed / 5 skipped** (1323 + the 4 new ids); out-of-scope gate (fp8 / mxfp8 / SM107-file-on-SM100) **467 passed / 62 skipped** (the #1145 baseline on develop `5567e2733` was 467 / 58; the base moved to `b03c3f332`).
- `black --line-length 160` clean on the ten files.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer · worktree /data/yanxu/worktrees/scratch-gpu/fe-wt-int64 (branch yanxu/frost-sdpa-int64-strides) · evidence /home/scratch.yanxu_gpu/fe-sites/pr1132_backup/q13 + fe-sites/rubin_int64</sub>


YangXu1990uiuc:

RomanAnders90 (approved):

YangXu1990uiuc (commented):

Reviewed `21b28b1a47fd` against base `b03c3f3321f5`.

No new findings. The Int64 width is carried consistently through compile placeholders, the pointer host ABI and operand-layout construction, including Rubin's gate strides. I independently reproduced the pre-fix silent batch aliasing on SM107 and verified the same regression passes at this head. The source-digested compiled-cache environment also prevents reuse of the old Int32 artifacts after this source change. This is a comment-only review of my account's PR; the existing independent approval remains in place.

Validation: All four new large-stride regression cases passed independently on both SM100 and SM107, without skips, using PyTorch 2.13 / CuTe DSL 4.8 / cuDNN 9.25.1. Running the exact new kv_2p28_b2_d256 test against base b03c3f3321f5 on SM107 failed numerically: 25,642 of 524,288 elements exceeded tolerance, maximum absolute error 0.132446. The head passes, so the regression genuinely discriminates the fix. Additional small-allocation ABI probes passed for all four flavors on each architecture: d128, d192/d128, d256 and d512. After a normal API launch, the actual compiled host frame received 2^31+32 for Q/K/V/O/LSE batch strides with B=1, then produced correct O and LSE. Maximum O error was 3.95e-5 and LSE error 9.54e-7. The same d128 probe against the SM100 base was rejected by TVM-FFI because q_strides[0] was still Int32. These probes deliberately bypass singleton-stride canonicalization and verify the host ABI, not arbitrary public-layout admission. Focused neighboring suites: SM100 23 passed (THD Stats, strided Stats and compiled-cache tests), with 4 Rubin-only gate cases skipped as intended; SM107 16 passed with no skips, including those four gate cases plus THD and strided Stats. Imported source paths and architecture-specific kernel paths were checked. SM100 ABBA component timing, FP16 B=2,H=16,S_q=S_kv=256,D=64 with Stats and the d128 prefill route: base CPU enqueue medians 52.401/51.880 us versus head 51.183/52.305 us; base CUDA Graph GPU medians 13.084/13.064 us versus head 13.140/13.146 us. No material regression was observed in this spot check. Analytic O/LSE and poisoned-output capture/replay checks passed in all four legs.

Limitations: Focused numerical validation covers local SM100 and SM107 on DSL 4.8; the performance measurement covers one SM100 workload and is not a full architecture or performance sweep. The owner-started current-head FROST/python_tests pipeline 68707048 is still running. No duplicate CI was triggered, and pending CI is not being reported as passed.

PR #1151, opened 2026-09-20
Reduce prepared SDPA execution overhead

YangXu1990uiuc:
## Before submitting

- [x] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches and my changes comply.
- [x] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*`.

## Affected area

Python API or bindings; FE OSS kernels or CuTeDSL.

## Summary

Reduce the host work in an already built FROST SDPA plan, including small-query split-KV execution. Previously split decode stayed on the tensor adapter and rebuilt buffer facts and workspace views on every call. The prepared path now binds the caller's partial workspace and final outputs directly, then launches attention and combine on the same stream.

- Migrate SM100 D128 and D256 f16/bf16 decode to the existing explicit pointer ABI. Delete the obsolete f16 tensor host, shape/stride-specialized fake-tensor compile branches, and adapter compile/execute fallback. Keep each kernel's host and pipeline explicit.
- Specialize D256 contiguous page-table columns and common O/Stats inner strides in the compiled host, with per-call branches retaining the general Int64 stride path. The device pipeline and public ABI remain unchanged.
- Prepare split-KV launches with independent per-call frames. Partial O uses the kernel's required dtype; partial LSE stays natural-log even without user Stats. The combine writes the actual final O/LSE strides, including padded layouts, without an O scratch/copy-back.
- Preserve Int64 stride leaves through annotations and compile placeholders, including all seven sibling prefill page-table slots. Page-table indices remain int32.
- Cache pure THD/dense geometry and project immutable facts directly from the native variant pack. Remove the unused Python `facts_of_pack` and native `facts()` projection; tests use an independent oracle.
- Reject mixed K/V pool layout kinds during standalone support checking. Validate output span/device/alignment/overlap and compiled shapes before launching; metadata inputs also check the producer storage span after shape overrides.

- Make FE-owned resource destruction safe during unrelated CUDA Graph capture: restore the thread capture mode after cleanup, avoid retaining backend exception tracebacks, and use existing aligned buffers for unused prepared ABI slots. Zero-KV THD descriptor bindings require no device allocation or initialization.

[...]

PR #1147, opened 2026-09-19
perf(dsa): optimize SM100 indexer score forward and recompute

NVIDIA-JerryChen:
## Before submitting

- [ ] I agree to license this contribution under the terms of [LICENSE.txt](https://github.com/NVIDIA/cudnn-frontend/blob/develop/LICENSE.txt).
- [x] I ran `pre-commit run` and committed any formatting changes.
- [x] I reviewed the Hard Rules in the `AGENTS.md` for each directory this PR touches (see [root AGENTS.md § Reviewing a PR](https://github.com/NVIDIA/cudnn-frontend/blob/develop/AGENTS.md#reviewing-a-pr-human-or-agent)) and my changes comply, or I explain the exception below.
- [ ] I added GitHub labels: one `cat-*`, one or more `area:*` / `op:*`, and one `orig-*` (see [label list](https://github.com/NVIDIA/cudnn-frontend/labels)).

Requested labels: `cat-perf-bug`, `area:sparse_attention`, `op: DSA`, `orig-nv-eng`. The submitting account cannot add labels to the upstream repository; a maintainer needs to apply them.

## Affected area

FE OSS kernels or CuTeDSL — DSA indexer dense forward and dense score/LSE recompute on SM100/SM103.

## Summary

Improve the shared BF16/MXFP8 indexer score kernels:

- Use M128 for D128 H32/H64; H32 processes four queries per tile to reuse K.
- Shorten FP32 head reduction with two independent partial sums where beneficial, retaining four sums for BF16 H64/ratio-one LSE and MXFP8 H32/ratio-one LSE.
- Tune MXFP8 TMEM reads, accumulator slots, and KV pipeline depth for the existing static configurations.
- Extend existing varlen/tail tests across H32/H64, ratios 1/4, and forward/recompute.

## Why

Indexer scoring reuses one KV head across 32 or 64 query heads. More query reuse and a shorter reduction epilogue reduce score computation time.

The following historical kernel-only measurements compare the optimized implementation with the original production baseline. Each entry aggregates 32 representative training cases with zigzag ring CP16, D128, Hkv=1, using manifest weights. Speedup = sum(weight × baseline time) / sum(weight × optimized time). Each cell is forward / recompute.

| Input / Hq / ratio | GB300 elapsed-time speedup | GB200 elapsed-time speedup |
|---|---:|---:|
| BF16 / 32 / 1 | 1.585× / 1.557× | 1.531× / 1.452× |
[...]

PR #1152, opened 2026-09-20
Rank FROST SDPA plans by workload

YangXu1990uiuc:
## Before submitting

## Affected area

Python graph planning, FROST SDPA forward, native block-mask support checks, benchmarks and regression tests.

## Summary

Eligible SM100 and SM120 FP16/BF16 SDPA forward graphs participate in ordinary graph planning. Workload rules place FROST candidates before or after the backend plans, so `heur_mode.A` callers receive the selection without API changes.

**Merge after #1151.** This branch includes its prepared executor and capture-lifecycle fixes. Override-enabled graphs retain FROST only for configurations supported by that prepared executor; tensor-adapter configurations decline during planning. Execution also rejects override arguments on legacy Python plans instead of silently dropping them. Prepared runtime shape/stride and workspace-envelope checks remain in force.

- Families can emit an internal `BACKEND` placement marker. The planner expands it once per mode block; `OPENSOURCE` omits native entries. Existing hooks without a marker, fallback and deduplication retain their contracts. The marker is not a new public `cudnn.engines` export.
- SM100 rules favor eligible multi-token decode and selected long-cache D256/D512/prefill regimes. Multi-token decode and D512 prefill require at least 2048 KV tokens before receiving the affected FROST-first preference. D128 single-token decode and unmeasured SM100 THD/paged-prefill regimes generally keep the backend first. Sinks at Sq1 are an exception and prefer FROST.
- The D512 single-token head-count shortcut requires KV >= 131072, matching its historical evidence. SM120 uses its existing architecture rules with the same long-cache restriction; this PR does not add new SM120 timing evidence.
- Other engines remain opt-in. `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1` still prefers FROST where supported; it bypasses neither capability checks nor the native block-mask guard.
- Native benchmark lanes exclude Python engines. Block-mask test inputs use a per-case generator so preceding tests cannot change their masks.

## Why

The provider crossover depends on query length, head dimensions and KV parallelism. Eligibility determines what can execute; this policy determines which eligible provider is tried first. Internal tile/scheduler/GQA/split selection remains in the existing forward heuristics and is refreshed separately.

A targeted native check rejects **SM10.x UNIFIED block-mask SDPA on cuDNN < 9.26.0**. A released 9.25 backend can consume stale TMEM when the initial KV tiles are masked. Official 9.26.0.51 GA passes the reproducer. The check adds no execute-time work, does not affect ordinary attention, and does not gate FROST participation on the backend version. FROST and COMPOSITE do not support this block-mask input.

Historical provider fitting used September 18 B200/RTX PRO 6000 BF16 measurements with cuDNN 9.27 and DSL 4.7.1. Raw records, source provenance and report-only ranking comparisons are archived separately. Public tests verify planning and execution contracts with synthetic policy outcomes and limits; they do not freeze historical winners, thresholds or timing tables.

No public signature or engine-identity changes. Default ordering changes, so positional plan indices must be resolved again. Explicit engine/knob selection and autotuning retain their semantics. FP16/BF16 forward SDPA native rejection can now surface at `create_execution_plans()` rather than `validate()` on every architecture, allowing an eligible Python engine to serve the graph.

Prepared override support is bounded by the compiled plan. Unsupported tensor-adapter configurations remain available to ordinary graphs and fall back to native planning for override-enabled graphs. SM120 THD/paged-prefill preference and decode-shaped D192x128 preference retain existing extrapolation; they are not newly measured cases.

The native version guard uses released version lines; it does not certify arbitrary prerelease builds with the same numeric version. Mask contents are mutable device data, so planning cannot safely special-case an all-visible mask. Older affected block-mask graphs report an actionable unsupported error.

Ordering targets GPU time relevant to captured execution. It does not guarantee lower eager latency, and no FlashInfer/vLLM or full-model gain is claimed.

## Testing

- Current head `bf55b1ed`: 175 focused host tests pass (3 skips, 1 deselected), plus 11 prepared-admission contract cases. Six negative controls expose the earlier blanket decline or legacy acceptance. The publication tree is identical to the tested `0fa93ae9` tree; only unpublished commit metadata was cleaned.
- Combined #1151/#1152/internal-heuristics correctness gate: **880 passed, 3 skipped, 1 deselected**. This includes override-enabled D128/D256 split and unsplit execution, THD bounded overrides, legacy override rejection, capture/GC lifetime and public-knob roundtrip tests. Native sources were verified against the freshly built extension.
- Dedicated production B200, official cuDNN 9.26.0.51 GA, FROST flag unset: D128 and D256 bounded KV overrides retain the prepared FROST route and produce O=1 / LSE=ln(1024), rather than the declared-length O=5 / LSE=ln(2048). A tensor-only synthesized-length case takes the native route and honors the same override. Full-capacity allocations and loaded library identities are checked.
- Four short-cache default-route cases separately pass full O/LSE reference checks on B200 and select the backend. The synthetic-limit regression checks the planning contract without fixing the production tuning threshold in a test.
- Changed-file formatting and license hooks pass. API-marker documentation and public export checks are corrected; no historical ranking-table tests or CSVs were restored. Fresh full CI is requested for this head.

Previously recorded native block-mask evidence remains applicable: official 9.25 zero/NaN/zero TMEM controls produce O NaN counts 0 → 196608 → 0 on identical inputs, while official 9.26 passes 36 fill controls and eight finite-sentinel O/LSE controls. The negative control did not reproduce on the dedicated B200, demonstrating its process/device dependence rather than proving a safe older backend. The guard was separately checked across version/architecture boundaries. No local performance is reported.

Earlier production B200 provider measurements passed 38 FP16/BF16 route/reference settings plus 16 D512 settings. They are historical snapshots, not performance measurements of this final combined head. Representative BF16 GPU times from that comparison:

| B / Hq / Hkv / Sq / Skv / D | Native | FROST |
|---|---:|---:|
| 8 / 32 / 8 / 1 / 8192 / 128 | 55.4 µs | 60.6 µs |
| 8 / 32 / 8 / 2 / 8192 / 128 | 224.5 µs | 61.4 µs |
| 2 / 16 / 4 / 1 / 16384 / 256 | 54.7 µs | 47.7 µs |
| 1 / 32 / 1 / 1 / 8192 / 512 | 33.9 µs | 74.9 µs |

Those provider timings use public 9.26.0.51 GA and same-process paired ordering with L2 flushed outside timing; they exclude host enqueue and planning. Raw datasets stay outside the public source tree and are not CI ranking gates.

## Summary by CodeRabbit

- **New Features**
  - FROST SDPA forward engines for SM100 and SM120 now participate by default for FP16/BF16 workloads, with ranking based on measured shape categories.
  - Enable `CUDNN_FRONTEND_ENABLE_FROST_ENGINES=1` to prioritize all available FROST engines.
  - Prepared SDPA plans support more compatible runtime shape and stride overrides, including split-KV and supported THD workloads.

- **Bug Fixes**
  - Added validation for unsupported block-mask configurations on older cuDNN versions and SM10x hardware.
  - Legacy UID-map execution now clearly rejects runtime shape or stride overrides instead of ignoring them.


YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `db71b8886b94` against base `667fe4ce8ce4`.

The placement-marker design preserves the existing mode/fallback structure, but default activation exposes one confirmed high-risk P1: a supported execute-time shape override can now silently compute the declared extent instead. Please address it before merging. I also found a bounded short-cache performance regression outside the fitted benchmark domain. This is a comment-only review of the author's own PR; no changes requested through GitHub's blocking review state.

**[P1 · blocks approval] Preserve shape overrides before making the FROST row a default candidate** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/db71b8886b944a5e7999fee57b8a24e620a70f1c/python/cudnn/engines/manifest.py#L242))

Removing the opt-in gate also changes override-enabled graphs. Reproduction with the flag unset: BF16, B=1/Hq=8/Hkv=1/Sq=4/Skv=2048/D=128, BSHD storage, generate_stats=True, is_override_shape_enabled=True, modes A+FALLBACK. Set Q=K=0 and V to 1 for keys 0..1023 and 9 thereafter; execute with K/V shapes overridden to Skv=1024 while retaining the full allocations and physical strides. Develop selects native and returns O=1, LSE=log(1024). This head automatically selects FROST SPLIT_KV=8 and returns O=5, LSE=log(2048), without an error. This plan has takes_variant_pack=False, so _pygraph.execute sends only the raw uid map and drops the override triple. Please keep override-enabled graphs on a compatible provider until every offered plan honors the overrides, or implement the missing propagation and capability checks. This is high-risk P1 because an existing legal call changes from correct native execution to silent wrong results; the underlying legacy executor limitation predates this PR, but the default-route regression is introduced here.

Evidence: Independently reproduced on a production B200 and two local SM100 boards with cuDNN 9.25.1.1 and DSL 4.8. On the same inputs the current-head native-pinned control and exact develop 667fe4ce8 baseline both return O=1/LSE=6.9314718; current-head automatic FROST returns O=5/LSE=7.6246190. All buffers retain declared capacity, so the reproduction requires no out-of-bounds access. Attribution: new default dispatch exposes an existing override-dropping executor path

**[P2] Bound the short-query rule by the measured KV domain** ([source](https://github.com/NVIDIA/cudnn-frontend/blob/db71b8886b944a5e7999fee57b8a24e620a70f1c/python/cudnn/sdpa/fwd/placement.py#L116))

The 2<=Sq<=16 branch returns LEAD for arbitrarily short KV, although its cited measurements start at 2k. On B200, BF16, bottom-right causal, B=1/Sq=4/Skv=64, the default/native CUDA-Graph device times are 5.00/3.08 us for Hq=8/Hkv=1/D=128, 8.39/4.74 us for Hq=32/Hkv=2/D=256, and 13.95/6.47 us for Hq=32/Hkv=1/D=512. The unconditional D512 prefill branch has the same boundary: B=1/Hq=8/Hkv=1/Sq=Skv=64 is 13.98/6.67 us. Please include short caches in the calibration or conservatively retain native below a supported boundary. The corresponding 2k cases do benefit from FROST, so this calls for a domain boundary rather than reversing the general ordering. These are bounded microsecond regressions, hence nonblocking P2.

Evidence: Exact-head B200 (148 SMs), cuDNN 9.25.1.1, DSL 4.8, full FP64 O/LSE checks and changed-input/poisoned-output replay controls passed. Six alternating timing rounds compare auto/native/forced-FROST on identical inputs; each GPU observation amortizes 640 executions. Native and kernel implementation files are unchanged for these ordinary-attention cases. CPU enqueue is measured separately: the short cases are approximately 30-38 us for FROST versus 9-10 us native; the PR already explicitly excludes an eager-wall improvement guarantee. Attribution: new placement policy extrapolates beyond its fitted KV domain

Validation: Built native bindings from db71b8886. Focused placement/heuristics/native-lowering tests: 128 passed, 2 skipped, 2 failed. Both failures reproduce identically on exact develop 667fe4ce8 (candidate-count and plan-name assertions), so they are not attributed to this PR. Shared dispatch, import-boundary and deterministic mask-seed tests: 75 passed. The new AUTO/UNIFIED x default/opt-in native block-mask rejection cases pass on cuDNN 9.25.1.1. Production B200: all 15 non-override shapes passed automatic/native/forced-FROST route checks, full FP64 O/LSE comparisons, and changed-input/poisoned-output CUDA-Graph replay. Host enqueue and captured device timings were measured separately. The separate analytic override probe reproduces the P1. Current-head pipeline 68841401 was already running; its mirrored commit message identifies db71b8886 exactly. Two inspected analysis failures need cleanup: the API index lacks the newly exported cudnn.engines.BACKEND, and guardwords_scan rejects absolute machine paths embedded in committed CSV failure tracebacks. Regenerate the intentional API entry and sanitize those diagnostic cells. No duplicate pipeline was triggered.

Limitations: The independent GPU measurements use public cuDNN 9.25.1.1, which remains inside the proposed default-routing domain; they do not revalidate the author's separate 9.26 GA measurements or the historical 9.27 CSV timings. No independent SM120 or Rubin GPU validation in this review. Both locally available Blackwell boards report SM100; architecture identity was rechecked rather than inferred from their names. The full current-head CI pipeline is still pending. The high-risk correctness finding makes this revision unsuitable for automatic approval regardless of the remaining CI outcome.

YangXu1990uiuc:
CI follow-up for `db71b8886`: pipeline **68841401** finished with **24 passed, 9 failed, 18 skipped, 5 manual**. FROST SDPA passed on **SM80/SM100/SM103/SM120**; SM100 GEMM and linear also passed.

I inspected all nine failing job summaries:

- Two PR-specific analysis failures: update the API index for `cudnn.engines.BACKEND`, and sanitize the machine paths in the committed CSV failure tracebacks.
- All six Python jobs fail the same four cuTile GDN/KDA correlated-key cases also observed in #1151's pipeline 68830210.
- SM107 has the same 12 ragged-decode native NVRTC compilation failures, plus a backward `test86` launch failure/worker crash. That backward failure remains **unattributed**; I have not independently rerun the matching nightly baseline.

The [independently reproduced shape-override P1](https://github.com/NVIDIA/cudnn-frontend/pull/1152#pullrequestreview-5259511061) remains unresolved despite the passing SDPA lanes. No additional CI run was started.

YangXu1990uiuc (commented):

Reviewed `4042742e0039` against base `667fe4ce8ce4`.

The test/documentation cleanup looks reasonable: synthetic placement verdicts verify the planner contract without freezing historical performance rankings. I found no new issue in this delta. The earlier high-risk P1 remains approval-blocking: on this exact head the safe KV override probe still returns O=5 and LSE=log(2048) through automatic FROST, while the native-pinned control returns the correct O=1 and LSE=log(1024). Please address it before merging. This revision changes no executable placement or dispatch logic, so the earlier short-cache P2 also remains. Comment-only review of the author's own PR.

- **[P1 · blocks approval] Still open:** [Preserve shape overrides before making the FROST row a default candidate](https://github.com/NVIDIA/cudnn-frontend/pull/1152#pullrequestreview-5259511061).

- **[P2] Still open:** [Bound the short-query rule by the measured KV domain](https://github.com/NVIDIA/cudnn-frontend/pull/1152#pullrequestreview-5259511061).

Validation: Exact-head placement, shared dispatch, import-boundary and deterministic-mask tests: 88 passed. The replacement placement tests preserve recommendations, marker position/count, explicit opt-in behavior and empty eligibility. Verified identical executable placement AST and byte-identical manifest, graph dispatcher, forward engine and heuristic implementation versus db71b8886; reused its matching native extension without changing a shared installation. Independently repeated the analytic shape-override probe on SM100 with public cuDNN 9.25.1.1 and CuTe DSL 4.8: auto still fails numerically; same-head native control is exact. Every allocation retains declared capacity. Removing the committed CSVs removes the specific guardwords inputs reported in the previous CI run. The previously reported cudnn.engines.BACKEND API-index omission is still present: its baseline is unchanged. Current-head pre-commit passed; no new broad CI run was requested because the correctness blocker remains.

Limitations: Previous B200 performance and full-suite evidence are reused only for unchanged runtime code; this round did not rerun production-GPU timing, SM120 or Rubin tests. The previous pipeline 68841401 belongs to db71b8886, not this head. Removing the offending CSV inputs is a source check, not a new successful guardwords CI result.

vedaanta (approved):

YangXu1990uiuc:

YangXu1990uiuc (commented):

Reviewed `bf55b1ed7c10` against base `b100e34239d4`.

Both findings from my earlier review are addressed. The default split-KV route now honors the demonstrated KV override through the prepared VariantPack executor, and all four reported short-cache shapes return to native-first selection. Sharing prepared admission between planning and lowering, plus rejecting overrides on legacy executors, is a sound fix. No remaining approval blocker found in this follow-up. Please carry the nonblocking [native lifetime-test layout fix reported on #1151](https://github.com/NVIDIA/cudnn-frontend/pull/1151#pullrequestreview-5261340734) when updating that dependency; preserve the stated merge-after-#1151 order. Comment-only, retaining the existing approval.

Validation: Rebuilt bf55b1ed's native binding and verified both loaded source paths. With FROST opt-in disabled, the original BF16 B1/Hq8/Hkv1/Sq4/Skv2048/D128 override reproducer now automatically selects FROST split8 with takes_variant_pack=True and returns exact O=1 and FP32 LSE=log(1024), matching the native-pinned control. Allocations retain their full declared capacity. SM100, cuDNN 9.26.0.51, CuTe DSL 4.7.0: 7 focused prepared tests passed, covering D128/D256 split/unsplit bounded KV overrides, THD bounded overrides and collection of a retired plan during capture. Planning admission, shared dispatch and placement suites: 233 passed, including early legacy-plan rejection and compatible split-candidate preservation. All four originally reported KV64 shapes now select the same native plan as explicit native selection: D128/D256/D512 at Sq4 and D512 at Sq64. Full FP64 O/LSE references and changed-V/poisoned-output CUDA Graph replay pass for both routes in all four cases. A separate local SM100 comparison of default/native/forced-FROST at D128 Sq4 and KV64/256/2048 passed full O/LSE and changed-input replay checks. Default execution matches the native route and its CPU/GPU timing at the two short extents, while KV2048 retains the prepared FROST route. This is a bounded component check, not a final B200 performance claim. Verified that the inherited prepared/buffer/kernel changes match the #1151 code independently validated immediately before this review. Reused that evidence while testing the newly shared admission predicate and default-route behavior here.

Limitations: Owner-triggered pipeline 68922439 remains running; its mirror identifies the full bf55b1ed head. The inspected SM90 job has four previously observed cuTile failures and the two inherited lifetime-fixture failures above. This is not a clean-CI result or an independent nightly-baseline rerun. No duplicate CI was started. No independent target-GPU SM120 or Rubin rerun in this follow-up. Host tests exercise unsupported-override admission for those paths, but do not establish their device execution or performance.

YangXu1990uiuc:
Pipeline **68922439** has finished for **bf55b1ed7c10**. SM80/SM100/SM103/SM120 FROST SDPA, SM100 linear/GEMM, builds/analysis and C++ sanitizer lanes passed.

The inspected Python failures are four existing cuTile correlated-key cases on SM100 (**5,270 passed / 2,717 skipped**) and those four plus the two inherited native C=1 lifetime-test fixture failures on SM90 (**5,363 passed / 2,622 skipped**). Please carry the supported-layout test fix from [#1151's review](https://github.com/NVIDIA/cudnn-frontend/pull/1151#pullrequestreview-5261340734) into this branch; the CI summary's “Old Failure” label does not distinguish those two new tests. Rubin has the same 12 native ragged-Stats NVRTC compilation failures, with **3,229 passed / 1,963 skipped**.

No additional placement/override regression was identified in these inspected results. The earlier review's two findings remain resolved; the inherited lifetime-test fixture issue remains actionable.

YangXu1990uiuc:
Merged `develop` (e5e59a761, the #1151 squash plus #1147/#1153/#1154) into this branch as c2d5ea2ed. develop's #1151 squash is byte-identical to the #1151 branch head this PR already carried, so the three textual conflicts (`python/cudnn/AGENTS.md`, `python/cudnn/sdpa/fwd/engines.py`, `test/python/sdpa/frost/test_sdpa_prepared_thd.py`) were resolved by keeping this branch's versions; the merge differs from bf55b1ed7 only in develop's 19 bsa/dsa files, and `git diff develop` is exactly this PR's 34-file payload. Rebuilt the native extension from the merged tree and ran the FROST/placement/prepared/capture set on SM100: 824 passed, 3 skipped (the two remaining failures are the pre-existing L1-only `test_split_kv_plan_pinned_by_name_matches_reference` / `test_runner_up_sched_plan_builds_and_matches_the_winner`, which expect lower-case knob names, are unchanged on develop and are not run by CI's `-m L0` default).

f3ce2b0ee fixes the two `test_cuda_capture_lifetime.py` failures on the SM80/SM90 lanes: the native ReLU graph used a 4-D `[1, 1, 32, 32]` tensor that the runtime-fusion pointwise engine refuses before SM100 (`getDimA()[1] == 1`; only the SM10x-only TensorIR MemBound engine accepted it), so the test failed at `graph.build()` before the capture it exists to exercise. A 3-D `[1, 32, 32]` tensor builds everywhere: 4/4 twice each on A100, L40S, H100 and SM100. develop carries the same test through the #1151 squash, so its SM80/SM90 lanes are red for the same reason until this lands.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer · workspace /data/yanxu/worktrees/scratch-gpu/fe-wt-1152</sub>

YangXu1990uiuc:

YangXu1990uiuc:
Pipeline 68961155 on f3ce2b0ee: every red job is develop-known except two, and neither is this PR's:

- `frost-sdpa:cutlass-rel:sm100` failed with `job_execution_timeout` (1h) at 83% of 7860 items with **0 test failures** before the kill (the three `node down` / `F` lines are the workers being terminated). This lane runs 52-58 min on every recent pipeline (develop's own nightlies: 58.5 and 60.0 min on `cudnn-dev-zenith-22-04`); the same tree minus the lifetime-test layout fix passed this lane in 57:12 on `cudnn-dev-abeam-22-04` (pipeline 68922439, 5956 passed), and the fixed lifetime test passes 4/4 on SM100 locally. The lane sits at its timeout edge and needs a longer `timeout` in `ci/stages/frost_tests/jobs.yml`, independent of this PR.
- `py_test:rel:sm80`: `test_matmul_fuzzer.py::test_matmul_fuzz[t1441_...]` is a `torch.OutOfMemoryError` on the 40 GB A100 while a sibling xdist worker held 9.6 GiB (fuzzer input allocations, not a numerical failure).

SM80/SM90 `py_test` are back to the four develop-known cuTile ids (the two `test_cuda_capture_lifetime` failures are gone), `frost-sdpa:cutlass-rel:sm103` / `sm120` / `frost-linear` / `frost-gemm` passed.

<sub>note to self: claude::61d24ed2-7c90-4a97-9cbb-b91ae18136fd — "flashinfer frost prefill/GEMM enablement" · cwd /home/scratch.yanxu_libs/flashinfer</sub>

